PR #62, opened 2019-03-14
Starting "static-app" test package

ef4:
This is an app configured for maximum static optimization. This first test verifies that it can consume two helpers out of ember-composable-helpers without pulling in any unused ones.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -21,6 +21,9 @@ function main {
   "macro-addon-classic")
     cd test-packages/macro-sample-addon && CLASSIC=true yarn test
     ;;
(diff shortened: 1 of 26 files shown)


stefanpenner:
nice!

ef4:
Can't tell yet why this is failing in CI, I can't make it fail locally.

(This is also the blocker for #63, which builds off this branch.)

ef4:
Oh, nevermind, I see. It's a problem with gitignore, I didn't actually commit all the files.

(It's slightly confusing because the main packages are all typescript and don't check in the javascript, but the test apps are Javascript.)

Issue #49, opened 2019-03-06
v1-addon css trees also get js and template treatment


stefanpenner:
Although the v2 world has a much improved approach here, I think for performance / parity reasons v1 should not:

https://github.com/embroider-build/embroider/blob/0c38b916ad7d2af2ef15bdc9b1fb6f1db7593898/packages/compat/src/v1-addon.ts#L195

* Now css trees will must also babel/hbs transform, which results in much IO
* introduces new functionality, but I think we for legacy compat we should likely not

PR #63, opened 2019-03-15
App styles handling

ef4:
This introduces a second synthetic package: `@embroider/synthesized-styles`. 

Our first synthetic package was `@embroider/synthesized-vendor`. It exists to be the place where vendor trees from all addons plus the app can get mashed together in the classic way, while exposing the resulting contents to Embroider in a spec-compliant way.

`@embroider/synthesized-styles` serves a similar purpose for the app styles tree handling. We need to mash together the results from all addons plus the app before transpiling, because that is how things like sass imports work. The synthetic package ends up containing the `assets/your-app-name.css` file, and it uses the `public-assets` metadata to ensure that file (and any accompanying sourcemap) ends up in the app.

This PR also fixed a bug in the way we make imports from your own package relative, when you have a customized addon broccoli tree.

### DIFF
--- a/packages/compat/src/v1-addon.ts
+++ b/packages/compat/src/v1-addon.ts
@@ -315,11 +315,11 @@ export default class V1Addon implements V1Package {
 
   protected treeForAddon(): Tree|undefined {
     if (this.customizes('treeForAddon', 'treeForAddonTemplates')) {
-      let tree = this.invokeOriginalTreeFor('addon');
+      let tree = this.invokeOriginalTreeFor('addon', { neuterPreprocessors: true });
       if (tree) {
-        return new MultiFunnel(tree, {
+        return this.transpile(new MultiFunnel(tree, {
           srcDirs: [this.moduleName, `modules/${this.moduleName}`]
-        });
+        }));
       }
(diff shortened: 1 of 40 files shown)


packages/compat/src/v1-addon.ts
```diff
@@ -374,23 +374,24 @@ export default class V1Addon implements V1Package {
...
-      // common pattern as far as I can tell.
-      //
-      // TODO: detect people doing the right thing (namespacing with their own
-      // package name) and send them down the happy path. Their styles can
-      // just ship inside the package root and be importable at the same name
-      // as before. Detect people doing anything other than that and yell at
-      // them and set up a fallback.
-      built.trees.push(
-        this.transpile(this.stockTree('styles', {
+      // the user's tree returns their own styles with no "app/styles" wrapping
```
stefanpenner: @ef4 I believe https://github.com/embroider-build/embroider/issues/49 is fixed by this change, is that correct?

ef4: This avoids running babel over the app styles, but we need to do the same for addon styles too.

PR #64, opened 2019-03-15
Fix macro if presets

ef4:
This refactors our macros so they are more eager. This keeps them correct even when there are also aggressive presets configured.


ef4:
We now run every macro test four times: babel 6 and 7, and with and without concurrent IE11 presets.

PR #76, opened 2019-03-19
[Failing Test] Adding test for addon pod component

kiwiupover:
An addon with a pod component throws a build error.

Includes a failing test.

Passes:
```
cd test-packages/macro-tests && CLASSIC=true yarn test
```

Fails
```
cd test-packages/macro-tests && yarn test
```

Failure
`Module not found: Error: Can't resolve './template' in '/private/var/folders/59/6y4_b2y15ys0bk910x5rfgg00000gn/T/embroider-DP1uY8/test-packages/macro-sample-addon/components/pod-component'`

### DIFF
--- a/test-packages/macro-sample-addon/addon/components/classic-component.js
+++ b/test-packages/macro-sample-addon/addon/components/classic-component.js
@@ -1,0 +1,6 @@
+import Component from '@ember/component';
+import layout from '../templates/components/classic-component';
+
+export default Component.extend({
+  layout
+});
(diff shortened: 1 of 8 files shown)


ef4:
Thanks, we definitely don't have working pod support and tests like this are the first step.

ef4:
As of #129 these tests now pass.

I'm not going to keep them because there is now other test coverage for this feature, and these ones don't really belong in the macro-tests packages because they're not about macros.

But thanks again for providing this runnable test case so I could verify that I covered your bug.

kiwiupover:
Thanks heaps Ed!

PR #78, opened 2019-03-19
tsconfig: Remove trailing comma

Turbo87:
Trailing commas in JSON are invalid, unfortunately 😉

/cc @stefanpenner @ef4

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -19,6 +19,6 @@
     "packages/*/dist",
     "test-packages/*/tmp",
     "test-packages/*/dist",
-    "tmp",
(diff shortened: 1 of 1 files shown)


ef4:
tsconfig cheats and isn't really `json` despite using the file extension. It's allowed to contain comments, for example.

So I think the comma is actually allowed by typescript. But I am also fine with merging this PR because I think it's weird to have non-json in a `json` file.

Turbo87:
yeah, I noticed that it wasn't breaking, but my IDE still complained about the extra comma 😉

PR #79, opened 2019-03-19
CI: Run `node-test` script too

Turbo87:
apparently we only ran `yarn test` and `yarn lint` until now 😱

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,5 +16,6 @@ install:
 
 script:
   - yarn lint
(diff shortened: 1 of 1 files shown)


ef4:
I think we already run this via:

`yarn test` -> `./scripts/test.sh` -> invokes 7 different test suites including `yarn node-test`

Turbo87:
hmm, indeed. looks like I missed that...

Issue #80, opened 2019-03-19
Refactor test runner


Turbo87:
Our test runner seems somewhat special and also does not run on Windows at all. From my first glance this is the current situation:

- `yarn node-test` is running QUnit and the unit tests in the `packages` folders
- `yarn test` is running `parallel` which runs `yarn test` (`ember test`) for most of the folders in `test-packages` in parallel

I'm wondering if instead of using `parallel` we should use a proper test runner that runs tests in parallel by default, e.g. Jest.

Refactoring the TS unit tests to Jest seems quite straight forward and running `ember test` on the `test-packages` should also be feasible using something like https://github.com/sindresorhus/execa/

@stefanpenner @ef4 I'd be happy to look into that if you're okay with the direction.

ef4:
Yes, that sounds good. My main concern would be to keep the output from each suite nicely grouped and labeled as it is now, rather than interleaved.

PR #81, opened 2019-03-19
Replace custom test runner with Jest 

Turbo87:
This is a first step, as discussed in #80 

The next step will involve porting the existing QUnit test suites to Jest and enabling Jest to run TS code directly

/cc @ef4


ef4:
Thanks! This looks nice.

Within each file, are the tests parallel or serial? For example, these two tests:

```js
const TESTS = [
  ['macro-addon', 'test', { cwd: `${__dirname}/macro-sample-addon` }],
  ['macro-addon-classic', 'test', { cwd: `${__dirname}/macro-sample-addon`, env: { CLASSIC: 'true' } }],
];

for (let [testName, command, options] of TESTS) {
  test(testName, async () => {
    jest.setTimeout(60000);

    try {
      await execa('yarn', [command], options);
    } catch (error) {
      throw error;
    }
  });
}
```

It's probably not a big deal either way but I'd like to make sure I understand.

Turbo87:
If I understand correctly Jest parallelizes per-file. In other words: the tests within one file are run sequentially, which is why I added 2ad4268 to make them parallelize at least a little bit more.

PR #88, opened 2019-03-21
webpack: Remove unused `node-range` module

Turbo87:
The module is not exporting any thing and is not imported from anywhere


ef4:
Oh yes, this should never have been committed. It was an early version of the file that became `html-placeholder.ts`.

PR #89, opened 2019-03-21
Convert TSLint rules to ESLint

Turbo87:
This PR converts some of the enabled TSLint rules to ESLint rules instead. I could not find a suitable replacement for all of them yet and some of the ESLint replacements are a little more strict, but this is the subset that could be converted without issues.


Turbo87:
@ef4 looks like a "race condition" while merging caused CI to currently fail on `master`. the latest commit in this PR should fix it.

PR #94, opened 2019-03-21
Move `test-support` package into `packages` folder

Turbo87:
This separates test helper code from test fixtures, which slightly simplifies the TypeScript configuration.


ef4:
My original intent was that `packages` is for real NPM packages that we publish, and test-packages is for anything else that is local to the monorepo. `@embroider/test-support` never publishes to NPM.

I see what you mean about simplifying the typescript config, but I think it's taking advantage of a coincidence. Another future test-package may want to use typescript, but it would still be appropriate for that to stay in test-packages.

PR #99, opened 2019-03-25
avoid port collisions

ef4:
This should really be ember-cli's default.

It can cause sporadic test failures when we're running multiple test suites in parallel.

### DIFF
--- a/test-packages/sample-transforms/package.json
+++ b/test-packages/sample-transforms/package.json
@@ -17,7 +17,7 @@
     "lint:hbs": "ember-template-lint .",
     "lint:js": "eslint .",
     "start": "ember serve",
-    "test": "ember test",
+    "test": "ember test --test-port=0",
     "test:all": "ember try:each"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


stefanpenner:
I usually set me `$HOME/.ember-cli` to have `port: 0` but ya, I think that that we no longer have both a livereload port and an http port, we should totally use `port: 0` as a default.

@rwjblue has been bugging me about this for ages...

stefanpenner:
@sparshithNR if you happen to have any free cycles, thoughts on making the above change to ember-cli?

SparshithNR:
If `--port` is not defined we just default it to auto find the port?

stefanpenner:
@SparshithNR today, if `--port` is omitted by the developer we try `4200`, if instead the default became `0` we would automatically just find an open port.

Turbo87:
@stefanpenner ideally we should check 4200 first though, otherwise a lot of people will be confused ;)

SparshithNR:
@stefanpenner @ef4 Created a PR https://github.com/ember-cli/ember-cli/pull/8578

stefanpenner:
@Turbo87 thats not a bad idea

SparshithNR:
@stefanpenner We never had 4200 port for testing. What we had before was `7357`. `4200` is for `ember s` not for `ember test`

PR #102, opened 2019-03-26
add workspace-reused options to default pipeline

ef4:


ef4:
For those playing along at home, this allows us to benchmark how much of the build cost goes into v1 addon compatibility. If you build first with `SAVE_WORKSPACE=/tmp/my-workspace` and then build again with `REUSE_WORKSPACE=/tmp/my-workspace`, your second build will skip rewriting all addons to v2 format. 

(Running in this mode you will miss some live-updating behaviors that v1 addons rely on. For example, ember-svg-jar won't be able to emit new SVGs into your app until you let it rebuild.)

PR #105, opened 2019-03-28
 implement cacheKeyForTree

ef4:
This implements the `cacheKeyForTree` optimization used in stock ember-cli so that we don't rebuild multiple copies of the same addon's trees when that addon is used multiple times. For us it's only relevant in possibly speeding up stage1.

I tested the impact of this in ember-observer and it didn't make a measurable difference. One possible reason is that the trees we are caching here are smaller and simpler -- they don't recurse the way the stock `treeFor()` does.

### DIFF
--- a/packages/compat/src/v1-addon.ts
+++ b/packages/compat/src/v1-addon.ts
@@ -4,7 +4,6 @@ import { dirname } from 'path';
 import { sync as pkgUpSync } from 'pkg-up';
 import { join } from 'path';
 import { existsSync, pathExistsSync } from 'fs-extra';
-import resolvePackagePath from 'resolve-package-path';
 import Funnel, { Options as FunnelOptions } from 'broccoli-funnel';
 import { UnwatchedDir } from 'broccoli-source';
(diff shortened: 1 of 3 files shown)


stefanpenner:
cc @rwjblue (you mentioned a relevant fix dropping soon, figure I would loop you in)

ef4:
I would appreciate review from either of you. 

The overall issue here is that I don't want to call `addonInstance.treeFor()` because it recurses and I really just want the addon's own tree. So I call `_treeFor()`, which has the right content, but lacks the caching, so I'm adding back the caching in this PR.

rwjblue:
> I tested the impact of this in ember-observer and it didn't make a measurable difference.

FWIW, the impact is much more noticeable on apps that use engines where multiple engines depend on the same addons (before the cacheKeyForTree each engine would rebuild the css/js for the dependency only to throw it away :sob:).

stefanpenner:
relevant thing @rwjblue brought up today: https://github.com/ember-cli/calculate-cache-key-for-tree/pull/14 (fix to how the cache keys are calculated)

PR #109, opened 2019-03-29
Per route code splitting

ef4:
This introduces the `splitAtRoutes` option. It lets you choose whatever set of routes you'd like to be split points, and it splits them and their children into lazily loaded bundles.

It also introduces the `@embroider/router` package, which is a very small extension to the stock Ember router that causes it to notice the existence of these lazy bundles and load them at the appropriate time.


villander:
I'm loving this hard work on Embroider, tree-shaking & code splitting are arriving. It's will be very helpful solving problems that Ember Engines don't can't solve alone.

Anxious to see Ember Engines + Embroider

thank you so much @ef4

PR #126, opened 2019-04-12
linter pass

givanse:
These are the results of executing `yarn lint` with the config changes found in https://github.com/embroider-build/embroider/pull/127

PR #127, opened 2019-04-12
single linter config

givanse:
- deleted all the `.eslintrc.js` except the one at the root
 - added some missing packages (`import/no-extraneous-dependencies'`)
 - in `packages/router` deleted eslint related packages and scripts entry

 - [ ] https://github.com/embroider-build/embroider/pull/126


packages/router/package.json
```diff
@@ -15,13 +15,20 @@
...
-    "ember-cli-babel": "^7.1.2"
+    "@ember/test-helpers": "^1.5.0",
+    "@embroider/compat": "^0.3.3",
+    "@embroider/macros": "^0.3.3",
+    "@embroider/webpack": "^0.3.3",
+    "ember-cli": "^3.9.0",
+    "ember-cli-babel": "^7.1.2",
+    "ember-qunit": "^4.4.1",
+    "ember-source-channel-url": "^1.1.0",
+    "qunit": "^2.9.2"
```
ef4: This is not correct. We cannot have `dependencies` on these things, they're really supposed to be devDependencies.

Presumably the linter told you to do this, but that's because some override got dropped that tells eslint that certain folders contain tests and are allowed to use devDependencies.

I think eslint also respects .npmignore, so if you make sure the tests are in .npmignore that would also address the spurious lint errors.

givanse: Yeah, doesn't make sense to have `qunit` there, should have noticed it. I'll look into this.

ef4:
Closing for inactivity, happy to reopen if someone has time to pick this up.

Issue #134, opened 2019-04-13
Could not find module `ember-power-calendar-utils`


josemarluedke:
Now that I have been unblocked by my previous issue, I got a new one.

This is related to ember-power-calendar, as it is doing some dynamic package under the hood. This works in classic Ember CLI, so that's why I'm opening it here. Pinging @cibernox to be in the loop here as well.

**The error:**

It only appears in the browser, not in the terminal as similar errors.

```
loader.js:247 Uncaught Error: Could not find module `ember-power-calendar-utils` imported from `(require)`
    at missingModule (loader.js:247)
    at findModule (loader.js:258)
    at requireModule (loader.js:24)
    at w._embroider_ (my-embroider-test.js:10)
    at eval (webpack:///external_%22_embroider_(:4200/\%22ember-power-calendar-utils\%22)%22?:1:18)
    at Object.ember-power-calendar-utils (chunk.8f1de7ceb06e8364c248.js:164)
    at __webpack_require__ (chunk.8f1de7ceb06e8364c248.js:79)
    at eval (power-calendar.js:4)
    at Module../node_modules/ember-power-calendar/components/power-calendar.js (chunk.5f9c7a14021a58709b9c.js:1695)
    at __webpack_require__ (chunk.8f1de7ceb06e8364c248.js:79)
```

**Embroider Version:**
v0.3.4

**Reproduction:**
https://github.com/josemarluedke/my-embroider-test/tree/power-calendar

cibernox:
To clarify what weird thing I'm doing.

In ember-power-calendar I need to do date manipulation. Doing date manipulation without any date library is the shortest way to quit programming and become a farmer that there is, so I need to use one.

My favourite date libraries are moment.js and luxon.js, the latter being nicer than the former and lighter. I could have chosen one, say, luxon.js, but that'd mean I have to ship a heavy library to an app that might already be using moment.js.

What I did is create a handful of library-agnostic date helpers, and implement them in them using both libraries in `ember-power-calendar-moment` and `ember-power-calendar-luxon`. In both I do a little bit of [broccoli trickery](https://github.com/cibernox/ember-power-calendar-luxon/blob/master/index.js#L11-L22) to make them export those helpers in the same import path, so the users gets to choose what date-library is their favorite based without bloat the size of the app.

If there is a way of refactoring this to make it easier for embroider to digest I'm all ears.

ef4:
[...]

PR #136, opened 2019-04-16
Prevent addonTree from escaping own module namespace

ef4:
We already had support for noticing if `treeForAddonTestSupport` was trying to escape its package's own namespace and rewriting it so it doesn't.

Apparently `treeForAddon` can also do that, so this applies the same fix there.

This PR changes the role of MultiFunnel from what it was doing before. Before it was doing two jobs -- handling the possible wrapping `modules` directory, and stripping off the addon's own name directory. Now the addon's own name directory is handled by rewriteAddonTree instead, and MultiFunnel only deals with `modules`.


ef4:
Fixes #134.

Issue #152, opened 2019-04-30
Embroider clobbers local `this.options` in ember-inline-svg


patocallaghan:
If if try to use embroider on an app that uses ember-inline-svg I get the following error: `this.options is not a function`.

It appears when using embroider the local `this.options` function is clobbered by an object hash

![image](https://user-images.githubusercontent.com/685034/56994397-db4c7c00-6b96-11e9-9d7d-70c48aa510a6.png)

Whereas without embroider it has the correct `this.options` function

![image](https://user-images.githubusercontent.com/685034/56994329-ab04dd80-6b96-11e9-9c13-ebc3bb85815e.png)

The relevant `this.options` code in ember-inline-svg is here https://github.com/minutebase/ember-inline-svg/blob/ca91b01daf73b7c714b83053b5e0057dee1a2ffd/index.js#L20-L32

I've also created a sample app which replicates the problem [here](https://github.com/patocallaghan/embroider-test-app/tree/f97fa30b7167e960f0f901c740f29e408db50739)

**Error log:**
```
=================================================================================

ENV Summary:

  TIME: Tue Apr 30 2019 22:14:51 GMT+0100 (Irish Standard Time)
  TITLE: ember
  ARGV:
  - /Users/pat/.nvm/versions/node/v10.15.0/bin/node
  - /Users/pat/hacking/embroider-test-app/node_modules/.bin/ember
  - build
  EXEC_PATH: /Users/pat/.nvm/versions/node/v10.15.0/bin/node
  TMPDIR: /var/folders/mk/5t937px5627_y_j551f8gbkm0000gn/T
  SHELL: /bin/zsh
  PATH:
  - /var/folders/mk/5t937px5627_y_j551f8gbkm0000gn/T/yarn--1556658883945-0.39350346415838433
  - /Users/pat/hacking/embroider-test-app/node_modules/.bin
  - /Users/pat/.config/yarn/link/node_modules/.bin
  - /Users/pat/hacking/embroider-test-app/node_modules/.bin
  - /Users/pat/.config/yarn/link/node_modules/.bin
  - /Users/pat/.nvm/versions/node/v10.15.0/libexec/lib/node_modules/npm/bin/node-gyp-bin
  - /Users/pat/.nvm/versions/node/v10.15.0/lib/node_modules/npm/bin/node-gyp-bin
  - /Users/pat/.nvm/versions/node/v10.15.0/bin/node_modules/npm/bin/node-gyp-bin
[...]

PR #153, opened 2019-04-30
Don't clobber `this.options` in v1-addons

patocallaghan:
Potential fix for https://github.com/embroider-build/embroider/issues/152

Looking at emberobserver.com, while not many addons do it, there is a pattern of some addons [declaring `this.options` to be a function](https://emberobserver.com/code-search?codeQuery=options%3A%20function()).

When we are setting up the `options` for v1-addons we clobber `this.options` to be a hash. This has a consequence of causing these addons to not work with Embroider. We get a `this.options is not a function` error.

Here I've added a potential fix to not clobber the existing `this.addonInstance.options` if it already exists.

I wasn't entirely sure how to test but I added a problematic addon, `ember-inline-svg` to the static-app dependencies. Before fixing, the static-app tests failed like so

image

but after adding the fix the test suite passes.

### DIFF
--- a/packages/compat/src/v1-addon.ts
+++ b/packages/compat/src/v1-addon.ts
@@ -178,29 +178,32 @@ export default class V1Addon implements V1Package {
   protected get options() {
     if (!this.addonInstance.options) {
       this.addonInstance.options = {};
-    } else {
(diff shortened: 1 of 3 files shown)


ef4:
Thanks, this looks good to me.

PR #166, opened 2019-05-10
fix externals cache lifetime

ef4:
We generate our shim externals modules as a side-effect of babel. But that means they need to live as long (on disk) as our persistent babel cache, or we can end up in a state where they are missing.

### DIFF
--- a/packages/core/src/app.ts
+++ b/packages/core/src/app.ts
@@ -23,6 +23,7 @@ import { TemplateCompilerPlugins } from '.';
 import TemplateCompiler from './template-compiler';
 import { Resolver } from './resolver';
 import { Options as AdjustImportsOptions } from './babel-plugin-adjust-imports';
+import { tmpdir } from 'os';
 
 export type EmberENV = unknown;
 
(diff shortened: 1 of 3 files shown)


packages/core/src/app.ts
```diff
@@ -322,7 +323,11 @@ export class AppBuilder<TreeNames> {
     let adjustOptions: AdjustImportsOptions = {
       rename,
       extraImports: this.adapter.extraImports(),
-      externalsDir: join(this.root, 'node_modules', '@embroider/externals'),
+
+      // it's important that this is a persistent location, because we fill it
+      // up as a side-effect of babel transpilation, and babel is subject to
+      // persistent caching.
+      externalsDir: join(tmpdir(), 'embroider', 'externals'),
```
stefanpenner: just heads-up, the OS is allowed to remove files from tmp if they haven't been accessed in some time, if they are missing will we restore them?

ef4: We will restore them *if* the corresponding babel cache in `$TMPDIR/embroider/webpack-babel-loader` is also missing.

The potential problem is if the babel cache exists and the externals cache does not. In that case, we have no opportunity to rediscover what needs to go in the externals cache because babel will not run again.

ef4: Entries in the two caches will have very similar access times, so my guess is that this mostly works, and I'm willing to wait for bug reports to find out if it doesn't.

And the cost of trying to do a fancier solution here is pretty high, because it would introduce a new API surface for all stage3 packagers.

PR #177, opened 2019-05-16
Bugfix - Handle single file renaming

jenweber:
Co-authored-by: Edward Faulkner <edward@eaf4.com>


jenweber:
Adds test coverage to go with #168

Issue #160, opened 2019-05-07
Embroider build replaces Fastboot generated package.json


dnalagatla:
My PR - https://github.com/ember-fastboot/ember-cli-fastboot/pull/690 moved the implementation form the postProcessTree to treeForPublic hook. But after doing that, I noticed embroider build doesn't add the package.json generated by `ember-cli-fastboot` addon. Instead, it keeps the package.json created by enbroider build in the dist folder. I am thinking some diff code is assuming the package.json generated as part of Fastboot tree is similar to package.json generated by embroider build and keeps embroider build related package.json

To Repro the issue 
1. clone - https://github.com/dnalagatla/my-embroider-test.git
2. run `yarn install`
3. ember s
4. launch http://localhost:4200

page will show following error:
```
Error: /var/folders/tv/fgbbtzvd1f1dbv45y7r8l09m000ty_/T/broccoli-82821i0tVejhn4W1b/out-176-packager_runner_embroider_webpack/package.json was malformed or did not contain a manifest. Ensure that you have a compatible version of ember-cli-fastboot.
    at EmberApp.readPackageJSON (/Users/dnalagat/DEVELOPMENT/ember-cli-fastboot/node_modules/fastboot/src/ember-app.js:362:13)
    at new EmberApp (/Users/dnalagat/DEVELOPMENT/ember-cli-fastboot/node_modules/fastboot/src/ember-app.js:34:23)
    at FastBoot._buildEmberApp (/Users/dnalagat/DEVELOPMENT/ember-cli-fastboot/node_modules/fastboot/src/index.js:114:17)
    at new FastBoot (/Users/dnalagat/DEVELOPMENT/ember-cli-fastboot/node_modules/fastboot/src/index.js:52:10)
    at app.use (/Users/dnalagat/DEVELOPMENT/ember-cli-fastboot/index.js:318:29)
    at Layer.handle [as handle_request] (/Users/dnalagat/DEVELOPMENT/test-pemberly-bpr/test-ember/my-embroider-test/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (/Users/dnalagat/DEVELOPMENT/test-pemberly-bpr/test-ember/my-embroider-test/node_modules/express/lib/router/index.js:317:13)
    at /Users/dnalagat/DEVELOPMENT/test-pemberly-bpr/test-ember/my-embroider-test/node_modules/express/lib/router/index.js:284:7
[...]

PR #178, opened 2019-05-17
Merge existing package.json to package created by embroider build

dnalagatla:
https://github.com/embroider-build/embroider/issues/160

### DIFF
--- a/packages/core/src/app.ts
+++ b/packages/core/src/app.ts
@@ -138,7 +138,7 @@ class BuiltEmberAsset {
 
 class ConcatenatedAsset {
   kind: 'concatenated-asset' = 'concatenated-asset';
-  constructor(public relativePath: string, public sources: (OnDiskAsset | InMemoryAsset)[]) { }
+  constructor(public relativePath: string, public sources: (OnDiskAsset | InMemoryAsset)[]) {}
   get sourcemapPath() {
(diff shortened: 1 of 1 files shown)


ef4:
Thanks, very nice work.

ef4:
@dnalagatla this had to get reverted, see #342.

st-h:
@ef4 is there any other means to fix this or a work around? I am seeing the error described in #160, which was supposed to be fixed by this PR

ef4:
@st-h I reopened #160 to cover this.

We can still merge the package.json, but it needs to be implemented differently. This implementation was merging from the wrong spot in the build process.

st-h:
@ef4 thanks for the clarification. I am just trying to give embroider a spin every now and then to see if I can make it work with my app. But I can hold off, if it's still too early. Unfortunately that I can't dedicate more time to fix stuff right now.

dnalagatla:
> @st-h I reopened #160 to cover this.
> 
> We can still merge the package.json, but it needs to be implemented differently. This implementation was merging from the wrong spot in the build process.

hi, @ef4  I am going to look into this issue. Can you please provide details regarding where the merging of package.json should happen. I can take a look into that.

ef4:
See https://github.com/embroider-build/embroider/pull/383#issuecomment-584386920

PR #179, opened 2019-05-17
Add compatAdapter for ember-inflector

ef4:
ember-inflector throws if its included more than once, which can be hard to avoid when initially adopting embroider or when using `yarn link` to things that depend on ember-inflector.

So this adapter makes it tolerant of multiple inclusion by predefining all the shared properties as configurable.

Co-Authored-By: Jen Weber <j@jenweber.me>

### DIFF
--- a/packages/compat/src/compat-adapters/ember-inflector.ts
+++ b/packages/compat/src/compat-adapters/ember-inflector.ts
@@ -1,0 +1,32 @@
+import V1Addon from '../v1-addon';
+import AddToTree from '../add-to-tree';
+import { Tree } from 'broccoli-plugin';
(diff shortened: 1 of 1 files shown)


jenweber:
I opened a PR to patch ember-inflector in order to support yarn linking etc for use cases outside embroider https://github.com/emberjs/ember-inflector/pull/183

ef4:
@jenweber the timeouts are actually caused by a gnarly bug we exposed here. We are accidentally editing the original copy of ember-inflector in node_modules, due to broccoli symlinking. 

That's why it fails even in the test suites that don't use embroider -- node_modules/ember-inflector has been corrupted.

PR #181, opened 2019-05-18
detect customized trees for add-ons have their own base-class

stefanpenner:
This fixes an issue, where not all customized tree's where correctly detected. Breaking compatibility if an add-on-main happened to export a class rather then a pojo.

- [x] tests


ef4:
@stefanpenner I have some questions about the custom base class you hit:

1. Does it ultimately extend from ember-cli's addon.js? 
2. Does it have a `_treeFor` method?
3. If it doesn't extend the build-in addon.js, do its customized hooks (like treeForAddon) have the same recursive behavior that the standard base class has? Does treeForAddon run the `treeForAddon` in each next-level-down addon?

The issue here is that on the standard base class we call `_treeFor` to avoid the recursion. But if we don't have a `_treeFor`, we probably need to go through the "front door" instead and call the real hook (like `treeForAddon`). But we would really prefer that people in those custom hooks aren't recursing.

It seems likely to me that people implementing a custom base class either probably didn't both recursing, or defer to the underlying standard implementation. In either of those cases we'd be OK.

ef4:
Also, the real public API from ember-cli's perspective is not the individual hooks like `treeForAddon`, it's `treeFor('addon')`.

So when we encounter a custom base class, it doesn't necessarily make sense to check `customizes()` at all, because the methods we're looking for may not even be relevant.

If we're being strict, addons that customize `treeFor` are black boxes to us. More typically, people don't do anything very wacky in `treeFor` and usually just use it as a filter to turn off the other hooks under certain conditions by deciding whether or not to invoke `_super`.

stefanpenner:
> 1) Does it ultimately extend from ember-cli's addon.js?

No

> 2) Does it have a _treeFor method?

yes, one that can be removed (they should be using `isEnabled`):

```
  treeFor: function() {
    // Only include assets in app if top level instance
    if (!this.parent.parent) {
      return this._super.treeFor.apply(this, arguments);
    } else {
      return null;
    }
  },
```

> 3)

PR #186, opened 2019-05-20
Handle implicit hbs files and fix regex bug

jenweber:


jenweber:
Paired & reviewed with Ed

PR #197, opened 2019-05-26
chore: Remove extra files

dcyriller:
This tiny PR removes redundant files. It introduces two changes:
- removes an extraneous `.editorconfig` file
- removes some `.npmignore` files (a whitelist of files to publish seems easier to maintain than a blacklist of files not to publish)

I've not edited the `router` package's files as it's a classic ember addon.


ef4:
Thanks, looks good to me.

Issue #194, opened 2019-05-24
Need a componentRule for expressing argument passthrough


ef4:
We need a componentRule similar to `acceptsComponentArguments` that instead describes which input arguments are mapped directly to yielded output values, unchanged.

Motivating example:

```hbs
{{#from-elsewhere "modal" as |modalComponent|}}
  {{! assume we already used a yieldsSafeComponents rule 
      for from-elsewhere, so we know modalComponent is safe. 
   }}
  {{#liquid-bind modalComponent as |currentModalComponent|}}
    {{! but we need to know that currentModalComponent is safe.
         It is, if we know that liquid-bind passes its argument through.
    }}
    {{component currentModalComponent}}
  {{/liquid-bind}}
{{/from-elsewhere}}
```

void-mAlex:
closing as yieldsArguments was added

PR #208, opened 2019-06-04
Added "yieldsArguments" to packageRules

ef4:
This lets you declare that a component yields one or more of its arguments. You may need to do that in order for embroider to trace the origin of a dynamic component invocation.


ef4:
Closes #194.

ef4:
(This was merged as part of #209, which included a test timeout change that turned CI green.)

PR #210, opened 2019-06-04
Use the copied options in v1-addons

2hu12:


ef4:
Thanks @2hu12, you are correct and we weren't using the copied options.

But we were relying on the buggy behavior because we do need to also alter `this.addonInstance.options`. I figured out what was going on and added that as well, and now CI is passing.

Issue #214, opened 2019-06-11
@babel/plugin-proposal-class-properties issue


stefanpenner:
An app that functions without embroider enabled, currently fails when embroider is enabled when `classProperties` are used.  

```js
Build Error (PackagerRunner) in app.js:7

/private/var/folders/4r/whc65vwj1xggvvky3yy1cp9m000mw4/T/embroider/2e1ac4/app.js: Support for the experimental syntax 'classProperties' isn't currently enabled (7:16):

   5 |
   6 | class App extends Application {
>  7 |   modulePrefix = config.modulePrefix;
     |                ^
   8 |   podModulePrefix = config.podModulePrefix;
   9 |   Resolver = Resolver;
  10 | }

Add @babel/plugin-proposal-class-properties (https://git.io/vb4SL) to the 'plugins' section of your Babel config to enable transformation.


Stack Trace and Error Report: /var/folders/4r/whc65vwj1xggvvky3yy1cp9m000mw4/T/error.dump.33e3fb7de16d52c299f3b39be89e429c.log
```

```js
Add @babel/plugin-proposal-class-properties (https://git.io/vb4SL) to the 'plugins' section of your Babel config to enable transformation.
        at PackagerRunner (@embroider/webpack)
  - errorType: Build Error
  - location:
    - column: [undefined]
    - file: app.js
    - line: 7
    - treeDir: [undefined]
  - message: app.js:7: /private/var/folders/4r/whc65vwj1xggvvky3yy1cp9m000mw4/T/embroider/2e1ac4/app.js: Support for the experimental syntax 'classProperties' isn't currently enabled (7:16):

   5 | 
   6 | class App extends Application {
>  7 |   modulePrefix = config.modulePrefix;
     |                ^
   8 |   podModulePrefix = config.podModulePrefix;
   9 |   Resolver = Resolver;
  10 | }
```
Interesting files:
* [Error Log](https://gist.github.com/stefanpenner/e6e8e0f71c2f03095178f6d8eb29f3fa)
* [_babel_config_.js](https://gist.github.com/stefanpenner/ec59bb84d455d24c1580875af88c2161)

ef4:
How is the babel plugging getting added? Directly by the app, or an addon mutating the app's babel config?

stefanpenner:
@ef4 investigating now.

stefanpenner:
[...]

PR #215, opened 2019-06-11
[DONT MERGE] Mega hax

stefanpenner:
This is mega haxor working around: https://github.com/embroider-build/embroider/issues/214

### DIFF
--- a/packages/compat/src/v1-app.ts
+++ b/packages/compat/src/v1-app.ts
@@ -215,6 +215,11 @@ export default class V1App implements V1Package {
       plugins.push([ModulesAPIPolyfill, { blacklist }]);
     }
 
+    // TODO: mega haxor
+    if (typeof babelInstance._addDecoratorPlugins === 'function') {
+      babelInstance._addDecoratorPlugins(plugins);
+    }
+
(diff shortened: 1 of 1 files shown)


packages/compat/src/v1-app.ts
```diff
@@ -215,6 +215,11 @@ export default class V1App implements V1Package {
       plugins.push([ModulesAPIPolyfill, { blacklist }]);
     }
 
+    // TODO: mega haxor
```
stefanpenner: as described above:

```js
   // this is reproducing what ember-cli-babel does. It would be nicer to just
    // call it, but it require()s all the plugins up front, so not serializable.	    // call it, but it require()s all the plugins up front, so not serializable.
    // In its case, it's mostly doing it to set basedir so that broccoli caching	    // In its case, it's mostly doing it to set basedir so that broccoli caching
    // will be happy, but that's irrelevant to us here.	    // will be happy, but that's irrelevant to us here.
```

The solution is to see if we can get ember-cli-babel to provide us a serializable variant of its plugins, that we nicely consume here. That would avoid the need for us to do this.

PR #218, opened 2019-06-12
[Fixes #214] User ember-cli-babel’s public API to discover config (pl…

### LINKED ISSUE #214: @babel/plugin-proposal-class-properties issue

stefanpenner:
…ugins/preset)

- [x] verified in offending app
- [x] confirm with @ef4 
- [x] make tests green

### DIFF
--- a/packages/compat/src/v1-app.ts
+++ b/packages/compat/src/v1-app.ts
@@ -7,7 +7,6 @@ import { WatchedDir } from 'broccoli-source';
 import resolve from 'resolve';
 import V1Package from './v1-package';
 import { Tree } from 'broccoli-plugin';
-import get from 'lodash/get';
 import { V1Config, WriteV1Config } from './v1-config';
 import { PackageCache, TemplateCompiler, TemplateCompilerPlugins, AddonMeta, Package } from '@embroider/core';
 import { writeJSONSync, ensureDirSync, copySync, readdirSync, pathExistsSync } from 'fs-extra';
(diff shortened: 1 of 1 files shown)


packages/compat/src/v1-app.ts
```diff
@@ -184,55 +183,35 @@ export default class V1App implements V1Package {
...
+    });
 
-    // this is reproducing what ember-cli-babel does. It would be nicer to just
-    // call it, but it require()s all the plugins up front, so not serializable.
-    // In its case, it's mostly doing it to set basedir so that broccoli caching
-    // will be happy, but that's irrelevant to us here.
+    let plugins = babelConfig.plugins as any[];
+    let presets = babelConfig.presets;
+
+    // install embroider macros
```
ef4: This comment is wrong. `debugMacrosPlugin` is a stock ember-cli, it's not related to embroider macros.

packages/compat/src/v1-app.ts
```diff
@@ -184,55 +183,35 @@ export default class V1App implements V1Package {
...
-        // even if the app was using @embroider/macros, we drop it from the config
-        // here in favor of our globally-configured one.
-        return (
-          !isEmbroiderMacrosPlugin(p) &&
-          // similarly, if the app was already using
-          // ember-cli-htmlbars-inline-precompile, we remove it here because we
-          // have our own always-installed version of that (v2 addons are
-          // allowed to assume it will be present in the final app build, the
-          // app doesn't get to turn that off or configure it.).
-          !TemplateCompiler.isInlinePrecompilePlugin(p)
```
ef4: We can't just drop this filtering. It's incorrect for either `@embroider/macros` or `hbs` to run in stage1. 

If `@embroider/macros` run here it will do so with the wrong configuration.

If `hbs` runs here, we will lose visibility into component and template dependencies that come from within those `hbs` templates.

stefanpenner:
> Is the idea here that we just give up on babel config serializability?

The config should be serializable: 

* https://github.com/babel/ember-cli-babel/blob/95d7c73fb0ff0b1aa029411e315c93efa5989a35/index.js#L358
* https://github.com/babel/ember-cli-babel/blob/95d7c73fb0ff0b1aa029411e315c93efa5989a35/index.js#L358
* etc

packages/compat/src/v1-app.ts
stefanpenner: oops

packages/compat/src/v1-app.ts
stefanpenner: Oops, removing this was actually by mistake. Good catch. I guess i screwed up a rebase or something.

ef4:
Ah, thanks. I went looking at this has indeed already been fixed for quite a while. I think the original commit was https://github.com/babel/ember-cli-babel/commit/a2c478474401e04b0702d61feba7d43f2c94ecd2

So that's good then, I'm glad to use the real API.

stefanpenner:
@ef4 ya, ensuring these are serializable in ember-cli-babel is important, so it has been fixed. Likely just after you made your adaptation here.

@ef4 do you mind if I merge? If you do, let me know what I can work to address said concerns.

packages/compat/src/v1-app.ts
```diff
@@ -184,55 +183,48 @@ export default class V1App implements V1Package {
...
-        );
-      });
-    }
+    const babelAddon = (this.app.project as any).findAddonByName('ember-cli-babel');
+    const babelConfig = babelAddon.buildBabelOptions({
+      'ember-cli-babel': {
+        includeExternalHelpers: true,
+        compileModules: false,
+        disableDebugTooling: true,
```
stefanpenner: I think we needed this to be false not true...

PR #222, opened 2019-06-14
Treat router as not-settled while loading lazy routes

ef4:
@rwjblue can I please get your opinion on whether this is a good idea?

The tl;dr is: should I `registerWaiter` while loading the code for lazy routes? Is there a precedent for how lazy engines do this?

### DIFF
--- a/packages/router/addon/index.js
+++ b/packages/router/addon/index.js
@@ -2,6 +2,8 @@
   This code is adapted from ember-engines/addon/-private/router-ext.js.
 */
 import EmberRouter from '@ember/routing/router';
+import { registerWaiter } from '@ember/test';
+
 let newSetup = true;
 
 function lazyBundle(routeName) {
@@ -12,6 +14,12 @@ function lazyBundle(routeName) {
 }
 
 export default EmberRouter.extend({
+  init(...args) {
+    this._super(...args);
+    this._inFlightLazyRoutes = 0;
+    registerWaiter(this._doneLoadingLazyRoutes.bind(this));
+  },
+
   // This is necessary in order to prevent the premature loading of lazy routes
   // when we are merely trying to render a link-to that points at them.
   // Unfortunately the stock query parameter behavior pulls on routes just to
@@ -55,10 +63,22 @@ export default EmberRouter.extend({
       if (!bundle || bundle.loaded) {
         return original(name);
       }
-      return bundle.load().then(() => {
-        bundle.loaded = true;
-        return original(name);
-      });
+      this._inFlightLazyRoutes++;
+      return bundle.load().then(
+        () => {
+          this._inFlightLazyRoutes--;
+          bundle.loaded = true;
+          return original(name);
+        },
+        err => {
+          this._inFlightLazyRoutes--;
+          throw err;
+        }
+      );
     };
+  },
+
+  _doneLoadingLazyRoutes() {
+    return this._inFlightLazyRoutes < 1;
   },
 });


ef4:
Thanks @villander for pointing out the [current story for lazy engines](https://github.com/ember-engines/ember-engines/tree/fix-parallel#testing-for-standalone-engines), which is that you're kinda on your own to preload them.

villander:
as well as ember-engines, I think embroider will need a `setupEmbroiderTest ()` for lazy loading tests. Or move this to ember-test-helpers.

villander:
@ef4 some test to make sure this works?

ef4:
> I think embroider will need a setupEmbroiderTest() for lazy loading tests. 

I disagree.

Ember's router already deals in asynchrony. The exact same time that asynchronous model hooks are running, asynchronous code can be loading. The test system deals with the asynchronous models already, dealing with the code is no different and shouldn't require anything more than we already have.

villander:
> Ember's router already deals in asynchrony. The exact same time that asynchronous model hooks are running, asynchronous code can be loading. The test system deals with the asynchronous models already, dealing with the code is no different and shouldn't require anything more than we already have.

Yeah, sorry I thought better and erased my comment, we do not need it here. We just need to notify ember-test-helpers when this is done. Your solution looks good, but we need a use case to test.

ef4:
> some test to make sure this works?

Yeah, this is pretty annoying to put a reliable automated test around. We're proving a negative (that a race condition doesn't exist), and to do that we'd need to hack into the underlying webpack code loader. Which we otherwise deliberately treat as a black box, for very good reasons (it can be swapped out with any other bundles that knows how to compile `import()`).

I verified this implementation by tracing the settled state of a whole app and saw that this definitely covered a gap where no other unsettled things were present. That plus a second or third set of eyes on the code is good enough for me. I'm willing to trust the semantics of `registerWaiter` without testing them end-to-end.

ef4:
(I'm still interested in feedback on the API in this PR, but I'm going to merge it tentatively. Our API is not stable yet, so we can still change our minds.)

packages/router/addon/index.js
```diff
@@ -12,6 +14,12 @@ function lazyBundle(routeName) {
 }
 
 export default EmberRouter.extend({
+  init(...args) {
+    this._super(...args);
+    this._inFlightLazyRoutes = 0;
+    registerWaiter(this._doneLoadingLazyRoutes.bind(this));
```
rwjblue: At the least you need to guard this with an `if (DEBUG)` flag, otherwise prod builds will be borked.

Also, registering in `init` and not unregistering in `willDestroy` will leave tests in a pretty weird /bad state.

packages/router/addon/index.js
```diff
@@ -55,10 +63,22 @@ export default EmberRouter.extend({
       if (!bundle || bundle.loaded) {
         return original(name);
       }
-      return bundle.load().then(() => {
-        bundle.loaded = true;
-        return original(name);
-      });
+      this._inFlightLazyRoutes++;
+      return bundle.load().then(
```
rwjblue: IMHO, we really want to leverage `ember-test-waiters` (what I've been working on to replace the `registerWaiter` system) which has a `waitForPromise` API that would fit this need precisely (and provide actual feedback in the console when delayed). At the moment that requires ember-test-helpers@1.6.0 (just released) and I haven't polyfilled it just yet.

```js
import { waitForPromise } from 'ember-test-waiters';

// ...snip...
return waitForPromise(
  bundle.load().then(() => {
    bundle.loaded = true;

    return original(name);
  })
);
```

[ember-test-waiters](https://github.com/rwjblue/ember-test-waiters)

packages/router/addon/index.js
rwjblue: See where we setup `Ember.Test.registerWaiter` here: 

https://github.com/emberjs/ember.js/blob/v3.10.1/packages/ember/index.js#L641-L650

Note, there is no `Ember.Test` in production builds.

PR #224, opened 2019-06-15
Added support to load config module for FastBoot environment

dnalagatla:
Added support to load config module for FastBoot environment. 

As Embroider is not supporting `config-module` type in `contentFor` hook. Added this check to support fastboot config module.

@stefanpenner @ef4


ef4:
We do plan to support `config-module` though, see the list in the spec here: https://github.com/embroider-build/embroider/blob/master/SPEC.md#contentfor

Perhaps instead of adding this fastboot-specific code to embroider we can implement `config-module`? That would fix this case and also fix any other addons that are trying to use `config-module`.

Embroider already configures pattern replacement [here](https://github.com/embroider-build/embroider/blob/21d0082da5603c687359492a7aaa43fa5dac60b9/packages/compat/src/v1-app.ts#L170-L181). If we replace the hard-coded `metaLoader()` code with `{{content-for "config-module"}}` and then ensure we run the output through the pattern replacer, that should do the trick.

dnalagatla:
> Perhaps instead of adding this fastboot-specific code to embroider we can implement `config-module`?

I agree. Let me see what is required here for supporting `config-module`. I will update the PR. Thank you @ef4

dnalagatla:
> Perhaps instead of adding this fastboot-specific code to embroider we can implement `config-module`?

I agree. Let me see what is required here for supporting `config-module`. I will update the PR. Thank you @ef4

dnalagatla:
@ef4  I added the support for `config-module` type in content-for hook. I noticed the ember-cli implementation expects the content injected is of AMD format and not ES6 JS module format. As a result, I updated `metaLoader()` method to return the following template
`
define('${appName}/config/environment', [], function() {
        {{content-for 'config-module'}}
    });
`
Adding the above changes didn't break the build flow and supported fastboot specific config update without any issues.

cc: @stefanpenner

packages/compat/src/v1-config.ts
```diff
@@ -49,16 +49,12 @@ export class WriteV1Config extends Plugin {
 }
 
 function metaLoader(appName: string) {
+  // Added contentFor config-module support.
+  // ember-cli adds content in AMD module format and
+  // not as ES6 JS modules. Changed template to match ember-cli
```
stefanpenner: Although ember-cli produces AMD here, could we continue to produce ES6 modules? Or is there a technical limitation preventing us from doing so.

In general, the idea of AMD in embroider is only for compatibility. If core features such as `config/environment` cannot be implemented in the modern way, maybe embroider is missing an important primitive.

packages/compat/src/v1-app.ts
```diff
@@ -135,6 +135,16 @@ export default class V1App implements V1Package {
     return this.app.options.storeConfigInMeta;
```
stefanpenner: Could we ensure the new code-path is tested?

packages/compat/src/v1-config.ts
```diff
@@ -49,16 +49,12 @@ export class WriteV1Config extends Plugin {
...
-  try {
-    metaName = '${appName}/config/environment';
-    let rawConfig = document.querySelector('meta[name="' + metaName + '"]').getAttribute('content');
-    config = JSON.parse(unescape(rawConfig));
-  }
-  catch(err) {
-    throw new Error('Could not read config from meta tag with name "' + metaName + '".');
-  }
-  export default config;
+    define('${appName}/config/environment', [], function() {
```
stefanpenner: Although ember-cli produces AMD here, could we continue to produce ES6 modules? Or is there a technical limitation preventing us from doing so.

In general, the idea of AMD in embroider is only for legacy compatibility. If core features such as config/environment cannot be implemented in the modern way, maybe embroider is missing an important primitive.

dnalagatla: I think the technical limitation is related to what ember-cli returns as content for `config-module`. The Content is expected to return an export object as show [here](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/app-config.js) and replaces a placeholder in this [AMD module ](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/app-config.js).

dnalagatla: @stefanpenner as we discussed I updated the Code to use the immediate invocation within the ES6. Also, updated the comments.

packages/compat/src/v1-app.ts
dnalagatla: I added the tests as part of macro-sample-addon to test the code path for config-module.
@stefanpenner @ef4

Issue #146, opened 2019-04-24
Windows 7 - Build fails just after enabling embroider on an existing app


HerveSeger:
Hi.

I have a working application using `ember v3.4.8` and I have followed the instructions in the `readme` to try `embroider`.
So I have updated the file `ember-cli-build.js` and run `ember build`.

It fails with:
```
Build Error (OneShot)

[BABEL] unknown: ._parallelBabel is not a valid Plugin property
        at ImportParser (embroider:core:import-parser)
-~- created here: -~-
    at new Plugin (c:\development\projects\git\insurance-online-ui\node_modules\broccoli-plugin\index.js:7:31)
    at new ImportParser (c:\development\projects\git\insurance-online-ui\node_modules\@embroider\compat\src\import-parser.js:36:9)
    at V1Addon.parseImports (c:\development\projects\git\insurance-online-ui\node_modules\@embroider\compat\src\v1-addon.js:187:16)
    at V1Addon.buildAddonTestSupport (c:\development\projects\git\insurance-online-ui\node_modules\@embroider\compat\src\v1-addon.js:481:42)
    at V1Addon.build (c:\development\projects\git\insurance-online-ui\node_modules\@embroider\compat\src\v1-addon.js:617:14)
    at V1Addon.<anonymous> (c:\development\projects\git\insurance-online-ui\node_modules\typescript-memoize\dist\memoize-decorator.js:67:52)
    at V1Addon.get v2Trees (c:\development\projects\git\insurance-online-ui\node_modules\@embroider\compat\src\v1-addon.js:335:45)
    at V1Addon.<anonymous> (c:\development\projects\git\insurance-online-ui\node_modules\typescript-memoize\dist\memoize-decorator.js:67:52)
    at V1Addon.get v2Tree [as v2Tree] (c:\development\projects\git\insurance-online-ui\node_modules\@embroider\compat\src\v1-addon.js:326:52)
    at oldPackages.map.pkg (c:\development\projects\git\insurance-online-ui\node_modules\@embroider\compat\src\build-compat-addon.js:42:48)
-~- (end) -~-


Stack Trace and Error Report: c:\DEVELO~1\tmp/error.dump.fe5e29e0269d89f4157130e26526ae3f.log
```

The stack trace:
```
ERROR Summary:

  - broccoliBuilderErrorStack: Error: [BABEL] unknown: ._parallelBabel is not a valid Plugin property
[...]

Issue #74, opened 2019-03-19
Windows error - node_modules\@ember-decorators contains invalid WIN32 path characters.


lifeart:
C:\Users\lifeart\AppData\Local\Temp\embroider-cbSrw5\C:\Users\lifeart\Documents\repos\dreamcatcher-web-app\node_modules\@ember-decorators contains invalid WIN32 path characters.

```
ERROR Summary:

  - broccoliBuilderErrorStack: Error: C:\Users\lifeart\AppData\Local\Temp\embroider-cbSrw5\C:\Users\lifeart\Documents\repos\dreamcatcher-web-app\node_modules\@ember-decorators contains invalid WIN32 path characters.
    at mkdirsSync (C:\Users\lifeart\Documents\repos\dreamcatcher-web-app\node_modules\@embroider\compat\node_modules\fs-extra\lib\mkdirs\mkdirs-sync.js:18:22)
    at Object.copySync (C:\Users\lifeart\Documents\repos\dreamcatcher-web-app\node_modules\@embroider\compat\node_modules\fs-extra\lib\copy-sync\copy-sync.js:30:35)
    at forEach (C:\Users\lifeart\Documents\repos\dreamcatcher-web-app\node_modules\@embroider\compat\src\compat-addons.js:59:24)
    at Array.forEach (<anonymous>)
    at CompatAddons.build (C:\Users\lifeart\Documents\repos\dreamcatcher-web-app\node_modules\@embroider\compat\src\compat-addons.js:58:47)
    at WaitForTrees.build (C:\Users\lifeart\Documents\repos\dreamcatcher-web-app\node_modules\@embroider\core\src\wait-for-trees.js:49:21)
    at resolve (C:\Users\lifeart\Documents\repos\dreamcatcher-web-app\node_modules\broccoli\lib\wrappers\transform-node.js:29:35)
    at new Promise (<anonymous>)
    at TransformNodeWrapper.build (C:\Users\lifeart\Documents\repos\dreamcatcher-web-app\node_modules\broccoli\lib\wrappers\transform-node.js:21:12)
    at Promise.resolve.then.then.then (C:\Users\lifeart\Documents\repos\dreamcatcher-web-app\node_modules\broccoli\lib\builder.js:111:28)
  - code: [undefined]
[...]

PR #228, opened 2019-06-17
fix windows file paths issues

lifeart:
fix for https://github.com/embroider-build/embroider/issues/146
https://github.com/embroider-build/embroider/issues/74

### DIFF
--- a/packages/compat/src/moved-package-cache.ts
+++ b/packages/compat/src/moved-package-cache.ts
@@ -1,4 +1,4 @@
-import { join, dirname, resolve } from 'path';
+import { join, dirname, resolve, sep, isAbsolute } from 'path';
 import { ensureSymlinkSync, readdirSync, readlinkSync, realpathSync, lstatSync } from 'fs-extra';
 import { Memoize } from 'typescript-memoize';
 import { PackageCache, Package, getOrCreate } from '@embroider/core';
@@ -171,8 +171,8 @@ function symlinksInNodeModules(path: string): { source: string; target: string }
 }
 
 function pathSegments(filename: string) {
-  let segments = filename.split('/');
-  if (segments[0] === '/') {
+  let segments = filename.split(sep);
+  if (isAbsolute(filename)) {
     segments.shift();
   }
   return segments;
--- a/packages/core/src/app.ts
+++ b/packages/core/src/app.ts
@@ -25,6 +25,7 @@ import TemplateCompiler from './template-compiler';
 import { Resolver } from './resolver';
 import { Options as AdjustImportsOptions } from './babel-plugin-adjust-imports';
 import { tmpdir } from 'os';
+import { sep } from 'path';
 import { explicitRelative } from './paths';
 
 export type EmberENV = unknown;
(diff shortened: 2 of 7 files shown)


ef4:
Thanks for working on this.

Is `.charAt(1) === ':'` safe? It seems like it would also match legal unix paths. 

Can we use [path.isAbsolute](https://nodejs.org/dist/latest-v10.x/docs/api/path.html#path_path_isabsolute_path) instead? I think that would replace the need for both kinds of checks: `path[0] === '/'` and `path.chatAt[1] === ':'`.

lifeart:
> Thanks for working on this.
> 
> Is `.charAt(1) === ':'` safe? It seems like it would also match legal unix paths.
> 
> Can we use [path.isAbsolute](https://nodejs.org/dist/latest-v10.x/docs/api/path.html#path_path_isabsolute_path) instead? I think that would replace the need for both kinds of checks: `path[0] === '/'` and `path.chatAt[1] === ':'`.

I think it's really rare case for second symbol = `:` in unix. I don't like this slicing approach at all, it can be an better way, but, embroider not working in windows at all (after this pr I have ussies): 

image

incorrect `define` paths for some deps, full windows file path in `moduleName` property of compiled template.


@ef4 thank for feedback! I'm not against proposed updates, but, I think we can delay it until we get embroider happy path for windows.

ef4:
A critical thing to sort out is that some paths are *filesystem* paths and therefore need OS-dependent handling, but other paths are *urls*, which should not get OS-dependent handling.

So it's probably not going to work if you just look for all the places that are using `/` explicitly and change them to `path.sep`. Some of those are URLs that are really supposed to always use `/`.

ef4:
Also I guess not just URLs, but also "runtime paths" as seen by Ember and loader.js, which always use `/` as the separator regardless of operating system.

lifeart:
Right, regarding fixes, 
`packages/compat/src/moved-package-cache.ts` - don't side `\\` in requrejs
`packages/core/src/app.ts` - don't side `\\` in requrejs
`packages/webpack/src/ember-webpack.ts` - don't side `\\` in requrejs

only posibble `siding` change in `packages/core/src/package-cache.ts`, but, it's critical change to get `embroider` working on windows.

looks like paths from `packages/core/src/package-cache.ts`, `ownerOfFile` - used in packaging and produce this side effects

lifeart:
@ef4 using this branch I able to build & serve ember app (https://github.com/krausest/js-framework-benchmark/tree/master/frameworks/keyed/ember) and it's working.

Issues: 

webpack failing in case of resolving absolute style paths, like 
`<link href="/css/currentStyle.css" rel="stylesheet"/>` - if there is no style at buld time, it's not reson to fail whole build, styles may be added later.

Template's `moduleName` full system paths leaking into production builds

image


Build failing with 
```
   staticHelpers: true,
   staticComponents: true
```
due to resolution erros

lifeart:
If current changes looks well, can we get it merged?

ef4:
Thanks, this looks great.

ef4:
> if there is no style at buld time, it's not reson to fail whole build, styles may be added later. (we can warn about it?)

Yes, we already warn about missing `<script>`, we can make styles be a warning too.

We already support putting a `data-embroider-ignore` attribute on `<script>` if you want embroider to ignore a script tag. We should make sure the same works for `<link rel="stylesheet">`.

ef4:
I think each of these deserves its own followup issue:

 - stylesheet link that's not available should be a warning
 - full system paths in moduleName
 - `can't resolve ..componentsmy-table.js`. This one is probably a windows path problem, it was trying to say `../components/my-table.js`

PR #223, opened 2019-06-14
teach embroider to also adjust AMD imports

stefanpenner:
As it turns out, webpack can ingest both AMD and es6 modules, this teaches embroider to also prepare AMD modules it discovers the same way it prepares es6 modules.

Currently embroider adjust modules in the following way:

* ensure renamed packages are correctly renamed by consumers
* ensure when appropriate resolution fallbacks are removed
* ensure renamed modules are correctly renamed by consumers
* self-imports should become relative

- [x] detect `define` statements
- [x] tests for detecting define statements
- [x]  harness to adjust define statements
- [x] actually adjust AMD define statements
- [x] flesh out tests
- [x] test in real app
- [x] get feedback

PR #234, opened 2019-06-18
Remove bad `only`

ef4:
An `only` snuck in from PR #223.

Once removed, one of the new tests added in that PR wasn't passing. I refactored it and cleaned up the usage of `@babel/types` so typescript can see what's going on.

This way of using `@babel/types` isn't obvious and I have no idea how one would discover it from documentation. I picked it up by reading what other typescript babel plugins were doing.

### DIFF
--- a/packages/core/src/babel-plugin-adjust-imports.ts
+++ b/packages/core/src/babel-plugin-adjust-imports.ts
@@ -13,6 +13,15 @@ import {
   Program,
   returnStatement,
   stringLiteral,
+  isStringLiteral,
+  isArrayExpression,
+  isFunction,
+  CallExpression,
+  StringLiteral,
+  ArrayExpression,
+  ExportNamedDeclaration,
+  ImportDeclaration,
+  ExportAllDeclaration,
 } from '@babel/types';
 import PackageCache from './package-cache';
 import Package, { V2Package } from './package';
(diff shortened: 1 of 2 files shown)


packages/core/tests/babel-plugin-adjust-imports-test.ts
```diff
@@ -48,7 +47,7 @@ test('isDefineExpression works', function(assert) {
   assert.equal(isDefineExpressionFromSource(`define('apple'); import foo from 'foo'`), false);
 });
 
-only('main', function(assert) {
+test('main', function(assert) {
```
stefanpenner: OOOPS sorry

packages/core/src/babel-plugin-adjust-imports.ts
```diff
@@ -255,7 +270,7 @@ function makeHBSExplicit(specifier: string, sourceFile: AdjustFile) {
   return specifier;
 }
 
-export default function main({ types: t }: { types: any }) {
```
stefanpenner: I thought the convention was to use the `t` provided during usage of the plugin rather then via what is imported

stefanpenner: Ultimately, I'm fine with it either way.

ef4: Yeah, that is the traditional way I've seen it done. But typescript seems happier this way. It gives really good feedback on which properties are legal on which nodes and paths.

Like I said, this is mostly based off me reading other typescript plugins to see how they do things.

PR #238, opened 2019-06-19
Fix #233 – Teach embroider about app.import + outputFile

### LINKED ISSUE #233: app.import(x, { outputFile: y }) is not yet supported
Specific motivating use-case: https://github.com/pzuraq/ember-cli-babel-polyfills/blob/2f0c79586d7f08c646635fe8214f3a1185e988ff/index.js#L43-L53

and https://github.com/pzuraq/ember-cli-babel-polyfills/blob/2f0c79586d7f08c646635fe8214f3a1185e988ff/index.js#L125

---

Relevant embroider stuff (from our notes)

* https://github.com/embroider-build/embroider/blob/master/packages/compat/src/v1-app.ts#L323
* https://github.com/embroider-build/embroider/blob/84784b2cdeec464bd995a750c153f7765581890e/packages/core/src/metadata.ts#L38-L40 
* https://github.com/embroider-build/embroider/blob/master/packages/core/src/app.ts#L562

stefanpenner:
[Fix #233] 

Today embroider does not understand `app.import`'s `outputFile` option. 

- [x] include motivating add-on into our test suite
- [x] failing tests
- [x] re-work based on @ef4's suggestion
- [x] fix
- [x] figure out why ember-cli-babel-polyfills doesn't appear to be running it AMD obfuscation code..
- [x] ensure more scenarios are tested
- [ ] get feedback cc @ef4 
- [x] figure out why testem.js isn't accessible from the browser...

### DIFF
--- a/packages/compat/src/v1-app.ts
+++ b/packages/compat/src/v1-app.ts
@@ -16,12 +16,40 @@ import { TransformOptions } from '@babel/core';
 import { isEmbroiderMacrosPlugin } from '@embroider/macros';
 import resolvePackagePath from 'resolve-package-path';
 import ExtendedPackage from './extended-package';
+import Concat from 'broccoli-concat';
 
 // This controls and types the interface between our new world and the classic
 // v1 app instance.
 
+type filePath = string;
+type OutputFileToInputFileMap = { [filePath: string]: filePath[] };
+
+interface EmberApp {
+  env: string;
+  name: string;
+  _scriptOutputFiles: OutputFileToInputFileMap;
+  _styleOutputFiles: OutputFileToInputFileMap;
+  legacyTestFilesToAppend: filePath[];
+  vendorTestStaticStyles: filePath[];
+  _customTransformsMap: Map<string, any>;
+  _nodeModules: Map<string, { name: string; path: filePath }>;
+  options: any;
+  tests: boolean;
+  trees: any;
+  project: any;
+  registry: any;
+  testIndex(): Tree;
+  getLintTests(): Tree;
+}
+
+interface Group {
+  outputFiles: OutputFileToInputFileMap;
+  implicitKey: '_implicitStyles' | '_implicitScripts';
+  vendorOutputPath: 'string';
+}
+
 export default class V1App implements V1Package {
-  static create(app: any, packageCache: PackageCache): V1App {
+  static create(app: EmberApp, packageCache: PackageCache): V1App {
     if (app.project.pkg.keywords && app.project.pkg.keywords.includes('ember-addon')) {
       // we are a dummy app, which is unfortunately weird and special
       return new V1DummyApp(app, packageCache);
@@ -30,7 +58,11 @@ export default class V1App implements V1Package {
     }
   }
 
-  protected constructor(protected app: any, protected packageCache: PackageCache) {
+  private _publicAssets: { [filePath: string]: string } = Object.create(null);
+  private _implicitScripts: string[] = [];
+  private _implicitStyles: string[] = [];
+
+  protected constructor(protected app: EmberApp, protected packageCache: PackageCache) {
     this.extendPackage();
   }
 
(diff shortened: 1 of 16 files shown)


stefanpenner:
@ef4 interesting, these tests suggest we are also unable to access `testem.js` at runtime. I'll investigate that as well.

stefanpenner:
@ef4 I think we may need to adjust the `public-files` format.

I have a "working solution" but the implementation is kinda off due to public-assets being a map of inputFile -> outputFile but the input data structure being `outputFile -> inputFile[]`. Where the ordering of inputFile[] is very important.

The current implementation I through together using `public-assets` likely has the following issues:

* loose ordering
* may not concat when multiple inputFile’s map to one outputFile

ef4:
That is not what `public-assets` is for. It's for designating files *that already exist in the package* that need to have URLs in the final output.

Making those files (by concatenation, respecting order) can be an internal concern of `@embroider/synthesized-vendor`.

Today, synthesized-vendor uses the `implicit-scripts` and `implicit-styles` features of the v2 format to handle the case of `app.import()` adding content to `vendor.js` and `vendor.css`. We're talking about adding the additional case where it also uses the `public-assets` feature to emit standalone files based off usage of `app.import(..., { outputFile })`.

stefanpenner:
@ef4 oh I see, so synthesized-vendor should have already flattened the artifacts, and this step `public-assets` should merely update `public-files` to map according to the already concatenated artifacts?

ef4:
Yes, that's right. 

You should be able to use typical broccoli-based concatenation.

stefanpenner:
@ef4 perfect, I'll work on that next.

packages/compat/src/v1-app.ts
```diff
@@ -290,11 +322,71 @@ export default class V1App implements V1Package {
...
+
+    // support: app.import / outputFile / using
+    for (let entry of outputGroups) {
+      const { outputFiles, implicitKey, vendorOutputPath } = entry;
+      for (let importPath of Object.keys(outputFiles)) {
+        const headerFiles = outputFiles[importPath];
+
+        if (headerFiles.length === 0) {
+          // something went really wrong
+          throw new Error('Embroider: EWUT');
```
stefanpenner: Interesting, looks like we do hit this. Investigating...

```
 FAIL  test-packages/static-app-rooturl.test.js (5.506s)
  ● static-app-classic
    Command failed: ember build
    Embroider: EWUT
```

stefanpenner: Apparently the `static-app` hits this, specifically when `importPath === vendorOutputPath`

stefanpenner:
I think I sorted out the strange error: https://github.com/pzuraq/ember-cli-babel-polyfills/pull/6

test-packages/macro-tests/package.json
```diff
@@ -52,5 +52,10 @@
   },
   "engines": {
     "node": "8.* || >= 10.*"
+  },
+  "dependencies": {
+    "broccoli-stew": "^3.0.0",
+    "core-js": "^3.1.4",
+    "ember-cli-babel-polyfills": "link:/Users/spenner/src/pzuraq/ember-cli-babel-polyfills"
```
stefanpenner: this depends on: https://github.com/pzuraq/ember-cli-babel-polyfills/pull/6

stefanpenner:
babel-pollyfils was my motivating usecase for this feature. But once implemented, I noticed the output files it omitted where invalid. After debugging I realized ember-cli-babel-polyfills itself was having issues with embroiders node_module topology. Specifically, it intended to use the core-js provided by ember-cli-babel but would in practice resolve whichever version was resolvable at the cwd. In our case here, it would simply resolve a very different and incompatible version.

The linked pr locks this down, and configures the appropriate basedir for rollup to use.

stefanpenner:
👯‍♂

PR #242, opened 2019-06-19
blank pod prefix support

lifeart:
* given: 

`podModulePrefix` = '';

`ember g component my-component --pod`

->

```
app/components/my-component/component.js
app/components/my-component/template.hbs

```


ef4:
Thanks, this looks good. 

Can we add a test to cover this? The resolver tests are [here](https://github.com/embroider-build/embroider/blob/master/packages/compat/tests/resolver-test.ts).

ef4:
Nice work.

PR #243, opened 2019-06-19
Fix fs path leaking for compiled templates

lifeart:
https://github.com/embroider-build/embroider/issues/245


lifeart:
@ef4 Do we need template's "moduleName" property in runtime?

In this PR I'm trying to convert "absolute" (fs path) to moduleName (runtime name) and use it for template compilation.
But, looks like it's not working well.

If we don't need "moduleName" for compiled output, I can remove it from compile result...

ef4:
Thanks for tackling this. I think I see what was tripping you up. There were two problems:

 - there is some hidden order dependence: `dependenciesOf` only works for a given module if you call it after `syntax.precompile`.
 - it's good that you're changing the `moduleName` that is passed to `syntax.precompile` to be a runtime name. That will make all other AST transform see what they were traditionally expecting. But our *own* AST transform (`resolver-transform.ts`) really needs the full path. So I added that explicitly as a separate property, and thankfully Glimmer passes it through.

PR #246, opened 2019-06-20
Conditionally use "one-shot" plugin

thoov:
If broccoli memoization is enabled then we do not need to use the one shot plugin. I tested this in a new default app (upgrading ember-cli to 3.11.0-beta.3) and conditionally setting: `BROCCOLI_ENABLED_MEMOIZE=true` & `EMBER_CLI_BROCCOLI_WATCHER=true`. 

cc @stefanpenner


stefanpenner:
LGTM. @ef4 thoughts?

ef4:
This looks good to me, but I think there's another change needed before this will really enable rebuilds of addons.

An embroider [Stage](https://github.com/embroider-build/embroider/blob/25671a8b43d8938d4c66d3de2fc92acd58563fd6/packages/core/src/stage.ts) is a broccoli plugin but has additional behaviors. 

[CompatAddons](https://github.com/embroider-build/embroider/blob/25671a8b43d8938d4c66d3de2fc92acd58563fd6/packages/compat/src/compat-addons.ts#L77) is the first stage. [This code](https://github.com/embroider-build/embroider/blob/25671a8b43d8938d4c66d3de2fc92acd58563fd6/packages/compat/src/compat-addons.ts#L76-L98) is responsible for taking the broccoli output for each addon (which as of your PR will rebuild correctly) and making sure they're copied into the `workspaceDir`. The `workspaceDir` can't be a regular broccoli tree because it really needs to use symlinks as symlinks, and also because it encompasses all of your entire node_modules, which would be sad to copy forward through many broccoli plugins.

The TODO in that code is what needs to change. Right now the whole stage only does work on the first build. Instead, on rebuilds it should use something like `fs-tree-diff` on only the addons with `mayRebuild` true and reflect any changes into the `workspaceDir`.

There is an example of code that already does a similar thing in the second stage ([App](https://github.com/embroider-build/embroider/blob/25671a8b43d8938d4c66d3de2fc92acd58563fd6/packages/core/src/app.ts)). It uses [AppDiffer](https://github.com/embroider-build/embroider/blob/25671a8b43d8938d4c66d3de2fc92acd58563fd6/packages/core/src/app-differ.ts) which uses [MultiTreeDiff](https://github.com/embroider-build/embroider/blob/25671a8b43d8938d4c66d3de2fc92acd58563fd6/packages/core/src/multi-tree-diff.ts) which uses `fs-tree-diff`. That use case is more complex than what we need in `CompatAddons`, because in our case there is no merging going on -- each addon has its own place in the `workspaceDir`. So a more typical direct use of `fs-tree-diff` would be fine here.

ef4:
Also, `synthesized-vendor` and `synthesized-styles` probably need to get treated as if they have `mayRebuild` true whenever at least one addon has `mayRebuild` true. Otherwise changes in the addon's vendor and appStyles won't be reflected.

thoov:
Created an issue to track the points that @ef4 brought up: https://github.com/embroider-build/embroider/issues/253

ef4:
Thanks, merging this since we have a separate open issue for the remainder.

PR #247, opened 2019-06-20
Ignore built-in components: Input, LinkTo, Textarea

cyk:
Ember 3.10 introduced a few ["built-in components"](https://blog.emberjs.com/2019/05/21/ember-3-10-released.html), this ignores them in `compat` resolver.

This avoids having to use workaround with `packageRules`:

```js
packageRules: [{
  package: 'my-app',
  components: {
    // TODO: remove when embroider ignores angle bracket built-ins
    '<LinkTo />': { safeToIgnore: true }
  }
}]
```

### DIFF
--- a/packages/compat/src/resolver.ts
+++ b/packages/compat/src/resolver.ts
@@ -75,6 +75,8 @@ const builtInHelpers = [
   'yield',
 ];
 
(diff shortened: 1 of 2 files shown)


ef4:
Thanks!

Issue #250, opened 2019-06-21
Incorrect requrejs definition for some modules (windows)


lifeart:
Have no idea there it's happening...

![image](https://user-images.githubusercontent.com/1360552/59934390-157a2180-9454-11e9-8380-ee16e5606eb0.png)

```
ember-temp/app
ember-temp/config/environment
ember-temp/initializers/app-version
ember-temp/initializers/container-debug-adapter
ember-temp/initializers/export-application-global
ember-temp/lint/app.lint-test
ember-temp/lint/templates.template.lint-test
ember-temp/lint/tests.lint-test
ember-temp/resolver
ember-temp/router
ember-temp/utils/benchmark-helpers
ember-temp/templates/index
ember-cli-app-version\initializer-factory
ember-cli-app-version\utils\regexp
ember-load-initializers\index
ember-resolver\features
ember-resolver\index
ember-resolver\resolver
ember-resolver\resolvers\classic\container-debug-adapter
ember-resolver\resolvers\classic\index
ember-resolver\utils\class-factory
ember-resolver\utils\make-dictionary
```

ef4:
I think the ones with `\` separators are all created by this code:

https://github.com/embroider-build/embroider/blob/c1adf0243b432741b474787410ec565d06533ed1/packages/compat/src/v1-addon.ts#L463-L469

If that needs to change, so will this code which does the same thing for test-support code:

https://github.com/embroider-build/embroider/blob/c1adf0243b432741b474787410ec565d06533ed1/packages/compat/src/v1-addon.ts#L525-L531

The actual metadata that's being written here should stay standardized on `/` separators. That is the format all v2 packages will use when publishing to NPM.

lifeart:
@ef4 I suppose `walkSync` already return normalized `/` paths (even on windows) @stefanpenner ?

PR #251, opened 2019-06-21
bugfix: requrejs definition fix for windows #250

lifeart:
https://github.com/embroider-build/embroider/issues/250


lifeart:
Amazing! Looks like now it's stable on windows!
Going to try `embroider` with typescript / addon

Issue #237, opened 2019-06-19
Build fails if rootUrl does not start with `/`


kandhavivekraj:
If the rootUrl does not have a leading `/`, build fails with the error below

```
Build Error (WaitForTrees)

could not find app javascript in index.html
```

lifeart:
@kandhavivekraj should be fixed by https://github.com/embroider-build/embroider/pull/252

Issue #148, opened 2019-04-26
Build Error: could not find app javascript in index.html


kevinansfield:
In Ghost we have a requirement to publish both development and production html/js/css files in a single package so we have a setup that generates files with different names and custom `{{content-for}}` hooks for inserting the right `<script>` and `<link>` tags.

- [in-repo-addon with custom {{content-for}} hook](https://github.com/TryGhost/Ghost-Admin/blob/master/lib/asset-delivery/index.js#L11-L27)
- [usage in index.html](https://github.com/TryGhost/Ghost-Admin/blob/master/app/index.html#L50)

Following the setup instructions in the readme, for a development build I see the following error:

```
>> Build Error (WaitForTrees)
>> could not find app javascript in index.html
>> Stack Trace and Error Report: /var/folders/30/3l01y1ld18l598kxr5srm7kr0000gn/T/error.dump.4359fa193c0e4e77c12dd665e1725452.log
```

<details>
<summary>Stack Trace and Error Report</summary>

```
=================================================================================

ENV Summary:

  TIME: Fri Apr 26 2019 08:54:16 GMT+0200 (Central European Summer Time)
  TITLE: ember
  ARGV:
  - /Users/kevinansfield/.nvm/versions/node/v10.15.1/bin/node
  - /Users/kevinansfield/code/ghost/Ghost/core/client/node_modules/.bin/ember
  - serve
  - --live-reload-base-url=/ghost/
  - --live-reload-port=4201
  EXEC_PATH: /Users/kevinansfield/.nvm/versions/node/v10.15.1/bin/node
  TMPDIR: /var/folders/30/3l01y1ld18l598kxr5srm7kr0000gn/T
  SHELL: /usr/local/bin/zsh
  PATH:
  - /Users/kevinansfield/.nvm/versions/node/v10.15.1/lib/node_modules/npm/node_modules/npm-lifecycle/node-gyp-bin
  - /Users/kevinansfield/code/ghost/Ghost/core/client/node_modules/.bin
  - /Users/kevinansfield/code/ghost/Ghost/core/client/node_modules/.bin
  - /Users/kevinansfield/code/ghost/Ghost/core/node_modules/.bin
  - /Users/kevinansfield/code/ghost/Ghost/node_modules/.bin
  - /Users/kevinansfield/code/ghost/node_modules/.bin
  - /Users/kevinansfield/code/node_modules/.bin
  - /Users/kevinansfield/node_modules/.bin
[...]

PR #252, opened 2019-06-21
fix for relative root url case #237

lifeart:
https://github.com/embroider-build/embroider/issues/237
https://github.com/embroider-build/embroider/issues/148 ?


stefanpenner:
thanks @lifeart !

PR #254, opened 2019-06-22
fix: show valid src/href paths inside warnings on windows

lifeart:
before:
`Warning: ... <... {src|href}="some\relative\windows\path.{js,css}" />`
after:
`Warning: ... <... {src|href}="some/relative/windows/path.{js,css}" />`


ef4:
Thanks. 

Related to this, I'm hoping to spend some time soon sketching out an API to formalize the difference between URLs and filesystem paths, and between relative and absolute versions of each. Then we could use the type system to make sure we are converting between them all in a way that works on all platforms.

PR #266, opened 2019-07-04
Copy ember-font-awesome font files

balinterdi:
The problem I bumped into is that the fonts ["vendored by"](https://github.com/martndemus/ember-font-awesome/blob/master/index.js#L35-L51) ember-font-awesome weren't copied to the build output when using it with Embroider.

As far as I could tell, the way to fix this is using a compat adapter that specifies those files as public-assets so that they are copied over.

### DIFF
--- a/packages/compat/src/compat-adapters/ember-font-awesome.ts
+++ b/packages/compat/src/compat-adapters/ember-font-awesome.ts
@@ -1,0 +1,16 @@
+import V1Addon from '../v1-addon';
+import { AddonMeta } from '@embroider/core';
+
+export default class EmberFontAwesome extends V1Addon {
+  get packageMeta(): Partial<AddonMeta> {
+    let meta = super.packageMeta || {};
+    meta['public-assets'] = {
+      'node_modules/font-awesome/fonts/FontAwesome.otf': '/fonts/FontAwesome.otf',
+    };
+    for (let extension of ['eot', 'svg', 'ttf', 'woff', 'woff2']) {
+      let fileName = `fontawesome-webfont.${extension}`;
+      meta['public-assets'][`node_modules/font-awesome/fonts/${fileName}`] = `/fonts/${fileName}`;
+    }
+    return meta;
+  }
+}
--- a/packages/compat/src/compat-adapters/ember-font-awesome.ts
+++ b/packages/compat/src/compat-adapters/ember-font-awesome.ts
@@ -1,15 +1,17 @@
 import V1Addon from '../v1-addon';
 import { AddonMeta } from '@embroider/core';
+import walkSync from 'walk-sync';
 
 export default class EmberFontAwesome extends V1Addon {
   get packageMeta(): Partial<AddonMeta> {
-    let meta = super.packageMeta || {};
-    meta['public-assets'] = {
-      'node_modules/font-awesome/fonts/FontAwesome.otf': '/fonts/FontAwesome.otf',
-    };
-    for (let extension of ['eot', 'svg', 'ttf', 'woff', 'woff2']) {
-      let fileName = `fontawesome-webfont.${extension}`;
-      meta['public-assets'][`node_modules/font-awesome/fonts/${fileName}`] = `/fonts/${fileName}`;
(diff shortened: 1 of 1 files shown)


packages/compat/src/compat-adapters/ember-font-awesome.ts
```diff
@@ -0,0 +1,16 @@
+import V1Addon from '../v1-addon';
+import { AddonMeta } from '@embroider/core';
+
+export default class EmberFontAwesome extends V1Addon {
+  get packageMeta(): Partial<AddonMeta> {
+    let meta = super.packageMeta || {};
+    meta['public-assets'] = {
+      'node_modules/font-awesome/fonts/FontAwesome.otf': '/fonts/FontAwesome.otf',
```
lifeart: wondering, should it be `os` specific path separators?

balinterdi: Ah, I think you're right. Also, it feels wrong to list the specific files instead of copying the whole fonts folder but

1) I think you need to list files, you cannot list folders in `public-assets`
2) Not sure how to do that as I can't use `const fs = require('fs'); fs.readdirSync(...)` and other node libraries here, can I?

lifeart: I think `walkSync` from https://www.npmjs.com/package/walk-sync may work for this case

lifeart: I see pefixed `@fontawesome` packages like - https://www.npmjs.com/package/@fortawesome/free-solid-svg-icons
https://github.com/FortAwesome/Font-Awesome/tree/master/js-packages/%40fortawesome
should current compat adapter handle it?

balinterdi: I pushed a commit with the `walkSync` improvement, thank you for the suggestion.

Regarding handling all the files in those packages, my understanding is that compat-adapters are looked up using the package name, so if we wanted to make them compatible with Embroider, too, then an adapter would have to be implemented for each of them.

I might be missing the idea, though.

balinterdi: I'm also not sure about the build error. It says

```
packages/compat/src/compat-adapters/ember-font-awesome.ts:14:7 - error TS2532: Object is possibly 'undefined'.
14       meta['public-assets'][path] = `/fonts/${fileName}`;
         ~~~~~~~~~~~~~~~~~~~~~
```

but there's this block 2 lines above that:

```js
    if (!meta['public-assets']) {
      meta['public-assets'] = {};
    }
```

I'm quite the newb when it comes to TS, though, so it's probably me.

ef4:
This shouldn't need a compat adapter (although that is a fine workaround in your own app until this issue is fixed).

ember-font-awesome uses the `app.import()` api [here](https://github.com/martndemus/ember-font-awesome/blob/c147bc04a4ad4a2f707c7e0ada641cc9414a3837/index.js#L183) with the `destDir` argument. I think we probably don't support `destDir` yet. The fix would be similar to https://github.com/embroider-build/embroider/pull/238, which added support for the `outputFile` argument.

ef4:
Closing in favor of https://github.com/embroider-build/embroider/issues/267

balinterdi:
@ef4 Thanks!

> This shouldn't need a compat adapter (although that is a fine workaround in your own app until this issue is fixed).

You mean just manually `app.import`-ing the files I need, correct?

ef4:
The addon already app.imports them, your app shouldn’t need to do anything.

balinterdi:
Well, but it didn't import the font files when I built the app with Embroider.

That was my reason of looking into how Embroider calls the `treeFor*` hooks (or what other means it uses to import files from add-ons).

ef4:
> Well, but it didn't import the font files when I built the app with Embroider.

Yes, because of #267 

My advice is: use your compatAdapter in your app (it's an option you can pass to `compatBuild()`, it doesn't need to be upstreamed into Embroider) for the present. And once we fix #267, you can remove the `compatAdapter`.

Issue #267, opened 2019-07-04
app.import(x, { destDir }) is not yet supported


ef4:
The `destDir` option to `app.import()` is not supported yet by embroider. Addons like ember-font-awesome rely on it to get files like fonts into the final build.

See https://github.com/embroider-build/embroider/pull/266#issuecomment-508509458

balinterdi:
This wasn't properly closed by your comment ("Fixed #267") on #278 but it should be.

PR #278, opened 2019-07-10
Add support for import with destDir option

balinterdi:


balinterdi:
@ef4 We put this PR together with @benedikt during a pair-programming session and it "just works".

A test still needs to be added and I wanted to ask you whether what we did in this PR is a correct way to do it. I'm aware of one issue: our new method, [`addOtherAssets`](https://github.com/embroider-build/embroider/pull/278/files#diff-fbcbb8d4c0d3117c2fb9db39fb54bf5dR401) only has side effects, it doesn't modify the inputTree it receives in any way, so the way it's being called now is not ideal.

We tried to call `addOtherAssets` from the `emberEntrypoints` method, before going through `classicEntrypoints` ([here](https://github.com/embroider-build/embroider/blob/4af5ab3b92/packages/compat/src/compat-app.ts#L147))because it seemed more apt there, but when we did that, the assets weren't copied.

Awaiting your guidance of what/how to improve. Thanks!

ef4:
Thanks, this is good. I added the test because I wanted to start a new test module for all these app.import cases anyway.

ef4:
Fixed https://github.com/embroider-build/embroider/issues/267

stefanpenner:
Nice!

PR #293, opened 2019-07-27
Added a global flag that allows disable AutoBoot during runtime

dnalagatla:
This flag allows disabling auto-boot in enviroments other that browser. For example: FastBoot


dnalagatla:
@stefanpenner @ef4 I added a global flag to disable autoboot during runtime. Let me know this change looks good.

packages/core/src/app.ts
```diff
@@ -944,7 +944,9 @@ let d = w.define;
 {{/if}}
 
 {{#if autoRun ~}}
-  i("{{js-string-escape mainModule}}").default.create({{{json-stringify appConfig}}});
+  if (!window.disableAutoBoot) {
```
stefanpenner: We should likely bike-shed the name of this environment variable.

My first thought would be: `EMBER_DISABLE_AUTO_BOOT`

Also, should we make this `window`? Or simply leave it unbound, allowing the wrapping environment to provide it without mutating a potential global.

stefanpenner: this will also need a test

dnalagatla: Makes sense, I will update the code to use `EMBER_DISABLE_AUTO_BOOT`. Thank you

dnalagatla: > Also, should we make this `window`? Or simply leave it unbound, allowing the wrapping environment to provide it without mutating a potential global.

In other parts of the template `window` is used. But I agree we should pull from global context without binding to the window object. I will update the code to below

`typeof EMBER_DISABLE_AUTO_BOOT === 'undefined' || ! EMBER_DISABLE_AUTO_BOOT`

dnalagatla: I didn't see any tests for the template generated here. Because of that didn't add a test. 

As I am not familiar adding tests for `broccoli-plugin` build, can you refer me to an example that I can use. 

I tested in the sample ember App by linking to the changes. And didn't see any issues.

dnalagatla: @stefanpenner I am not sure if there is an easy way of testing this global flag setup without adding a sample package with fastboot related changes. But Fastboot updates depends on this PR related to config-module - https://github.com/embroider-build/embroider/pull/224

if above PR looks good, Can we have above PR merged?

PR #297, opened 2019-07-30
[WIP] Enable addons to be rebuilt

thoov:
A POC of how addons could be selectively rebuilt if Broccoli passed change information into the plugin. The goal of this PR is to get overall feedback on the strategy and to confirm this is a viable path forward. I will add tests once we work out the higher level vision of this.

Overall Strategy:

* Broccoli passes change information to WaitForTrees (not yet added to broccoli).
* WaitForTrees translates that information down to the build callback in compat-addon
* The build callback uses tree-sync to conditionally "sync" to the workspaceDir based on if the addon has changed (for the first build all addons are considered "changed")

Perf:

Measuring the execution time of the build callback for a brand new app: 

With change tracking: ~50ms
Without: ~75ms 

I have not tested this in a large app but change tracking should stay relatively constant were as without will obviously grow with the number of addons.

Limitations:

This conditional checking would only be available for those that are using the version of broccoli were this is enabled. These changes are backwards compatible however there would be performance implications for those not on it. 

Open Questions for Reviewers:

* How should plugins and broccoli communicate change? In this POC broccoli passes change info down to the plugin or should it be where the plugin "calls" up and asks broccoli for this information?
* Should we only enable rebuilding for those only on the latest broccoli?

### DIFF
--- a/packages/compat/src/compat-addons.ts
+++ b/packages/compat/src/compat-addons.ts
@@ -10,12 +10,14 @@ import buildCompatAddon from './build-compat-addon';
 import Options, { optionsWithDefaults } from './options';
 import V1App from './v1-app';
 import { createHash } from 'crypto';
+import TreeSync from 'tree-sync';
 
 export default class CompatAddons implements Stage {
   private didBuild = false;
   private destDir: string;
   private packageCache: MovedPackageCache;
   private nonResolvableDeps: Package[];
+  private treeSyncMap: WeakMap<Package, any>;
   readonly inputPath: string;
   readonly tree: Tree;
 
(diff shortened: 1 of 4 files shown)


thoov:
cc: @stefanpenner & @ef4

packages/compat/src/compat-addons.ts
```diff
@@ -64,35 +67,58 @@ export default class CompatAddons implements Stage {
...
+      synthStyles,
+    }: {
+      movedAddons: string[];
+      synthVendor: string;
+      synthStyles: string;
+    },
+    changedMap?: Map<string, boolean>
+  ) {
+    // empty the directory only on the first pass
+    // TODO: should we still do this?
```
ef4: Yes, I think we should still clear the directory on the first build.

thoov:
@ef4 / @stefanpenner could you guys review again?

PR #300, opened 2019-08-01
pin development node/yarn version's with volta

stefanpenner:
@ef4 thoughts?

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,5 +30,9 @@
       "<rootDir>/packages/*",
       "<rootDir>/test-packages/"
     ]
+  },
+  "volta": {
+    "node": "10.16.1",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -30,5 +30,9 @@
       "<rootDir>/packages/*",
       "<rootDir>/test-packages/"
     ]
+  },
+  "volta": {
+    "node": "10.16.1",
```
stefanpenner: This version was chosen, as it seems to be compatible with the version of native extensions locked in the yarn file. We can bump this when the lockfiles are re-rolled

stefanpenner: https://github.com/embroider-build/embroider/pull/290 aims to re-roll the lockfile

PR #301, opened 2019-08-01
app.import + node_modules + outputFile failure.

stefanpenner:
The issue appears to only exist when all three are used together.
Error:
```
Build Error (WaitForTrees)

ENOENT: no such file or directory, stat '/private/var/folders/4r/whc65vwj1xggvvky3yy1cp9m000mw4/T/embroider/e81a6f/test-packages/macro-tests/node_modules/@embroider/synthesized-vendor/node_modules/highcharts/highcharts.src.js'
```

*Note: I’ve also added an example with also stresses the `using` option.*

### DIFF
--- a/test-packages/macro-tests/ember-cli-build.js
+++ b/test-packages/macro-tests/ember-cli-build.js
@@ -33,11 +33,38 @@ module.exports = function(defaults) {
     outputFile: 'apple.js'
   })
 
-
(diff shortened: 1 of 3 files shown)


stefanpenner:
Oh man, nice thank you! I didn't realize this was fixed, I was just about to start digging in... this is pleasant surprise.

stefanpenner:
I'll verify shortly.

Issue #312, opened 2019-08-20
Error: Can't reexport the named export 'BREAK' from non EcmaScript module (only default export is available)


knownasilya:
Stacktrace:
```
  - broccoliBuilderErrorStack: ModuleDependencyError: Can't reexport the named export 'BREAK' from non EcmaScript module (only default export is available)
    at Compilation.reportDependencyErrorsAndWarnings (/Users/ilya/workspace/crashco/web/node_modules/webpack/lib/Compilation.js:1368:21)
    at Compilation.finish (/Users/ilya/workspace/crashco/web/node_modules/webpack/lib/Compilation.js:1165:9)
    at hooks.make.callAsync.err (/Users/ilya/workspace/crashco/web/node_modules/webpack/lib/Compiler.js:548:17)
    at _done (<anonymous>:9:1)
    at _err0 (<anonymous>:20:22)
    at _addModuleChain (/Users/ilya/workspace/crashco/web/node_modules/webpack/lib/Compilation.js:1093:12)
    at processModuleDependencies.err (/Users/ilya/workspace/crashco/web/node_modules/webpack/lib/Compilation.js:1005:9)
    at processTicksAndRejections (internal/process/next_tick.js:74:9)
```

File:
Build Error (PackagerRunner) in ../../../../../../../../../Users/ilya/workspace/crashco/crashkit/node_modules/graphql/index.mjs

Using the default instructions in the readme.

knownasilya:
Looks like https://github.com/graphql/graphql-js/issues/1272#issuecomment-393903706 is the answer.

ef4:
Yes, most likely this is a problem with `mjs` files. (One of the more infuriating design decisions made by NodeJS.)

Try adding mjs to the start of this array:

https://github.com/embroider-build/embroider/blob/3755542b18fb9bf33cafd7f042d4e9415625da58/packages/compat/src/compat-app.ts#L144

knownasilya:
Doing that resolved that error.

PR #313, opened 2019-08-20
fix: add .mjs to allowed extensions

knownasilya:
Resolves https://github.com/embroider-build/embroider/issues/312


ef4:
Thanks.

PR #321, opened 2019-09-10
Ignore dotfiles to fix .eslintrc unexpected token error

### LINKED ISSUE #319: embroider attempting to parse `.eslintrc`
```
< Build Error (PackagerRunner) in .eslintrc:2
< Module parse failed: Unexpected token (2:17)
< You may need an appropriate loader to handle this file type.
< | {
< >   "parserOptions": {
< |     "sourceType": "module"
< |   },
< Stack T
```

thoov:
Fixes #319

### DIFF
--- a/packages/core/src/app-differ.ts
+++ b/packages/core/src/app-differ.ts
@@ -29,7 +29,7 @@ export default class AppDiffer {
             return {
               mayChange: addon.mayRebuild,
               walk() {
-                return walkSync.entries(definitelyDir);
+                return walkSync.entries(definitelyDir, { ignore: ['**/.*'] });
               },
             };
           }


packages/core/src/app-differ.ts
```diff
@@ -29,7 +29,7 @@ export default class AppDiffer {
             return {
               mayChange: addon.mayRebuild,
               walk() {
-                return walkSync.entries(definitelyDir);
+                return walkSync.entries(definitelyDir, { ignore: ['**/.*'] });
```
rwjblue: Hmm, seems _vaguely_ odd to somewhat arbitrarily exclude all dot files. Does this affect build output (e.g. will modules be _missing_ that would have been present before)?

ef4: I think we *do* need some special handling for dotfiles. Stock ember-cli refuses to include any dotfiles in the app, even if they have `.js` extensions and are sitting in directories that are normally part of the app.

But I agree this may be the wrong place to do it. It's not clear that we don't want the files present, it's more that we don't want them to be treated as implicit-modules.

ef4: @rwjblue I tested to confirm it's true, but maybe you know where it actually happens: why does `app/whatever.js` appear in `requirejs.entries` but `app/.whatever.js` does not?

stefanpenner: let me do a quick debugging.

stefanpenner: So @thoov and myself did some debugging, it appears that the first thing to filter out dotfiles currently is broccoli-concat.

specifically here:
https://github.com/broccolijs/broccoli-concat/blob/d08f478100ee920e7cac7a32accdd158f35fde08/concat.js#L228

As ` this.allInputFiles` contains `['app-name/**/*.js']` which does not match dotfiles.

```
> walkSync('.', ['dot-dot-dot/**/*.js']).filter(x => x.includes('bar') || x.includes('foo'))
[]
> walkSync('.', ['dot-dot-dot/**/*']).filter(x => x.includes('bar') || x.includes('foo'))
[]
> walkSync('.', ['dot-dot-dot/']).filter(x => x.includes('bar') || x.includes('foo'))
[]
> walkSync('.', ['dot-dot-dot/.*']).filter(x => x.includes('bar') || x.includes('foo'))
[ 'dot-dot-dot/.bar', 'dot-dot-dot/.foo.js' ]
```

stefanpenner: parity with classic builds may be tricky, since we must let the "babel" step see them, but not let the bundling step consider them as entry point. That being said, maybe there is an better place to put this logic...

stefanpenner: Chated with @ef4, TL;DR:

```
* all dotfiles are not considered as implicit-modules
* this PR is slightly overzealous.
```

stefanpenner: actually: https://github.com/embroider-build/embroider/blob/db57ada3aed6bd2fca6369f94856b89fc32250db/packages/compat/src/v1-addon.ts#L452 may be more relevant

stefanpenner:
@ef4 when you have a chance, your review/input would be good. We moved this slightly before implicit-modules, so that we can also ensure these modules do not appear in the `entryTemplate` but if that is too soon, we can move it to only not appear in implicit-modules list [here](https://github.com/embroider-build/embroider/pull/321/files#diff-81b62ae153aff221cffd2b75d7adcfc3R823)

ef4:
> although we thought changes would need to occur to v1-addon it turns out the way minimatch works in walkSync, the v1-addon code already does not match on dotfiles.

That makes sense, but it made me wonder why the same logic wasn't already taking care of the list you're filtering. That list also comes ultimately from walkSync, here:

https://github.com/embroider-build/embroider/blob/05ffe58f0c695241560de89c72c75589fd55032e/packages/core/src/app-differ.ts#L32

and here:

https://github.com/embroider-build/embroider/blob/05ffe58f0c695241560de89c72c75589fd55032e/packages/core/src/app-differ.ts#L43

I guess it's because there are no globs passed to walkSync in these places, so no minmatch logic runs.

ef4:
We need meaningful test coverage here, for both the addon and app cases. This seems very easy to break by accident, particularly since the addon case is being handled by a non-obvious feature of the glob matcher. 

You can make a new test file modeled on [this one](https://github.com/embroider-build/embroider/blob/05ffe58f0c695241560de89c72c75589fd55032e/packages/compat/tests/app-import-test.ts). It can have dotfiles in both the app and an addon, and it can assert that the dotfiles names don't appear in the synthesized entrypoint file.

stefanpenner:
> I guess it's because there are no globs passed to walkSync in these places, so no minmatch logic runs.

@ef4 ya exactly, most likely something  walkSync itself may need to change in a future version. Essentially it should respect minimatch's behavior even without globs, but allow via flag to opt into dotfiles. But that is likely for another day...

thoov:
@ef4: I added a test based on your recommendation. It tests that dot files will not show up in `my-app.js` nor will they show up as `implicit-modules` for addons.

PR #322, opened 2019-09-11
fastboot support + tests

stefanpenner:
I realized we don't yet have fastboot in our test suite, so I figured I would take the smallest step possible and add it to macro-test.

Once ember-cli-fastboot is added to an embroider app, that app will fail to build.
It will error with:

 `Error: attempted to set config after we have already emitted our config`

<details>
=================================================================================

ENV Summary:

  TIME: Wed Sep 11 2019 13:52:37 GMT-0600 (MDT)
  TITLE: ember
  ARGV:
  - /Users/spenner/.volta/tools/image/node/8.16.1/6.4.1/bin/node
  - /Users/spenner/src/embroider-build/embroider/test-packages/macro-tests/node_modules/.bin/ember
  - test
  - --test-port=0
  EXEC_PATH: /Users/spenner/.volta/tools/image/node/8.16.1/6.4.1/bin/node
  TMPDIR: /var/folders/4r/whc65vwj1xggvvky3yy1cp9m000mw4/T
  SHELL: /usr/local/bin/fish
  PATH:
  - /var/folders/4r/whc65vwj1xggvvky3yy1cp9m000mw4/T/yarn--1568231541278-0.06136183649502702
  - /Users/spenner/src/embroider-build/embroider/test-packages/macro-tests/node_modules/.bin
  - /Users/spenner/.config/yarn/link/node_modules/.bin
  - /Users/spenner/src/embroider-build/embroider/node_modules/.bin
  - /Users/spenner/src/embroider-build/embroider/test-packages/macro-tests/node_modules/.bin
  - /Users/spenner/.config/yarn/link/node_modules/.bin
  - /Users/spenner/src/embroider-build/embroider/node_modules/.bin
  - /Users/spenner/.volta/tools/image/node/8.16.1/6.4.1/libexec/lib/node_modules/npm/bin/node-gyp-bin
  - /Users/spenner/.volta/tools/image/node/8.16.1/6.4.1/lib/node_modules/npm/bin/node-gyp-bin
  - /Users/spenner/.volta/tools/image/node/8.16.1/6.4.1/bin/node_modules/npm/bin/node-gyp-bin
  - /Users/spenner/.volta/tools/image/node/8.16.1/6.4.1/bin
  - /Users/spenner/.volta/tools/image/yarn/1.16.0/bin
  - /Users/spenner/.cargo/bin
  - /Users/spenner/.fzf/bin
  - /Users/spenner/.yarn/bin
  - /Users/spenner/.local/bin
[...]

PR #325, opened 2019-09-18
re-roll lockfile

stefanpenner:
Doing so runs into some issues that run square into:

```
    attempted to set config after we have already emitted our config
```

Which I believe other add-ons such as ember-cli-fastboot also encounter.
 I took at stab at fixing this, at-least for ember-cli-fastboot (https://github.com/embroider-build/embroider/pull/322): https://github.com/embroider-build/embroider/pull/322/commits/5546c90819380edb1c3cb5b36420304315b4c1d9 Which likely needs further discussion.

### DIFF
--- a/test-packages/static-app/package.json
+++ b/test-packages/static-app/package.json
@@ -56,5 +56,8 @@
   },
   "engines": {
     "node": "8.* || >= 10.*"
+  },
+  "resolutions": {
+    "ember-auto-import": "stefanpenner/ember-auto-import#lazily-read-babel-config"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,147 +2,53 @@
 # yarn lockfile v1
 
 
-"@babel/code-frame@^7.0.0":
-  version "7.0.0"
-  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.0.0.tgz#06e2ab19bdb535385559aabb5ba59729482800f8"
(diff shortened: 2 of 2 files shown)


stefanpenner:
a quick test suggests https://github.com/embroider-build/embroider/commit/5546c90819380edb1c3cb5b36420304315b4c1d9 is not sufficient in fixing the above mentioned error:  `attempted to set config after we have already emitted our config/Users/spenner/src/embroider-build/embroider/test-packages/static-app/ember-cli-build.js`

Further investigation will be required.

stefanpenner:
Looks the above problem was being caused by ember-auto-import. We have a WIP fix up: https://github.com/ef4/ember-auto-import/pull/245

test-packages/static-app/package.json
```diff
@@ -56,5 +56,8 @@
   },
   "engines": {
     "node": "8.* || >= 10.*"
+  },
+  "resolutions": {
+    "ember-auto-import": "stefanpenner/ember-auto-import#lazily-read-babel-config"
```
stefanpenner: I suspect this resolution isn't working properly on CI, but appears to be fine locally. With this change, this CI work lands and we are good.

stefanpenner:
I'm going to move from this, while we wait for ember-auto-imports to be fixed, and onto improving the warnings/errors in MacroConfig to catch these issues early and with good messaging.

stefanpenner:
https://github.com/embroider-build/embroider/pull/325 is also a stab at a less brittle approach re: MacrosConfig

PR #329, opened 2019-09-26
Importing module name fails for fastboot

dnalagatla:
PR for testing the import module fails fastboot only assets, as there is module name mismatch.


stefanpenner:
@dnalagatla can you apply the patches I described in chat? That way we can see the actual issues.

stefanpenner:
@dnalagatla is the error: https://travis-ci.com/embroider-build/embroider/builds/129950655 the one you where seeing locally?

dnalagatla:
@stefanpenner that is not the error I was seeing locally. I think the tests are not taking assets from index.html. let me check quickly

dnalagatla:
> @dnalagatla is the error: https://travis-ci.com/embroider-build/embroider/builds/129950655 the one you where seeing locally?

@stefanpenner I finally got the tests fixed to show the moduleName mismatch. Thank you providing the patch. 

Sorry, I took longer than I expected after getting the patch to build the tests to show the issue. Had some gaps in getting my tests working with jest. 

Issue:
Addons can define fastboot specific code, that only runs in fastboot env and can have imports defined with moduleName used in ember-cli configs. Embroider during its build updates the module to module name based on location of the library in node_modules. Code in question - https://github.com/embroider-build/embroider/blob/master/packages/core/src/babel-plugin-adjust-imports.ts#L114

As a result, importing of a module from chunk-<app>.js  in <app>-fastboot.js throws following error during the initialization of Fastboot app:
`Could not find module `funky-sample-addon` imported from `macro-tests/instance-initializers/sample-fastboot-initalize``

Repro steps:
1. Check out this branch.
2. `yarn`
3. run `yarn test test-packages/macro-fastboot-import-issue.test.js`

stefanpenner:
> Sorry, I took longer than I expected after getting the patch to build the tests to show the issue. Had some gaps in getting my tests working with jest.

No problem, thank you for doing this.

ef4:
I understand this better now.

ember-cli-fastboot has its own entire build pipeline for the `treeForFastBoot` in each addon, plus the app's own `fastboot` directory. This is not OK. Embroider needs to be able to see and control that part of the module graph.

To fix this, we need to make stage2 understand fastboot-specific modules. It will need to layer them correctly onto the rest of the app JS. Where there are collisions (both fastboot and non-fastboot versions of the same module), we'll need to combine them and use the macro system to switch between the implementations. 

We'll also want to finish implementing a feature in the macro system that we haven't yet -- allowing the macros to be runtime-evaluated in development mode. That will let us have a single build that works for both browser and fastboot.

ef4:
This got fixed as of 0.16.0.

PR #330, opened 2019-09-26
Guard MacroConfig reads and writes

stefanpenner:
Add `finalization` concept to MacroConfigs to guard against reading
before writes are done or writing while reads are possible.

The timing ends up being:

1. MacroConfig is writable (not readable)
2. MacroConfig is finalized, and becomes readable but no longer writable
  In general the goal here was allow config writing until just before `treeFor*` hooks are invoked.

  * For Classic builds: once `app.toTree` is invoked, the MacosConfig is finalized, becoming frozen but readonable
  * For Embroider builds: about the same time but via the `default-pipeline` the MacrosConfig  is finalized, becoming frozen but readonable

For the classic app builds the `toTree` usage here is a POC, if we continue down this path I don't mind adding the corresponding public API to ember-cli.

--- 

paired with @thoov


packages/core/src/app.ts
```diff
@@ -259,9 +264,7 @@ export class AppBuilder<TreeNames> {
     private app: Package,
     private adapter: AppAdapter<TreeNames>,
     private options: Required<Options>
-  ) {
-    MacrosConfig.shared().setOwnConfig(__filename, { active: true });
-  }
+  ) {}
```
thoov: Can we remove the constructor completely?

packages/macros/src/ember-addon-main.ts
```diff
@@ -43,6 +56,7 @@ export = {
           },
         });
       });
+
```
thoov: Remove newline

packages/macros/src/macros-config.ts
```diff
@@ -59,9 +65,10 @@ export default class MacrosConfig {
   }
 
   private internalSetConfig(fromPath: string, packageName: string | undefined, config: unknown) {
-    if (this.cachedUserConfigs) {
-      throw new Error(`attempted to set config after we have already emitted our config`);
+    if (this._configWritable === false) {
```
thoov:
```js
if (!this._configWritable) {
```

packages/macros/src/macros-config.ts
```diff
@@ -161,9 +174,10 @@ export default class MacrosConfig {
   // this exists because @embroider/compat rewrites and moves v1 addons, and
   // their macro configs need to follow them to their new homes.
   packageMoved(oldPath: string, newPath: string) {
-    if (this.cachedUserConfigs) {
-      throw new Error(`attempted to call packageMoved after we have already emitted our config`);
+    if (this._configWritable === false) {
```
thoov:
```js
if (this._configWritable === false) {
```

packages/macros/tests/babel/helpers.ts
```diff
@@ -4,14 +4,17 @@ import { allBabelVersions as allBabel, runDefault } from '@embroider/test-suppor
...
 export { runDefault };
-
 export function allBabelVersions(createTests: (transform: (code: string) => string, config: MacrosConfig) => void) {
   let config: MacrosConfig;
 
   describe('without presets', function() {
     allBabel({
       babelConfig() {
         config = new MacrosConfig();
+        if (createTests.length < 2) {
```
thoov: We should probably add a comment about this

packages/macros/src/ember-addon-main.ts
```diff
@@ -25,8 +25,21 @@ export = {
     let babelOptions = (parentOptions.babel = parentOptions.babel || {});
     let babelPlugins = (babelOptions.plugins = babelOptions.plugins || []);
     babelPlugins.unshift(MacrosConfig.shared().babelPluginConfig(source));
+
+    // TODO: LOL MEGA HACK
```
thoov: We should discuss this

thoov:
Do we want to add the ability to turn this into warnings?

packages/core/src/app.ts
stefanpenner: ya

packages/macros/src/ember-addon-main.ts
stefanpenner: yes we must

packages/macros/src/ember-addon-main.ts
thoov: Fixed

packages/macros/src/macros-config.ts
thoov: Fixed

packages/macros/src/macros-config.ts
thoov: Fixed

packages/macros/src/ember-addon-main.ts
stefanpenner: If we decide this is the correct timing, I don't mind adding the appropriate hook to ember-cli proper, and using this as a temp polyfil.

cc @ef4

stefanpenner:
> Do we want to add the ability to turn this into warnings?

Let's wait, and let a use-case drive that. I would prefer to keep it strict, as other "downstream" issues may be pretty gnarly to debug.

packages/macros/tests/babel/helpers.ts
stefanpenner: ya

stefanpenner: doing that now.

packages/macros/src/ember-addon-main.ts
```diff
@@ -3,28 +3,46 @@ import { MacrosConfig } from '.';
...
+    //
+    // In general, we do not condone this type of monkey patching and our plan
+    // is if this indeed remains needed, to add the appropriate public API in
+    // ember-cli. One interesting tid-bit, once users use embroider primarily
+    // this hack will have no impact, and then adding the API to ember-cli
+    // may be wasted. We will see how this plays out, and do the appropriate
+    // thing.
+    const originalToTree = parent.toTree;
+    parent.toTree = function() {
+      debugger;
```
kratiahuja: We should remove this debugger

packages/macros/src/macros-config.ts
```diff
@@ -9,38 +9,70 @@ export type Merger = (configs: unknown[]) => unknown;
...
+  // }
+
+  static for(key: any): MacrosConfig {
+    let config = CONFIGS.get(key);
+    if (config) {
+      return config;
     }
-    return localSharedState;
+    COUNTER;
+    debugger;
```
kratiahuja: Ditto.

Issue #183, opened 2019-05-19
Embroider + Fastboot: Build not invoking `app-boot` in contentFor hook. Causing Fastboot app to fail


dnalagatla:
Embroider document indicates contentFor hook for `app-boot` type is deprecated. Ember-cli-fastboot relies on that type to insert Fastboot app-factory module and during runtime, the fastboot relies on this module. 

https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/index.js#L109

As this module is not present Fastboot throws following error upon loading the app:
```
Error: Could not find module `~fastboot/app-factory` imported from `(require)`
    at missingModule (/Users/dnalagat/DEVELOPMENT/test-pemberly-bpr/test-ember/my-embroider-test/dist/assets/node_modules/@embroider/synthesized-vendor/vendor/loader/loader.js:247:1)
    at findModule (/Users/dnalagat/DEVELOPMENT/test-pemberly-bpr/test-ember/my-embroider-test/dist/assets/node_modules/@embroider/synthesized-vendor/vendor/loader/loader.js:258:1)
    at Object.requireModule (/Users/dnalagat/DEVELOPMENT/test-pemberly-bpr/test-ember/my-embroider-test/dist/assets/node_modules/@embroider/synthesized-vendor/vendor/loader/loader.js:24:1)
    at Object.<anonymous> (/Users/dnalagat/DEVELOPMENT/test-pemberly-bpr/test-ember/my-embroider-test/node_modules/fastboot/src/ember-app.js:179:18)
    at VMSandbox.run (/Users/dnalagat/DEVELOPMENT/test-pemberly-bpr/test-ember/my-embroider-test/node_modules/fastboot/src/vm-sandbox.js:18:15)
    at EmberApp.createEmberApp (/Users/dnalagat/DEVELOPMENT/test-pemberly-bpr/test-ember/my-embroider-test/node_modules/fastboot/src/ember-app.js:178:30)
    at EmberApp.retrieveSandboxedApp (/Users/dnalagat/DEVELOPMENT/test-pemberly-bpr/test-ember/my-embroider-test/node_modules/fastboot/src/ember-app.js:203:17)
    at new EmberApp (/Users/dnalagat/DEVELOPMENT/test-pemberly-bpr/test-ember/my-embroider-test/node_modules/fastboot/src/ember-app.js:59:21)
[...]

PR #335, opened 2019-10-03
Added support for custom app-boot type in contentFor

dnalagatla:
Fix for #183 

This includes the following:
1. Adds support to use custom `app-boot` of contentFor.
2. Updates that fixes Macro-config error came when building test app with fastboot support. Applied the patch provided by @stefanpenner 
3. Fixed vendor files in Macro-test app.
4. Test that asserts application is booting using custom boot logic added by ember-cli-fastboot addon as part of a separate fastboot app. @stefanpenner.
5. Added test for custom app-boot as part of Macro-test app.
6. Verified by linking to ember app using the ember-cli-fastboot add-on.

cc: @stefanpenner @ef4

### DIFF
--- a/packages/compat/src/compat-app.ts
+++ b/packages/compat/src/compat-app.ts
@@ -48,6 +48,7 @@ function setup(legacyEmberAppInstance: object, options: Required<Options>) {
   let htmlTree = oldPackage.htmlTree;
   let publicTree = oldPackage.publicTree;
   let configTree = oldPackage.config;
+  let appBootTree = oldPackage.appBoot;
 
   if (options.extraPublicTrees.length > 0) {
     publicTree = mergeTrees([publicTree, ...options.extraPublicTrees].filter(Boolean));
@@ -58,6 +59,7 @@ function setup(legacyEmberAppInstance: object, options: Required<Options>) {
     htmlTree,
     publicTree,
     configTree,
+    appBootTree,
   };
 
   let instantiate = async (root: string, appSrcDir: string, packageCache: PackageCache) => {
@@ -212,6 +214,10 @@ class CompatAppAdapter implements AppAdapter<TreeNames> {
 
   autoRun(): boolean {
     return this.oldPackage.autoRun;
+  }
+
+  appBoot(): string | undefined {
+    return this.oldPackage.appBoot.readAppBoot();
   }
 
   mainModule(): string {
--- a/packages/compat/src/v1-app.ts
+++ b/packages/compat/src/v1-app.ts
@@ -8,6 +8,7 @@ import resolve from 'resolve';
 import V1Package from './v1-package';
 import { Tree } from 'broccoli-plugin';
 import { V1Config, WriteV1Config } from './v1-config';
+import { WriteV1AppBoot, ReadV1AppBoot } from './v1-appboot';
 import { PackageCache, TemplateCompiler, TemplateCompilerPlugins, AddonMeta, Package } from '@embroider/core';
 import { writeJSONSync, ensureDirSync, copySync, readdirSync, pathExistsSync } from 'fs-extra';
 import AddToTree from './add-to-tree';
@@ -170,6 +171,22 @@ export default class V1App implements V1Package {
 
   get autoRun(): boolean {
     return this.app.options.autoRun;
+  }
+
+  @Memoize()
+  get appBoot(): ReadV1AppBoot {
+    let env = this.app.env;
+    let appBootContentTree = new WriteV1AppBoot();
+
+    let patterns = this.configReplacePatterns;
+
+    appBootContentTree = new this.configReplace(appBootContentTree, this.configTree, {
+      configPath: join('environments', `${env}.json`),
+      files: ['config/app-boot.js'],
+      patterns,
+    });
+
+    return new ReadV1AppBoot(appBootContentTree);
   }
 
   private get storeConfigInMeta(): boolean {
--- a/packages/compat/src/v1-appboot.ts
+++ b/packages/compat/src/v1-appboot.ts
@@ -1,0 +1,39 @@
+import Plugin, { Tree } from 'broccoli-plugin';
+import { join } from 'path';
+import { readFileSync, outputFileSync } from 'fs-extra';
+
+export class WriteV1AppBoot extends Plugin {
+  private lastContents: string | undefined;
+  constructor() {
+    super([], {
+      persistentOutput: true,
+    });
+  }
+  build() {
+    let filename = join(this.outputPath, 'config/app-boot.js');
+    let contents = `{{content-for "app-boot"}}`;
+    if (!this.lastContents || this.lastContents !== contents) {
+      outputFileSync(filename, contents);
+    }
+    this.lastContents = contents;
+  }
+}
+
+export class ReadV1AppBoot extends Plugin {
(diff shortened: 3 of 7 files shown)


stefanpenner:
This will depend on us landing: https://github.com/embroider-build/embroider/pull/330 (I'll see if I can wrap that up today)

test-packages/macro-fastboot-tests.test.js
```diff
@@ -0,0 +1,37 @@
...
+  // Fastboot's `visit` API.
+  try {
+    const fastboot = new FastBoot({
+      distPath,
+    });
+
+    const resp = await fastboot.visit('/');
+    expect(resp._body).toContain('<script type="x/boundary" id="fastboot-body-start">');
+    expect(resp._body).toContain('<script type="x/boundary" id="fastboot-body-end">');
+  } catch (e) {
```
stefanpenner: This try/catch appears only re-throw something which is already thrown. This leads me to believe it is redundant, can we remove it?

test-packages/macro-tests/vendor/prepend/four.js
```diff
@@ -1 +1 @@
-self.ORDER.push('prepend/four');
+self.ORDER && self.ORDER.push('prepend/four');
```
stefanpenner: This guard seems to be masking another issue, please ensure the root cause is addressed.

stefanpenner: My existing PR #322 addresses this issue: https://github.com/embroider-build/embroider/pull/322

Specifically, this commit: https://github.com/embroider-build/embroider/pull/322/commits/382d9e5b32748c7a5d6666f1e73596cc8d7158b0

test-packages/macro-fastboot-tests.test.js
```diff
@@ -0,0 +1,37 @@
...
+const FastBoot = require('fastboot');
+
+test('FastBoot - macro-test app boots in FastBoot environment using custom app-boot', async () => {
+  jest.setTimeout(120000);
+
+  const buildProcess = execa('ember', ['build'], {
+    cwd: `${__dirname}/macro-tests`,
+  });
+
+  const buildCompletePromise = new Promise(res => {
```
stefanpenner: wouldn't `await execa('ember', ['build'], { ... }` also tell us the build is complete, at which point we know `dirname + '/macro-tests/dists'` is ready to be consumed by fastboot

test-packages/macro-fastboot-tests.test.js
```diff
@@ -0,0 +1,37 @@
+const execa = require('execa');
+const FastBoot = require('fastboot');
+
+test('FastBoot - macro-test app boots in FastBoot environment using custom app-boot', async () => {
+  jest.setTimeout(120000);
+
+  const buildProcess = execa('ember', ['build'], {
+    cwd: `${__dirname}/macro-tests`,
```
stefanpenner: This is fine for now, although I know ed would prefer us to create a fresh test-app for fastboot, but if you don't have cycles I can totally help.

test-packages/macro-fastboot-tests.test.js
dnalagatla: sure, I will fix that.

dnalagatla:
I just realized, My change for {{content-for app-boot}} didn't go in and the tests were not capturing the correct scenarios.

Also, noticed that changes for supporting app-boot are more involved than I expected, as I cannot update `entryTemplate` in core/app.ts  with an else condition to include custom app-boot content.

https://github.com/embroider-build/embroider/blob/master/packages/core/src/app.ts#L953

My current approach based on my understanding of embroider flow is to:

-  Update entry template to take else condition [here](https://github.com/embroider-build/embroider/blob/master/packages/core/src/app.ts#L953)  

- update the Adapter interface to include `appBoot: string` property. 

- And then update `compat/v1-app.ts` to have a getter for `appBoot` that returns custom  app-boot content when autoRun is false or else return an empty string.

- The `appBoot` getter loops through `this.configReplacePatterns` to get the content for `app-boot` type.

The issue I am seeing is the signature of [compile](https://github.com/embroider-build/embroider/blob/master/packages/core/src/app.ts#L924) doesn't accept custom property and get typescript compilation error. 

Working around ^ compilation issue by making changes to JS code, I can see my approach works. But I am not happy with that.  

 @stefanpenner @ef4 can you recommend a good a way to support `app-boot`.

dnalagatla:
@stefanpenner @ef4 I added the changes to support app-boot. As this implementation should be included as part of entryTemplate, I updated the code with following flow:

- Update entry template to take else condition here

- update the Adapter interface to include appBoot: string property.

- And then update compat/v1-app.ts to have a getter for appBoot that returns custom app-boot content when autoRun is false or else return an empty string.

- The appBoot getter logic involves looping through this.configReplacePatterns to get the content for app-boot type.

Please let me know if you have any suggestions.

test-packages/macro-fastboot-tests.test.js
dnalagatla: I created a separate fastboot-app for fastboot related testing.

dnalagatla: will be useful for any future changes related to Fastboot add-on

test-packages/macro-fastboot-tests.test.js
dnalagatla: I cleaned up the code use await on the build command.

test-packages/macro-tests/vendor/prepend/four.js
dnalagatla: removed changes and fixed this issue as part o theis PR - https://github.com/embroider-build/embroider/pull/336

test-packages/macro-fastboot-tests.test.js
dnalagatla: Fixed it.

stefanpenner:
@dnalagatla tests appear to be failing, could you take a look?

ef4:
Thanks for updating.

This reimplements config replacement and it doesn't do it the same way the existing config replacement code (in broccoli-config-replace) works. For example, the replacer is not required to be a function, it can also be a string.

Let's continue to use broccoli-config-replace instead of maintaining a separate implementation. There is already a place [here](https://github.com/embroider-build/embroider/blob/117e8451c47257135f17dab975fda1565898118b/packages/compat/src/compat-app.ts#L44) to setup broccoli trees that will be consumed by `CompatAppAdapter`.

Also, this PR still has stray whitespace changes in unrelated files.

dnalagatla:
> Thanks for updating.
> 
> This reimplements config replacement and it doesn't do it the same way the existing config replacement code (in broccoli-config-replace) works. For example, the replacer is not required to be a function, it can also be a string.
> 
> Let's continue to use broccoli-config-replace instead of maintaining a separate implementation. There is already a place [here](https://github.com/embroider-build/embroider/blob/117e8451c47257135f17dab975fda1565898118b/packages/compat/src/compat-app.ts#L44) to setup broccoli trees that will be consumed by `CompatAppAdapter`.
> 
> Also, this PR still has stray whitespace changes in unrelated files.

Hi @ef4, thanks for the input. I made the changes based on your suggestion. Thanks to @sparshithNR for helping me with broccoli-tree flow. I have verified my tests are passing.

ef4:
Thank you, nice work.

PR #332, opened 2019-09-30
WIP: memoize template compiler parse/transform/loads

stefanpenner:
POC fix for #331

Gotta run to a meeting, will make this not terrible shortly.

PR #346, opened 2019-10-24
Fix remaining TODOs for memoizing template compilation

kratiahuja:
This PR is continuation of @stefanpenner work in this [PR](https://github.com/embroider-build/embroider/pull/332). 

This PR also fixes an issue with checking the entry values against `mtime` which returns a date. With this change, it is actually hitting the cache correctly.

Also tested with two LI apps, the build shows visible improvements.

cc: @stefanpenner

### DIFF
--- a/packages/core/src/template-compiler.ts
+++ b/packages/core/src/template-compiler.ts
@@ -37,9 +37,18 @@ interface GlimmerSyntax {
   print(ast: AST): string;
   defaultOptions(options: PreprocessOptions): PreprocessOptions;
   registerPlugin(type: string, plugin: unknown): void;
-  precompile(templateContents: string, options: {
-    contents: string; moduleName: string; filename: string, plugins?: any
-  }): string;
+  precompile(
+    templateContents: string,
+    options: {
+      contents: string;
+      moduleName: string;
+      filename: string;
+      plugins?: any;
+      parseOptions?: {
+        srcName?: string;
+      };
+    }
+  ): string;
   _Ember: { FEATURES: any; ENV: any };
   cacheKey: string;
 }
@@ -80,14 +89,11 @@ function getEmberExports(templateCompilerPath: string): EmbersExports {
     if (
       currentStat.mode === entry.stat.mode &&
       currentStat.size === entry.stat.size &&
-      currentStat.mtime === entry.stat.mtime
+      currentStat.mtime.getTime() === entry.stat.mtime.getTime()
     ) {
-      console.count('HIT');
       return entry.value;
     }
   }
-
-  console.count('MISS');
 
   let replacedVar = false;
   let stat = statSync(templateCompilerPath);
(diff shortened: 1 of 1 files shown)


packages/core/src/template-compiler.ts
```diff
@@ -37,9 +37,15 @@ interface GlimmerSyntax {
   print(ast: AST): string;
   defaultOptions(options: PreprocessOptions): PreprocessOptions;
   registerPlugin(type: string, plugin: unknown): void;
-  precompile(templateContents: string, options: {
-    contents: string; moduleName: string; filename: string, plugins?: any
-  }): string;
+  precompile(
+    templateContents: string,
+    options: {
```
rwjblue: FWIW, it also accepts `parseOptions?: { srcName?: string }`

https://github.com/ember-cli/ember-cli-htmlbars/blob/4fdc8949917a3f79a33a038fa581f183cf74bd1d/lib/template-compiler-plugin.js#L75-L77

kratiahuja: Thanks for catching that, I'll add this as well.

packages/core/src/template-compiler.ts
```diff
@@ -80,15 +89,12 @@ function getEmberExports(templateCompilerPath: string): EmbersExports {
     if (
       currentStat.mode === entry.stat.mode &&
       currentStat.size === entry.stat.size &&
-      currentStat.mtime === entry.stat.mtime
+      currentStat.mtime.getTime() === entry.stat.mtime.getTime()
```
stefanpenner: good catch!!

packages/core/src/template-compiler.ts
```diff
@@ -304,6 +300,11 @@ export default class TemplateCompiler {
     return { syntax, cacheKey };
   }
 
+  @Memoize()
+  private getReversedASTPlugins(ast: unknown[]): unknown[] {
```
stefanpenner: Two notes:

* we should comment why `reversed` is the correct order
* we should likely considered changing the word from `reversed` to `getOrderedASTPlugins` or simply `astPlugins`

packages/core/src/template-compiler.ts
```diff
@@ -316,12 +317,19 @@ export default class TemplateCompiler {
...
 
     let opts = this.syntax.defaultOptions({ contents, moduleName });
-    opts.plugins!.ast = [...this.params.plugins.ast!.slice().reverse(), ...opts.plugins!.ast!];
+    if (this.params.resolver) {
+      let transform = this.params.resolver.astTransformer(this);
+      if (transform) {
+        this.params.plugins.ast!.push(transform);
+      }
+    }
+    opts.plugins!.ast = [...this.getReversedASTPlugins(this.params.plugins.ast!), ...opts.plugins!.ast!];
```
stefanpenner: A brief comment describing the exact precedence/ordering here would be good. It is a little tricky, so us sharing that knowledge will help future travelers.

ef4:
Thanks @kratiahuja!

Issue #347, opened 2019-10-30
Cannot find module '...node_modules/broccoli-ember-hbs-template-compiler'


danwenzel:
Got this error trying to `yarn start` with embroider:

```
Cannot find module '.../node_modules/broccoli-ember-hbs-template-compiler'
```

After some digging, it turned out to be the presence of [`ember-cli-stripejs`](https://github.com/layervault/ember-cli-stripejs), which uses very old versions of ember-cli dependencies. `broccoli-ember-hbs-template-compiler` itself is listed as a `devDependency` of that addon.

Pull this branch to see the issue: https://github.com/danwenzel/embroider-issues/tree/ember-cli-stripejs

I realize embroider probably doesn't support addons with ember cli dependencies this old, but do you think there could be a way to catch a better error for this?

ef4:
Thanks for sharing a working reproduction.

ef4:
This is now fixed on master.

PR #353, opened 2019-11-11
short-circuit template compilation

ef4:
Classic ember-cli uses this check to avoid instantiating any template preprocessors for addons that have no templates. In addition to being an optimization, there are addons in the ecosystem that rely on this for correctness, because they have broken template preprocessors that throw on instantiation, but no actual templates.

We checked and `shouldCompileTemplates` has been stable since before ember-cli 3.0 through current master.

Co-authored-by: Jen Weber <j@jenweber.me>


ef4:
Fixes https://github.com/embroider-build/embroider/issues/347

Issue #354, opened 2019-11-11
version 0.8.0 breaks import of `emojis` package, imported via ember-auto-import


NullVoxPopuli:
Just upgraded to 0.8.0 to see if my previous test issue was resolved, new issue though:

```
ERROR Summary:

  - broccoliBuilderErrorStack: ModuleNotFoundError: Module not found: Error: Can't resolve './characters' in '/home/lprestonsegoiii/Development/NullVoxPopuli/emberclear/packages/frontend/node_modules/emojis'
    at factory.create (/home/lprestonsegoiii/Development/NullVoxPopuli/emberclear/packages/frontend/node_modules/webpack/lib/Compilation.js:821:10)
    at factory (/home/lprestonsegoiii/Development/NullVoxPopuli/emberclear/packages/frontend/node_modules/webpack/lib/NormalModuleFactory.js:397:22)
    at resolver (/home/lprestonsegoiii/Development/NullVoxPopuli/emberclear/packages/frontend/node_modules/webpack/lib/NormalModuleFactory.js:130:21)
    at asyncLib.parallel (/home/lprestonsegoiii/Development/NullVoxPopuli/emberclear/packages/frontend/node_modules/webpack/lib/NormalModuleFactory.js:224:22)
    at /home/lprestonsegoiii/Development/NullVoxPopuli/emberclear/packages/frontend/node_modules/neo-async/async.js:2830:7
    at /home/lprestonsegoiii/Development/NullVoxPopuli/emberclear/packages/frontend/node_modules/neo-async/async.js:6877:13
    at normalResolver.resolve (/home/lprestonsegoiii/Development/NullVoxPopuli/emberclear/packages/frontend/node_modules/webpack/lib/NormalModuleFactory.js:214:25)
    at doResolve (/home/lprestonsegoiii/Development/NullVoxPopuli/emberclear/packages/frontend/node_modules/enhanced-resolve/lib/Resolver.js:184:12)
    at hook.callAsync (/home/lprestonsegoiii/Development/NullVoxPopuli/emberclear/packages/frontend/node_modules/enhanced-resolve/lib/Resolver.js:238:5)
    at _fn0 (<anonymous>:15:1)
    at resolver.doResolve (/home/lprestonsegoiii/Development/NullVoxPopuli/emberclear/packages/frontend/node_modules/enhanced-resolve/lib/UnsafeCachePlugin.js:37:5)
[...]

PR #355, opened 2019-11-11
keep json and wasm as valid resolvable extensions

ef4:
This aligns us with ember-auto-import (which uses webpack's default resolve.extensions), and I think it's the best default in terms of wider ecosystem compatibility.

We will need to clarify in the v2 spec that importing these extensions is legal.


ef4:
Fixes #354

PR #359, opened 2019-11-19
Docs - Add common options to the example

jenweber:
I found myself hunting through other apps to get the block of text to copy-paste that would turn on the more strict features. This PR adds them right into the options example.


lifeart:
@jenweber I'm wondering Is it make sense to add some webpack options and  route-splitting as additional examples?

ef4:
Those additional suggestions would be great too but I'm merging this as is. Thanks.

PR #365, opened 2019-12-06
fix: css reorder issue

### LINKED ISSUE #337: The order of CSS included by app.import is different from ember-cli
Consider a class `dropdown`. In ember-cli, if there was an addon that had styles for this as

```css
.dropdown {
    color: red;
}
```

and if there was an `app.import('vendor/style.css')` that had contents as

```css
.dropdown {
    color: blue;
}
```

the color of the contents of `dropdown` were `blue`. This is because, the `vendor.css` was generated as:

```css
.dropdown {
    color: red;
}
.dropdown {
    color: blue;
}
```

In embroider, this order seems to have changed.
```css
.dropdown {
    color: blue;
}
.dropdown {
    color: red;
}
```

kandhavivekraj:
fixes #337

Embroider creates a new addon with all CSS files imported in `ember-cli-build.js`. While doing this, the styles included using `app.import` get added after the styles of the addon(consuming app) itself. Due to this, the styles break as their order has changed and there is no way to override any third party styles in consuming apps.

This PR fixes it by pushing all CSS files of each addon into a new array and finally unshifts those to the result array.

### DIFF
--- a/packages/core/src/app.ts
+++ b/packages/core/src/app.ts
@@ -302,12 +302,21 @@ export class AppBuilder<TreeNames> {
   }
 
   private impliedAddonAssets(type: keyof ImplicitAssetPaths): string[] {
-    let result = [];
+    let result: Array<string> = [];
     for (let addon of sortBy(this.adapter.allActiveAddons, this.scriptPriority.bind(this))) {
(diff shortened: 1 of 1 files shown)


ef4:
Thanks.

PR #378, opened 2020-01-29
workaround a style rebuilding bug in ember-cli

ef4:
when using BROCCOLI_ENABLED_MEMOIZE


ef4:
https://github.com/ember-cli/ember-cli/issues/9020

PR #379, opened 2020-01-30
Component template colocation

ef4:
This gives us support for octane's template colocation feature.

My current thinking on the spec is that v2 formatted addons don't get any automatic template colocation, they need to use `setComponentTemplate` explicitly. Their build step should take care of that before npm publish.


buschtoens:
Does this have implications for template-only components placed in a co-location structure in v2 addons? How should the `app` re-export look like?

ef4:
Yes, a V2 addon with a template-only component will need to produce the corresponding JS module like this before doing `npm publish`:

```js
import { setComponentTemplate } from "@ember/component";
import templateOnlyComponent from "@ember/component/template-only";
import TEMPLATE from './your-template.hbs';
export default setComponentTemplate(TEMPLATE, templateOnlyComponent());
```

Or alternatively:

```js
import { setComponentTemplate } from "@ember/component";
import templateOnlyComponent from "@ember/component/template-only";
const TEMPLATE = hbs`Your template here`;
export default setComponentTemplate(TEMPLATE, templateOnlyComponent());
```

In either case, that JS module will be what the app sees as the component, so that JS module would be the thing to reexport in app.

buschtoens:
Thanks for the precise explanation!

PR #383, opened 2020-02-04
Merge exisiting package.json to package.json created by emberoider build

SparshithNR:
This uses the same logic from #178. I have reversed the order in which merging is done. 
In #178 We merged `new package.json` content with `old package.json` content. That's why rebuild failed. If we just reverse the input values to the `merge` call it will fix the issue.

### DIFF
--- a/packages/core/src/app.ts
+++ b/packages/core/src/app.ts
@@ -4,7 +4,7 @@ import { compile } from './js-handlebars';
 import Package, { V2AddonPackage } from './package';
 import resolve from 'resolve';
 import { Memoize } from 'typescript-memoize';
-import { writeFileSync, ensureDirSync, copySync, unlinkSync, statSync } from 'fs-extra';
+import { writeFileSync, ensureDirSync, copySync, unlinkSync, statSync, readJSONSync } from 'fs-extra';
 import { join, dirname, sep, resolve as resolvePath } from 'path';
 import { todo, debug, warn } from './messages';
 import cloneDeep from 'lodash/cloneDeep';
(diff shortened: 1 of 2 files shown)


ef4:
This is still the wrong place to do this. It allows a data cycle. Each time the build runs, we read and write from *the same file*. So future builds get state mixed in from earlier builds.

This is the right location to write, but the reading of package.json needs to come earlier, before it gets written out into this location.

The `package.json` from fastboot is present in one of the `inputPaths`, and the `AppDiffer` is what's copying it into this directory.

SparshithNR:
> but the reading of package.json needs to come earlier, before it gets written out into this location.
@ef4 I am not clear about this. Is there any recommendation where it can be done?

ef4:
```js
```

ef4:
This `build` method has `inputPaths` as an argument. For each path in `inputPaths`, look for a `package.json` file in that path. If it exists, load it. Merge together all the ones you find. Finally, merge on top of that `this.app.packageJSON` as before, and write out to `pkgPath`.

SparshithNR:
@ef4 I have updated the review with merging all package.json into one package.json file.

packages/core/src/app.ts
```diff
@@ -680,6 +697,12 @@ export class AppBuilder<TreeNames> {
     }
     pkg['ember-addon'] = Object.assign({}, pkg['ember-addon'], meta);
     const pkgPath = join(this.root, 'package.json');
+    if (existsSync(pkgPath)) {
```
ef4: This is still reading from the output directory.

ef4:
^ That was just a rebase on master before I make a few small changes. Overall this is pretty close.

ef4:
Thanks. There were just a few small things needed:

 - using `require` for the package.json is likely to run afoul of node's require cache. Any rebuild changes that cause the package.json to change might not get picked up. So I used readJSONSync from fs-extra instead.
 - the `assets:any` type was too general, and was masking a bug. Assets are of multiple possible kinds, which needed to be handled.
 - I added regression test coverage.

PR #398, opened 2020-03-04
More macros system implementation

ef4:
This adds runtime implementations for the JS `each`, `getConfig`, `getOwnConfig`, and `macroCondition` macros. This will let us defer optimization to runtime in development, so a single build can (for example) run in both fastboot and browser. Then in production you can still use the build-time macro implementation to produce two separate optimized builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 ---
 language: node_js
 node_js:
-  - "10"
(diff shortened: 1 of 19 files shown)


ef4:
yarn is messing with me here. It hoisted incorrectly so that a peerDep is lost.

PR #401, opened 2020-03-13
Template resolver for addons in stage 1

SparshithNR:
Provide template resolver so that transforms to get the correct path as moduleName.

- [x] Tests to be added.


ef4:
I think this is waiting on the change we discussed that would replace `class V1AddonCompatResolver extends CompatResolver` with something like `class V1AddonResolver implements Resolver`.

ef4:
Thanks, looks good.

Issue #397, opened 2020-02-26
Complete macro implementation from rfc


ef4:
This tracks bringing `@embroider/macros` up to date with [RFC 507](https://github.com/emberjs/rfcs/pull/507).

Here's a table to track progress.
 - "implemented macro" means that the build-time implementation is done and matches the RFC. Some of the unchecked ones may have implementations but need review to compare against the latest RFC
 - "implemented types & runtime" means that the run-time implementation is done and matches the RFC. This also provides the typescript types that users of the macros will see.

| name                             | type| implemented macro | implemented types & runtime | documented |
| --------------------- |----- |---------------------|-------------------------------|--------------|
| dependencySatisfies | js      | ✔️                               |                                                       |                        |
|                                       | hbs   |                                     |                                                       |                        |
| each                             | js      | ✔️                               |                                                       |                        |
|                                       | hbs   |                                     |                                                       |                        |
| failBuild                        | js      | ✔️                               |                                                       |                        |
|                                       | hbs   |                                     |                                                       |                        |
| getConfig                    | js      | ✔️                               |                                                       |                        |
|                                       | hbs   |                                     |                                                       |                        |
| getOwnConfig              | js      | ✔️                               |                                                       |                        |
|                                       | hbs   |                                     |                                                       |                        |
| importSync                  | js      | ✔️                               |                                                       |                        |
|                                       | hbs   |                                     |                                                       |                        |
| macroCondition         | js      | ✔️                               |                                                       |                        |
|                                       | hbs   |                                     |                                                       |                        |
| moduleExists              | js      |  ✔️                               |                                                       |                        |
|                                       | hbs   |                                     |                                                       |                        |
| maybeAttrs                 | js      | not applicable             |   not applicable                          |  not applicable |
|                                       | hbs   |                                     |                                                       |                        |

simonihmig:
Here's a first attempt at updating the hbs macros: #404

rwjblue:
The grid probably needs tweaking now that #404 was landed.

simonihmig:
> The grid probably needs tweaking now that #404 was landed.

Yes. As [said before](https://github.com/embroider-build/embroider/pull/404#issuecomment-603428496) I wanted to do that, but was a bit unsure if I missed anything from the RFC. Assuming this is not the case, I updated the grid now!

simonihmig:
I updated the grid further with what my understanding is of the runtime implementations, but please double check!
* Babel:
[...]

PR #404, opened 2020-03-22
Update Handlebars Macros according to RFC

simonihmig:
Drop `macroIf`, add `macroCondition`. 
Internal refactoring: extracted `macroMaybeAttrs` to its own (test) module.

AFAICT this should match now what is proposed in the RFC, but happy to iterate on this in case it is not yet!


ef4:
Wow, really nice work @simonihmig!

simonihmig:
Thank you @ef4! 🙏

 As said above, I was a bit unsure if anything is missing for the hbs macros to make them match the RFC, but I *think* they should. Can you confirm? If so, I can update the tracking issue #397...

PR #419, opened 2020-04-01
support nullish coalescing inside macroCondition

### LINKED ISSUE #406: Support nullish coalescing inside macroCondition
My plan for making it easy to access config from optional packages is to make sure the macro system understands nullish coalescing syntax:

```js
if (macroCondition(getConfig('ember-cli-fastboot')?.isRunning)) {
  // fastboot-only code goes here
}
```

This is needed for fastboot support because I want to use it just like that within embroider itself. Without it there's not a clear way to access config from packages that may or may not be present.

SparshithNR:
PR includes
- Fixes #406 
- Babel upgrade
- Bumping node heap size 
- updating @embroider/macro to 0.14.0


ef4:
Here's my example from our discussion about next steps for this PR: https://github.com/embroider-build/embroider/tree/collapse

SparshithNR:
@ef4 If this PR is doing way more things, 
We can bump heap size in #400 and merge that. Then I can rebase so that we can reduce around 10-12 files in here.

ef4:
Your latest change fixes part of what I pointed out but not all of it. Here are more failing tests.

ef4:
In 244ac3254843579d0d78a4758eccd882f90e34c7 I changed the types so that typescript will enforce that we are only using `value` on confident results.

SparshithNR:
I added more test cases and tried covering all the branching paths. Update the PR with change in code and tests.

ef4:
Thanks, the tests are looking good.

This still doesn't address the stuff we talked about in this example commit: https://github.com/embroider-build/embroider/commit/c78e76a2f09c026079764d7f5c3f48128569e033

`collapse` should only call `evaluate`, it shouldn't need its own special handling for MemberExpression and OptionalMemberExpression.

ef4:
Thanks @SparshithNR! This turned into a pretty big PR with lots of good work. I was able to address the last few issues we discussed by refactoring the way macros can call each other, which lets us run see things before present-env. So now this PR increases the static analysis simultaneously from two different directions -- more cases are handled by your expanded evaluation code, and we get more information by running sooner.

Issue #414, opened 2020-03-30
Webpack Bundle Analyzer stats server is started even when only building the app


ohcibi:
I tried to use webpack-bundle-analyzer with my app and it basically works and gives me all the stats that I expect. But when I only run `ember build` the server is still started and the process tries to open that page in the browser. So for every `ember build` a new browser tab pops up, stealing focus from the app that is currently used. Not really an error but a bit annoying.

NullVoxPopuli:
webpack-bundle-analyzer can be configured to not open:
```ts
new BundleAnalyzerPlugin({
              generateStatsFile: true,
              openAnalyzer: false,
              statsFilename: path.join(
                process.cwd(),
                'concat-stats-for',
                'ember-auto-import.json',
              ),
            }),
```

ohcibi:
>idk if this helps, but webpack-bundle-analyzer can be configured to not open: ...

@NullVoxPopuli yeah I know but I would expect `ember build` to do this automatically, hence this issue. Neither opening the browser nor staring up the analyze server makes sense for `ember build` when that server is stopped after the build process ends.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

ef4:
The thing is, if you’re passing custom config to webpack you are — by definition — doing something ember-cli doesn’t do automatically for you. That is the trade off.

ohcibi:
@ef4 that "custom config" ist documented within the embroider docs without mentioning that the analyzer server (in fact its not even mentioned that there is one) will be pointlessly started by `ember build`. If you can explain why and how to disable it for `ember build` only but not for `ember serve`, it would be beneficial if that "trade off" would be added to the docs that introduce it, as well. I would PR it myself but I know nothing about that regarding embroider except for the snippet @NullVoxPopuli posted.

void-mAlex:
we now have documentation on how to configure this so closing this issue

PR #423, opened 2020-04-06
Add info about disabling webpack bundle analyzer result browser

ohcibi:
Regarding #414 this adds the snippet from @NullVoxPopuli to disable the webpack bundle anlyzer server to the analyzer readme.


ef4:
Thanks!

PR #425, opened 2020-04-08
Move to github actions

thoov:
This has the added benefit of using volta in CI such that we are consistent with our local environments.


thoov:
@ef4 you will need to enable actions for this repo for this to work

ef4:
Thanks, I have been wanting to do this.

Issue #428, opened 2020-04-09
Support in-repo addons inside addons


ef4:
Right now they're supported inside the app, but not inside addons.

This came up first in the context of engines, which are addons that are more likely to use in-repo addons. But I think they're technically allowed inside *any* addon, not just an engine.

mehulkar:
Some related conversation here that seems worth cross linking https://github.com/ember-cli/ember-cli/issues/9016

ef4:
Sketch of where to fix: https://github.com/embroider-build/embroider/tree/in-repo-addon-in-addon

thoov:
I am working on this

PR #429, opened 2020-04-11
Support in-repo addons within other addons

thoov:
Fixes: #428 

This PR enables in-repo addons of addons to be supported. To achieve this the PR:

- enables V1Addons to be extended (just like V1App) such that their dependencies include in-repo addons

- rewrites the addon's package.json to include in-repo addons as direct dependencies instead of as ember-addon.paths (ie removes the "emberness" from them)

- Symlink the in-repo addons into the addons node_module such that they can be naturally discovered like normal dependencies

### DIFF
--- a/packages/compat/src/build-compat-addon.ts
+++ b/packages/compat/src/build-compat-addon.ts
@@ -8,21 +8,27 @@ import Funnel from 'broccoli-funnel';
 import { UnwatchedDir } from 'broccoli-source';
 import EmptyPackageTree from './empty-package-tree';
 
-export default function cachedBuildCompatAddon(originalPackage: Package, v1Cache: V1InstanceCache): Tree {
-  let tree = buildCompatAddon(originalPackage, v1Cache);
+export default function cachedBuildCompatAddon(
+  originalPackage: Package,
+  v1Cache: V1InstanceCache
+): { tree: Tree; nonResolvableDeps: Package[] } {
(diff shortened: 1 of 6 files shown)


packages/compat/src/build-compat-addon.ts
```diff
@@ -36,16 +42,26 @@ function buildCompatAddon(originalPackage: Package, v1Cache: V1InstanceCache): T
...
-    return new EmptyPackageTree();
+    return { tree: new EmptyPackageTree(), nonResolvableDeps: [] };
+  }
+
+  let nonResolvableDependencies: Set<Package> = new Set();
+  for (let pkg of oldPackages) {
+    for (let dep of pkg.nonResolvableDependencies()) {
+      nonResolvableDependencies.add(dep);
+    }
   }
```
thoov, line 53: @ef4 This is the only spot I am confused about. In your branch you left a comment about how we need to loop over oldPackages and add each ones nonResolvableDependencies. Can you help describe a test case I can add that will have one of the oldPackages contain a different in-repo addon. Right now each oldPackage is basically the same so this is really no different than: `oldPackages[0].nonResolvableDependencies();`

ef4: I think it's pretty hard to contrive an example where they would differ. I'm OK with not having a test for this, it's enough to review this case and see that it's doing the right thing even if they did differ in the future.

Issue #113, opened 2019-03-30
Broken app tests


josemarluedke:
When using Embroider, the application tests fails. 

Here is the reproduction: https://github.com/josemarluedke/embroider-pods
Just start the server and visit `/tests`.

<img width="1680" alt="Screen Shot 2019-03-30 at 9 42 04 AM" src="https://user-images.githubusercontent.com/230476/55278951-24778a80-52d0-11e9-8ba2-3ba23622afbd.png">

Note that tests pass when not using embroider. To see that, just start the server with: `CLASSIC=true ember s` and visit `/tests`.

ef4:
Does it work if you run “ember test”?

It is technically not really supported to run tests via “ember s”, although
it often traditionally works by accident.

josemarluedke:
Yep, tests passes when running with `ember test`.

ef4:
It would be nice to fix this anyway, since people do like to run their
tests that way.

I think the bug is that we layer the app vs test code differently than
classic ember cli, and you’re getting the app bundle twice.

NullVoxPopuli:
I'm getting this error:
```
not ok 2 Chrome 77.0 - [undefined ms] - Global error: Uncaught Error: Assertion Failed: The tests file was not loaded. Make sure your tests index.html includes "assets/tests.js". at http://0.0.0.0:7357/assets/vendor.js, line 37262
    ---
        browser log: |
            testContext: [object Object]
            ERROR: Uncaught Error: Assertion Failed: The tests file was not loaded. Make sure your tests index.html includes "assets/tests.js". at http://0.0.0.0:7357/assets/vendor.js, line 37262
```

sandstrom:
I'm doing some issue gardening 🌱🌿 🌷 and came upon this issue. Since it's quite old I just wanted to ask if this is still relevant? If it isn't, maybe we can close this issue?

By closing some old issues we reduce the list of open issues to a more manageable set.

PR #431, opened 2020-04-15
Enable tests to run from the browser with ember s command

SparshithNR:
https://github.com/ember-cli/ember-cli/blob/627934f91b2aa0e19b041fdb1b547873c1855793/lib/broccoli/default-packager.js#L994-L1013

ember-cli adds all these to `test-support.js`, not to `tests.js` file. Moving the code fixes issue #113


ef4:
Thanks, glad to have this fixed.

PR #432, opened 2020-04-20
Engines compatibility

### LINKED ISSUE #412: appTrees need to merge into the nearest engine
Right now we combine the appTrees from all active addons into the app's package namespace in stage2. That is how most ember apps work.

But engines are different. When you have an engine, all the addons beneath that engine need to merge their app trees into the engine's package namespace, not the app's.

 - we need to mark engines as always needing to be relocated, even if they happen to be natively v2, because we're going to be rewriting them regardless.
 - the `AppDiffer` either needs to manage multiple engines simultaneously, or we move to having recursive `AppDiffers` at each level.

ef4:
Fixes #412.

This implements the correct merging strategy for combining addons into engines. As far as I know, this completes all engine compatibility, with the caveat that lazy engines aren't automatically lazy yet, there is a small bit of remaining work to plug them into embroider's existing lazy route support (which itself was based on how lazy engines do it, so there should be a close match).


ef4:
An architectural note for those playing along: engines are addons, so they must go through stage1 to get compiled into v2 addons. But engines are also app-like in that at stage2, all an engine's addons' appTrees need to merge into the engine's module namespace. In order to handle this distinction, engines get build into specially renamed directories in stage1, which lets us reserve their real destination directories for the output of stage2. This is what engine-manger.ts is doing.

Issue #437, opened 2020-04-28
'original is not a function' with embroider's router


thoov:
When using embroider's router (`import EmberRouter from '@embroider/router';`) with a classical app the following error happens:

```
Uncaught TypeError: original is not a function
    at PrivateRouter.<anonymous> (index.js:83)
    at PrivateRouter.getRoute (index.js:83)
    at UnresolvedRouteInfoByParam.fetchRoute (router_js.js:826)
    at UnresolvedRouteInfoByParam.get route [as route] (router_js.js:738)
    at URLTransitionIntent.applyToState (router_js.js:1346)
    at PrivateRouter.getTransitionByIntent (router_js.js:1490)
    at PrivateRouter.transitionByIntent (router_js.js:1445)
    at PrivateRouter.doTransition (router_js.js:1579)
    at PrivateRouter.handleURL (router_js.js:2041)
    at Class._doURLTransition (router.js:419)
```

thoov:
This happens when the app is using ember-engines as that has router-ext which is modifying the router and is colliding with us.

PR #440, opened 2020-04-29
Enable embroider's router to be backwards compatible with ember-engines

thoov:
Fixes: #437 

TLDR: When using embroider's router in a classical build, ember-engines would cause issues as ember-engines is reopening the router. This does not cause a problem with an embroider build because we specially handle ember-engines/-private/router-ext.


packages/router/addon/index.js
```diff
@@ -28,6 +28,20 @@ function lazyBundle(routeName, engineInfoByRoute) {
...
+  // builds. Currently, this collides with ember-engines/-private/router-ext's
+  // overrides. To overcome this, during embroider builds we remove that file
+  // from being included as apart of the build (and thus will not reopen the class).
+  // However, during classical builds this is not possible so here we check if that
+  // module is included and if it is we know that we are in a classical build and
+  // that the router is being reopened. If it is being reopened then we can simply
+  // call into ember-engines router as we dont need to do anything.
+
+  // eslint-disable-next-line no-undef
+  return !!requirejs.entries['ember-engines/-private/router-ext'];
```
thoov, line 42: @ef4 This is the interesting piece of the whole PR. I have a couple of questions / thoughts about this for you.

1) Do you think this is the best approach for determining if ember-engines has reopened the router? I could also check the instance for things like `this._enginePromies` (which ember-engine adds) but that seems too brittle.

2) Should we simply noop all of our changes in the case that you are in classical mode?

ef4:
We don't need a new test-package to cover this -- we can run engines-host-app with the `CLASSIC` environment variable. I would rather do that anyway, because it ensures that the things we're exercising in that app all work the same in both kinds of builds, and I originally intended that we would keep the classic build in the test suite rather than replace it with only the embroider build.

For example:

```js
// test-packages/engines-host-app-classic.test.js
const execa = require('execa');

test('engines-host-app-classic', async () => {
  jest.setTimeout(120000);

  await execa('yarn', ['test'], {
    cwd: `${__dirname}/engines-host-app`,
    env: { CLASSIC: 'true' },
  });
});
```

ef4:
I think the way the test is written to look at a path containing `_app_` is incorrect. That path is an internal implementation detail of embroider.

PR #441, opened 2020-04-29
Add support for namespaced (aka index.{hbs,js,ts}) components

NullVoxPopuli:


packages/compat/tests/resolver.test.ts
```diff
@@ -95,6 +95,46 @@ describe('compat-resolver', function() {
     ]);
   });
 
+  describe('bare namespaced', function() {
+    test('dasherized component, js only', function() {
```
NullVoxPopuli, line 99: I wonder why a template is matched in this test?
```

      Array [
        Object {
          "path": "../components/hello-world/index.js",
          "runtimeName": "the-app/components/hello-world",
        },
    +   Object {
    +     "path": "../components/hello-world/index.js",
    +     "runtimeName": "the-app/templates/components/hello-world",
    +   },
      ]

      100 |       let findDependencies = configure({ staticComponents: true });
      101 |       givenFile('components/hello-world/index.js');
    > 102 |       expect(findDependencies('templates/application.hbs', `{{hello-world}}`)).toEqual([
          |                                                                                ^
      103 |         {
      104 |           path: '../components/hello-world/index.js',
      105 |           runtimeName: 'the-app/components/hello-world',

      at Object.<anonymous> (packages/compat/tests/resolver.test.ts:102:80)
```

packages/compat/tests/resolver.test.ts
```diff
@@ -95,6 +95,46 @@ describe('compat-resolver', function() {
...
+      let findDependencies = configure({ staticComponents: true });
+      givenFile('components/hello-world/index.js');
+      givenFile('components/hello-world/index.hbs');
+      expect(findDependencies('templates/application.hbs', `{{hello-world}}`)).toEqual([
+        {
+          path: '../components/hello-world/index.js',
+          runtimeName: 'the-app/components/hello-world',
+        },
+        {
+          path: '../components/hello-world/index.hbs',
```
ef4, line 131: I think this might be incorrect.  Co-located templates don't need to be resolvable. 

When you have co-located js and hbs, the js file get rewritten such that it directly imports the hbs. That means consumers of the component only depend directly on the js, they don't need to worry about the hbs.

Similarly, when you have only hbs and no js (inside `app/component` or `addon/components`), we synthesize the `templateOnlyComponent` javascript file and carry on as before. In this case also, consumers of the component only depend directly on the js, not the hbs.

ef4: So I think your change that adds a check for `app/components/foo/index.{js,ts}` is good. But you can probably remove the check for `app/components/foo/index.hbs` and it will work just the same.

PR #445, opened 2020-05-04
Support nested in-repo addons

thoov:
When initial traversing the list of dependencies of an application we were simply casting them as Packages. This results is an incomplete view of all dependencies that need to be moved. Here we detect if the package is actually a Package or an ExtendedPackage and cast it as such.


thoov:
@ef4 I have tested this on our internal app and things look good. There are 2 parts of interest in this PR:

1. Moving extended-package from compat to core. I did this to colocate next to package and as it harder to import compat stuff into core than it is the other way around.

2. The changes in package-cache. The only thing that doesn't feel right is that we do this path check in a few places now and I'm not sure if there is a cleaner way.

ef4:
Closing in favor of #447. I checked that the test case in this PR also passes under that PR.

PR #452, opened 2020-05-13
transpile syntax that webpack doesn't support yet

ef4:
Webpack doesn't support optional chaining or nullish coalescing yet (because acorn doesn't).

Depending on your preset-env, you might leave those untranspiled (modern browsers do have native implementations already).

Therefore, in embroider/webpack we need to always make sure we transpile that syntax so webpack will not explode.

### DIFF
--- a/packages/webpack/package.json
+++ b/packages/webpack/package.json
@@ -25,6 +25,8 @@
   },
   "dependencies": {
     "@babel/core": "^7.8.7",
+    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.8.3",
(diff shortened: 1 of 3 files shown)


josemarluedke:
FYI, this breaks apps using `ember-cli-typescript` as it adds `@babel/plugin-proposal-optional-chaining` (https://github.com/typed-ember/ember-cli-typescript/blob/master/package.json#L42-L43).

ef4:
Thanks, you're right, it has been reported in #453 as well.

PR #454, opened 2020-05-18
Invalidate in-repo-addons based on ember-cli's logic

thoov:
Ember-cli will reject or invalidate in repo addons that:
 - do not exist
 - have no package.json
 - have a malformed package.json
 - do not have a main file

However, instead of an error ember-cli will simply print a warning to the console and continue as if that in-repo addon was not specified. This PR simply removes them from being considered as the warning will already be on the screen. More background info about [how ember-cli does the check](https://github.com/ember-cli/ember-cli/blob/627934f91b2aa0e19b041fdb1b547873c1855793/lib/models/package-info-cache/index.js#L427).

### DIFF
--- a/packages/core/src/package.ts
+++ b/packages/core/src/package.ts
@@ -1,6 +1,6 @@
 import { Memoize } from 'typescript-memoize';
-import { readFileSync } from 'fs';
-import { join } from 'path';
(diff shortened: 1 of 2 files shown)


ef4:
Thanks.

I tweaked this slightly so that instead of directly loading the package.json, we do it through the packageCache so that we won't do it twice.

PR #457, opened 2020-05-21
updates to match the new v5 fastboot schema

ef4:


ef4:
As part of this change, I refactored so that the stage2 output is already in terms of the app's configured `rootURL`. Stage3 was already expected to understand `rootURL`, but now instead of needing to move all URLs into that path, they come already in that path. This works better for things (like `<fastboot-scripts>`) that stage3 doesn't necessarily need to know about or touch.

Issue #463, opened 2020-05-28
Adjust default-pipeline.ts for fastboot production builds


ef4:
In dev and test builds, `@embroider/macros` are left in place to runtime, which lets us produce a single build that will work in multiple contexts (including browser vs fastboot). But in production, `@embroider/macros` are resolved at build time to produce optimized output.  See https://github.com/embroider-build/embroider/blob/e35bafa1efca396a65b395c08925a043e31c1a49/packages/core/src/app.ts#L737

So in production builds when an app uses fastboot, to get valid output for both fastboot and browser we need to run stage3 twice, with different settings.

There are a couple possibilities for what we do with both outputs, which really comes down to what's more convenient for people's deployment situations. 

Option 1: We could put each build in a different subdirectory, so you'd get a shape like:

```
dist
├── browser
│   ├── assets
│   │   └── chunk-123.js
│   └── index.html
└── fastboot
    ├── assets
    │   └── chunk-456.js
    └── index.html
```

Or Option 2: we could use the `data-fastboot-src` attributes in index.html to combine both into one build, so that where there's different build output between the two environments the HTML would say:

```
<script src="/assets/chunk-123.js" data-fastboot-src="/assets/chunk-456.js"></script>
```

Option 2 is probably better if you're *always* serving your HTML from fastboot. Option 1 may be better if you only use fastboot for some things, and still want pristine browser-facing HTML the rest of the time, without relying on a fastboot server to strip out the fastboot-only metadata. Option 1 makes it easier to integrate other stage3 packagers because they can just emit entire builds into separate directories without learning about fastboot.

rwjblue:
@ef4 - Just curious, do you have a general picture of the work/effort for both options?

rwjblue:
Does Option 1 set us up for other future features a bit better (e.g. separate builds by browser)? On the surface, it _seems_ like it might...

rwjblue:
For my part, I'd personally prefer the first option (assuming there is not a massive difference in effort required).

ef4:
The first option is the easier one, I think. But yes, I think we can make sure this is a general purpose matrix-like feature that would let you do multiple flavors of builds for reasons other than fastboot.

ef4:
It turns out that it's not super useful, in the usual case, to do completely separated fastboot and browser builds, because you want the HTML that the fastboot app emits to the browser to contain URLs to assets from the browser build.

[...]

PR #465, opened 2020-06-02
Build variants

ef4:
This creates an API for third-stage packagers to support multiple simultaneous build variants. The immediate motivation is to produce fastboot- and browser-optimized production builds. It should also be able to grow to allow the targeting of different sets of browsers.


ef4:
Closes https://github.com/embroider-build/embroider/issues/463

PR #472, opened 2020-06-14
Identify component in snippet rule when ast type is "Template"

GCheung55:
ember-source@3.17.0 introduced a change to `ember-source/vendor/ember/ember-template-compiler` that is parsing `{{foo-bar}}` AST type as a "Template".

### DIFF
--- a/packages/compat/src/dependency-rules.ts
+++ b/packages/compat/src/dependency-rules.ts
@@ -116,7 +116,7 @@ export type ArgumentMapping =
     };
 
 // A component snippet is a string containing valid HBS that is a single
-// component invocation. We use it to refer to compoanents in a way that doesn't
+// component invocation. We use it to refer to components in a way that doesn't
 // require any new syntax or rules, and that's necessarily supported by whatever
 // build-time template resolver is in use.
 //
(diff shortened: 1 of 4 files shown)


GCheung55:
Failure when installing dependencies. How do I try the build again?

```
2/4] Fetching packages...
error An unexpected error occurred: "https://registry.yarnpkg.com/jsonfile/-/jsonfile-4.0.0.tgz: ENOENT: no such file or directory, open '/home/runner/.cache/yarn/v4/npm-jsonfile-4.0.0-8771aae0799b64076b76640fca058f9c10e33ecb/node_modules/jsonfile/.yarn-tarball.tgz'".
```

GCheung55:
I've updated the PR with the fix, but to be sure we're on the same page, https://github.com/embroider-build/embroider/runs/774461248 shows the failure after updating `test-packages/support/package.json` with `ember-source@v3.17.0`.

ef4:
Thanks, nice work.

I wondered at first whether we should keep separate test coverage for the older ember-source version, but I think we can skip that for the present. Once there's a stable 1.0 of Embroider, from that point forward we can test a matrix of ember versions but tracking latest stable ember is sufficient right now.

PR #475, opened 2020-06-16
fix renamed-modules when used with languages that transpile to js

ef4:
The `rewriteAddonTree` transformation captures any modules that are trying to escape their own package namespace and records their names as our `renamed-modules` in the addon metadata. But if you are using typescript or another preprocessor that changes the file extensions, we currently save the renamed-modules and then transpile, which results in the names in the metadata not matching the actual files.

This switches the order so we preprocessor first and then record the renamed-modules.


ef4:
(I need to come back around and add test coverage here.)

PR #478, opened 2020-06-19
moduleName must include the extension in precompile options

SparshithNR:
moduleName in template.__meta previously included an extension of the compiled file.
Embroider build stripped it off during the build process. Adding it back again.


ef4:
Thanks. I wasn't sure at first whether this should move upstream into absPathToRuntimeName, but after reviewing I think you are correct to put it here. It's only the template compiler that wants to see the extension, all other users of that function don't.

PR #485, opened 2020-06-25
more robust customized tree detection

ef4:
There are addons in the wild that directly patch the methods on other addon's instances. We don't currently catch those as customized. This hopefully does.


ef4:
Ugh. ember-engines is apparently one of the addons that uses this behavior, so now we have started detecting custom behavior there that we were missing before. It's bad custom behavior we don't want, so we probably need another way to patch that out.

PR #492, opened 2020-07-10
Remove compat adapter for ember-window-mock

simonihmig:
The new release v0.7.0 should be compatible now, through https://github.com/kaliber5/ember-window-mock/pull/191


Turbo87:
I'm seeing:

> message: Module not found: Error: Can't resolve '../../../ember-window-mock' in '/tmp/embroider/6bfa8c/node_modules/ember-window-mock/test-support/-private'

with the new embroider release and ember-window-mock v0.7.1 🤔

simonihmig:
Hm, strange. I have a test scenario for embroider (that ran without the compat adapter), not sure why this pops up now. Will have to investigate, thanks for reporting!

simonihmig:
@Turbo87 I tested this again, once in the addon itself using its embroider test scenario, and once by manually creating a minimal app, using e-w-m both in app code as well as test code, which all worked well! 🤔 

If you have something reproducible, then please file an issue at e-w-m's repo!

PR #507, opened 2020-08-13
Use runtime loader for tests

ef4:
We statically import all tests into the tests entrypoint. Which makes them invisible to ember-cli-test-loader, which shouldn't actually matter since we're ensuring they get evaluated directly.

But in practice, there is a [hack in ember-cli-test-loader](https://github.com/ember-cli/ember-cli-test-loader/blob/d24046ea834f8d9b5b3e2f0d54f131e83eaed0b3/addon-test-support/index.js#L62-L63) that guards against some kind of memory problem, and without that it seems that embroider may have worse memory consumption patterns in huge test suites.

We can use the more traditional pattern of registering the tests with the runtime loader and then letting ember-cli-test-loader find and evaluate them, though this PR may not be sufficient by itself, because webpack's runtime loader is also going to retain these modules.

Actually unloading a module from memory violates the ES module spec, so it's not really a thing that we should be relying on. (The spec says you have to guarantee that a module will only run once, and if you unload a module that would allow it to run a second time if requested again.) IMO, this `require.unsee` thing is probably papering over deeper issues in some test suites.

### DIFF
--- a/packages/core/src/app.ts
+++ b/packages/core/src/app.ts
@@ -1117,29 +1117,30 @@ export class AppBuilder<TreeNames> {
     // the app.
     let engine = engines[0];
 
-    let { appFiles } = engine;
     const myName = 'assets/test.js';
-    let testModules = appFiles.tests
-      .map(relativePath => {
-        return `../${relativePath}`;
-      })
-      .filter(Boolean) as string[];
 
     // tests necessarily also include the app. This is where we account for
     // that. The classic solution was to always include the app's separate
     // script tag in the tests HTML, but that isn't as easy for final stage
     // packagers to understand. It's better to express it here as a direct
     // module dependency.
-    testModules.unshift(explicitRelative(dirname(myName), this.topAppJSAsset(engines, prepared).relativePath));
+    let eagerModules: string[] = [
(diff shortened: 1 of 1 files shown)


stefanpenner:
> thing is probably papering over deeper issues in some test suites.

For context: 

The original motivation was not due to memory leaks in test-suites or tools but for very large test suites, simply loading + retaining the code related to thousands or tens of thousands of large test files alone created hefty memory pressure.

It also happened to paper over improper cleanup of module scope'd state, and test infra. Although we historically did go and fix all our test runners repeatedly to fix these leaks, it wouldn't surprise me if something has once again regressed (and should be investigated).

Some future-proof strategy for running large test suites of large apps will be important. But likely should be accomplished by test partitioning, something similar to what ember-exam does (but without loading all test files upfront), which has additional benefits such as parallelization etc.

---

> Actually unloading a module from memory violates the ES module spec,

This is true, although the constraint is specifically per realm, or per loader context. So if the realm, or loader context is discarded entirely, the imported module is also discarded. This doesn't violate your above rule, but may enable an additional mechanism to implement a spec inspired solution to the "memory pressure" described above. Although it's unclear of its value, relative to a good parallel test runner / good partitioning.

I know a holistic approach to this problem will result in a much better solution. I'm largely sharing the above as context, hopefully it is of some value.

PR #514, opened 2020-09-10
support {{component "x"}} snippet

patricklx:


packages/compat/src/resolver.ts
```diff
@@ -223,6 +223,9 @@ export default class CompatResolver implements Resolver {
       throw new Error(`unable to parse component snippet "${snippet}" from rule ${JSON.stringify(rule, null, 2)}`);
     }
     if ((ast.type === 'Program' || ast.type === 'Template') && ast.body.length > 0) {
+      if (first.type === 'MustacheStatement' && first.path.type === 'PathExpression' && first.path.original === 'component') {
+        return first.params[0].value;
```
lifeart, line 227: I think we have to check for `params[0].type === "StringLiteral"`;

packages/compat/tests/resolver.test.ts
```diff
@@ -321,6 +321,23 @@ describe('compat-resolver', function() {
...
+
+  test('component rules can be expressed via component helper', function() {
+    let findDependencies = configure({
+      staticComponents: true,
+      staticHelpers: true,
+      packageRules: [
+        {
+          package: 'the-test-package',
+          components: {
+            '{{component "this-one"}}': { safeToIgnore: true },
```
lifeart, line 333: @patricklx, @ef4 is it make sence to add support for case like:

```hbs
{{#let (component "foo-bar") as |MyComponent|}}
   <MyComponent />
{{/let}}
```

```hbs
{{#let (hash item=(component "my-component")) as |ctx|}}
   <ctx.item />
{{/let}}
```

ef4: What is the use case here?

I think this snippet parsing code is only used when processing `packageRules`. I don't think there's a reason for someone to write their packageRule that way.

lifeart: I have usecase from one of my production projects (long component names, simplified using such notation)

image

lifeart: same for eagier component aruments setup (to cleanup view)

image

lifeart: @ef4 

Let's imagine we have to use component with 5+ arguments, and we have to use it few times on a page, with few different args.

```hbs
<BubbleRenderer @target="foo" @name="foo" @type="zoo" @field="key1" />
```

we could write is as this:

```hbs
<BubbleRenderer @target="foo" @name="foo" @type="zoo" @field="key1" />
<BubbleRenderer @target="foo" @name="foo" @type="zoo" @field="key2" />
<BubbleRenderer @target="foo" @name="foo" @type="zoo" @field="key3" />
<BubbleRenderer @target="foo" @name="foo" @type="zoo" @field="key3" />
<BubbleRenderer @target="foo" @name="foo" @type="zoo" @field="key5" />
```

or

```hbs
{{#let (component 'bubble-renderer' target="foo" name="foo" type="zoo") as |Bubble|}}
	<Bubble @field="key1" />
	<Bubble @field="key2" />
	<Bubble @field="key3" />
	<Bubble @field="key4" />
	<Bubble @field="key5" />
{{/let}}
```

this is 10 times easier to read and maintain

ef4: Sure, I understand why you'd use these patterns *in your templates*. But that's not what this code is for. The snippets in packageRules are only used to resolve a component at build time in order to identify which component you're making rules about.

If you're saying you want to make sure embroider knows some packageRules for `Bubble` in an example like this:

```hbs
{{#let (component 'bubble-renderer' target="foo" name="foo" type="zoo") as |Bubble|}}
	<Bubble @field="key1" />
	<Bubble @field="key2" />
	<Bubble @field="key3" />
	<Bubble @field="key4" />
	<Bubble @field="key5" />
{{/let}}
```

that's a good idea, but it doesn't involve writing any packageRules differently. You'd still make a rule for the snippet  `<BubbleRenderer/>`. The change that would be needed (if this isn't already implemented, I can't remember) is to make the template analysis smart enough to notice that `<Bubble/>` is really invoking `<BubbleRenderer/>` and therefore those are the rules to use.

ef4: I accepted the original PR to support `{{component "x"}}` in snippets only because I can imagine there are cases where some component names are weird and easier to express in that form. But don't be mislead into thinking that a rule written against `{{component "x"}}`  will only match an invocation like `{{component "x"}}`.  It will also match`<X/>` or `{{#x}} {{/x}}` too. Those syntactic differences don't matter to the rules system. It cares about which module on disk we're talking about, not how you spell it.

lifeart: Thank you for answers and clarifications! :)

PR #517, opened 2020-09-12
Sort package.json contents

Turbo87:
see https://github.com/keithamus/sort-package-json

This is also used by ember-cli internally for sorting the contents.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,11 +8,20 @@
     "types/@babel__traverse"
   ],
   "scripts": {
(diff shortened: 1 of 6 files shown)


ef4:
Is this going to give us more conflicts when trying to run ember-cli-update?

Turbo87:
not that I'm aware of. afaik ecupdate is using a special JSON merging algo to avoid such issues. and I think they use this sorting order too.

PR #521, opened 2020-09-13
Support static apps using @embroider/macros

### LINKED ISSUE #476: Build error with @embroider/macros and static helpers
When adding `staticHelpers: true` to the embroider config for `ember-bootstrap` (which uses `@embroider/macros` as a v1 addon), as suggested in the addon guide, I get this build failure:

```
Error: Missing helper macroCondition in /private/var/folders/q3/yzggf_l965zbzps_bk8xr37c0000gn/T/embroider/9d4dc2/templates/components/bs-accordion.hbs
    at CompatResolver.dependenciesOf (/Users/simonihmig/Projects/ember-bootstrap/node_modules/@embroider/compat/src/resolver.js:184:31)
    at TemplateCompiler.precompile (/Users/simonihmig/Projects/ember-bootstrap/node_modules/@embroider/core/src/template-compiler.js:246:49)
    at TemplateCompiler.compile (/Users/simonihmig/Projects/ember-bootstrap/node_modules/@embroider/core/src/template-compiler.js:255:47)
    at Object.hbsLoader (/Users/simonihmig/Projects/ember-bootstrap/node_modules/@embroider/webpack/src/webpack-hbs-loader.js:10:16)
```

See [this PR](https://github.com/kaliber5/ember-bootstrap/pull/1125) and its [build error](https://github.com/kaliber5/ember-bootstrap/pull/1125/checks?check_run_id=778459435#step:5:69)

simonihmig:
Spike to fix #476.

The first commit adds a test, that fails in the way described in #476.

The second seems to fix the problem by ignoring the embroider macros in the resolver just like Ember's built-ins. Though as stated in https://github.com/embroider-build/embroider/issues/476#issuecomment-691524287, I am not sure if that's the preferred way to do it...

### DIFF
--- a/packages/core/src/template-compiler.ts
+++ b/packages/core/src/template-compiler.ts
@@ -285,19 +285,20 @@ export default class TemplateCompiler {
     }
 
     let opts = this.syntax.defaultOptions({ contents, moduleName });
-    if (this.params.resolver) {
-      let transform = this.params.resolver.astTransformer(this);
-      if (transform) {
-        this.params.plugins.ast!.push(transform);
-      }
-    }
-    opts.plugins!.ast = [...this.getReversedASTPlugins(this.params.plugins.ast!), ...opts.plugins!.ast!];
+    let plugins: Plugins = {
+      ...opts.plugins,
+      ast: [
+        ...this.getReversedASTPlugins(this.params.plugins.ast!),
+        this.params.resolver && this.params.resolver.astTransformer(this),
+        ...opts.plugins!.ast!,
+      ].filter(Boolean),
+    };
 
     let compiled = this.syntax.precompile(stripBom(contents), {
       contents,
       moduleName: runtimeName,
       filename: moduleName,
-      plugins: opts.plugins,
+      plugins,
     });
 
     if (this.params.resolver) {
--- a/test-packages/static-app/app/router.js
+++ b/test-packages/static-app/app/router.js
@@ -10,6 +10,7 @@ Router.map(function() {
   this.route('helpers-example');
   this.route('components-example');
   this.route('static-component-rules-example');
+  this.route('macros-example');
 });
 
 export default Router;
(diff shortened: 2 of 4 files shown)


ef4:
Thanks for working on this.

Instead of making the resolver ignore the macros, I would like to compile away the macros before the resolver sees them.

Within the final babel config (the one assembled in `packages/core/src/app.ts`), both `packages/macros/src/glimmer/ast-transform.ts` and `packages/compat/src/resolver-transform.ts` are running. I think if we can ensure that the macros one runs before the resolver one, it will fix this issue.

simonihmig:
I hope I found the correct place to fix this, according to your suggestion. The template-compiler seems to put all plugins together, which is where I tweaked the order. This made the added test pass. Hope that it also didn't cause any regressions. Running all tests locally is often failing due to timeouts, and CI is currently broken here as I just rebased this to the latest master, which is also broken due to #526. Will rebase again once master is green again...

simonihmig:
> Will rebase again once master is green again...

All green! 😅 
@ef4 please re-review!

packages/core/src/template-compiler.ts
```diff
@@ -285,19 +285,20 @@ export default class TemplateCompiler {
...
-    }
-    opts.plugins!.ast = [...this.getReversedASTPlugins(this.params.plugins.ast!), ...opts.plugins!.ast!];
+    let plugins: Plugins = {
+      ...opts.plugins,
+      ast: [
+        ...this.getReversedASTPlugins(this.params.plugins.ast!),
+        this.params.resolver && this.params.resolver.astTransformer(this),
+        ...opts.plugins!.ast!,
+      ].filter(Boolean),
+    };
```
simonihmig, line 295: I chose to refactor this a bit, as the previous implementation had a side-effect of mutating `opts.plugins.ast` and `params.plugins.ast`, which seems undesirable.

ef4:
Thanks!

PR #526, opened 2020-09-15
Fix broken style-loader dependency

simonihmig:
Seems the latest release commit (0.23.0 to 0.24.0) also accidentally bumped the version of this package, which coincidentally also had a version of 0.23.0, to a non-existing version of 0.24.0. (Is this a manual find/replace type operation, or automated?)

This broke CI on master. And will certainly break when folks install the latest release, so a quick follow up release of 0.24.1 seems required! @ef4


thoov:
👍

ef4:
Oops!

Yeah, it's a search and replace. I manually review to look for this kind of problem but missed this one.

PR #529, opened 2020-09-16
Fix for addons using ember-cli-typescript 4x

### LINKED ISSUE #528: Addons using ember-cli-typescript@4 do not get their TypeScript transpiled
We discovered this while testing a new version of ember-a11y-testing that uses `ember-cli-typescript@4`. During the Embroider build the `.ts` files (located in `addon-test-support`) are not transpiled into `.js` in Embroider's working directory.

I'm making a tracking issue here so we can work towards a solution...

jamescdavis:
Fixes #528

We need to trigger a custom Babel run for any addons using ember-cli-typescript 4.

### DIFF
--- a/packages/compat/src/v1-addon.ts
+++ b/packages/compat/src/v1-addon.ts
@@ -199,6 +199,18 @@ export default class V1Addon {
   }
 
   private needsCustomBabel() {
+    if (
+      this.addonInstance.addons.find(
+        (a: any) =>
+          a.name === 'ember-cli-typescript' &&
+          semver.satisfies(semver.coerce(a.pkg.version) || a.pkg.version, '>=4.0.0-alpha.1')
+      )
+    ) {
+      // This addon is using ember-cli-typescript 4, which relies on
+      // ember-cli-babel to add the TypeScript transform Babel plugin.
+      return true;
+    }
+
     let babelConfig = this.options.babel as TransformOptions | undefined;
     if (babelConfig && babelConfig.plugins && babelConfig.plugins.length > 0) {
       // this addon has custom babel plugins, so we need to run them here in


packages/compat/src/v1-addon.ts
```diff
@@ -429,6 +444,10 @@ export default class V1Addon {
     if (this.templateCompiler) {
       babelConfig.plugins.push(this.templateCompiler.inlineTransformsBabelPlugin());
     }
+
+    if (this.needsTypeScriptTransform) {
+      babelConfig.plugins.push(require.resolve('@babel/plugin-transform-typescript'));
```
ef4, line 449: I didn't think this would be necessary, because ember-cli-babel adds the typescript plugin when it sees that ember-cli-typescript 4x is present?

Your fix above in `needsCustomBabel` should ensure that we allow ember-cli-babel to do its usual thing (other than the changes we make to its own settings like `compileModules: false`).

If this line is really needed to make it work, we should debug why because then I don't understand it.

ef4: Also, if we do keep this line, we should probably guard against double insertion of the plugin. If anybody else has already inserted it, babel will throw when it sees the duplicate.

For an example of where we already do this, see https://github.com/embroider-build/embroider/blob/3d48872ece01cedb9dd81da754dc5d75a9052b21/packages/webpack/src/ember-webpack.ts#L500-L502

ef4: Also, this is resolving the plugin relative to `@embroider/compat`, which doesn't have a dependency on `@babel/plugin-transform-typescript`. It needs to resolve relative to `ember-cli-babel`'s root, because AFAIK that is the package that depends on the plugin. But again, this is one more reason I think it's suspicious to add the plugin here -- I think ember-cli-babel should probably be doing it.

ef4 (changes requested):
Thanks for working on this!

packages/compat/src/v1-addon.ts
jamescdavis: You're totally right. All it needed was https://github.com/embroider-build/embroider/pull/529/commits/701009a54b39500fcffaac22bd142d21753218eb!

I'm not sure how I ended up down the rabbit hole where I thought I needed to have Embroider add the plugin. ¯\\\_(ツ)\_/¯

ef4:
This is marked draft still but I think it's ready?

jamescdavis:
@ef4 I've been trying to add an app that repros the issue to `test-packages`, but having some issues getting it to properly fail without this fix. I *was* able to basically airlift https://github.com/scalvert/a11y-embroider-repro/tree/broken and it fails the build without the fix (with the same issue seen in that repro) and builds correctly with the fix, but the tests end up failing due to other issues (can't find `@ember/destroyable`). I'll go ahead and take this out of draft and work on getting a proper test-case in later.

ef4:
I'm hesitant to keep adding separate test apps anyway. It's a lot to maintain and it's slow. People keep doing it because it's more obvious, but I'd encourage more tests that look like these:

https://github.com/embroider-build/embroider/blob/master/packages/compat/tests/app-import.test.ts
https://github.com/embroider-build/embroider/blob/master/packages/compat/tests/addon-styles.test.ts

ef4 (approved):

jamescdavis:
Yeah, it's *very* slow. Those tests are a much better strategy.

Issue #530, opened 2020-09-16
Avoiding `Testem.hookIntoTestFramework()` invocation with `ember-qunit@5.0.0-beta.3`+


rwjblue:
As ember-qunit is progressing towards its 5.0.0 release, I'd like to ensure that it is compatible with Embroider. The goal of that release is to play nicely with the goals of Embroider. Some of the major changes there:

* Move `qunit` and `@ember/test-helpers` to be `peerDependencies` (along with validation that the app provides them)
* Remove usage of `content-for` to emit the DOM structure required for `tests/index.html` (instead have it included in the file normally)
* Avoid `app.import`ing **any** JavaScript assets (including `qunit` itself and our custom configuration)

Part of the issue that we run into with the last item above is that `qunit` is no longer evaluated directly in the root of the `test-support.js` file, it is now in a module (by way of either ember-auto-import or Embroider magic). Unfortunately, both ember-cli and Embroider have baked in the assumption that by the end of `test-support.js` `window.QUnit` will be defined. This is no longer true as of `ember-qunit@5.0.0-beta.3`.

Specifically, the issue is in this Embroider code:

https://github.com/embroider-build/embroider/blob/38a4030a8490db8211370c45fc8184f7d9645d6a/packages/core/src/app.ts#L256-L258

[`223e0cf` (#752)](https://github.com/emberjs/ember-qunit/pull/752/commits/223e0cf3cad9ec3e700125e81da40c3dbd588c67) is specifically working around this issue in a classic ember-cli build pipeline (by emitting a `vendor/ember-cli/test-support-suffix.js` that no longer contains the `Testem.hookIntoTestFramework()` invocation which clobbers the one provided by ember-cli itself). Obviously that work around doesn't work here (and its clearly haxxxors anyways).

rwjblue:
https://github.com/embroider-build/embroider/pull/531 seems like the simplest fix (and one that I can land in ember-cli itself also), what do you think @ef4?

PR #531, opened 2020-09-16
Avoid `Testem.hookIntoTestFramework` unless `QUnit` or `Mocha` are defined.

rwjblue:
Fixes https://github.com/embroider-build/embroider/issues/530

### DIFF
--- a/packages/core/src/app.ts
+++ b/packages/core/src/app.ts
@@ -253,8 +253,8 @@ export class AppBuilder<TreeNames> {
         relativePath: '_testing_suffix_.js',
         source: `
         var runningTests=true;
-        if (window.Testem) {
-          window.Testem.hookIntoTestFramework();
+        if (typeof Testem !== 'undefined' && (typeof QUnit !== 'undefined' || typeof Mocha !== 'undefined')) {
+          Testem.hookIntoTestFramework();
         }`,
       });
 


packages/core/src/app.ts
```diff
@@ -253,8 +253,8 @@ export class AppBuilder<TreeNames> {
         relativePath: '_testing_suffix_.js',
         source: `
         var runningTests=true;
-        if (window.Testem) {
-          window.Testem.hookIntoTestFramework();
+        if (typeof Testem !== 'undefined' && (typeof QUnit !== 'undefined' || typeof Mocha !== 'undefined')) {
+          Testem.hookIntoTestFramework();
         }`,
```
rwjblue, line 258: FWIW, I think this is transitionary. Once I can roll out ember-qunit@5.0.0 more broadly, we can remove this completely (and release a breaking change release).

rwjblue:
I've confirmed (by locally editing `node_modules/@embroider/core/src/app.js`) that https://github.com/emberjs/ember-qunit/pull/756 passes with these changes.

ef4:
Thanks.

If QUnit or Mocha is not defined at this point, where does the `hookIntoTestFramework()` thing happen instead?

rwjblue:
> If QUnit or Mocha is not defined at this point, where does the hookIntoTestFramework() thing happen instead?

`ember-qunit` invokes it when it knows for sure that QUnit is defined (because it itself has imported it):

https://github.com/emberjs/ember-qunit/blob/32c4cee9dd266bf36f76274d33c713306feada61/addon-test-support/index.js#L8-L10

(Added in https://github.com/emberjs/ember-qunit/pull/752/commits/223e0cf3cad9ec3e700125e81da40c3dbd588c67)

PR #534, opened 2020-09-16
Update TypeScript to latest 4.0

simonihmig:


.eslintrc.js
```diff
@@ -23,11 +22,18 @@ module.exports = {
     '@typescript-eslint/class-name-casing': 'error',
     '@typescript-eslint/no-angle-bracket-type-assertion': 'error',
     '@typescript-eslint/no-inferrable-types': 'error',
-    '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
+    // @todo re-enable when ESLint has been updated
+    // '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
```
simonihmig, line 26: I had to update `@typescript-eslint/parser`, but didn't want to mess with all the other ESLint dependencies in the same PR, as changed rules could have caused a lot of churn. Unfortunately it seems this rule freaks out now on unused variables in `types`, which uh, makes no sense.

I would try to update the ESLint stuff later in a separate effort, I expect that we can activate this rule then again!

packages/compat/src/options.ts
```diff
@@ -92,6 +92,7 @@ const defaults = Object.assign(coreWithDefaults(), {
   extraPublicTrees: [],
   workspaceDir: null,
   optionalComponents: [],
+  packageRules: [],
```
ef4, line 95: There was a real bug here that the new type error was revealing.

packages/core/src/build-stage.ts
```diff
@@ -35,7 +35,7 @@ export default class BuildStage<NamedTrees> implements Stage {
         this.packageCache = packageCache;
         this.active = await this.instantiate(outputPath, this.prevStage.inputPath, packageCache);
       }
-      delete treePaths.__prevStageTree;
+      delete (treePaths as any).__prevStageTree;
```
ef4, line 38: Here we are intentionally changing a thing of one type into another type.

The types were both correct, we don't want `__prevStageTree` to be optional. Instead, we need to escape the type system just during the middle of the conversion between types. Everybody outside these functions will only see consistent types.

packages/core/src/portable-plugin-config.ts
```diff
@@ -145,8 +145,9 @@ export class PortablePluginConfig {
             return buildBroccoli(placeholder);
           case 'htmlbars-parallel':
             return buildHTMLBars(placeholder);
+          default:
+            assertNever(placeholder);
```
ef4, line 149: Re-introducing this safety check. It's intended to error if a new `placeholder.type` ever gets added without adding it to this switch.

simonihmig: Yeah, I understood that, but was not able to silence TS complaining about unreachable code. I wonder why TS treats this call in the `default` case differently than below the switch statement, as semantically there are the same. 🤔

packages/core/src/build-stage.ts
simonihmig: :+1:

packages/compat/src/options.ts
simonihmig: Ah, good catch. I didn't see that, and was misinterpreting TS complaining that `options` would allow to pass something like `{ packageRules: undefined, ... }`. In fact this is possible now, and while it satisfies all type constraints, it would lead to a wrong result (returning a partial `Options`), as `Object.assign()` would ignore missing properties, but would still override the default ones with `undefined` in my example above. AFAIK TS is not able to distinguish between an object with a missing property vs. an object with a property value of `undefined`...

PR #535, opened 2020-09-18
Update ESLint dependencies and rules

simonihmig:
Following up on #534, this updates all ESLint deps to their latest version, and re-enables `@typescript-eslint/no-unused-vars`. Some of the old `@typescript-eslint` rules have been removed, so I updated them to their (more or less) equivalent new counterparts (see inline comment).

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,12 +25,12 @@
   },
   "devDependencies": {
     "@types/jest": "^24.0.11",
-    "@typescript-eslint/eslint-plugin": "^1.5.0",
+    "@typescript-eslint/eslint-plugin": "^4.1.1",
     "@typescript-eslint/parser": "^4.1.1",
-    "eslint": "^5.15.3",
-    "eslint-config-prettier": "^4.1.0",
-    "eslint-plugin-import": "^2.16.0",
-    "eslint-plugin-prettier": "^3.0.1",
+    "eslint": "^7.9.0",
+    "eslint-config-prettier": "^6.11.0",
+    "eslint-plugin-import": "^2.22.0",
+    "eslint-plugin-prettier": "^3.1.4",
     "execa": "^1.0.0",
     "jest": "^24.5.0",
     "prettier": "1.16.4"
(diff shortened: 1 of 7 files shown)


packages/compat/src/v1-app.ts
```diff
@@ -23,18 +23,18 @@ import { isEmberAutoImportDynamic } from './detect-ember-auto-import';
 // This controls and types the interface between our new world and the classic
 // v1 app instance.
 
-type filePath = string;
-type OutputFileToInputFileMap = { [filePath: string]: filePath[] };
+type FilePath = string;
+type OutputFileToInputFileMap = { [filePath: string]: FilePath[] };
```
simonihmig, line 27: Under the new `@typescript-eslint/naming-convention` rule settings, PascalCase is required for all so called `typeLike` things, including type aliases. We could fine tune this further to only include classes and interfaces as it was before, but this seems more consistent to me, and matches the default config of that rule, and needed changes only in these two files.

simonihmig:
Speaking of rules, I did not even try to enable new rules, but just wanted to more or less keep the existing rules in place, with the updated libraries. 

However I would propose - in separate PRs - to extend the rules config to eventually include all of `eslint:recommended` and `plugin:@typescript-eslint/recommended`. This is probably unfeasible in a single effort, due to the vast code changes needed all over the place, so maybe enabling on rule after the other in separate PRs until we basically land at the recommended settings?

Thoughts?

package.json
```diff
@@ -25,12 +25,12 @@
...
   "devDependencies": {
     "@types/jest": "^24.0.11",
-    "@typescript-eslint/eslint-plugin": "^1.5.0",
+    "@typescript-eslint/eslint-plugin": "^4.1.1",
     "@typescript-eslint/parser": "^4.1.1",
-    "eslint": "^5.15.3",
-    "eslint-config-prettier": "^4.1.0",
-    "eslint-plugin-import": "^2.16.0",
-    "eslint-plugin-prettier": "^3.0.1",
+    "eslint": "^7.9.0",
```
rwjblue, line 30: With this bump you can remove the `--ext` flag to `yarn lint` script. Since we have an override for `**/*.ts`, all `.ts` files will automatically get linted (this was one of the nicer changes in 7.0.0 IMHO).

simonihmig: Oh, good to know. Removed that!

Reminds me of my early TS endeavors, being glad that linting immediately passes without any errors, just to realize later that it was not even running on .ts files! 😝

ef4:
Thanks!

PR #536, opened 2020-09-18
Add compatibility for ember-element-helper

simonihmig:
It does some dynamic component magic that Embroider is not able to understand. Specifically it is returning the component name (`-dynamic-element` or `-dynamic-element-alt`) from the [`-element` helper](https://github.com/tildeio/ember-element-helper/blob/main/addon/helpers/-element.js), that is passed to `{{component}}`, all after its own AST transformation. This rule should make Embroider aware of the dependencies, to pull those two components in.

I tested this rule successfully in a local Embroider-build addon (adding it to the Embroider setup in `ember-cli-build.js`). I didn't see test coverage for other compat adapters here, so do we need tests for this here?

/cc @chancancode

### DIFF
--- a/packages/compat/src/addon-dependency-rules/ember-element-helper.ts
+++ b/packages/compat/src/addon-dependency-rules/ember-element-helper.ts
@@ -1,0 +1,13 @@
+import { PackageRules } from '..';
+
+let rules: PackageRules = {
+  package: 'ember-element-helper',
+  addonModules: {
+    'helpers/-element.js': {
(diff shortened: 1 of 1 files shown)


ef4:
Thanks, this looks correct.

No, I don't think we need to necessarily have test coverage for all these rules here. They are not really a long-term solution, they're just supposed to try to keep people as unblocked as possible while we can switch addons to patterns that don't need to be annotated.

PR #540, opened 2020-09-18
Fix loader.js issue by disabling makeDefaultExport option

### LINKED ISSUE #539: Interoperability issue of embroider and loader.js
*seems I have a talent for the ugly issues (after #509), just spent most of the day pulling my hair out on this one...*

Tests were failing for me randomly in an addon, but only when building with Embroider. It turned out that a component could be wrongly resolved, depending on the timing.

Having components A and B, and B extending from A, if B was resolved before A, everything seems fine. The other way around however, when resolving B you would get back the module for A.

After a long debugging session, I think I have found the reason. `ember-resolver` would try to get the default export of a module if available [here](https://github.com/ember-cli/ember-resolver/blob/master/addon/resolvers/classic/index.js#L471-L473).

This works fine in a classic build, when you look at the compiled vendor.js, then each module in its AMD form has an explicit default export, like this:

```js
define(..., function(...) {
  let MyComponent = ...
  _exports.default = MyComponent;
});
```

This is however not the case with a Webpack build. 

Now the problem lies with `loader.js`, which mutates a module export by creating a default export [here](https://github.com/ember-cli/loader.js/blob/master/lib/loader/loader.js#L115-L122), if it is not already present (remember, it *is* present in a classic build) and the `makeDefaultExport` option is set (which it is by default). 

But in the case of a webpack-build module that exports a component, this *creates* a `default` property on the exported class! 😱

So when B is resolved first, it creates `B.default` (pointing to `B` again). If then `A` is resolved, `A.default` does not exist, so also `A.default` is created (pointing to `a`). This side-effect of extending the classes seems ugly enough, but at least it works so far.

Now the other way around: `A` is resolved first, `A.default` is created, referencing `A` again. Now it is the turn for `B`, and here this mess breaks everything: so as `B` extends from `A`, `B.default` is already defined, but it still points to `A`! So no `default` is created for `B` by loader.js, as it already inherited it from `A`. But `ember-resolver` will now use `B.default` (same place [here](https://github.com/ember-cli/ember-resolver/blob/master/addon/resolvers/classic/index.js#L471-L473)) which still refers to `A`. So by trying to resolve component `B`, you get a reference to class `A`! 🙈

When patching my local `loader.js`, specifically setting `makeDefaultExport` to false [here](https://github.com/ember-cli/loader.js/blob/master/lib/loader/loader.js#L76), all the failing tests were gone. So that confirms my observation.

Tbh, the choice to also extend a function as the module export [here](https://github.com/ember-cli/loader.js/blob/master/lib/loader/loader.js#L118) seems odd to me! 🤔

simonihmig:
This is just a quick PoC, which seems to fix #539 for me. ~However that seems to be more like a dirty workaround tbh. Just wanted to post this here to discuss future directions for how to fix this more properly.~

Unfortunately there seems to be no way to tell loader.js to not set makeDefaultExport by default. We could extend the addon, but then we don't have control over the version the user's app uses, as it's the app's dependency, and not ours. Or maybe we can tell Webpack somehow to export a `default` key as a classic build does? (see #539)

### DIFF
--- a/packages/core/src/app.ts
+++ b/packages/core/src/app.ts
@@ -244,6 +244,12 @@ export class AppBuilder<TreeNames> {
         relativePath: '_ember_env_.js',
         source: `window.EmberENV=${JSON.stringify(emberENV, null, 2)};`,
       });
+
+      result.push({
+        kind: 'in-memory',
+        relativePath: '_loader_.js',
+        source: `loader.makeDefaultExport=false;`,
+      });
     }
 
     if (type === 'implicit-test-scripts') {


simonihmig:
> IMHO, landing the fix in your other PR is totally fine, but I want to also land the loader.js changes so that we attack the issue from both angles and eventually as projects upgrade the divergence is reduced.

Given Rob's comment [here](https://github.com/embroider-build/embroider/issues/539#issuecomment-696818203), @ef4 what do you think about merging this as-is?

Issue #545, opened 2020-09-22
Movable package cache should tolerate missing deps


ef4:
If a dep (especially a peerDep) is missing, today things error quite early in the movable package cache, because it's trying to identify all ember addons and rewrite them to v2.

It's probably better to tolerate missing deps at this stage, so that error message can move closer to the point of use. This will also make it easier for us to correctly handle optional peerDeps.

PR #546, opened 2020-09-22
Handle case where peerDep is unresolvable

thoov:
Instead of throwing an early error during the movable phase if a dependency was not found instead do nothing and let the error happen further down the line.

Fixes: #545


ef4:
Thanks, looks good.

Issue #544, opened 2020-09-22
Unresolvable dynamic imports should be runtime errors, not static errors


ef4:
When a static import can't be resolved, it's a build error.

But when a dynamic `import()` (or our own `importSync()`, which can be used dynamically) can't be resolved, it may not really be an error because the programmer has guarded against calling it. So we should emit it as a runtime error instead.

In the case of webpack integration, this probably means we emit our own stub module that throws, because I don't think webpack natively understands this distinction. The implementation will look a lot like `makeExternal` in babel-plugin-adjust-imports.ts, except instead of calling window.require it would just throw a helpful error message.

To handle the importSync case, we also need to resolve #543.

Issue #543, opened 2020-09-22
adjust-imports babel plugin doesn't handle importSync


ef4:
I think our internal adjust-imports babel plugin never sees specifiers passed to `importSync`. This will cause them to 
 - not get externals support
 - not get renaming support

We should add explicit handling for `importSync(...)` in babel-plugin-adjust-imports.ts. This might also require changing the relative order of that plugin and the macros plugin, so the macros happen second (I would prefer not to operate on the output of the importSync macro, but rather on the importSync callExpression itself).

PR #547, opened 2020-09-23
Handle dynamic import or importSync of missing module

thoov:
If a module is imported via `import('foo')` or `importSync('foo')` but `foo` cannot be resolved, instead create a dummy module that throws a runtime error stating that the module could not be found. The reasoning behind this change is that it is acceptable to have "guarded imports". An example is take `ember-exam` which does not explicitly depend on either ember-qunit or ember-mocha but has code like the following:

```js
let emberTestFramework;

if (macroCondition(dependencySatisfies('ember-qunit', '*'))){
  emberTestFramework = importSync('ember-qunit');
} else if (macroCondition(dependencySatisfies('ember-mocha', '*'))) {
  emberTestFramework = importSync('ember-mocha');
}
```

During dev builds we do not strip out the conditionals that are falsy and thus when `macros-babel-plugin` runs it will try and resolve these imports (and cause a build-time error). This differs from 2 points of perspective: 1) classical system allow requiring a module that does not exist and you will not see any errors until that require is encountered at runtime (or if it is guarded it may never error). 2)  The behavior of the macro system in production mode would hide the fact that this module was not found since it striped out the import branch which is different from how it would behave under dev mode.

Additional background on ember-exam's change: https://github.com/ember-cli/ember-exam/pull/599

Fixes: #543
Fixes: #544

### DIFF
--- a/packages/core/src/babel-plugin-adjust-imports.ts
+++ b/packages/core/src/babel-plugin-adjust-imports.ts
@@ -66,7 +66,12 @@ type DefineExpressionPath = NodePath<CallExpression> & {
 };
 
 export function isImportSyncExpression(path: NodePath<any>) {
-  if (!path.isCallExpression() || path.node.callee.type !== 'Identifier' || path.node.callee.name !== 'importSync') {
+  if (
+    !path ||
+    !path.isCallExpression() ||
+    path.node.callee.type !== 'Identifier' ||
+    path.node.callee.name !== 'importSync'
+  ) {
     return false;
   }
 
@@ -75,7 +80,7 @@ export function isImportSyncExpression(path: NodePath<any>) {
 }
 
 export function isDynamicImportExpression(path: NodePath<any>) {
-  if (!path.isCallExpression() || path.node.callee.type !== 'Import') {
+  if (!path || !path.isCallExpression() || path.node.callee.type !== 'Import') {
     return false;
   }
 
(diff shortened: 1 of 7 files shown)


thoov:
Still need to add test coverage

packages/core/src/babel-plugin-adjust-imports.ts
```diff
@@ -261,6 +302,15 @@ export default function main() {
         },
       },
       CallExpression(path: NodePath<CallExpression>, state: State) {
+        if (isImportSyncExpression(path) || isDynamicImportExpression(path)) {
+          const [source] = path.get('arguments');
+          let { opts } = state;
+          opts.dynamicImports.push((source.node as any).value);
```
ef4, line 308: Instead of using `opts.dynamicImports` to store this state, I think it would be better to add an `isDynamic: boolean` argument to the `adjustSpecifier()` function. Especially because a specifier can be used both statically and dynamically, and the static locations should still be a static error, only the dynamic locations should compile to a runtime error.

packages/core/src/babel-plugin-adjust-imports.ts
```diff
@@ -64,6 +65,29 @@ type DefineExpressionPath = NodePath<CallExpression> & {
   };
 };
 
+export function isImportSyncExpression(path: NodePath<any>) {
```
ef4, line 68: See `referencesImport` as used in https://github.com/embroider-build/embroider/blob/d5167811ef396d6cc0e0212f3f2edd44f9812cc3/packages/macros/src/babel/macros-babel-plugin.ts#L91 for a better way to implement `isImportSyncExpression`. Specifically, if somebody says 

```js
import { importSync as i } from "@embroider/macros"; 
i('foo')
```

we still want to detect that case.

ef4 (changes requested):
Overall looking good, if you can incorporate these two suggestions and get the tests green.

packages/core/src/babel-plugin-adjust-imports.ts
thoov: fixed

thoov:
@ef4 this should be ready to merge

ef4 (approved):

PR #554, opened 2020-09-25
Update @embroider/macros examples in the readme

thoov:
Not the full list of macros but a first pass on the big ones

### DIFF
--- a/packages/macros/README.md
+++ b/packages/macros/README.md
@@ -8,61 +8,179 @@ Traditionally, Ember addons have a lot of power to run arbitrary code during the
 
 The [Embroider package spec](../../SPEC.md) proposes fixing this by making Ember addons much more static. But they will still need the ability to change themselves in certain ways at app compilation time. Hence this package.
 
-This package works in both Embroider builds and normal ember-cli builds, so that addon authors can switch to this newer pattern without disruption.
-
-## Examples
-
-TODO
-
-## The Macros
-
-### dependencySatisfies
-
-Tests whether a given dependency is present and satisfies the given semver range.
+This package works in both Embroider and Classical builds, so that addon authors can switch to this newer pattern without disruption.
 
 ## Setting Configuration: from an Ember app
 
 1. Add `@embroider/macros` as `devDependency`.
 2. In `ember-cli-build.js`, do:
 
-   ```js
-   let app = new EmberApp(defaults, {
-     '@embroider/macros': {
-       // this is how you configure your own package
-       setOwnConfig: {
-         // your config goes here
-       },
-       // this is how you can optionally send configuration into your
-       // dependencies, if those dependencies choose to use
-       // @embroider/macros configs.
-       setConfig: {
-         'some-dependency': {
-           // config for some-dependency
-         },
-       },
-     },
-   });
-   ```
+```js
+let app = new EmberApp(defaults, {
+  '@embroider/macros': {
+    // this is how you configure your own package
+    setOwnConfig: {
+      // your config goes here
+    },
+    // this is how you can optionally send configuration into your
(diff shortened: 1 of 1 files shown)


packages/macros/README.md
```diff
@@ -8,61 +8,179 @@ Traditionally, Ember addons have a lot of power to run arbitrary code during the
...
-         setConfig: {
-           'some-dependency': {
-             // config for some-dependency
-           },
-         },
-       },
-     },
-   };
-   ```
+- [ember-exam](https://github.com/ember-cli/ember-exam)
```
simonihmig, line 185: I looked at this to find out what they are using macros for, but couldn't find any usage!? 🤔

packages/macros/README.md
```diff
@@ -8,61 +8,179 @@ Traditionally, Ember addons have a lot of power to run arbitrary code during the
...
+- [ember-bootstrap](https://github.com/kaliber5/ember-bootstrap)
```
simonihmig, line 186: If you want to add another example, I use macros in tests only for https://github.com/kaliber5/ember-stargate, basically as a replacement for `ember-compatibility-helpers`

packages/macros/README.md
```diff
@@ -8,61 +8,179 @@ Traditionally, Ember addons have a lot of power to run arbitrary code during the
...
+}
+```
+
+```hbs
+{{macroDependencySatisfies 'qunit' '^2.8.0'}}
+```
+
+### getOwnConfig, getConfig, and getGlobalConfig
+
+A common pattern is to have a set of configuration properties that you define (or a consumer defines for you) which you base certain build time conditions around. This is achieved via the `getOwnConfig`, `getConfig`, and `getGlobalConfig` macros (depending on which config you want to read).
```
simonihmig, line 148: Should we explain those in a bit more detail? Like `getGlobalConfig` was new to me, and not mentioned in the RFC, so it's not obvious what a global config is.

simonihmig (commented):
👍

packages/macros/README.md
thoov: Its coming with my open PR: https://github.com/ember-cli/ember-exam/pull/599

packages/macros/README.md
thoov: 👍  will add

packages/macros/README.md
```diff
@@ -8,61 +8,180 @@ Traditionally, Ember addons have a lot of power to run arbitrary code during the
 
 The [Embroider package spec](../../SPEC.md) proposes fixing this by making Ember addons much more static. But they will still need the ability to change themselves in certain ways at app compilation time. Hence this package.
 
-This package works in both Embroider builds and normal ember-cli builds, so that addon authors can switch to this newer pattern without disruption.
+This package works in both Embroider and Classical builds, so that addon authors can switch to this newer pattern without disruption.
```
acorncom, line 11: Might read more consistently across the ecosystem to call these Classic builds?

packages/macros/README.md
```diff
@@ -8,61 +8,180 @@ Traditionally, Ember addons have a lot of power to run arbitrary code during the
...
+    },
+  },
+};
+```
 
 ## The Macros
 
+### macroCondition
+
+The `macroCondition` macro allows branch level code isolation (and deletion in the case of production builds). Generally macroConditions are viewed as a foundation macro and are combined with others marcos (detailed below) to create more complex scenarios. `macroCondition` takes a single argument which must be statically known or another macro which will compile down to a static value.
```
acorncom, line 64: combined with other **macros**

rwjblue (approved):
I'd have loved these docs when I recently tried to use `@embroider/macros` instead of `ember-compatibility-helpers`...

🙏

PR #560, opened 2020-10-04
Fix runtime getConfig

### LINKED ISSUE #559: MacroCondition throws when used as a class field initializer
Works:

```js
export default class TestComponent extends Component {
  @someRandomDecorator
  myClass = macroCondition(getOwnConfig().isRed) ? 'red' : 'blue';
}
```

Works:

```js
const klass = macroCondition(getOwnConfig().isRed) ? 'red' : 'blue';
export default class TestComponent extends Component {
  myClass = klass
}
```

Throws `the first argument to macroCondition must be statically known`:

```js
export default class TestComponent extends Component {
  myClass = macroCondition(getOwnConfig().isRed) ? 'red' : 'blue';
}
```

The `macroCondition` function in `/src/babel/macro-condition.ts` first sees the original code, on a second invocation it sees the modified runtime-mode code, which it fails to regard as "confident":

image

Working on a PR...

simonihmig:
Fixes #559

### DIFF
--- a/packages/macros/src/babel/evaluate-json.ts
+++ b/packages/macros/src/babel/evaluate-json.ts
@@ -347,22 +347,16 @@ export class Evaluator {
   // confidence without actually forcing the value.
   private maybeEvaluateRuntimeConfig(path: NodePath<CallExpression>): EvaluateResult {
     let callee = path.get('callee');
-    if (callee.isMemberExpression()) {
-      let prop = assertNotArray(callee.get('property'));
-      if (prop.isIdentifier() && prop.node.name === '_runtimeGet') {
-        let obj = callee.get('object');
-        if (
-          obj.isIdentifier() &&
-          (obj.referencesImport('@embroider/macros', 'getConfig') ||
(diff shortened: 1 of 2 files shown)


packages/macros/src/babel/evaluate-json.ts
```diff
@@ -347,22 +347,16 @@ export class Evaluator {
   // confidence without actually forcing the value.
   private maybeEvaluateRuntimeConfig(path: NodePath<CallExpression>): EvaluateResult {
     let callee = path.get('callee');
-    if (callee.isMemberExpression()) {
-      let prop = assertNotArray(callee.get('property'));
-      if (prop.isIdentifier() && prop.node.name === '_runtimeGet') {
```
simonihmig, line 352: This `_runtimeGet` thing has been removed in #444, but in seems it has been forgotten here. 

As shown in #559, the runtime version which is now something like `config('path/to/package')` was seen by this but not recognized as such.

ef4:
Thanks, nice work!

PR #561, opened 2020-10-04
Run yarn-deduplicate

simonihmig:
Decreased `node_modules` from 2.2GB to 575MB, due to a lot of duplicated dependencies.

`yarn` in CI: from 6m to 1m


ef4:
Thanks!

PR #565, opened 2020-10-10
guard against stringly broccoli trees

ef4:
We need to make sure all the broccoli trees we're passing out of v1-app are actually Trees and not strings. Previously, we were just passing vendor and public trees through to the output, which could let strings escape. And it turns out, dummy apps all use a string by default for their public tree, which meant we were dropping their public assets.


ef4:
(I should probably check for path existence too, similar to what the default buildTreeFor does in ember-cli.)

PR #566, opened 2020-10-10
adding support for addons that use treeFor() to suppress various trees

ef4:
In general, we can't run custom `treeFor()` hooks because they expect to recurse into all their child addons and have visibility into the output of all their child addons, which we forbid.

But the most common use case for `treeFor()` is quite simple, and we *can* opportunistically handle that case. It's the case where treeFor is returning either falsy or the unchanged output of `_super.treeFor()`.

This lets us detect those cases and suppress trees that are supposed to get suppressed.

### DIFF
--- a/packages/compat/src/v1-addon.ts
+++ b/packages/compat/src/v1-addon.ts
@@ -313,7 +313,11 @@ export default class V1Addon {
   }
 
   @Memoize()
-  private hasStockTree(treeName: string) {
+  private hasStockTree(treeName: string): boolean {
+    if (this.suppressesTree(treeName)) {
+      return false;
+    }
+
     // we need to use this.addonInstance.root instead of this.root here because
     // we're looking for the classic location of the stock tree, and that
     // location is influenced by a customized ember-addon.main in package.json,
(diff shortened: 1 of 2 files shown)


ef4:
(Needs tests and needs deletion of several built-in compat adapters that should no longer be needed after this change.)

packages/compat/src/v1-addon.ts
```diff
@@ -919,3 +956,10 @@ function babelPluginAllowedInStage1(plugin: PluginItem) {
 function notColocatedTemplate(path: string) {
   return !/^\.\/components\/.*\.hbs$/.test(path);
 }
+
+const markedEmptyTree = new UnwatchedDir(process.cwd());
+const returnMarkedEmptyTree = {
```
thoov, line 961: @ef4 I ran into an issue with this on an addon that calls `this._super` instead of `this._super.treeFor`. Currently the way we are mocking this as a POJO results them getting an error with: `Uncaught TypeError: this._super is not a function`. Obviously, this can easily be fixed by changing their call to `this._super.treeFor` but not sure how wide spread that is or if that is even something we should support.

ef4: Ok, yeah, we can support that. It looks like `core-object` supports both, which is... ugh.

I filed https://github.com/embroider-build/embroider/issues/569

PR #570, opened 2020-10-15
Use the namespace over the default for extra imports

thoov:


thoov:
@ef4 this should be good to go

ef4:
Thanks, looks good.

Issue #487, opened 2020-06-26
Dynamic Component Hole Discussion


ef4:
I'm opening this issue to track discussions with the glimmer-vm and ember core framework teams whose help we need to collaborate on this problem.

# Problem Statement

1. To make embroider robust under `staticComponents: true`, we want addon authors to begin porting away from dynamic component invocation patterns that can't be analyzed.

2. We should make sure we're asking them to port to an API that is ergonomic and long-term supported. We shouldn't make them port a second time any time soon.

## Motivating Example

Today a lot of popular addons do this:

```js
// hypothetical-button.js
export default class extends Component {
  get labelComponent() {
    return this.args.labelComponent || "hypothetical-button-default-label";
  }
}
```

```hbs
{{! hypothetical-button.hbs }}
<button>
  {{component this.labelComponent}}
</button>
```

Callers of the component can do all of these:

```hbs
{{! default case: get the default implementation from the addon }}
<HypotheticalButton />

{{! string case: pass the string name of our own component }}
<HypotheticalButton @labelComponent="custom-label" />

{{! definition case: pass the component definition of our own component }}
<HypotheticalButton @labelComponent={{component "custom-label"}} />
```

How should this addon author refactor so that their code doesn't introduce any static analysis hole? Today we have a hole because we can't be sure what `{{component this.labelComponent}}` refers to.

## Embroider's current state-of-the-art

Today, Embroider has a powerful system of rules for annotating what other code is doing and assisting the user in incrementally tracing where dynamic components come from. We can annotate that the `labelComponent` argument must be a valid, statically-known component definition, which will cause Embroider to trigger new assertions at the *callsites* of `<HypotheticalButton/>`, which can then be resolved through more rules or refactoring, etc, until all dynamic component invocations bottom out on something we can understand statically (example: `{{component "a-string-literal"}}` is safe and common), or the user opts out of safety and says they'll manage it themselves.

This is only scaffolding. It's not a good experience, and it's fragile because the annotations live outside the packages they're talking about. Rather than create first-party annotations inside addons, I would rather give addons better patterns to refactor into.

## The Difficulty Today

For an addon author who wants to do the right thing, there are two awkward problems:

 1. We need to make a clear static dependency on the default implementation. The example getter code:

    ```js
    get labelComponent() {
      return this.args.labelComponent || "hypothetical-button-default-label";
    }
    ```

    has an implicit dependency on a component that we can't discover statically. It would be better if we could say:

    ```js
[...]

PR #571, opened 2020-10-16
ensureSafeComponent helper

ef4:
This introduces a helper function we can tell addon authors to use to make their dynamic component invocations more future-proof.


ADDON-AUTHOR-GUIDE.md
```diff
@@ -88,3 +88,130 @@ To avoid these limitations, you will want to have a build step that allows you t
 - When automatically upgrading addons to **v2**, `@embroider/compat` already has the full application available to examine, so any configuration-dependent transformations will run with the right configuration. For example, any polyfills that behave differently under different Ember versions will see the app's Ember version and do the right thing. But when publishing a native **v2** package, you _don't_ have that information available so if you want to emit different code for different Ember versions the polyfill needs to use `@embroider/macros`.
 
 While it's important that we ship good shared tooling, a nice thing about having a stable **v2** format standard is that the community is free to experiment with alternative addon build tooling, and it won't impact compatibility or stability.
+
+## Replacing the {{component}} helper
+
+Using the `{{component}}` helper (with one small exception which we'll get to) prevents an addon from achieving the **Optimized Embroider Safe** support level. So addons should stop using it. But how?
```
simonihmig, line 94: I believe this is not *quite* true, i.e. there are at least two valid exceptions: the one you mention (static string), but also when yielding contextual components you will still need the `{{component}}` helper to curry arguments. And this will still work with static components, as long as the component positional argument itself is guaranteed to refer to a component that embroider was able to "see" during its analysis. 

Example: https://github.com/kaliber5/ember-bootstrap/pull/1251/files#diff-c3c2155cba3ac783f3443fa7014f24ced9dac371164096f02cb9d1c0767a651eR8-R9. 

Admittedly there is a loophole here as `@titleComponent` could be a string, but I guess I could now wrap that with the `ensureSafeComponent` *helper*, right? Anyways, as the example shows you still need the `component` helper to curry arguments. FWIW, I have that pattern quite a lot in ember-bootstrap, where contextual components are yielded, but they can *also* be passed as arguments to exchange the used component class.

simonihmig (commented):
This is great! 

*I noticed that something like that was planned in #487, but didn't expect this to be available *that* soon. otherwise I would have waited for this before doing https://github.com/kaliber5/ember-bootstrap/pull/1251, where I basically pulled all component references from JS to templates (in a statically analyzable way) to enable static components mode...*

ADDON-AUTHOR-GUIDE.md
ef4: This is a good point. We should expand this section of the guide to help people with those cases. While there are plenty of patterns that are safe, there really is only one that embroider can *see* is safe, and that is a string literal argument.

We need to think about what the future of currying looks like in Ember, because I suspect it doesn't need to continue being a special feature of the `{{component}}` helper.

Once we're free to pass around component classes and then invoke them, "currying" can just be `class extends`.

Also, I think named blocks replaces many cases where you would have passed a curried component as an argument:

```hbs
<Menu @titleBar={{component "fancy-bar" mode=this.mode}} />
{{!-- vs --}}
<Menu>
  <:titleBar>
    <FancyBar @mode={{this.mode}} />
  </:titleBar>
</Menu>
```

Notice that this also solves the problem of "how do I curry HTML attributes?".

packages/addon/app/helpers/ensure-safe-component.js
```diff
@@ -0,0 +1,7 @@
+import { ensureSafeComponent } from '@embroider/addon';
+import Helper from '@ember/component/helper';
+export default class extends Helper {
+  compute([value]) {
+    return ensureSafeComponent(value, this);
+  }
+}
```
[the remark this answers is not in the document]
ef4: I agree that would be more conventional, and I should probably make the change you suggest. This started out as a functional helper, in which case there really was basically no implementation to worry about. But then it became a class-based helper, and now has a bunch of syntax that *might* in theory be problematic for an app.

But I don't think this actually breaks under any real conditions. The thing about the app folder isn't that it doesn't get transpiled, it's that it gets transpiled under the app's babel config and not the addon's. But what is the minimum level of Javascript features we can assume will work in the app? We already know that the app *must* support ES modules because people put reexport syntax in the app tree. Are there really any apps that support ES modules but not class syntax or array destructuring? Probably not.

ADDON-AUTHOR-GUIDE.md
ef4: Your comment makes me realize I do need to explain the rule more precisely.

You can use the component helper, but only if the first argument is: 
 - a string literal
 - or a call to the `ensure-safe-component` helper.

I didn't say so explicitly, but embroider should be made to not worry about this pattern:

```hbs
{{component (ensure-safe-component @whatever) }}
```

I didn't mention that case because I was assuming it would never be needed (use angle brackets instead), but you're right that if you want to curry arguments onto it but not invoke it, you'd need that pattern.

ef4: This is now [implemented](https://github.com/embroider-build/embroider/pull/608) and [documented](https://github.com/embroider-build/embroider/pull/611)

Issue #569, opened 2020-10-14
Support direct `this._super` calls in treeFor


ef4:
We try to infer whether `treeFor` is passing out the value returned by its super method unchanged, but right now we only support `this._super.treeFor()` and not `this._super()`, when apparently core-object supports both.

See https://github.com/embroider-build/embroider/pull/566#discussion_r504863610

While doing this, we should probably wrap the entire invocation of the custom `treeFor` in a catch block, because the whole thing is very "best-effort". In the most general case we can't make these methods see exactly what they would have seen in a classic build, we're really just looking for the most common two special cases (returning null or returning the super implementation unchanged).

ef4:
The core-object implementation is here https://github.com/ember-cli/core-object/blob/480c831a19935860b0786307a94edba48adf3044/lib/assign-properties.js#L15-L35

rwjblue:
> apparently core-object supports both

😭

PR #572, opened 2020-10-17
Support direct _super call in treeFor hook

thoov:
Fixes: #569

### DIFF
--- a/packages/compat/tests/stage1.test.ts
+++ b/packages/compat/tests/stage1.test.ts
@@ -372,7 +372,7 @@ describe('stage1 build', function () {
         `
         treeFor(name) {
           if (name !== 'app') {
-            return this._super.treeFor(name);
+            return this._super(name);
           } else {
             return undefined;
           }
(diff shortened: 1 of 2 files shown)


packages/compat/tests/stage1.test.ts
```diff
@@ -372,7 +372,7 @@ describe('stage1 build', function () {
         `
         treeFor(name) {
           if (name !== 'app') {
-            return this._super.treeFor(name);
+            return this._super(name);
```
rwjblue, line 375: Does this replace the only test for `this._super.treeFor` (shouldn't we test both things)?

thoov: No, I left the other test that @ef4 created: https://github.com/embroider-build/embroider/pull/572/files/0ba72bf537c5bd0fc6073751f7c6796d04ffdda2#diff-4397bded80f536069b8084a839df928911762318b820e9f6976baa1cc3f40538L353 which uses `this._super.treeFor`

rwjblue: Awesome, thank you!

ef4:
Thanks!

PR #573, opened 2020-10-20
Add Fastboot smoke tests

simonihmig:
The `fastboot-app` package uses FastBoot, but its tests just run in the browser. This adds the same set of tests as in `tests/acceptance/basic-test.js`, but for the server rendered output.

### DIFF
--- a/test-packages/fastboot-app/config/environment.js
+++ b/test-packages/fastboot-app/config/environment.js
@@ -17,6 +17,8 @@ module.exports = function(environment) {
       }
     },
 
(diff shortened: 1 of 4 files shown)


simonihmig:
Damned, overlooked that there actually were Fastboot tests, just not where I expected them (when using `ember-cli-fastboot-testing`) 🤦‍♂️

ef4:
Sorry, the tests are definitely confusing right now. There are several ways to do things.

PR #574, opened 2020-10-20
Fix missing hostWhitelist in FastBoot's package.json manifest

simonihmig:
Adds accessing `request.host` to `fastboot-app` that reproduces the missing FastBoot package.json data (here `hostWhitelist`) ~reported in #160~. This triggers the `You must provide a hostWhitelist to retrieve the host` error making the app's FastBoot tests fail. 

The second commit adds a fix, exposing the `hostWhitelist` manifest field, that got lost in the custom compat adapter, but still seems to be required and expected by `fastBoot` [here](https://github.com/ember-fastboot/fastboot/blob/429eb1d1c3d9353f3085196e71cfde416f154cba/src/fastboot-schema.js#L94).

### DIFF
--- a/packages/compat/src/compat-adapters/ember-cli-fastboot.ts
+++ b/packages/compat/src/compat-adapters/ember-cli-fastboot.ts
@@ -119,6 +119,7 @@ class RewriteManifest extends Plugin {
       schemaVersion: 5,
       htmlEntrypoint: 'index.html',
       moduleWhitelist: json.fastboot.moduleWhitelist,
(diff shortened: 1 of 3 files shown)


simonihmig:
@ef4 this is now containing both the reproduction of the problem I had (and wrongly contributed to #160) and a simple fix. Hope this looks ok!

wagenet:
I was hitting the same problem. This fix works for me.

ef4:
Thanks, nice work.

PR #576, opened 2020-10-21
Timestamp changelog

sandstrom:
Timestamps makes it easy to understand release cadence, see if something has changed recently, etc.

Only added for the previous 3, but would be great if this could be added to future updates.

Thanks for an awesome tool! 🏅


ef4:
Thanks, good idea.

PR #579, opened 2020-10-21
Update Babel related dependencies.

rwjblue:
- Update @babel/* and @types/babel* related dependencies.
- Fix typings to account for StringLiteral's as `ExportSpecifier`s (new changes to the babel types in 7.12)

### DIFF
--- a/packages/compat/package.json
+++ b/packages/compat/package.json
@@ -21,11 +21,11 @@
     "test": "jest"
   },
   "dependencies": {
-    "@babel/core": "^7.8.7",
+    "@babel/core": "^7.12.3",
     "@babel/plugin-syntax-dynamic-import": "^7.8.3",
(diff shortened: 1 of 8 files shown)


rwjblue:
Rebased to account for changes in #575, but kept my original fix for the `ExportSpecifiers` as I believe it is more correct than to cast the types.

rwjblue:
@ef4 - CI is definitely much faster since #575

ef4:
Yeah, thanks, your fix for those type errors is better -- in truth I didn't mean to commit mine, that was a quick hack while trying to focus elsewhere. Yarn unstability means you sometimes get these type differences when you didn't want them.

The CI is about twice as fast now. And hopefully it should run reliably locally too, and pretty fast.

Issue #577, opened 2020-10-21
Installation warnings due to outdated package dependencies.


rwjblue:
The following warnings are emitted by `yarn` when running a `yarn install` where you have the following dependencies:

```json
{
  devDependencies: {
    "@embroider/compat": "^0.27.0",
    "@embroider/core": "^0.27.0",
    "@embroider/macros": "^0.27.0",
    "@embroider/webpack": "^0.27.0",
  }
}
```

You get the following warnings on a fresh `yarn install` (without a prior lockfile):

```
warning @embroider/compat > typescript-memoize > core-js@2.4.1: core-js@<3 is no longer maintained
 and not recommended for usage due to the number of issues. Please, upgrade your dependencies to t
he actual version of core-js@3.
warning @embroider/compat > babel-core > babel-register > core-js@2.6.11: core-js@<3 is no longer
maintained and not recommended for usage due to the number of issues. Please, upgrade your depende
ncies to the actual version of core-js@3.
warning @embroider/compat > babel-core > babel-runtime > core-js@2.6.11: core-js@<3 is no longer m
aintained and not recommended for usage due to the number of issues. Please, upgrade your dependen
cies to the actual version of core-js@3.
warning @embroider/compat > jsdom > request@2.88.2: request has been deprecated, see https://githu
b.com/request/request/issues/3142
warning @embroider/compat > jsdom > request-promise-native@1.0.9: request-promise-native has been
deprecated because it extends the now deprecated request package, see https://github.com/request/r
equest/issues/3142
warning @embroider/compat > jsdom > request > har-validator@5.1.5: this library is no longer suppo
rted
warning @embroider/compat > broccoli > findup-sync > micromatch > snapdragon > source-map-resolve
> resolve-url@0.2.1: https://github.com/lydell/resolve-url#deprecated
warning @embroider/compat > broccoli > findup-sync > micromatch > snapdragon > source-map-resolve
> urix@0.1.0: Please see https://github.com/lydell/urix#deprecated
warning @embroider/webpack > @types/source-map@0.5.7: This is a stub types definition for source-m
ap (https://github.com/mozilla/source-map). source-map provides its own type definitions, so you d
on't need @types/source-map installed!
warning @embroider/webpack > webpack > watchpack > watchpack-chokidar2 > chokidar@2.1.8: Chokidar
2 will break on node v14+. Upgrade to chokidar 3 with 15x less dependencies.
[...]

PR #580, opened 2020-10-21
Remove babel-core dependency from @embroider/compat.

rwjblue:
This dependency seems unused within the `@embroider/compat` package, and causes a number of installation warnings.

Related to https://github.com/embroider-build/embroider/issues/577


ef4:
Yes, I think the code that needed this moved into `@embroider/test-support`, at `test-packages/support`. Thanks.

PR #582, opened 2020-10-21
Explicitly declare Node version support: 10, 12, 14+.

rwjblue:
Updates each of our published packages' `package.json` files to have the following engines field value:

```json
{
  "engines": {
    "node": "10.* || 12.* || >= 14"
  }
}
```

This is the list of currently supported Node releases (based on [nodejs.org's release page](https://nodejs.org/en/about/releases/)).

We _may_ want to consider this a breaking change (due to the `@embroider/router` package having had engines defined that allowed Node 8), but for most of the packages there simply was no declaration of support at all which IMHO means adding `engines` is "safe".


packages/router/package.json
```diff
@@ -67,7 +67,7 @@
     }
   },
   "engines": {
-    "node": "8.* || >= 10.*"
+    "node": "10.* || 12.* || >= 14"
```
rwjblue, line 70: This change should probably be considered "breaking".

ef4:
(I merged master in because I want to see that my recent test suite changes are working well in a PR like this.)

PR #583, opened 2020-10-22
Update to released version of fixturify-project.

rwjblue:
The original fork was done in order to support scoped packages (https://github.com/stefanpenner/node-fixturify-project/pull/25), and that PR has landed.


ef4:
Thank you.

Issue #497, opened 2020-07-26
dependencySatisfies hbs and js implementations differ


ef4:
https://github.com/embroider-build/embroider/issues/121#issuecomment-663995521

Issue #121, opened 2019-04-10
dependencySatisfies macro has confusing semantics


ef4:
@rwjblue pointed out that `semver.satisfies()` is not really the semantics you usually want for conditional code inclusion.

The problem is that you want to write things like 

```js
if (!semver.satisfies(emberVersion, '> 3.8')) {
  // add some polyfill
}
```

But  `semver.satisfies('3.9.0-beta.0', '> 3.8')` returns false, which was really not your intent.

`ember-dependency-helpers` solves this by exposing `gte` and `lte` instead of `satisfies`. But that is less expressive than being able to use complete semver range expressions.

We could choose to replace our `dependencySatisfies` macro with `dependencyGTE` and friends.

Alterantively, I *think* `semver.satisfies(semver.coerce(pkg.version), pattern)` has the semantics we actually want, but I'm opening this issue to get feedback on that idea.

rwjblue:
I think coercing is probably fine, but I _vaguely_ worry about diverging behaviors between `semver.satisfies` and `dependencySatisfies`.

ef4:
I can understand your suspicion of reusing the terminology "satisfies". 

Maybe it would help if we rename ours to `hasDependency('ember-source', '> 3.8')`?

ef4:
Or maybe `hasPackage`, since I guess technically `ember-source` is rarely a dependency of the code that actually wants to use these macros.

ef4:
(Thanks for your input BTW, I really appreciate it!)

rwjblue:
@ef4 - I just took a look at this again (trying to use `@embroider/macros` instead of `ember-compatibility-helpers` for a new package) and have a few concerns:

* https://github.com/embroider-build/embroider/pull/122 only resolved this for templates, the original issue is still present for the JS version (and we now have divergent behavior between the Babel side and Glimmer side)...
* Coercing means (AFAICT) that it is impossible to ask if something is at least `3.20.0-beta.4` and have that **not** match `3.20.0-beta.1`. 

I'm falling back to `ember-compatibility-helpers` for now here, but I'd like to circle around and make that package obsolete (to aid future Embroider adoption)...

PR #584, opened 2020-10-22
Fix issues with dependencySatisifies with prereleases.

rwjblue:
Based on https://github.com/npm/node-semver#prerelease-tags:

> If a version has a prerelease tag (for example, 1.2.3-alpha.3) then it will only be allowed to satisfy comparator sets if at least one comparator with the same [major, minor, patch] tuple also has a prerelease tag.

That means that if an addon uses:

```js
dependencySatisfies('ember-source', '^3.0.0')
```

When the project actually has `3.2.0-beta.1` installed `semver.satisifies` will _by default_ return `false`.

For our purposes however, we need to ensure that is allowed.  Mostly because the "end user" (the host project, or the package that included the package that uses `dependencySatisfies`) has explicitly asked for a prerelease version.

Fixes embroider-build/embroider#121 properly (without requiring `semver.coerce`) using new feature added to `semver` since the original fix
Fixes https://github.com/embroider-build/embroider/issues/497
Fixes https://github.com/ember-cli/ember-exam/issues/616


test-packages/support/index.ts
```diff
@@ -11,7 +11,10 @@ interface RunDefaultOptions {
 
 export function toCJS(code: string): string {
   return transform7(code, {
-    plugins: ['@babel/plugin-transform-modules-commonjs', '@babel/plugin-transform-typescript'],
+    plugins: [
+      require.resolve('@babel/plugin-transform-modules-commonjs'),
+      require.resolve('@babel/plugin-transform-typescript'),
+    ],
```
rwjblue, line 17: This change was needed because using the non absolute string versions will try to resolve from `process.cwd()` so if any tests that need this functionality run with a different `process.cwd()` (like the tests added in this PR) these babel packages will not be found.

rwjblue:
@ef4 - This fixes an issue blocking usage of `ember-exam` in the current beta's of ember-cli (https://github.com/ember-cli/ember-exam/issues/616), would love a review (and release if it seems ok). I _could_ work around the issue by removing usage of `@embroider/macros` over in `ember-exam` but I'd really prefer not to (as I think we should be moving **towards** `@embroider/macros` and away from `ember-compatibility-helpers` generally speaking).

ef4:
Thanks!

PR #586, opened 2020-10-23
[WIP] Allow staticOptions via packageRules

thoov:
This enables all of the static flags to be individually changed on a per package basis via the packageRules option. This will greatly help with apps onboarding to embroider as it will give them the flexibility to "opt out" of an optimization for a bad addon.

### DIFF
--- a/packages/compat/src/dependency-rules.ts
+++ b/packages/compat/src/dependency-rules.ts
@@ -9,7 +9,7 @@ export interface PackageRules {
   package: string;
   semverRange?: string;
 
-  components: {
+  components?: {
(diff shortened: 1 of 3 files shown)


ef4:
I think `staticAddonTrees` and `staticTestSupportTrees` make sense here. They control only the *pushing* of files into the build, they don't control any pulling.

But `staticHelpers` and `staticComponents` can't really be scoped per package, because they control both sides. For them to work, both the package that provides a component and the package that consumes it must agree that it's being handled statically, so the provider can stop pushing and the consumer can start pulling.

ef4:
Closing this until someone wants to pick it back up.

PR #434, opened 2020-04-22
Lazy loading engine JS

thoov:
- [x] Add tests
- [x] Clean up code
- [x] Review new "lazy-import" w/ @ef4 
- [x] Validate engine name lookup logic

PR #591, opened 2020-10-28
Support CSS from lazy engines

thoov:
Lazy engine CSS is now imported via their own entry javascript file. Given the work that was done for #434 we now get as a by-product lazy CSS (for non lazy engines their styles will not be lazy). The fundamental change here is that CSS is able to be imported into javascript via webpack's style loader plugin and now during construction of each lazy engine's entry file we stick CSS imports at the top.

An example of what a lazy engine's entry file looks like after this change is:

```js
import { importSync as i, macroCondition, getGlobalConfig } from '@embroider/macros';
let w = window;
let d = w.define;

if (macroCondition(getGlobalConfig().fastboot?.isRunning)) {}
else {
  // STYLES ARE NOW IMPORTED HERE
  i("../../../lazy-engine/lazy-engine.css");
}

d("lazy-engine/_app_/helpers/duplicated-helper", function(){ return i("../../../lazy-engine/_app_/helpers/duplicated-helper");});
d("lazy-engine/config/_environment_browser_", function(){ return i("../../../lazy-engine/config/_environment_browser_");});
// ...
```

Here is the corresponding app's entry file (we can see how the above lazy entry file is not loaded until invoked by embroider's router):

```js
import { importSync as i, macroCondition, getGlobalConfig } from '@embroider/macros';
let w = window;
let d = w.define;


d("engines-host-app/app", function(){ return i("../app");});
d("engines-host-app/config/asset-manifest", function(){ return i("../config/asset-manifest");});
//...
d("ember-test-waiters/wait-for-promise", function(){ return i("../../../node_modules/ember-test-waiters/wait-for-promise");});
d("ember-test-waiters/waiter-manager", function(){ return i("../../../node_modules/ember-test-waiters/waiter-manager");});

i("./_engine_/eager-engine.js");


w._embroiderEngineBundles_ = [
  {
    names: ["lazy-engine"],
    load: function() {
      return import("./_engine_/lazy-engine.js");
    }
  },
]


if (typeof FastBoot === 'undefined') {
  if (!runningTests) {
    require('engines-host-app/app')['default'].create({"name":"engines-host-app","version":"0.0.0+f34a4df5"});
  }
}
```

### DIFF
--- a/packages/core/src/app.ts
+++ b/packages/core/src/app.ts
@@ -218,8 +218,12 @@ export class AppBuilder<TreeNames> {
     return extensionsPattern(this.adapter.resolvableExtensions());
   }
 
-  private impliedAssets(type: keyof ImplicitAssetPaths, emberENV?: EmberENV): (OnDiskAsset | InMemoryAsset)[] {
-    let result: (OnDiskAsset | InMemoryAsset)[] = this.impliedAddonAssets(type).map(
+  private impliedAssets(
+    type: keyof ImplicitAssetPaths,
+    engine: Engine,
+    emberENV?: EmberENV
+  ): (OnDiskAsset | InMemoryAsset)[] {
+    let result: (OnDiskAsset | InMemoryAsset)[] = this.impliedAddonAssets(type, engine).map(
       (sourcePath: string): OnDiskAsset => {
         let stats = statSync(sourcePath);
         return {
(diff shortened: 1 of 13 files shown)


packages/core/src/app.ts
```diff
@@ -1015,6 +1015,17 @@ export class AppBuilder<TreeNames> {
       requiredAppFiles.push(appFiles.helpers);
     }
 
+    let styles = [];
+    if (!engine.parent) {
```
thoov, line 1019: This needs to be cleaned up

packages/core/src/app.ts
```diff
@@ -1195,16 +1206,24 @@ export class AppBuilder<TreeNames> {
 }
 
 const entryTemplate = compile(`
-import { importSync as i } from '@embroider/macros';
+import { importSync as i, macroCondition, getGlobalConfig } from '@embroider/macros';
 let w = window;
 let d = w.define;
 
+{{#if styles}}
+  if (macroCondition(getGlobalConfig().fastboot?.isRunning)) {}
```
thoov, line 1214: @ef4 this might be a uses case for having a falsy macro or do you have a suggestion here?

packages/core/src/app.ts
```diff
@@ -427,8 +429,6 @@ export class AppBuilder<TreeNames> {
       }
     }
 
-    html.insertStyleLink(html.styles, `assets/${this.app.name}.css`);
```
thoov, line 430: We need to confirm that if this is what we want to do from the parent engine perspective

packages/core/src/app.ts
ef4: (we discussed this out of band, but I'm commenting for other readers)

You can't say `if (!macroCondition(whatever))` but you can say `if (macroCondition(!whatever))`.

We could choose to make the other case work too. It would mean that the presence of `macroCondition` *anywhere* in the predicate requires the whole predicate to be statically determined (and generate a build error if it's not).

ef4 (approved):

ef4:
Thanks, looking good.

Issue #470, opened 2020-06-09
Components don't seem to be split into route bundles


GavinJoyce:
I'm trying out embroider at https://github.com/GavinJoyce/embroider-spike (things are going pretty well, I'm very excited).

Perhaps I'm missing something, but I can't seem to get components to split into their appropriate route bundles. Helpers seem to be splitting just fine.

Here's a [PR where I enable `splitAtRoutes`](https://github.com/GavinJoyce/embroider-spike/pull/1). The `player` helper splits correctly, but the `<Player />` component does not.

Here's my config:

[`ember-cli-build.js`](https://github.com/GavinJoyce/embroider-spike/blob/6841433183b89e8b4178cb9640625281eb5f9cae/ember-cli-build.js):

```js
return require("@embroider/compat").compatBuild(app, Webpack, {
  staticAddonTestSupportTrees: true,
  staticAddonTrees: true,
  staticHelpers: true,
  staticComponents: true,
  packagerOptions: {
    webpackConfig: {},
  },
  splitAtRoutes: ["teams", "players", "table"],
});
```

Here's a gif showing that the `<Player>` component is present in the main app bundle:

![embroider-loading](https://user-images.githubusercontent.com/2526/84034285-0c9a9c80-a992-11ea-8e3f-b05406df4977.gif)

ef4:
I think this is fixed by #594.

GavinJoyce:
I'm trying to verify your fix, but I still seem to be seeing the same behaviour:

https://github.com/GavinJoyce/embroider-spike/pull/31

ef4:
As I commented on that issue, I think this is working correctly, but the presence of translations in the main bundle was making it look like it wasn't.

GavinJoyce:
Ah, thanks - I missed that detail. I'll take a second look now and verify that everything is working as expected

PR #594, opened 2020-11-03
fix unnecessary inclusion of co-located templates

### LINKED ISSUE #589: Tree-shaking for code in /app not working
I have this [reproduction app](https://github.com/simonihmig/embroider-playground) that shows a couple of issues with the way modules are distributed across chunks, which basically boils down to code in the `/app` folder not being tree-shaken as one would expect.

The app does the following things:
* the `foo` route renders a `foo` component
* the `bar` route renders a `bar` component
* the `addon` route renders an external addon's component (Bootstrap button)
* clicking the above button should lazy load the `/app/lazy.js` module through a dynamic import
* there is a `unused` component and a `/app/unused.js` simple module
* the app is configured with the most aggressive settings ( `static*` all `true`)

## First run: *without* route-based code splitting

When commenting out [this line](https://github.com/simonihmig/embroider-playground/blob/master/ember-cli-build.js#L33) to disable `splitAtRoutes`, you get these results: 

✅  `foo` route renders a `foo` component
✅  `bar` route renders a `bar` component
✅  `addon` route renders addon's button component
✅  dynamically importing the `lazy.js` module resolves the correct module
❌  dynamically importing the `lazy.js` module loads a new chunk lazily
❌  the `lazy.js` module is not contained in any of the base chunks (`<script>` in `index.html`)
❌   the `unused` component is not contained in any chunk
❌   the `unused.js` module is not contained in any chunk
✅  unused components of the addon are not contained in any chunk

## Second run: *with* route-based code splitting

✅  `foo` route renders a `foo` component
✅  transitioning to `foo` lazy loads a new chunk
✅  that chunk contains the route's class and template
❌  that chunk contains the `foo` component (see #470)
❌  the `foo` component is not contained in any of the base chunks (`<script>` in `index.html`)

✅  `bar` route renders a `bar` component
✅  transitioning to `bar` lazy loads a new chunk
✅  that chunk contains the route's class and template
❌  that chunk contains the `bar` component (see #470)
❌  the `bar` component is not contained in any of the base chunks (`<script>` in `index.html`)

✅  `addon` route renders addon's button component
✅  transitioning to `addon` lazy loads a new chunk
✅  that chunk contains the route's class and template
✅  that chunk contains the `bs-button` component (*actually two separate chunks, which ideally should be only one!?)
✅  the `bs-button` component is not contained in any of the base chunks (`<script>` in `index.html`)

✅  dynamically importing the `lazy.js` module resolves the correct module
❌  dynamically importing the `lazy.js` module loads a new chunk lazily
❌  the `lazy.js` module is not contained in any of the base chunks (`<script>` in `index.html`)
❌   the `unused` component is not contained in any chunk
❌   the `unused.js` module is not contained in any chunk
✅  unused components of the addon are not contained in any chunk

## Conclusions

It seems as if the old bundling behavior of ember-cli is de-facto still present, i.e. everything in `/app` is going into the (base) bundles unconditionally, whether the modules are used or not. Only for external modules (coming from an addon) we can see the bundling behavior we expect.

ef4:
Fixes #589.

The hbs files of co-located templates in the app were being treated as `otherAppFiles`, meaning uncategorized and therefore unsafe to leave out of the build.

Really we just want to always leave them out here -- they are not, by themselves, resolvable. They're directly references by their corresponding JS (which we will have already synthesized if they are template-only).

### DIFF
--- a/packages/compat/tests/template-colocation.test.ts
+++ b/packages/compat/tests/template-colocation.test.ts
@@ -131,3 +131,91 @@ describe('template colocation', function () {
     assertFile.get(['ember-addon', 'implicit-modules']).doesNotInclude('./components/component-two.hbs');
   });
 });
+
+describe('template colocation with staticComponents', function () {
+  jest.setTimeout(120000);
+  let expectFile: ExpectFile;
+  let build: BuildResult;
+  let app: Project;
+
+  throwOnWarnings();
+
+  beforeAll(async function () {
+    app = Project.emberNew();
+
+    merge(app.files, {
+      config: {
+        'targets.js': `module.exports = { browsers: ['last 1 Chrome versions'] }`,
+      },
+      app: {
+        templates: {
+          'index.hbs': `
+              <HasColocatedTemplate />
+              <TemplateOnlyComponent />
+            `,
+        },
+        components: {
+          'has-colocated-template.js': `
+            import Component from '@glimmer/component';
+            export default class extends Component {}
+            `,
+          'has-colocated-template.hbs': `<div>{{this.title}}</div>`,
+          'template-only-component.hbs': `<div>I am template only</div>`,
+        },
+      },
+    });
+
+    let addon = app.addAddon('my-addon');
+    merge(addon.files, {
+      app: {
+        components: {
+          'component-one.js': `export { default } from 'my-addon/components/component-one';`,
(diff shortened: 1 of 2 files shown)


ef4:
Fixes #470.

PR #600, opened 2020-11-05
fix snippet resolving

ef4:
#593 broke some cases when you're parsing component snippts in packageRules at an early point in the build when the app itself doesn't have a package.json yet. This fixes the problem by ensuring we don't try to discover the owningPackage for our own stuff when we know it's owned by the app.

### DIFF
--- a/packages/compat/src/resolver.ts
+++ b/packages/compat/src/resolver.ts
@@ -328,8 +328,8 @@ export default class CompatResolver implements Resolver {
     return extensionsPattern(this.params.adjustImportsOptions.resolvableExtensions);
   }
 
-  absPathToRuntimeName(absPath: string) {
-    let pkg = PackageCache.shared('embroider-stage3').ownerOfFile(absPath);
+  absPathToRuntimeName(absPath: string, owningPackage?: { root: string; name: string }) {
+    let pkg = owningPackage || PackageCache.shared('embroider-stage3').ownerOfFile(absPath);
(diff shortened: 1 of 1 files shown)


simonihmig:
This seems to also fix a real issue in an app, not only related to snippets in package rules. Wehen updating an app to 0.30.0 from 0.28.0, the build started to fail with `The "path" argument must be of type string. Received undefined`, due to [packageRuntimeName](https://github.com/embroider-build/embroider/blob/master/packages/compat/src/resolver.ts#L359) being undefined. `pkg` was an object, with `dependencies`, `fastboot` and `embroider-fastboot` being the only keys there, but no `name`. Patching my `node_modules` with the changes here fixed the issue.

`absPath` was referring to https://github.com/tildeio/ember-element-helper/blob/main/app/components/-dynamic-element.js, don't know if there was something specific to that or not, as I didn't investigate this further. Just fyi!

ef4:
Yup, that's the same bug. We have a built-in rule that has a snippet for that component: https://github.com/embroider-build/embroider/blob/4a58720b35b3c9c440fa1c1099df0c86fd5ff961/packages/compat/src/addon-dependency-rules/ember-element-helper.ts#L7

simonihmig:
> We have a built-in rule that has a snippet for that component:

Yup, I added that! :) 

However I was not sure how this plays together. But thanks for confirming!

Issue #563, opened 2020-10-06
Cannot find module './vendor/ceibo/index.js'


hergaiety:
Hey, brand new to Emberoider just following the basic getting started installation steps. I'm seeing this error regardless of how much of our ember-cli-build.js I comment out to narrow it down.

```
⠏ building... [SassCompiler]Building into /tmp/embroider/f5e48f
Build Error (WaitForTrees)

Cannot find module './vendor/ceibo/index.js' from '/tmp/embroider/f5e48f/node_modules/@embroider/synthesized-vendor'


Stack Trace and Error Report: /tmp/error.dump.d9374aa702aeca563a507908f5b50fa9.log
```

In our `yarn.lock` the only reference I can find to `ceibo` is in a dependency of a dependency of a dependency for ultimately `ember-table` (which we use at the latest stable 2.x release)

Log: [error.dump.d9374aa702aeca563a507908f5b50fa9.log](https://github.com/embroider-build/embroider/files/5334611/error.dump.d9374aa702aeca563a507908f5b50fa9.log)

I usually like to provide more information for GitHub issues, but I'm not familiar enough with Embroider or webpack to debug this further. Appreciate any suggestions~

ef4:
Hi, thanks for testing. I tried to quickly reproduce by adding ember-table to a new app (including rendering the table and using the provided test-support code), but that all worked fine.

I see `vendor/ceibo/index.js` is getting included by `ember-cli-page-object` which uses `ember-cli-node-assets` to load it. I have had [other bug reports](https://github.com/embroider-build/embroider/issues/500) about `ember-cli-node-assets` not working under embroider, but I have been unable to reproduce them.

What versions these packages are present in your app?
 - `@embroider/*`
 - `ember-cli-node-assets`
 - `ember-cli`
 - `ember-cli-page-object`

To poke around more, look in `/tmp/embroider/f5e48f`. That is where Embroider is outputting your app and addons after converting them to the v2 format, and before they get combined by webpack. The way I would expect `vendor/ceibo/index.js` to get handled is that it will end up in the `/tmp/embroider/f5e48f/node_modules/@embroider/synthesized-vendor` package. That package is a dumping ground for all the shared vendor files injected by addons. 

[...]

Issue #500, opened 2020-08-02
Incompatible with ember-cli-node-assets


Turbo87:
It seems that addons using ember-cli-node-assets are currently incompatible with embroider. I've seen this on other addons in the past too, and currently I'm running into it with https://github.com/shipshapecode/ember-flatpickr:

```
Build Error (WaitForTrees)

Cannot find module './vendor/flatpickr/flatpickr.js' from '/private/var/folders/jb/3qmm1y051xdcv33pjkh4b01c0000gn/T/embroider/42038c/node_modules/@embroider/synthesized-vendor'
```

/cc @shipshapecode @ef4

ef4:
I have seen it work, but maybe something in either embroider or the addon changed.

GCheung55:
@Turbo87 did you exclude `flatpickr` in `autoImport` config? That's what I needed to do to get things building again.

Turbo87:
hmm... I'm not importing `flatpickr` myself, only using it through the components provided by the addon. Shouldn't ember-auto-import ignore imports from addons that don't bring it as a dependency? 🤔

GCheung55:
@Turbo87 Maybe, but how should ember-auto-import know?

ef4:
I continue to be unable to reproduce this problem. I see ember-cli-node-assets working correctly under embroider in several apps. Can anybody share a reproduction where it does not?

Turbo87:
the flatpickr addon is no longer using `ember-cli-node-assets` since https://github.com/shipshapecode/ember-flatpickr/pull/535. you might need to use an older release, I guess.

ef4:
Fixed in #607

Issue #442, opened 2020-05-01
Cannot find module for ember-fullcalendar


iamareebjamal:
After doing the first step listed in the README, embroider starts throwing error:

```
Build Error (WaitForTrees)

Cannot find module './vendor/fullcalendar/dist/fullcalendar.js' from '/tmp/embroider/606e06/node_modules/@embroider/synthesized-vendor'
```

Stacktrace:

```
  - broccoliBuilderErrorStack: Error: Cannot find module './vendor/fullcalendar/dist/fullcalendar.js' from '/tmp/embroider/606e06/node_modules/@embroider/synthesized-vendor'
    at Function.resolveSync [as sync] (/home/iamareebjamal/git/open-event-frontend/node_modules/resolve/lib/sync.js:89:15)
    at AppBuilder.impliedAddonAssets (/home/iamareebjamal/git/open-event-frontend/node_modules/@embroider/core/src/app.js:135:55)
    at AppBuilder.impliedAssets (/home/iamareebjamal/git/open-event-frontend/node_modules/@embroider/core/src/app.js:99:27)
    at AppBuilder.insertEmberApp (/home/iamareebjamal/git/open-event-frontend/node_modules/@embroider/core/src/app.js:207:36)
    at AppBuilder.prepareAsset (/home/iamareebjamal/git/open-event-frontend/node_modules/@embroider/core/src/app.js:333:18)
    at AppBuilder.prepareAssets (/home/iamareebjamal/git/open-event-frontend/node_modules/@embroider/core/src/app.js:343:18)
    at AppBuilder.updateAssets (/home/iamareebjamal/git/open-event-frontend/node_modules/@embroider/core/src/app.js:411:27)
    at AppBuilder.build (/home/iamareebjamal/git/open-event-frontend/node_modules/@embroider/core/src/app.js:505:38)
    at WaitForTrees.buildHook (/home/iamareebjamal/git/open-event-frontend/node_modules/@embroider/core/src/build-stage.js:37:31)
    at runMicrotasks (<anonymous>)
  - code: [undefined]
  - codeFrame: Cannot find module './vendor/fullcalendar/dist/fullcalendar.js' from '/tmp/embroider/606e06/node_modules/@embroider/synthesized-vendor'
  - errorMessage: Cannot find module './vendor/fullcalendar/dist/fullcalendar.js' from '/tmp/embroider/606e06/node_modules/@embroider/synthesized-vendor'
        at WaitForTrees (@embroider/compat/app)
[...]

PR #607, opened 2020-11-10
fix caching bug

ef4:
The default implementation of `cacheKeyForTree` checks whether there are any customized methods that back the given tree name.

We use that same system for our own v2Trees, because it interacts correctly with addon's own customnized cache key implementations.

But we have been using the cacheKey for the `addon` tree only, and that is not correct, because if the addon tree is not customized, ember-cli will give us a key for it, even though the addon may have other customized trees that should cause the v2Tree to really not be cacheable.

### DIFF
--- a/packages/compat/src/v1-addon.ts
+++ b/packages/compat/src/v1-addon.ts
@@ -458,7 +458,25 @@ export default class V1Addon {
   }
 
   get v2Tree(): Tree {
-    return this.throughTreeCache('addon', 'v2Tree', () => mergeTrees(this.v2Trees, { overwrite: true }));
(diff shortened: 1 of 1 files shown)


ef4:
This was the cause of the elusive ember-cli-node-assets problem. It was elusive because it only happened if you had multiple different addons using the same copy of ember-cli-node-assets. The first one that ran would incorrectly end up cached, so that the second one would not get the right assets.

ef4:
Fixes #500 
Fixes #563 
Fixes #442

Turbo87:
nice, thanks for looking into this 👍

PR #621, opened 2020-11-19
Fingerprint scripts and styles in production

### LINKED ISSUE #289: Is there a way to fingerprint vendor.js?
Is there a way to fingerprint vendor.js and other assets like images and css?

simonihmig:
Fixes #289

Tested it locally in an app (`ember s --prod`). 
Do we have any test coverage here for the stage 3 part, especially the production variant? AFAICT we only have one test-package that runs its [tests in a production build](https://github.com/embroider-build/embroider/blob/master/test-packages/fastboot-app/package.json#L22) in CI.

### DIFF
--- a/packages/webpack/src/ember-webpack.ts
+++ b/packages/webpack/src/ember-webpack.ts
@@ -208,7 +208,7 @@ const Webpack: Packager<Options> = class Webpack implements PackagerInstance {
   }
 
   private async writeScript(script: string, written: Set<string>, variant: Variant) {
-    if (!variant.optimizeForProduction || script.match(/testem/)) {
+    if (!variant.optimizeForProduction) {
       this.copyThrough(script);
       return script;
     }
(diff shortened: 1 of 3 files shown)


simonihmig:
There is one issue here, that I could observe locally. However I believe this was just uncovered by the renaming of files, but the root cause is elsewhere. Created this issue: #622

simonihmig:
> AFAICT we only have one test-package that runs its tests in a production build in CI.

And it seems exactly this is failing here 🙈
Will need to investigate...

simonihmig:
> And it seems exactly this is failing here

Seems we have to explicitly exclude `testem.js` from fingerprinting, otherwise testing in a  production environment does not work. Note that `ember s --prod` works (when opening `/tests` in the browser), it seems just the testem integration is broken when fingerprinting everything.

A quick Google search revealed a similar fix for ember-cli in this old PR: https://github.com/ember-cli/ember-cli/pull/1230/files#diff-8098951b282ee3680dd2afa8d34b95c7e542d90d65961bb60cbf897779c54eadR411

lifeart (approved):

ef4:
Thanks. I used the existing `data-embroider-ignore` feature to avoid needing to special-case testem. I think it's reasonable to require people to annotate scripts that they don't want to go through the build.

PR #624, opened 2020-11-21
Fix ensureSafeComponent to not reuse registered component across owners

simonihmig:
See https://github.com/tildeio/ember-element-helper/pull/34#discussion_r528217743 for context.


lifeart (approved):

ef4:
Thanks for figuring this out.

Can we make this more conventional by keeping our state in an Ember Service instead?

Then we would only need a single level of WeakMap, because the WeakMap itself could live in the Service and therefore get destroyed when the app is destroyed.

simonihmig:
> Can we make this more conventional by keeping our state in an Ember Service instead?

@ef4 done!

PR #627, opened 2020-11-23
Support preloading of splitted route chunks in FastBoot

simonihmig:
This is just a dysfunctional first attempt, to start a conversation...

The idea is that when using route-based code splitting together with FastBoot, you could instruct the browser to immediately start loading the route chunk with a `<link rel="preload">` element, instead of letting the browser first load all the base chunks, evaluate and run them only to realize much later that another chunk needs to be loaded (by `@embroider/router`)  before any rendering can happen. 

Integrating that into the router is rather easy, as this WIP PR shows (this part is basically working), *however* the hard part is how to find out what the URL of the route chunk actually is. At first I naively thought I could just add the path to the global `_embroiderRouteBundles_`, but that fails miserably as that gives us just the original module path like `_route_/split-me`, and not what webpack emits as the actual loadable js file. (the `import()` part is transformed by webpack, so that works of course)

Given that webpack handles this opaquely, the problem seems considerably harder than initially thought...

It seems possible to make the chunk files have [deterministic names](https://v4.webpack.js.org/plugins/split-chunks-plugin/#splitchunksname), but that prevents them to be effectively fingerprinted, so not really an option.

There exists some built-in support for preloading and prefetching, using plugins based on their [HtmlWebpackPlugin](https://v4.webpack.js.org/plugins/html-webpack-plugin/), but we don't use this plugin here, and I also don't see how this would integrate well with FastBoot and our router.

The closest I could find in the wild was the `Loadable Components` library for React-based SSR, which AFAIU seems to generate a stats file using a custom webpack plugin, collecting the necessary information about chunks and what not, and importing that again in their actual app code to extract the necessary bits: See this: https://loadable-components.com/docs/server-side-rendering/#prefetching

Any thoughts?

### DIFF
--- a/packages/core/src/app.ts
+++ b/packages/core/src/app.ts
@@ -1247,6 +1247,7 @@ w._embroiderRouteBundles_ = [
   {{#each lazyRoutes as |route|}}
   {
     names: {{{json-stringify route.names}}},
+    path: "{{js-string-escape route.path}}",
     load: function() {
       return import("{{js-string-escape route.path}}");
     }
@@ -1260,6 +1261,7 @@ w._embroiderEngineBundles_ = [
   {{#each lazyEngines as |engine|}}
   {
     names: {{{json-stringify engine.names}}},
+    path: "{{js-string-escape engine.path}}",
     load: function() {
       return import("{{js-string-escape engine.path}}");
     }
--- a/packages/router/addon/index.js
+++ b/packages/router/addon/index.js
@@ -5,6 +5,7 @@ import EmberRouter from '@ember/routing/router';
 import { registerWaiter, unregisterWaiter } from '@ember/test';
 import { DEBUG } from '@glimmer/env';
 import { macroCondition, getGlobalConfig } from '@embroider/macros';
+import { getOwner } from '@ember/application';
 
 let Router;
 
@@ -30,6 +31,15 @@ if (macroCondition(getGlobalConfig()['@embroider/core']?.active)) {
     }
 
     return false;
+  };
+
+  const preloadBundle = function preloadBundle(document, url) {
+    let link = document.createElement('link');
+    link.setAttribute('rel', 'preload');
+    link.setAttribute('href', url);
+    link.setAttribute('as', 'script');
+
+    document.head.appendChild(link);
   };
 
   Router = EmberRouter.extend({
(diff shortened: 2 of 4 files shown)


ef4:
Yes, I think we should use fastboot to compute which bundles are needed to do the initial render of a given route.

The benefit of this approach is that it can be robust to whatever techniques people use for splitting code. It's not webpack specific, and it's not specific only to route splitting (you can put arbitrary `await import` throughout your app and we will still find those dependencies).

Practically, what's needed is extending the fastboot environment to be a bit more browser-like so that real bundle loading works in fastboot, through a polyfilled `fetch` API that is instrumented to record which things were actually fetched. Right now, we always preload all chunks in fastboot because webpack's bundle loader doesn't work.

simonihmig:
> Yes, I think we should use fastboot to compute which bundles are needed to do the initial render of a given route.

I needed a bit to think about it. At first thought it sounded like a really clever idea. At second I am not so sure anymore though tbh. Basically my concern is if that is not *too clever*. More specifically...

First a pragmatic concern: it seems like a long journey to get there. Like extending FastBoot (I haven't really looked into it, but maybe this requires bigger changes, breaking changes, an RFC, etc.?), providing that fetch polyfill with recording functionality (a new project on its own), bringing all together. Given how OSS works (authoring things just when your (mostly spare) time allows, getting people on board to these changes, waiting for PR reviews), it seems we are talking about months at least.

Second, I am not sure if we are introducing too much "magic" and basing that approach on too many assumptions, that might work for many cases, but not all. The thing is that when recording all fetch requests, we have to apply some heuristics for what to do with these requests, as we don't really know what these recorded requests are all about (their semantics):
* should all be preloaded? Or only scripts?
* if we want to also preload generic fetch requests like API requests? What if they depend on some headers being sent, like auth tokens?
* what should the [`as`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-as) be derived from, the response's MIME type?
* what if the user has already through some other means added preload links for the same URLs? (account for absolute vs. relative URLs)
* can we really derive reliably all what is necessary for a valid preload link, e.g. the `crossorigin` attribute?
* what if we `import()` conditionally based on FastBoot vs. browser? Take this example:

```js
if (this.fastboot.isFastBoot) {
  return await import('some-fastboot-only-module');
} else {
  return await import('some-browser-only-module');
}
```

During FastBoot's `visit()` it would only see the fetch request for `some-fastboot-only-module.js`. When I understand you correctly, this would make FastBoot render a `<link href="/some-fastboot-only-module.js" rel="preload" as="script"/>` link, right? However that is not relevant for the browser, may not even work there (uses `FastBoot.require`), be potentially huge and so worsen the performance. What we really want is for the browser to preload `some-browser-only-module.js`, but FastBoot will never see a request for this.

> We actually have a similar real world use case, that triggered my desire for lazy loading, where we render a simple HTML component in FastBoot for the quick initial paint, and then in the browser we lazy load the fancy component (that imports a big browser library) for a progressive enhancement like effect.

In contrast for the approach that I *tried* here, the scope is much more limited. Everything is handled within Embroider. We preload only the route-splitted chunk, which *we* own, which *we* `import()` in *our* router, which *we* know for sure what it is and how it is used. For the app, this is a black box, an implementation detail, so we can also be pretty sure that the app is not interfering somehow and we are not interfering with the app. So I think none of the potential problems listed above apply here.

ef4:
I don't think this is as hard as it sounds.

The API we'd need to hook into (fetch and/or XMLHttpRequest) already has Node implementations with spec-compliant API. Fastboot already supports this stuff, which is why routes can fetch API data when rendering in Fastboot. We're not adding anything there.

We control not just the client but also the "server" for these requests. So determining which requests are hitting our own Javascript bundles, and therefore deserve preload tags, is not hard. We can get very precise about it -- we know the naming pattern for the chunk files, because we picked it, and we only need to preload those chunk files, nothing else.

Issue #568, opened 2020-10-13
Support for disabled in-repo addons


mehulkar:
If you ave an in-repo addon that sets `isEnabled() { return false }`, Embroider fails on a somewhat innocuous line of code:

https://github.com/embroider-build/embroider/blob/c4b20e5a0ee16f1c23b7f0c3baf5eaf42780214f/packages/compat/src/compat-addons.ts#L172

Debugging a bit, it looks like the addon is made available in the tmp directory, but its `package.json` file has an empty json object in it `{}`, causing the `path.join` to fail with:

```js
Build Error (WaitForTrees)

The "path" argument must be of type string. Received type undefined
```

mehulkar:
Haven't tried if this also reproduces for external addons.

iamareebjamal:
Receiving the same error which I think is same as #302

mehulkar:
Looks very similar, thanks! The only difference is that #302 is repro'ing with the `blacklist` feature of Ember CLI, whereas this is with the `isEnabled()` hook. 

Which makes me thing the fix is actually for ember-cli or even broccoli to exclude folders of addons that aren't going to get built. @rwjblue @stefanpenner do you have any immediate thoughts? Or can you confirm my understanding?

Issue #302, opened 2019-08-02
in-repo-addon blacklisting issues


Windvis:
Hey!

I'm trying out Embroider in our project and I encountered some build error which I'm not sure how to resolve.

```
The "path" argument must be of type string. Received type undefined
```
This error is triggered by [this line of code](https://github.com/embroider-build/embroider/blob/7b7be78744150e79de101ea3189898da9ee29688/packages/compat/src/compat-addons.ts#L107), because the name property is undefined, but only for our in-repo-addons (it seems).

I tried debugging to find some clues, and it looks like it has something to do with the [`this.packageCache.moved`](https://github.com/embroider-build/embroider/blob/7b7be78744150e79de101ea3189898da9ee29688/packages/compat/src/compat-addons.ts#L103) map. The name property is defined in the key object, but not in the value. I tried following the breadcrumbs through the codebase to see where this `moved` map is created, but I feel it's way out of my league, and it's getting late as well :sleepy: :smile: .

Any pointers about what the problem could be?

lifeart:
Hi @Windvis! Thank you for report! Could you please, create minimal repoduction repo?

Windvis:
Hmm, I just tried setting up a new ember app + in-repo-addon + Embroider, and it builds perfectly fine.. So the in-repo-addon error might be a symptom of a different problem. Any pointers how to debug this further?

ef4:
Do your in-repo addons have names? They need a valid package.json with a name. The generator gives them one, which may be why your new reproduction didn't fail.

Windvis:
Yea, they do. That's one of the first things I checked as well. When I debugged the code I also saw that they all had names. It just seems to get lost somewhere when "moving" them. The keys in the `packageCache.moved` map contain the name, but they are undefined in the value objects. I couldn't find where that map was being built. Anyway, since the repro project didn't have any problems I guess it's something we do wrong in our project. It's just hard to find out what :smile:.

Windvis:
@ef4 I did some more digging and I found out that it is related to the blacklisting of in-repo-addons.

We created an in-repo-addon wrapper around bugsnag which we blacklist during development. Once I removed the blacklist everything compiled as usual. (I didn't test if this also affects blacklisting of normal addons).

I created a [minimal reproduction repo](https://github.com/Windvis/embroider-blacklist-example-repo) for the in-repo-addon blacklist scenario. Building the master branch will fail with the aforementioned error, while building on the `pre-blacklist` branch will work.

PR #639, opened 2020-11-26
Handle disabled in-repo addons

mydea:
This fixes https://github.com/embroider-build/embroider/issues/568 and https://github.com/embroider-build/embroider/issues/302.

The change might be a bit simplistic, but it works. Not sure if that is acceptable, I tried it in our app and the build worked fine.
Only in-repo addons seem to be affected, it seems to be fine with regular disabled addons.

### DIFF
--- a/packages/compat/src/build-compat-addon.ts
+++ b/packages/compat/src/build-compat-addon.ts
@@ -10,7 +10,7 @@ import EmptyPackageTree from './empty-package-tree';
 
 export default function cachedBuildCompatAddon(originalPackage: Package, v1Cache: V1InstanceCache): Node {
   let tree = buildCompatAddon(originalPackage, v1Cache);
-  if (!originalPackage.mayRebuild) {
+  if (!originalPackage.mayRebuild && !(tree instanceof EmptyPackageTree)) {
     tree = new OneShot(tree);
   }
   return tree;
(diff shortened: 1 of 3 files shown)


ef4:
Thanks for taking a stab at this. 

> The change might be a bit simplistic, but it works. Not sure if that is acceptable,

Yeah, my concern would be that this is relying on an accidental behavior (why do disabled in-repo addons have no name?) and we don't have a way to notice if we break this in the future.

A place where you could try to more directly check for disabled addons would be here:

https://github.com/embroider-build/embroider/blob/cf26ead2070f687b60e2361f897d7d237772f4d5/packages/compat/src/build-compat-addon.ts#L30

Right now it checks for addons that never got instantiated at all, but it could be extended to also check for addons that are instantiated but returning `false` from `isEnabled()`. Something like this:

```js
// extending the condition I linked to above:
if (oldPackages.length === 0 || oldPackages.every(p => !p.isEnabled)) {
    return new EmptyPackageTree();
}

// and in class V1Addon, we need to expose `isEnabled`:
class V1Addon {
  get isEnabled(): boolean {
    return this.addonInstance.isEnabled();
  }
}

```

mydea:
I will take a look and see if I get somewhere ;)

mydea:
OK, it seems that is already happening! For blacklisted or disabled addons, `oldPackages.length` is 0.

I have made a second stab, this time by checking if the tree is an empty tree and handling it as disabled for this case. Is that a better direction? It's rather tricky as everything is so intertwined :sweat_smile:

ef4:
Thanks for working on this, I found a simpler solution in #659

Issue #644, opened 2020-12-01
Build fails with ember engines + embroider in a monorepo configuration


chriseppstein:
I've got an issue that is blocking me from landing embroider support to http://github.com/linkedin/eyeglass

I've reproduced the issue here (without eyeglass involved): 

Steps to reproduce issue:

1. `git clone https://github.com/chriseppstein/ember-engines-and-embroider-build-issue.git`
2. `cd ember-engines-and-embroider-build-issue`
3. `yarn install`
4. `cd packages/my-embroider-app`
5. `yarn build`
6. If Step 5 succeeds, run it again.

thoov:
@chriseppstein Thanks for the steps to reproduce!

TLDR: this is a bug introduced with my lazy engine CSS PR (https://github.com/embroider-build/embroider/pull/591/)

The error produced is: 

```
Build Error (PackagerRunner) in assets/_engine_/my-lazy-engine.js

Module not found: Error: Can't resolve '../../../my-lazy-engine/my-lazy-engine.css' in '$TMPDIR/embroider/fbe523/packages/my-embroider-app/assets/_engine_'
```

Looking at the output of stage 2 we see:

```js
import { importSync as i, macroCondition, getGlobalConfig } from '@embroider/macros';
let w = window;
let d = w.define;

if (macroCondition(!getGlobalConfig().fastboot?.isRunning)) {
  i("../../../my-lazy-engine/my-lazy-engine.css");
}

d("my-lazy-engine/config/_environment_browser_", function(){ return i("../../../my-lazy-engine/config/_environment_browser_");});
d("my-lazy-engine/config/environment", function(){ return i("../../../my-lazy-engine/config/environment");});
d("my-lazy-engine/engine", function(){ return i("../../../my-lazy-engine/engine");});
d("my-lazy-engine/package", function(){ return i("../../../my-lazy-engine/package");});
d("my-lazy-engine/templates/application", function(){ return i("../../../my-lazy-engine/templates/application");});
```

The issue is that:

```js
i("../../../my-lazy-engine/my-lazy-engine.css");
```

Was added via: 
https://github.com/embroider-build/embroider/pull/591/files#diff-23d24d20889abc4014e995cea006d0f41c5fc66151a977385522474151162448R1049 even though the engine does not contain styles.

chriseppstein:
[...]

PR #645, opened 2020-12-01
Conditionally import lazy engine css

thoov:
Fixes: #644

### DIFF
--- a/packages/compat/tests/stage2.test.ts
+++ b/packages/compat/tests/stage2.test.ts
@@ -657,4 +657,74 @@ describe('stage2 build', function () {
       assertFile.matches(/console\.log\(true\)/);
     });
   });
+
+  describe('engines', function () {
+    let build: BuildResult;
+    let expectFile: ExpectFile;
+
+    beforeAll(async function () {
+      let app = Project.emberNew();
+      let buildOptions: Partial<BuildParams> = {
+        stage: 2,
+        type: 'app',
+        emberAppOptions: {
+          tests: false,
+          babel: {
+            plugins: [],
+          },
+        },
+        embroiderOptions: {},
+      };
+
+      app.linkPackage('ember-engines');
+
+      let lazyEngine = app.addEngine('lazy-engine', true);
+      let eagerEngine = app.addEngine('eager-engine', false);
+
+      merge(lazyEngine.files, {
+        addon: {
+          styles: {
+            'addon.css': `.lazy { background-color: red; }`,
+          },
+        },
+      });
+
+      merge(eagerEngine.files, {
+        addon: {
+          styles: {
+            'addon.css': `.eager { background-color: blue; }`,
+          },
+        },
+      });
+
+      build = await BuildResult.build(app, buildOptions);
+      expectFile = expectFilesAt(build.outputPath);
+    });
+
+    afterAll(async function () {
+      await build.cleanup();
+    });
+
+    test('lazy engines appear in _embroiderEngineBundles_', function () {
+      expectFile('assets/my-app.js').matches(`w._embroiderEngineBundles_ = [
+  {
+    names: ["lazy-engine"],
+    load: function() {
+      return import("./_engine_/lazy-engine.js");
+    }
+  },
+]`);
+    });
+
+    test('lazy engine css is imported', function () {
+      expectFile('assets/_engine_/lazy-engine.js')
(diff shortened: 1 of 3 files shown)


stefanpenner (changes requested):
Thanks for jumping on this so quick. Is it possible to add test coverage for this scenario, to help ensure the issue is fixed and prevent accidental future regression?

chriseppstein:
I left a note on #644 as well, but I don't think this fix will fix my particular issue. I've updated my reproduction to include a css file -- the error still occurs in that case.

thoov:
> I left a note on #644 as well, but I don't think this fix will fix my particular issue. I've updated my reproduction to include a css file -- the error still occurs in that case.

@chriseppstein: I left a comment on the issue but this PR will still fix the new reproductions

>  Is it possible to add test coverage for this scenario, to help ensure the issue is fixed and prevent accidental future regression?

@stefanpenner I will chat with @ef4 about this. Right now its kinda heavy handed (we have to create a new engine). I will look into if there is a lighter weight method way we can test this.

thoov:
@stefanpenner I updated with tests if you want to review again

@ef4 This should be good to go

packages/compat/tests/stage2.test.ts
```diff
@@ -657,4 +657,111 @@ describe('stage2 build', function () {
...
+    names: ["lazy-engine"],
+    load: function() {
+      return import("./_engine_/lazy-engine.js");
+    }
+  },
+]`);
+    });
+
+    test('lazy engine css is imported', function () {
+      expectFile('assets/_engine_/lazy-engine.js')
```
stefanpenner, line 720: So, one way I test css stuff in an acceptance test Is to start the app, and inspect the relevant element's style property. Rather than parsing/matching content in output files. Is this relevant here?

The current approach indirectly tests the system, based both on what appears to be build specific internals but we have little confidence that the css is loaded, and loaded in the right order.

stefanpenner (changes requested):

packages/compat/tests/stage2.test.ts
stefanpenner: Speaking with travis, it seems this PR should add an acceptance test scenario for this case. And (as time permits, or maybe as a follow up) adjust how we test these intermediate stage 2 outputs.

For example, today we are testing that a file contains a strong, but that isn't a very durable mechanism for testing.

Rather, we should execute these files in a carefully constructed environment and ensure they have the side-affects we expect. It seems like these files are not one-off, and having this machinery will enable testing all of them.

One thought would be to (with a modern version of node). Context a vmcontext that is able to execute these files.

https://nodejs.org/api/vm.html#vm_class_vm_module
Given that the vm context now has, although experimental, support for SourceTextModules creation, linking, and evaluation. see: https://nodejs.org/api/vm.html#vm_class_vm_module.

My thought here is, we invest a little in such a harness then we can execute actual tests against them. Ones who do not care about the actual source-code, but only the side-affects.

Only risk I see, is that the node API is currently unstable. If that is a concern, this could be implemented as a code-transform instead, although I would likely recommend falling back to a code-transform only if the node folks fundamentally break this api.

ef4: This is a misunderstanding of how our tests are structured.

If you want to go end-to-end with actual execution in the browser, use one of the many test packages that run entire ember test suites. Don’t go adding VM execution to these tests, it’s fake and inferior to browser testing anyway. 

Tests line the ones in stage2.test.ts are deliberately testing a static view of the build, not a runtime one.

stefanpenner: @ef4 the above is two thoughts, does your comment apply to both points or just the vm thought?

Specifically, are you concerned with:

> Speaking with travis, it seems this PR should add an acceptance test scenario for this case [blocker].

thoov: @ef4 / @stefanpenner Should I add an additional engine that doesn't contain any styles or are we happy with theses tests

ef4: I would be OK with adding an end-to-end test but don't consider it a blocker.

What I was trying to say in reply to @stefanpenner above is that I most definitely do not want any tests that try to evaluate browser code inside node's vmcontext. Any time someone feels tempted to do that, they should move the test into one of our many ember apps to test in the browser instead. These node tests are deliberately looking at a state view of the output, not a runtime view. It's important to have both views, because there are things you can't prove with only a runtime view (for example -- that dead code isn't present).

PR #653, opened 2020-12-09
allow publicAssetURL to differ from rootURL

ef4:
This serves as a replacement for the classic `prepend` in broccoli-asset-rev.


ef4:
This also slightly optimizes the HTML when not using fastboot, because before we were always emitting `<fastboot-script>` for lazy chunks, even when there's no fastboot in use.

PR #661, opened 2021-01-04
v2 addon shim

ef4:
This will depend on a concurrent release of ember-auto-import, because that is where we polyfill most of the behavior for classic builds. The tl;dr is: v2 addons will just be auto-importable NPM packages that get handled by ember-auto-import in classic builds. This involves extending ember-auto-import so it supports all the v2 features (`@embroider/hbs-loader`, `@embroider/macros`) in those packages.


ef4:
Corresponding ember-auto-import changes are https://github.com/ef4/ember-auto-import/pull/324

ef4:
Closing in favor of https://github.com/embroider-build/embroider/pull/732

PR #666, opened 2021-01-13
Update to Terser@5.5.1

### LINKED ISSUE #665: Terser outputting error as a string instead of throwing
Embroider is currently using Terser@^3.16.1, where an error in the code will have Terser's `minify` function return a string of the error.

Code here: https://github.com/embroider-build/embroider/blob/master/packages/webpack/src/ember-webpack.ts#L253

Example:

```
"error: Z [SyntaxError]: Unexpected token: punc (.)"
```
This would be if an error was thrown, but the error is a string. And the code output from Terser's `minify` function is undefined. This prevents the app from ever loading, since there is no code, and impossible to debug.

Starting with Terser@5, `minify` function is async and rejects when there's an error in the code it is processing.

The fix would be to update to `Terser@5` which involves updating the usage of `minify` to be async.

GCheung55:
Fixes #665 to surface errors from Terser `minify` function.

### DIFF
--- a/packages/webpack/package.json
+++ b/packages/webpack/package.json
@@ -39,7 +39,7 @@
     "pkg-up": "^3.1.0",
     "source-map-url": "^0.4.0",
     "style-loader": "^1.1.4",
-    "terser": "^3.16.1",
+    "terser": "^5.5.1",
     "thread-loader": "^1.2.0",
     "webpack": "^4.43.0"
   },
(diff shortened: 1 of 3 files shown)


ef4:
Thanks.

raido:
This also fixes null coalescing operator error with Ember.js v3.24 - https://github.com/emberjs/ember.js/blob/master/packages/@ember/-internals/runtime/lib/mixins/array.js#L1350

PR #669, opened 2021-01-16
Add `this` param to `ensureSafeComponent` import example

patocallaghan:
I was trying out this exact example myself and as written, it was throwing the following error: `Cannot read property 'Symbol(OWNER)' of undefined`. 

Adding `this` as a second param, similar to the examples above, makes it work.


simonihmig (approved):
Good catch! :+1:

PR #670, opened 2021-01-19
bugfixes for production webpack-handled css

ef4:
- MiniCssExtractPlugin emits css chunks from what were originally JS assets, so our placeholder rewriter needs to handle type switching
  - apps with NPM namespaces needed to created missing directories when writing production CSS.


mattmcmanus:
@ef4 this seems safe to merge, right?

PR #687, opened 2021-02-09
clear main fields from v1 addons when rewriting them

ef4:
v1 addons don't get to customize `main` in package.json. When rewriting them to v2, we always rewrite them to put their main at `index.js` anyway. The presence of these fields can confuse stage3 packagers.

### DIFF
--- a/packages/compat/src/rewrite-package-json.ts
+++ b/packages/compat/src/rewrite-package-json.ts
@@ -26,6 +26,13 @@ export default class RewritePackageJSON extends Plugin {
       this.getMeta()
     );
     pkg['ember-addon'] = meta;
+
+    // classic addons don't get to customize their entrypoints like this. We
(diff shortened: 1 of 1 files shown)


rwjblue:
> v1 addons don't get to customize main in package.json

I don't understand this. V1 addons definitely _can_ customize `main` (I have used htis a number of times to make `typescript` smart enough to know that `addon/index.ts` is the right entry point for example).

rwjblue:
Ah, I see. This is talking about the interim state while upgrading them to v2, sorry for the noise.

ef4:
Correct, this is about how they look after being auto-upgraded to v2. We do take into account any customized main as they’re being rewritten, but the output is never customized.

rwjblue:
Ya, sorry about that.

PR #690, opened 2021-02-12
Exclude 'babel-plugin-compact-reexports' during Stage 1 build

charlespierce:
Info
-----
* The plugin `babel-plugin-compact-reexports`, included by `ember-engines`, translates an ES Modules re-export statement (`export { foo } from './bar'`) into a single-line AMD declaration, which saves space compared to the multiple lines an explicit import & export would use.
* However, during the Stage 1 Embroider build, we want to leave the ES Modules exports as they are, so that they can be properly parsed during Stage 3.

Changes
-----
* Analogous to the existing code that removes other invalid plugins during Stage 1, added a section to detect and exclude `babel-plugin-compact-reexports` during a stage 1 build.

Tested
-----
* Performed a `STAGE1_ONLY` build and confirmed that the reexport statements are _not_ translated and are left as-is

### DIFF
--- a/packages/compat/src/detect-compact-reexports.ts
+++ b/packages/compat/src/detect-compact-reexports.ts
@@ -1,0 +1,13 @@
+import { PluginItem } from '@babel/core';
+
+export function isCompactReexports(item: PluginItem): boolean {
+  let pluginPath: string;
+  if (typeof item === 'string') {
+    pluginPath = item;
+  } else if (Array.isArray(item) && item.length > 0 && typeof item[0] === 'string') {
+    pluginPath = item[0];
+  } else {
+    return false;
+  }
+  return /(^|\/)babel-plugin-compact-reexports\//.test(pluginPath);
+}
--- a/packages/compat/src/v1-addon.ts
+++ b/packages/compat/src/v1-addon.ts
@@ -23,6 +23,7 @@ import modulesCompat from './modules-compat';
 import writeFile from 'broccoli-file-creator';
 import SynthesizeTemplateOnlyComponents from './synthesize-template-only-components';
 import { isEmberAutoImportDynamic } from './detect-ember-auto-import';
+import { isCompactReexports } from './detect-compact-reexports';
 import { ResolvedDep } from '@embroider/core/src/resolver';
 
 const stockTreeNames = Object.freeze([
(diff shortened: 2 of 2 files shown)


rwjblue (commented):
ember-engines doesn't use this plugin if module transpilation is disabled ([see here](https://github.com/ember-engines/ember-engines/blob/2a7a46a821db2ffc635dd144b25972722be533ad/packages/ember-engines/lib/utils/should-compact-reexports.js#L13)). Have you looked into why that flag isn't set (therefore preventing this plugin from being added in the first place)?

ef4:
We disable module compilation in [stage1 here](https://github.com/embroider-build/embroider/blob/7a916c3067e0f6d9a126482b9ab724a3f7125837/packages/compat/src/v1-addon.ts#L450) and [stage2 here](https://github.com/embroider-build/embroider/blob/7a916c3067e0f6d9a126482b9ab724a3f7125837/packages/compat/src/v1-app.ts#L222).

But some addons insist on invoking babel or rollup directly and eliminate modules before we get a chance to stop them. Maybe that is the case here.

charlespierce:
I think the issue here is actually just one of timing. It looks like the check for `compileModules` is performed within the `init()` method of the `ember-engines` mixin, so it happens _before_ embroider is invoked at all, and well before it gets the chance to update the configuration. At that point, it's adding `compact-reexports` to the babel plugins.

Then later on when Embroider updates `compileModules`, it's too late since the plugin was already added.

I'm definitely open to a different approach for avoiding that plugin if there's a cleaner way forward.

rwjblue (approved):
This seems good to me, I don't see how we can modify ember-engines to avoid adding the babel plugin. As @charlespierce mentioned above ember-engines does this in its `init`:

https://github.com/ember-engines/ember-engines/blob/2a7a46a821db2ffc635dd144b25972722be533ad/packages/ember-engines/lib/engine-addon.js#L481-L491

When this is ran (at addon `init`) Embroider hasn't started loading, and there is no way to introspect that we will _eventually_ be in an Embroider build.

rwjblue:
@charlespierce - Ready for a rebase (should fix canary failures).

rwjblue (approved):

PR #692, opened 2021-02-13
Add `{{macroMaybeModifier}}` template macro

### LINKED ISSUE #680: macroMaybeModifier needed?
I just came across a use case where I wanted to use `@embroider/macros` to conditionally apply a modifier. As we are in element space, we cannot simply use `{{#if}}` here. 

The same applied for setting attributes conditionally, but this is explicitly covered by `macroMaybeAttrs`. Do we eventually need an equivalent for this, `macroMaybeModifier`?

simonihmig:
Closes #680

### DIFF
--- a/packages/macros/src/glimmer/ast-transform.ts
+++ b/packages/macros/src/glimmer/ast-transform.ts
@@ -4,6 +4,7 @@ import dependencySatisfies from './dependency-satisfies';
 import { maybeAttrs } from './macro-maybe-attrs';
 import { macroIfBlock, macroIfExpression, macroIfMustache } from './macro-condition';
 import { failBuild } from './fail-build';
+import { maybeModifier } from './macro-maybe-modifier';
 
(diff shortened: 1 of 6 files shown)


simonihmig:
Just realized I should add some e2e tests in `macro-tests`. Will do that later...

Also no documentation yet, as we didn't have any for template macros yet, and I felt adding all of that was out of scope for this PR.

Also tests are failing for canary, but also for other PRs, so seems unrelated.

simonihmig:
> Just realized I should add some e2e tests in macro-tests. Will do that later...

Done!

ef4 (approved):
Thanks, this looks good.

simonihmig:
I marked this PR as draft to not accidentally merge this prematurely, as we should probably first try out this alternative approach before introducing a new API here: https://github.com/embroider-build/embroider/issues/680#issuecomment-779481710

rwjblue:
Replaced by #712

PR #694, opened 2021-02-19
Run importSync transform later, so ember-auto-import can support importSync

simonihmig:
This runs the transform of importSync in the exit hook, to allow ember-auto-import to do its own transformation *before* us. See https://github.com/ef4/ember-auto-import/pull/347

### DIFF
--- a/packages/macros/src/babel/macros-babel-plugin.ts
+++ b/packages/macros/src/babel/macros-babel-plugin.ts
@@ -86,12 +86,8 @@ export default function main(context: unknown): unknown {
           return;
         }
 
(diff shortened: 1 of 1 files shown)


simonihmig:
Unrelated canary failures here...

simonihmig:
@ef4 can we merge this?

rwjblue:
@simonihmig the canary failures have been fixed, mind rebasing?

rwjblue (approved):

simonihmig:
> the canary failures have been fixed, mind rebasing?

Done, green now! Thanks for the review!

PR #696, opened 2021-02-23
Resolve the template compiler from ember-source instead of project root

thoov:
We cannot assume this will be resolvable from the project root as it might be symlinked. Instead we should get the addon instance and grab the template compiler location directly from it. This is technically a more stable location of the compiler instead of hard coding a location.

This becomes relevant in ember-cli's test suite as they symlink the entire node_modules directory into a per test tmp directory and looking for a hard coded location fails.

### DIFF
--- a/packages/compat/src/compat-app.ts
+++ b/packages/compat/src/compat-app.ts
@@ -34,7 +34,6 @@ import { MacrosConfig } from '@embroider/macros/src/node';
 import bind from 'bind-decorator';
 import { pathExistsSync } from 'fs-extra';
 import { tmpdir } from 'os';
-import { dirname } from 'path';
 import { Options as AdjustImportsOptions } from '@embroider/core/src/babel-plugin-adjust-imports';
 
 interface TreeNames {
@@ -318,10 +317,6 @@ class CompatAppAdapter implements AppAdapter<TreeNames> {
     return this.configTree.readConfig().rootURL;
   }
 
-  private templateCompilerPath(): string {
-    return 'ember-source/vendor/ember/ember-template-compiler';
-  }
-
   strictV2Format() {
     return false;
   }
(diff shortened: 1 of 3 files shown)


rwjblue:
@thoov - Ready for a rebase (should fix canary failures).

thoov:
@rwjblue This is now green

packages/compat/src/compat-app.ts
```diff
@@ -317,10 +317,6 @@ class CompatAppAdapter implements AppAdapter<TreeNames> {
     return this.configTree.readConfig().rootURL;
   }
 
-  templateCompilerPath(): string {
-    return 'ember-source/vendor/ember/ember-template-compiler';
```
rwjblue, line 321: It seems like this is actually trying to resolve the path relative to the build itself. That is why it is `ember-source/vendor/ember-template-compiler.js` instead of where it actually is in the npm package `ember-source/dist/ember-template-compiler.js`.

This is where that is done:

https://github.com/emberjs/ember.js/blob/eff00a35624cd33f9f46f0991252182c367bb045/lib/index.js#L298-L300

packages/compat/src/compat-app.ts
```diff
@@ -377,6 +373,13 @@ class CompatAppAdapter implements AppAdapter<TreeNames> {
     };
   }
 
+  @Memoize()
+  resolveTemplateCompilerPath() {
+    let emberSource = this.oldPackage.app.project.findAddonByName('ember-source');
+    let templateCompilerPath = emberSource.absolutePaths.templateCompiler;
+    return templateCompilerPath;
```
rwjblue, line 380: This changes away from resolving the template compiler relative to the vendor asset output to resolving it in normal node modules resolution system. 

This means that it would be no longer possible for folks to provide a custom / overridden ember-template-compiler asset (before all they needed to do was put it in the right location in their vendor folder).

I think this might be perfectly fine, but I'm not 100% sure...

rwjblue:
I debugged this with @thoov this afternoon, and we figured out what is going on. In ember-cli's own test suite (in order to avoid multiple invocations of `npm install`) the setup of some of the tests symlinks the **entire** `node_modules` folder. Embroider already had code (in `packages/compat/src/moved-package-cache.ts`) that handles symlinked packages, but it only handles if the individual packages themselves were symlinked (not the `node_modules` folder, or a full `node_modules/@foo` scope folder). 

We have an alternative fix that works better in those cases (allowing the existing symlinked module dereferencing to be smarter).

PR #697, opened 2021-02-23
Refactor suite-setup-util to avoid knock on errors.

rwjblue:
This is a small refactor that does a few things:

* migrate the manual promise chaining to async/await
* unify the error handling (so that it is all done in the same way)
* migrate from `process.stderr.write` to `console.error` (`process.stderr.write` only accepts a string or buffer, if you pass it an instance of `Error` you get an error that actually masks the _real_ error)
* migrate to `process.exitCode` instead of `process.exit` (forcing exit with `process.exit()` _in general_ should be avoided, and doesn't seem required in this context)

### DIFF
--- a/test-packages/support/suite-setup-util.ts
+++ b/test-packages/support/suite-setup-util.ts
@@ -152,33 +152,29 @@ export async function emitDynamicSuites() {
   }
 }
 
+async function main() {
+  try {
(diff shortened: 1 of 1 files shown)


ef4 (approved):

rwjblue:
Rebased to get the canary build fixes...

PR #700, opened 2021-02-26
TemplateCompiler fixes / improvements (avoid monkey patch for Ember 3.27+)

rwjblue:
At a high level this PR does a few things:

- Avoid monkey patch template compiler on Ember >= 3.27
- Refactor template compiler patches to simplify and make it clearer to know when we can remove each patch
- Use Node's builtin `vm.createContext` to prevent global mutation during template compiler evaluation
- Remove references to the deprecated `registerPlugin` template compiler API (we were not using it)
- Update `TemplateCompiler#applyTransforms` to be safer for source -> source transformations (e.g. preserving entity encoding)

### DIFF
--- a/packages/core/src/patch-template-compiler.ts
+++ b/packages/core/src/patch-template-compiler.ts
@@ -15,124 +15,178 @@ import {
 import { NodePath } from '@babel/traverse';
 import { transform } from '@babel/core';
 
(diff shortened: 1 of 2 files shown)


ef4:
Nice, thanks!

PR #702, opened 2021-03-01
Handle case where node_modules are symlinked

thoov:
When moving packages the code handled the case where individual node modules were symlinked but did not handle the cases when all of node_modules (or sub directories) were symlinks themselves. This happens when ember-cli's test suite creates a cache of node_modules that it symlinks into individual test case tmp directories and tries to invoke builds from there.


rwjblue (approved):
Awesome, thanks for working through this! This definitely seems like the right solution here.

PR #703, opened 2021-03-02
Add 3.20 and 3.24 to list of matrix tests.

rwjblue:
Not all of our internal (or test) packages run `ember-try` scenarios, but this updates the ones that do to include ember-source@3.20 and ember-source@3.24.


thoov:
👍

PR #704, opened 2021-03-02
Avoid running CI jobs on both push and pull_request events.

rwjblue:
When pushing changes to a branch, CI would run **double** the jobs (one set of CI jobs for the branch push event, and another set of jobs for the `pull_request` event).

This change tweaks the CI hooks to only run for pull requests **targeting** `master` and branches named `master` (also allow `v*` style release branches, just in case we wanted them in the future).

Compare #703's (90 CI jobs) CI runs vs #694's (41 CI jobs).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,14 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+      - 'v*' # prior release branches (e.g. `v0.30.x` branch)
+    tags:
+      - 'v*'
+  pull_request:
+    branches: [ master ]
 
 jobs:
   discover_matrix:


thoov:
👍

ef4:
FWIW, this was intentional. They're supposed to be testing slightly different things. But I agree it's overkill.

rwjblue:
Hmmm. They run the same matrix as far as I could tell, how do they test different things? What was I missing?

Happy to revert, and also change to primarily pushing to my own fork to avoid the double running.

ef4:
I was under the impression that git's default PR checking tests against a new commit that merges the PR with the target branch, whereas the branch checking tests exactly the PR's head commit. But I guess I was just assuming that and didn't see any docs.

ef4:
Either way, I'm still fine with this change. I pushed a copy of this repo into a branch on a private repo and the single run of actions that resulted ate my entire free quota. 😅

PR #706, opened 2021-03-04
Remove `ember-cli-htmlbars` dependency in `@embroider/compat`.

rwjblue:
When introduced in 48fc86baff4cca03935435856468b318f4146724 this dependency was used (by `packages/compat/src/apply-ast-transforms.ts`) to import and re-use the broccoli plugin, but that code was removed in 07d3c461.

The only import we had for it was specifically for types, which we define internally (in `types/ember-cli-htmlbars/index.d.ts`).

These dependencies do not seem to be needed anymore.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -7247,7 +7247,6 @@ ember-cli-get-component-path-option@^1.0.0:
   integrity sha1-DXtZVVni+QUKvtgE8djv8bCLx3E=
 
 "ember-cli-htmlbars-3@npm:ember-cli-htmlbars@3", ember-cli-htmlbars@^3.0.0, ember-cli-htmlbars@^3.0.1, ember-cli-htmlbars@^3.1.0:
-  name ember-cli-htmlbars-3
   version "3.1.0"
   resolved "https://registry.yarnpkg.com/ember-cli-htmlbars/-/ember-cli-htmlbars-3.1.0.tgz#87806c2a0bca2ab52d4fb8af8e2215c1ca718a99"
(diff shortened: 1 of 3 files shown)


rwjblue:
Just to note the status here for anyone else stumbling by, this PR is not quite correct. There are tests in `packages/compat`'s test suite that do require `ember-cli-htmlbars-3`, and those tests **should** have failed with this change (but didn't). We need to figure out _why_ they still passed, then we can fix this PR by moving these down to `devDependencies` (which is the correct thing here, due to the tests).

yarn.lock
```diff
@@ -6832,7 +6821,18 @@ ember-cli-htmlbars-inline-precompile@^2.1.0:
     heimdalljs-logger "^0.1.9"
     silent-error "^1.1.0"
 
-ember-cli-htmlbars@^4.0.0, ember-cli-htmlbars@^4.0.8, ember-cli-htmlbars@^4.0.9, ember-cli-htmlbars@^4.2.0, ember-cli-htmlbars@^4.2.2, ember-cli-htmlbars@^4.2.3, ember-cli-htmlbars@^4.3.1:
+ember-cli-htmlbars@^3.0.0, ember-cli-htmlbars@^3.0.1, ember-cli-htmlbars@^3.1.0:
+  name ember-cli-htmlbars-3
```
rwjblue: This `name` alias is what is causing the tests to continue to pass. Someone in the build has `ember-cli-htmlbars@3` still, and when `yarn` was ran after removing the alias it did **not** remove the `name` field here.

Removing this `name` field makes the tests properly fail.

rwjblue (commented):
OK, so I've gotten to the bottom of why this was passing. Apparently, removing the aliasing in `packages/compat/package.json` then running `yarn` was "not enough" to remove the alias.

I've removed the `name` alias from `yarn.lock` (it seems newer versions of `yarn@1` do not preserve `name` if it isn't needed), confirmed tests failed properly, then brought it back.

PR #709, opened 2021-03-04
Fix the path to the on-disk styles file for in-repo engines

charlespierce:
Info
-----
* When using in-repo engines, the path generated to the engine's styles is malformed, resulting in an error like:
```
Can't resolve '../../../account-engine/account-engine/__COMPILED_STYLES__/account-engine.css' in '$TMPDIR/embroider/8840c8/assets/_engine_'
```
* There appear to be two issues: First, by passing the _filename_ into `explicitRelative`, the path has one too many `../` elements at the beginning (since it treats the file as a directory as well).
* Second, assuming that the engine will be in a top-level directory with the name equal to it's package name is wrong, as in-repo engines are contained in `lib/{package-name}`.

Changes
-----
* Updated the `implicit-styles` code to call `dirname` on the underlying `relativePath` (stripping off the filename from the path)
* Updated the `implicit-styles` code to use `engine.appRelativePath`, instead of hard-coding it to be the package name for the engine, making sure it includes the `lib/` if necessary.

Tested
-----
* Ran a build with an in-repo engine to confirm that the styles can be located by Webpack.
* Updated the test scenarios to include an in-repo addon and confirm that it builds and works as expected.

### DIFF
--- a/packages/compat/tests/stage2.test.ts
+++ b/packages/compat/tests/stage2.test.ts
@@ -719,7 +719,7 @@ describe('stage2 build', function () {
     test('lazy engine css is imported', function () {
       expectFile('assets/_engine_/lazy-engine.js')
         .matches(`  if (macroCondition(!getGlobalConfig().fastboot?.isRunning)) {
-i(\"../../../lazy-engine/lazy-engine.css\");
+i(\"../../node_modules/lazy-engine/lazy-engine.css\");
   }`);
     });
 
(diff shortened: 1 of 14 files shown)


rwjblue:
Can you add an in-repo engine with styles to the engines-host-app?

charlespierce:
Updated to include an in-repo addon in the `engines-host-app` project. Also manually confirmed that the test fails to build without the changes in this PR (for the reasons described).

PR #710, opened 2021-03-04
Fix case where autoRun = false

thoov:
When `this.appBoot` is an empty string this would result in a falsy condition when instead it should be considered truthy. This code path is uncovered when autoRun is set to false due to: https://github.com/embroider-build/embroider/blob/9d13424943e9023537ac6ddc330c6f0d0fb26f1c/packages/core/src/app.ts#L1017

### DIFF
--- a/packages/compat/src/v1-appboot.ts
+++ b/packages/compat/src/v1-appboot.ts
@@ -32,9 +32,8 @@ export class ReadV1AppBoot extends Plugin {
   }
 
   readAppBoot() {
-    if (!this.appBoot) {
-      throw new Error(`AppBoot not available until after the build`);
+    if (this.appBoot) {
+      return this.appBoot;
     }
-    return this.appBoot;
   }
 }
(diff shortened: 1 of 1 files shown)


rwjblue:
I know we poked at this together but I think we'll need a bit more details here before landing.

packages/compat/src/v1-appboot.ts
```diff
@@ -22,17 +22,19 @@ export class WriteV1AppBoot extends Plugin {
...
       persistentOutput: true,
     });
   }
   build() {
     this.appBoot = readFileSync(join(this.inputPaths[0], `config/app-boot.js`), 'utf8');
+    this.hasBuilt = true;
   }
 
   readAppBoot() {
-    if (!this.appBoot) {
```
rwjblue, line 35: FWIW, the prior guard had issues because in cases where no addon emits `{{content-for 'app-boot'}}` content the value will be `''`, which is going to evaluate to falsey.

rwjblue (approved):

PR #712, opened 2021-03-07
Allow macroCondition inside modifier

### LINKED ISSUE #680: macroMaybeModifier needed?

simonihmig:
Previously the macro-if transform would already allow to switch between two modifier implementations with a `(if (macroCondition ...) foo bar)` sub-expression inside a modifier. This now also completely removes the modifier if the sub-expression statically evaluates to undefined, allowing you to conditionally apply modifiers. See https://github.com/embroider-build/embroider/issues/680#issuecomment-779481710

Closes #680, supersedes #692.

Note this only works with Ember 3.25, which now allows you to use a sub-expression as the modifier path, see https://github.com/embroider-build/embroider/issues/680#issuecomment-779424024. Older versions would throw a parse error (not exactly sure at which version exactly the changes in the parser where introduced). Therefore I also had to replace the vendored template compiler (for tests) with that of Ember 3.25.

### DIFF
--- a/packages/macros/README.md
+++ b/packages/macros/README.md
@@ -89,6 +89,14 @@ Macros can also be used inside of templates:
 {{/if}}
 ```
 
+Starting with Ember 3.25 you can also use it to conditionally apply modifiers:
+
+```hbs
+<button {{(if (macroCondition true) on) "click" this.something}}>Submit</button>
+```
+
+However, in all cases the argument to `macroCondition` must be statically analyzable:
+
 ```js
 import { macroCondition } from '@embroider/macros';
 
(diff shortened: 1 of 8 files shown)


rwjblue (approved):
Nice, thanks @simonihmig!

PR #717, opened 2021-03-11
Explore a new Packager: ESBuild

NullVoxPopuli:
This is _super_ WIP, and I don't have planned dedicated time to work on this, so.. we'll see how long this takes :D 

ESBuild is still considered experimental (esp their plugin system), but I want to start a couple things with this PR:
 - track what remaining work needs to happen for ESBuild to be supported
 - Demonstrate how a packager can be implemented that isn't webpack
 - Potentially further extract shared utilities between multiple packagers


ef4:
This is good to work on.

My starting suggestion is to configure an ember app like this:

```js
  process.env.STAGE2_ONLY = 'true';
  return compatBuild(app, null, {
    staticAddonTrees: true,
    staticAddonTestSupportTrees: true,
    staticComponents: true,
    staticHelpers: true,
    workspaceDir: '../out-ember-app',
  });
```

and run `ember s`. Then you can work inside `../out-ember-app` which will be very close to consumable by esbuild. The main thing you'll need to adapt is that it will contain `_babel_config_.js` which you will need to translate into the esbuild equivalents, and `_template_compiler_.js` which you will need to install to handle .hbs files.

ef4:
I've just been going through this same exercise for snowpack and vite. 

You will most likely want the work on [this branch](https://github.com/embroider-build/embroider/tree/importSync-eager-mode) which I haven't had a chance to bring back into master yet.

NullVoxPopuli:
Gonna close this, because PR #759 from @alexlafroscia looks amazing! (and way more complete).
Also, Vite is on top of ESBuild! https://vitejs.dev/config/#esbuild (I had no idea!)

<3

PR #720, opened 2021-03-11
Improve error message when module cannot be found in webpack

thoov:
This fixes the "full path" of the source file of the error which allows it to be easier to copy and paste so you can inspect the file that is importing the module which does not exist.

Before:
Screen Shot 2021-03-05 at 10 29 37 AM

After:
Screen Shot 2021-03-11 at 10 32 01 AM

Notice how `schema-helpers.js` is missing in the first example so you would have to grep the whole directory to find out which file contained the bad import.


charlespierce (approved):
LGTM

rwjblue (approved):

Issue #724, opened 2021-03-16
Cannot `ember test --environment=production` an Embroider based application


rwjblue:
### Problem

At the moment, running `ember test --environment=production` does not work properly. This presents as the testem harness not realizing that the tests were properly loaded, and therefore timing out. The reason for htis is that when running tests the `/testem.js` asset **must** be loaded from the running testem middleware. However, as of https://github.com/embroider-build/embroider/pull/621 the fingerprinting that is done on `testem.js` breaks the test runners ability to communicate with the host environment.

In classic ember-cli builds, the `lib/broccoli/ember-app.js` instance had a `app.options.fingerprint.exclude` configuration which users could configure to **disable** fingerprinting for specific files or `app.options.fingerprint.enabled` field to disable fingerprinting for all files.

These settings are used by default in classic ember-cli builds here:

* https://github.com/ember-cli/ember-cli/blob/1301d24c8d7dec1f1559cffd0688d46bcd50284e/lib/broccoli/ember-app.js#L351-L355
* https://github.com/ef4/ember-auto-import/blob/4027f2a5d0d89d4af999ea09547f48876cd6cb64/packages/ember-auto-import/ts/auto-import.ts#L135-L148
* https://github.com/ember-cli/broccoli-asset-rev/blob/78f6047c15acb3bd348611f658b03bdd1041911f/index.js#L9

### Suggested Solution

Update `@embroider/webpack`'s fingerprinting setup to honor the `app.options.fingerprint` settings.

runspired:
I dug around on this a little tonight after realizing I couldn't use embroider for EmberData's test apps since we test prod builds. I was hoping it was just a matter of passing in the right webpack config to get it working but it looks like we've hard-coded fingerprinting https://github.com/embroider-build/embroider/blob/77ed393c11bf71eccd8c2ae1e5b8212880382264/packages/webpack/src/ember-webpack.ts#L342

PR #726, opened 2021-03-16
Respect fingerprinting options during stage 3

thoov:
We currently were fingerprinting all assets when in production mode which would break testing as testem.js would become fingerprinted and thus cause problems loading (ie if you did: `ember t --environment=production`). This threads through the fingerprinting options from the app and then determines if it should fingerprint or not from that. This makes the behavior inline with how classic builds work.

Fixes: #724

### DIFF
--- a/packages/core/src/packager.ts
+++ b/packages/core/src/packager.ts
@@ -2,8 +2,8 @@
 // are intended to be generic across Packagers, and it's up to Packager authors
 
 export interface FingerprintOptions {
-  enabled?: boolean;
-  exclude?: any[];
+  enabled: boolean;
+  exclude: any[];
 }
 
 // to support each option (or not).
--- a/packages/core/src/packager.ts
+++ b/packages/core/src/packager.ts
@@ -2,8 +2,8 @@
 // are intended to be generic across Packagers, and it's up to Packager authors
 
 export interface FingerprintOptions {
-  enabled: boolean;
-  exclude: any[];
+  enabled?: boolean;
+  exclude?: string[];
 }
 
 // to support each option (or not).
--- a/packages/core/src/packager.ts
+++ b/packages/core/src/packager.ts
@@ -1,5 +1,11 @@
 // This is a collection of flags that convey what kind of build you want. They
 // are intended to be generic across Packagers, and it's up to Packager authors
+
+export interface FingerprintOptions {
+  enabled?: boolean;
+  exclude?: any[];
+}
+
 // to support each option (or not).
 export interface Variant {
   // descriptive name that can be used by the packager to label which output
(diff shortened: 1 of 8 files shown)


rwjblue:
Getting a failure in `macro-tests` test-packages `test:ember`:

```
TypeError: Cannot read property 'exclude' of undefined
    at new Webpack (/home/runner/work/embroider/embroider/packages/webpack/src/ember-webpack.js:73:29)
    at PackagerRunner.build (/home/runner/work/embroider/embroider/packages/core/src/to-broccoli-plugin.js:28:33)
```

And the `static-app` production job that we added fails like:

```
not ok 1 Chrome - [undefined ms] - error
    ---
        message: >
            Error: Browser failed to connect within 90s. testem.js not loaded?
            Stderr: 
             
            DevTools listening on ws://127.0.0.1:46523/devtools/browser/41170ff8-1c66-43fa-88be-828d7339d0f6
```

packages/webpack/src/ember-webpack.ts
```diff
@@ -33,6 +34,7 @@ const debug = makeDebug('embroider:debug');
 // terser lazily so it's only loaded for production builds that use it. Don't
 // add any non-type-only imports here.
 import type { MinifyOptions } from 'terser';
+import { FingerprintOptions } from '@embroider/core/src/packager';
```
rwjblue, line 37: Since this is just for types, maybe use:

```suggestion
import type { FingerprintOptions } from '@embroider/core/src/packager';
```

test-packages/static-app/ember-cli-build.js
```diff
@@ -21,6 +21,12 @@ module.exports = function (defaults) {
     staticAddonTrees: true,
     staticComponents: true,
     staticHelpers: true,
+    skipBabel: [
+      {
+        package: 'qunit',
+        semver: '*',
+      },
+    ],
```
rwjblue, line 29: This is unfortunate, but at the moment Babel'ing QUnit results in `define` invocation here:

https://github.com/qunitjs/qunit/blob/f3d7f93f0ce23541559bace7d5d3af5a3f894ccc/src/export.js#L38-L41

And since our loader.js doesn't support anonymous AMD, it throws an error.

packages/core/src/packager.ts
```diff
@@ -1,5 +1,11 @@
 // This is a collection of flags that convey what kind of build you want. They
 // are intended to be generic across Packagers, and it's up to Packager authors
+
+export interface FingerprintOptions {
+  enabled?: boolean;
+  exclude?: any[];
```
rwjblue, line 6: I think its actually:

```suggestion
  exclude?: string[];
```

Not that that is a big deal though...

thoov: done

ef4:
I don't really think it's a good idea to pass these fingerprinting options to stage3. It assumes too much about stage3's implementation. The fingerprint options are expressed in terms of *output* filenames, but embroider gives you no guarantee any of those files will even exist in the output.

I think there is already a stage3 feature that covers this case: the `data-embroider-ignore` HTML attribute. Probably if we're going to read the classic fingerprint options and use them, we should use them to set that data attr. That would avoid the need to teach stage3 a new concept.

thoov:
@ef4 sounds good, I will implement something that adds those attributes.

Issue #725, opened 2021-03-16
Regression: Build Error: Don't know how to insertURL for sourcemaps


johanrd:
After 99a3202:
 
 1. Run `ember build` with sourcemaps:

``` 
webpackConfig: {
  devtool: 'source-map',
}
```
2. Experience Build Error:
```console
Build Error (PackagerRunner)
don't know how to insertURL /assets/chunk.9487b3713b2712e9dae6.js.map
```

Build does not break before 99a320288ac0b91ea9c8b962c73dace004f31fae. Probably caused by fallthrough condition `url.endsWith('.js')` for `js.map`. See https://github.com/embroider-build/embroider/commit/99a320288ac0b91ea9c8b962c73dace004f31fae#diff-f6871a419b7638a206e71923afc7910d1a3252cdfa5c563a5fb7e455063146b0L44-L54:

ef4:
Thanks for reporting. We should ignore these instead of erroring, because they're not supposed to be inserted into the HTML.

johanrd:
Ok, thanks, yes, that makes sense. 

That would also close #654 I guess.

naiyt:
I threw up a quick PR with a potential fix: https://github.com/embroider-build/embroider/pull/734

johanrd:
@naiyt Thanks. The fix is quite specific, but it does the job.

@ef4 If you don't see any obvious shortcomings to #734, it would be great to have it merged in, as now any project with sourcemaps is stranded on `v <=  35.1`.

johanrd:
Closing, as source maps does not seem to be referenced in `index.html` anymore when building with `v0.43`. 

Which means that #734 can be closed too I guess, @naiyt, I you experience the same as me.

PR #734, opened 2021-03-25
Do not try to insert source maps

naiyt:
This should resolve https://github.com/embroider-build/embroider/issues/725, which we recently ran into while trying out Embroider. The solution is to just ignore source maps, because those aren't included as tags in the HTML, and are just requested on demand by the browser.

## Replication

Create an Ember app with Embroider, and set this in your Embroider config in `ember-cli-build.js`:

```
        packagerOptions: {
            webpackConfig: {
                devtool: "source-map",
            },
        },
```

On master, running `ember serve` will error with `don't know how to insertURL <asset name>.js.map`. With these changes, it should build correctly (including the source maps).

### DIFF
--- a/test-packages/static-app/ember-cli-build.js
+++ b/test-packages/static-app/ember-cli-build.js
@@ -21,6 +21,13 @@ module.exports = function (defaults) {
     staticAddonTrees: true,
     staticComponents: true,
     staticHelpers: true,
+    packagerOptions: {
+      // Setting devtool to source-map to ensure a failure if there's a regression on
+      // https://github.com/embroider-build/embroider/issues/725
+      webpackConfig: {
+        devtool: 'source-map',
+      },
+    },
     packageRules: [
       {
         package: 'static-app',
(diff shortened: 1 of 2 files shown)


test-packages/static-app/ember-cli-build.js
```diff
@@ -21,6 +21,13 @@ module.exports = function (defaults) {
     staticAddonTrees: true,
     staticComponents: true,
     staticHelpers: true,
+    packagerOptions: {
+      // Setting devtool to source-map to ensure a failure if there's a regression on
+      // https://github.com/embroider-build/embroider/issues/725
+      webpackConfig: {
+        devtool: 'source-map',
+      },
+    },
```
naiyt, line 30: I wasn't sure of the best way to test this. It looks like there aren't unit tests for `html-placeholder.ts`, and I couldn't think of a better way other than "include the setting in one of the test apps, so the test suite will at least break if there's a regression". Let me know if there's a better way I could handle that.

naiyt:
Closing, since it looks like this was resolved as of v0.43

Issue #750, opened 2021-03-31
Incompatibility with Node 10 introduced in #748


rwjblue:
Looks like https://github.com/embroider-build/embroider/pull/748 introduced an incompatibility with Node 10 due to `Object.fromEntries` usage:

https://github.com/embroider-build/embroider/blob/031ce6d8a3137f5df6106be26c336b92ed948717/packages/compat/src/v1-addon.ts#L800-L802

Added in https://github.com/embroider-build/embroider/commit/56db384c516183dbbe6edd3c48cd2e574f0c4eb6.

PR #751, opened 2021-03-31
Use backwards compatible fromPairs over fromEntries for node 10 support

thoov:
Fixes: #750


ef4:
Thanks @thoov

ef4:
Published in 0.39.1.

PR #739, opened 2021-03-25
Add windows scenarios to CI matrix generator

rwjblue:
This runs out test suites for both Ubuntu and Windows (with the exception of linting).

This replicates the reported failures in https://github.com/embroider-build/embroider/issues/681 and https://github.com/embroider-build/embroider/issues/676.

PR #752, opened 2021-03-31
Windows CI matrix

thoov:
Contains the changes to get #739 passing under Windows.

### DIFF
--- a/packages/core/src/app.ts
+++ b/packages/core/src/app.ts
@@ -1271,8 +1271,9 @@ export class AppBuilder<TreeNames> {
           }
 
           let runtime = join(packageName, name).replace(this.resolvableExtensionsPattern, '');
-          if (renamedModules && renamedModules[runtime]) {
-            runtime = renamedModules[runtime];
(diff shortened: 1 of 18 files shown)


packages/core/src/app.ts
```diff
@@ -1271,8 +1271,9 @@ export class AppBuilder<TreeNames> {
           }
 
           let runtime = join(packageName, name).replace(this.resolvableExtensionsPattern, '');
-          if (renamedModules && renamedModules[runtime]) {
-            runtime = renamedModules[runtime];
+          let runtimeRenameLookup = runtime.split('\\').join('/');
```
thoov, line 1274: runtime will look like: `ember-source\ember\index` because of the join.

renamedModules will contain items like:

```js
{
...
'ember-source/ember/index:'ember/index'
...
```
which fails to do the lookup. This converts the lookup to be in the expected format. I didn't change the meta info format as I'm not sure the full ramifications of that but a longer term solution would be to look into that.

thoov:
@rwjblue / @ef4 could either of you review this

rwjblue (approved):

PR #753, opened 2021-04-01
Readme: Add splitAtRoutes to options

scottmessinger:
Clarify splitAtRoutes is an option that can be configured.


thoov:
Thanks, @scottmessinger

PR #770, opened 2021-04-21
Add compat adapter for `ember-get-config`

alexlafroscia:
I've been struggling to get `ember-get-config` to work with the Vite packager due to the fact that it's desired behavior is to re-export a module from the host app.

At @ef4's suggestion I totally replaced the original way of doing this with a new approach that is Embroider-compatible. We can look up the host app's config at build-time and serialize it into the module's `index.js`; this maintains the add-on's API without having the export from the host app.

One drawback is that the `config/environment` contents are now serialized twice in the payload, but I don't think that's necessarily a huge deal.

One other approach I had tried was using the `setOwnConfig`/`getOwnConfig` API from the `macros` package but

1. I couldn't get it working; I overrode the `options` getter and added the `@embroider/macros` property but `getOwnConfig` introduced into the module's `index.js` would return `undefined`
2. It ultimately was just extra steps; the approach taken here is simpler by cutting out that whole step

### DIFF
--- a/packages/compat/src/compat-adapters/ember-get-config.ts
+++ b/packages/compat/src/compat-adapters/ember-get-config.ts
@@ -1,0 +1,29 @@
+import V1Addon from '../v1-addon';
+import writeFile from 'broccoli-file-creator';
+import { join } from 'path';
+
+function createIndexContents(config: any): string {
+  return `export default ${JSON.stringify(config)};`;
+}
+
+/**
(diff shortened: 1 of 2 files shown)


alexlafroscia:
The CI failure seems like a browser timeout in both accounts, but I can't re-run CI myself

alexlafroscia:
Rebased from `master` in hopes of getting a passing build this time around

ef4:
Thanks for working on this.

PR #772, opened 2021-04-22
Allow compat adapter's to expose shouldApplyAdapter

thoov:
The use case is that if a compat adapter is found (either the built-in ones or custom ones) they will always be used. This method allows the compat adapters to opt out if say a newer version of a package fixes the issue.

### DIFF
--- a/packages/compat/src/v1-addon.ts
+++ b/packages/compat/src/v1-addon.ts
@@ -1010,7 +1010,7 @@ export interface V1AddonConstructor {
     orderIdx: number
   ): V1Addon;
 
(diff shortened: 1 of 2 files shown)


ef4:
Can we rename `isEnabled`? Because `V1Addon.prototype.isEnabled` looks like it controls whether the addon itself is enabled, not the compat adapter for the addon.

thoov:
@ef4 I agree, after I created the PR I did think that isEnabled isn't the best name for this. I changed it to shouldApplyAdapter.

Issue #780, opened 2021-04-27
`ember test` fails to run tests


timiyay:
We're experimenting with `embroider` adoption on a new Ember 3.24 app.

We're using 0.39, to avoid the issue where `@embroider/webpack` wants webpack 5, but other embroider packages or dependencies want webpack 4.
```
"@embroider/compat": "0.39.0",
"@embroider/core": "0.39.0",
"@embroider/webpack": "0.39.0",
"webpack": "^4.46.0",
```

Our test suite won't run when using `embroider` builds - the build completes but the tests never start.

```
$ ember test

Environment: test
⠋ Buildingbroccoli-babel-transpiler is opting out of caching due to a plugin that does not provide a caching strategy: `function TransformTestSelectorParamsToHashPairs() {
  this.syntax = null;
}`.
⠹ building... [OneShot > Babel: my-addon3 > applyPatches]broccoli-babel-transpiler is opting out of caching due to a plugin that does not provide a caching strategy: `function TransformTestSelectorParamsToHashPairs() {
  this.syntax = null;
}`.
⠦ building... [CSSModules]Building into /private/var/folders/1n/vdckwbw10md6ltblrlff8k6m0000gn/T/embroider/93e967
⠋ building... [@embroider/compat/app]WARNING: Your build is slower because some babel plugins are non-serializable
⠇ building... [@embroider/webpack][BABEL] Note: The code generator has deoptimised the styling of /Users/tasquith/Blake/my-ember-app/node_modules/axe-core/axe.js as it exceeds the max of 500KB.
⠴ building... [@embroider/webpack]Hash: aab2e6330492a12f11ae
Version: webpack 4.44.2
Child
    Hash: aab2e6330492a12f11ae
    Time: 13749ms
    Built at: 27/04/2021 10:21:57 am
                            Asset      Size                                                   Chunks                                Chunk Names
    chunk.46822e32daf463bfa117.js   322 KiB                                           assets/test.js  [emitted] [immutable]  [big]  assets/test.js
    chunk.52c43ca7ce4038c20152.js  6.41 KiB                         assets/my-ember-app.js  [emitted] [immutable]         assets/my-ember-app.js
    chunk.6c95fd2602af60083ec1.js  1.21 MiB                                   vendors~assets/test.js  [emitted] [immutable]  [big]  vendors~assets/test.js
[...]

PR #782, opened 2021-04-27
docs: for setting publicAssetUrl in non-production environments

timiyay:
This PR updates the `README.md` section for setting `publicAssetURL`, so the default example works for `development` and `test` builds.

This is a follow-up action for https://github.com/embroider-build/embroider/issues/780, which ended up being caused by using the default example in the current `README.md`.


ef4:
Thanks!

PR #789, opened 2021-04-27
Update README for webpack install requirement

bendemboski:
Now that apps need to bring their own webpack, add it to the "add dependencies" getting started step in the README


ef4:
Thanks, good catch.

PR #792, opened 2021-04-27
fix(test-setup): install `webpack` alongside `@embroider/webpack`

alexlafroscia:
Since `webpack` is a peer dependency to `@embroider/webpack` now, the `@embroider/test-setup` package should also install that as part of the "Ember Try" configuration that it generates.


ef4:
Thanks, good call.

PR #793, opened 2021-04-27
Fix test setup to include webpack

bendemboski:
Make it add webpack 5 so addons don't have to add it to their own devDependencies.


bendemboski:
Whoops! @alexlafroscia beat me to it in #792 😃

PR #795, opened 2021-04-28
Allow thread-loader configuration

bendemboski:
Currently `thread-loader` is not configurable, meaning that in some CI environments where the OS reports a very large number of CPUs, `thread-loader` will create a very large worker pool.

The only configuration that is currently available is setting `JOBS=1`, but the pool warmup call doesn't check `JOBS`, so always spawns the full number of workers.

So this commit:

* Skips warming the worker pool if `JOBS=1` and we know we won't be using `thread-loader`
* Adds a `thread-loader` configuration object to the options interface that allows applications to pass their own `thread-loader` configuration, or `false` to disable the use of `thread-loader` entirely.

### DIFF
--- a/packages/webpack/src/ember-webpack.ts
+++ b/packages/webpack/src/ember-webpack.ts
@@ -22,6 +22,7 @@ import makeDebug from 'debug';
 import { format } from 'util';
 import { tmpdir } from 'os';
 import { warmup as threadLoaderWarmup } from 'thread-loader';
+import Options from './options';
 import { HTMLEntrypoint } from './html-entrypoint';
 import { StatSummary } from './stat-summary';
 import crypto from 'crypto';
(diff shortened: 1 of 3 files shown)


bendemboski:
The context around this (for me) is that my embroider builds are failing in CI because the containers have 4GB of RAM, and that gets used up quickly spawning 35 or however many workers (which is unavoidable currently because of the worker pool warming). Just the change for `JOBS=1` to disable the warmup would fix the issue, but it'd also be great to be able to keep concurrency, just limit it to a reasonable number of workers.

But if there are any concerns about the configuration/options change, I'd be happy to split the change where `JOBS=1` disabled the worker pool warmup into a separate PR as that would unblock my CI builds.

packages/webpack/src/ember-webpack.ts
```diff
@@ -76,6 +66,7 @@ const Webpack: Packager<Options> = class Webpack implements PackagerInstance {
   private extraConfig: Configuration | undefined;
   private passthroughCache: Map<string, Stats> = new Map();
   private publicAssetURL: string | undefined;
+  private extraThreadLoaderOptions: object | false | undefined;
```
avihai-yosef, line 69: maybe you can use ThreadLoaderOptions Type instead of object

bendemboski: I couldn't find types for `thread-loader` anywhere, and wasn't sure we'd want to sign up for maintaining our own type. We certainly could though...

avihai-yosef: Oh, I couldn't find either, nvm..

ef4:
Thanks for digging into this, this looks great.

PR #797, opened 2021-04-29
Use configPath to locate the configuration file, instead of assuming a fixed path

charlespierce:
Info
-----
The `compatAdapter` for `ember-get-config` currently assumes that the configuration file is always in `config/environment.js`. However, it is possible to change the config directory with a setting in `package.json`. If that is done, the compat adapter will fail when it tries to require a file that doesn't exist.

Changes
-----
* Updated `V1App` to provide access to Ember's method of locating the `environment` file.
* Updated the `ember-get-config` adapter to use the above, instead of assuming that it's always at `config/environment.js`


ef4:
Thanks.

PR #766, opened 2021-04-15
Update to broccoli-funnel@3.0.5

rwjblue:
Related:

* https://github.com/broccolijs/broccoli-funnel/pull/136
* https://github.com/SparshithNR/fs-merger/pull/43
* https://github.com/SparshithNR/fs-merger/pull/45 (the real bug that was fixed)

Closes https://github.com/embroider-build/embroider/issues/764

PR #802, opened 2021-05-06
Upgrade broccoli-funnel

asakusuma:
Upgrade now that https://github.com/broccolijs/broccoli-funnel/pull/136 has been released


rwjblue:
Tracking the fix over in #766

PR #803, opened 2021-05-07
Ensure extracted CSS comes before app/vendor CSS

### LINKED ISSUE #801: Ability to load third-party CSS before app CSS
I have an addon that wraps a third-party library, and that library provides CSS via webpack-style `import`s in Javascript files. It supports UI customization by defining CSS variables that the consuming app can override.

I can't figure out how to get this to work with embroider because all webpack-extracted CSS ends up put in `script`/`link` tags after the `<app>.css`. Consuming apps could play CSS tricks with higher specificity to make it work, but it seems like there should be a way to supply webpack-imported CSS that comes before `<app>.css` without requiring apps to modify their `index.html`.

NOTE: I don't want to do this in the broccoli phase because then I would lose the tree-shaking aspect and have to include all the CSS.

bendemboski:
CSS that webpack loads from JS imports currently gets added after the vendor and app CSS. This makes it difficult for addons or apps to customize this CSS, e.g. by overriding CSS variables or rules, because they cannot rely on document order for such overrides, so have to play tricks with specificity or the dreaded `!important`.

This commit moves such CSS to just before the first in-`index.html` stylesheet, ensuring that it comes before both the vendor and app CSS.

This also splits out the CSS handling from the JS handling a little more to make the render() method easier to follow, mainly by making it clear that the fastboot replacement stuff only applies to JS and not CSS.

Fixes #801


bendemboski:
I have tested this manually in a few scenarios, but I'm not sure if/how to do automated testing of it. If someone can point me in the right direction, I'd be happy to write some tests.

bendemboski:
I need some help understanding what's happening with the CSS in the `engines-host-app` `lazy-engine` test (I don't have any experience with engines). Some of the CSS coming from addons is going through the `style-loader`, although I don't see it going through the `render()` code that I modified. That's causing the test to fail, and I'm not sure what to do about it.

When I change the expected order of the style rules in the test, the `CLASSIC` scenario fails, so my change is changing the order of the CSS in this case, which I don't think we want. But I can't figure out how/why it's going through the webpack pipeline, so I don't know what we can do about it.

Can anyone help me understand how that CSS is flowing through the build system?

Issue #800, opened 2021-05-03
template only components defined inside engine are not honoured


siva-sundar:
Template only components defined inside engine are not honoured.

```
Build Error (PackagerRunner) in lib/lazy-loading-engine/templates/application.hbs
Module Error (from ../../../home/travis/build/siva-sundar/embroider-template-only-component-error/node_modules/@embroider/hbs-loader/src/index.js):
Missing component: TemplateOnlyComponent in lib/lazy-loading-engine/templates/application.hbs
```

CI Link: https://travis-ci.com/github/siva-sundar/embroider-template-only-component-error/builds/224756553#L271

Repo: https://github.com/siva-sundar/embroider-template-only-component-error

siva-sundar:
On debugging, we have passed the appPackage instance to resolve the templates of the engine.

Shouldn't we resolve this by passing the file ?

```ts
cache.resolve(packageName, cache.ownerOfFile(from)!)
```

ef4:
Nice work.

I think you're correct about the location of the problem, but the fix is not `ownerOfFile(from)`, because unfortunately we're modeling Ember's traditional resolving here, not proper package-scoped resolving. Instead it needs to be the nearest lazy engine. *sigh*. 

Probably this resolver code will need to use the output of [partitionEngines](https://github.com/embroider-build/embroider/blob/ea6f2ddce1878a8597bea708f984e46292ba743b/packages/core/src/app.ts#L551).

siva-sundar:
Thanks for the details.

Just a clarification, can we access the package from package-cache make use of it here ?

```js
PackageCache.shared('embroider-stage3').ownerOfFile(from);
```

The same is being used [here](https://github.com/embroider-build/embroider/blob/92e04c356737081d8d1ba2903031ef1ca3a1e923/packages/compat/src/resolver.ts#L379) for  absPathToRuntimePath.

PR #808, opened 2021-05-11
WIP: [BUGFIX] Pass owner pkg to from file while resolving components

siva-sundar:


packages/compat/src/resolver.ts
```diff
@@ -437,16 +437,16 @@ export default class CompatResolver implements Resolver {
...
-      let cache = PackageCache.shared('embroider-stage3');
       let packageName = parts[0];
       let renamed = this.params.adjustImportsOptions.renamePackages[packageName];
       if (renamed) {
         packageName = renamed;
       }
       return this._tryComponent(parts[1], from, withRuleLookup, cache.resolve(packageName, cache.ownerOfFile(from)!));
     } else {
-      return this._tryComponent(path, from, withRuleLookup, this.appPackage);
+      return this._tryComponent(path, from, withRuleLookup, cache.ownerOfFile(from)!);
```
ef4, line 449: Thanks for working on this.

As I said [here](https://github.com/embroider-build/embroider/issues/800#issuecomment-832690721), it's not correct to use `cache.ownerOfFile(from)` here. That's not how component resolution works -- they're not resolved relative to the current package, they're resolved relative to the containing engine.

The code that understands how to find the containing engine is [partitionEngines](https://github.com/embroider-build/embroider/blob/ea6f2ddce1878a8597bea708f984e46292ba743b/packages/core/src/app.ts#L551).

ef4 (changes requested):

PR #812, opened 2021-05-13
Update thread-loader to get RegExp serialization

bendemboski:
Update to the latest thread-loader, mainly so we can [use RegExps](https://github.com/webpack-contrib/thread-loader/pull/102) in plugin options (which was release in v3.0.2).


ef4:
Thanks.

PR #828, opened 2021-05-29
Update addon-shim to use ember-auto-import v2 final

josemarluedke:


rwjblue (commented):
Hmm, https://github.com/ef4/ember-auto-import/blob/6b32fba32a9cb562a62e466e1094ba0c92bc9390/packages/ember-auto-import/package.json#L36 is a little bit difficult for us here.

Since `^0.40.0` can't drift to get `0.41.0` (current release), we have a bit of node_modules duplication / clashing.

@ef4 - Do we need to be concerned here?

ef4:
We could do a coordinated release of ember-auto-import and embroider so these don't disagree. But it's not a correctness problem. The shared internals are really a grab bag of useful utilities and they don't need to all be on one version.

This is another reason to do a 1.0 though, that would make these easier to manage.

rwjblue:
> But it's not a correctness problem. The shared internals are really a grab bag of useful utilities and they don't need to all be on one version.

That's great, thank you for confirming.

PR #830, opened 2021-06-01
Add assertion for ensureSafeComponent called with legacy component class

### LINKED ISSUE #825: ensureSafeComponent does not work for imported classic component
It seems using `ensureSafeComponent()` on an imported component class works only when that component uses the newer colocation layout. On a classic or pod component, no error is thrown, but also nothing is rendered. 

I assume this happens because the template is somehow not found, because when I use `setComponentTemplate()` to explicitly associate the template, then it works!

Is this expected?

I confirmed this behavior pretty consistently in an app, and I am confident I could produce a reproduction as a failing test here if needed.

simonihmig:
This adds a developer friendly assertion message when `ensureSafeComponent()` is being called with a component class without an associated template, which would have failed silently before.

Closes #825.


ef4:
I think this will be a false positive if they use `Ember.Component` with `layout`?

simonihmig:
Hm, right! 

I guess I could check for `klass.layout`, but wonder if we can safely assume that `layout` can only exist on the prototype? That's what I know from classic addon patterns, but would this be also valid?

```js
import template from 'app/templates/my-component';

export default class MyComponent extends Component {
  layout = template;
}
```

In which case we couldn't check this just by looking at the class, but only after instantiating the component. So `ensureSafeComponent()` couldn't do this, and this whole PR would be invalid, right?

rwjblue:
> I guess I could check for `klass.layout`, but wonder if we can safely assume that `layout` can only exist on the prototype? That's what I know from classic addon patterns, but would this be also valid?

Ember only checks for `layout` as an instance property (could be inherited from the prototype, or set in a class field / in `constructor` / in `init`) for `@ember/component`. We never check for a static `layout` property.

> 
> ```js
> import template from 'app/templates/my-component';
> 
> export default class MyComponent extends Component {
>   layout = template;
> }
> ```

Ya, this is technically valid (assuming `Component` there is an `@ember/component`).

rwjblue:
> In which case we couldn't check this just by looking at the class, but only after instantiating the component. So `ensureSafeComponent()` couldn't do this, and this whole PR would be invalid, right?

I think we could modify slightly to limit the missing assertion cases, but I agree that it is not trivially possible to handle all scenarios.

Something like:

```js
import EmberComponent, { getComponentTemplate } from '@ember/component';

function isValidComponent(klass) {
  if (klass.prototype instanceof EmberComponent) {
    // we can't really detect if layout will be late bound (e.g. class field, set in init/constructor, etc)
    return true;
  } else {
    return Boolean(getComponentTemplate(klass));
  }
}
```

Alternatively, we could re-wrap the class when its an `EmberComponent` and add the helpful assertion to its `init`. You'd still get a nicer message than today...

simonihmig:
Ok, thanks for confirming @rwjblue!

So we cannot really be sure if the component *instance* will have a template at the `ensureSafeComponent()` sees the component *class*. That means we cannot really throw an assertion as this PR was pretending, but maybe just a dev-only warning, like "You passed a component class "MyComponent" to ensureSafeComponent that does not have an associated template. This might fail, unless this component explicitly imports the template and sets it as `.layout`. Consider migrating to co-located components!".

simonihmig:
Oh, seems my comment and your last one overlapped...

So in case of an EmberComponent without template I could issue a warning, otherwise an assertion as it is currently, right?

rwjblue:
> So in case of an EmberComponent without template I could issue a warning, otherwise an assertion as it is currently, right?

Yes, you can make a warning when we find an `Ember.Component` and `klass.prototype.layout` is `undefined`, but I think you can additionally wrap the class to make it an assertion if it's ever invoked. So in that case, you'd get a warning eagerly (at the time we see `ensureSafeComponent`) so you are guaranteed to have some console output telling you a bad thing is happening, but then if you do end up invoking it we can make an assertion that is still helpful (vs today's situation where you have no clue why nothing rendered).

simonihmig:
Ok, I can try that!

Another thing I was unsure: I thought about checking whether the class is an Ember component as you did with an `instanceof` check, but is it actually a good thing to import from `'@ember/component'` within this addon? Like will there eventually be a future where you could build Ember apps without `Ember.Component` when they are not used, in which case our own import of it could actually lead to needlessly pulling in all that code? 🤔

rwjblue:
> Another thing I was unsure: I thought about checking whether the class is an Ember component as you did with an instanceof check, but is it actually a good thing to import from '@ember/component' within this addon?

Ya, good point. Personally, I don't think that we are particularly close to realistically shaking out `@ember/component` in normal sized apps yet. Maybe we could avoid that by checking for a brand on the class?

https://github.com/emberjs/ember.js/blob/42bd258a026fb6652ae7cb991b01c9738d2ab032/packages/%40ember/-internals/glimmer/lib/component.ts#L1110

We could basically look for a large combo of things we know are in an Ember.Component, where it is super unlikely that any other component class would have all of those things.

rwjblue:
Something like:

```js
function looksLikeEmberComponent(klass) {
  return klass.isComponentFactory === true &&
    Array.isArray(klass.positionalParams) &&
    klass.prototype.isComponent === true &&
    typeof klass.prototype.didReceiveAttrs === 'function' &&
    typeof klass.prototype.didRender === 'function' &&
    typeof klass.prototype.willRender === 'function' &&
    typeof klass.prototype.didUpdateAttrs === 'function' &&
    typeof klass.prototype.willUpdate === 'function' &&
    typeof klass.prototype.didUpdate === 'function' &&
    typeof klass.prototype.didUpdateAttrs === 'function' &&
    typeof klass.prototype.didUpdateAttrs === 'function';
}
```

simonihmig:
Ok... looks like duck typing taken to the extreme! 😝

simonihmig:
I turned the assertion into a warning, and assert at component invocation time now, as was discussed above.

It seems to work in general, tests were passing locally (added additional test cases), *however* when I run the `legacy component class value throws assertion` test in isolation, things behave differently...

This is `klass` (passed into `looksLikeEmberComponent`) when other tests have run before:
image

Everything looks as expected, and `looksLikeEmberComponent` returns true.

Now running that *same* test in isolation, `klass` is this now:

image

`looksLikeEmberComponent` obviously returns false now, making the test fail.

I have now idea what's going on here, do you @rwjblue?

Issue #819, opened 2021-05-20
`@types/loader-utils` results in type issues


mydea:
When trying to set this up for our typescript app, using latest versions of embroider, I get this error:

```
node_modules/@types/loader-utils/index.d.ts:12:10 - error TS2305: Module '"webpack"' has no exported member 'loader'.

12 import { loader } from 'webpack';
```

Which I guess is related to this:
https://github.com/webpack/loader-utils/issues/179

Is it possible to drop `@types/loader-utils`?

nightire:
Seeing this error after upgrade `embroider@0.42` with `ember-cli-typescript@4.2.0`:

![image](https://user-images.githubusercontent.com/1104018/122304176-5f108080-cf37-11eb-8e3a-00ba83920bf8.png)

NullVoxPopuli:
this is the work-around I use: https://github.com/NullVoxPopuli/limber/blob/main/package.json#L24

in package.json#scripts: `"postinstall": "rm -rf node_modules/@types/loader-utils"`

NullVoxPopuli:
Confirmed fixed in: https://github.com/NullVoxPopuli/limber/pull/219

PR #831, opened 2021-06-02
Replace loader-utils with built-in webpack 5 functionality

mydea:
Webpack has built-in types now, which seem to conflict with `@types/loader-utils`. I think we can just remove this dependency - see https://github.com/webpack/loader-utils/issues/179.

Fixes https://github.com/embroider-build/embroider/issues/819


mydea:
I rebased on latest master.

rwjblue:
Kicked off CI

chancancode:
We can't just remove the types. We currently use `getOptions` [here](https://github.com/embroider-build/embroider/blob/a2883a9764cbee1e538d03c10384bf76834cacaf/packages/hbs-loader/src/index.ts#L13). Removing the types package would cause the import itself to become a type error/any. Instead, I believe the recommendation was that we should remove the `loader-utils` package itself and switch from `getOptions` to using [the (supposedly, I haven't verified) equivalent provided by webpack 5](https://webpack.js.org/api/loaders/#thisgetoptionsschema).

@mydea would you work on that?

chancancode:
I think you can also address [this](https://github.com/embroider-build/embroider/blob/a2883a9764cbee1e538d03c10384bf76834cacaf/packages/hbs-loader/src/index.ts#L9-L10) while you are there. It seems like recent versions of webpack does have a type for [`LoaderContext`](https://github.com/webpack/webpack/blob/7102df3bb52a33529ff5db4fdf34484d2a359a49/types.d.ts#L5852), further, it takes a generic slot for the options so it could just be `function hbsLoader(this: LoaderContext<HbsLoaderConfig>, ...)` and `this.getOptions()` should Just Work™.

mydea:
I will give it a try!

mydea:
Hmm, does hbs-loader need to remain compatible with webpack 4? Meaning, do I need to build this in a way that it _either_ uses the new webpack loader, _or_ `loader-utils`, or can I update that package to webpack 5 and just use that? 

I would go with the latter (`@embroider/webpack` relies on 5.x already anyhow), but if it should be the former I'm not sure what approach I should take to achieve that.

mydea:
I made a PR updating webpack peerDependency to 5 for hbs-loader, and replacing loader-utils with the built-in functionality & the new types.

ef4:
Thanks!

PR #837, opened 2021-06-04
remove ember-cli-htmlbars-inline-precompile as it is no longer needed

stefanpenner:
remove ember-cli-htmlbars-inline-precompile as it is no longer needed


stefanpenner:
Once CI completes if green, I will attempt to remove all occurrences of inline-precompile.

stefanpenner:
bummer, problem still occurs. Even though my isolated reproduction appeared to be ok... mystery continues..

rwjblue (approved):

PR #838, opened 2021-06-07
Fix webpack.run callback

stefanpenner:
Ensure all `throws`, including the explicit throw, correctly reject the wrapping promise.

---

As far as I can tell, this isn't the root cause of any posted issues, but something I noticed while debugging a different issue.

### DIFF
--- a/packages/webpack/src/ember-webpack.ts
+++ b/packages/webpack/src/ember-webpack.ts
@@ -463,7 +463,7 @@ const Webpack: PackagerConstructor<Options> = class Webpack implements Packager 
             this.consoleWrite(stats.toString());
           }
           resolve(stats.toJson());
-        } catch(e) {
+        } catch (e) {
           reject(e);
         }
       });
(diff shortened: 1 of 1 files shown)


rwjblue (approved):

packages/webpack/src/ember-webpack.ts
```diff
@@ -443,27 +443,29 @@ const Webpack: PackagerConstructor<Options> = class Webpack implements Packager
...
-        }
-        if (stats.hasErrors()) {
-          // the typing for MultiCompiler are all foobared.
-          reject(this.findBestError(flatMap((stats as any).stats, s => s.compilation.errors)));
-          return;
-        }
-        if (stats.hasWarnings() || process.env.VANILLA_VERBOSE) {
-          this.consoleWrite(stats.toString());
+          resolve(stats.toJson());
+        } catch(e) {
```
rwjblue, line 466:
```suggestion
        } catch (e) {
```

PR #839, opened 2021-06-09
Fix Babel-Loader Caching for ember-template-compiler

stefanpenner:
Fix Babel-Loader Caching for ember-template-compiler

Today, when switching template-compiler versions, we can end up with global cache pollution. Although this can manifest many ways, it most commonly manifests when running ember-try across versions of ember with different template compilation semantics when also using inline templates.

To address this:

We now include the checksum of the template compiler in the inline-template-compiler babel configuration options.
This busts the cache, as any functioning caching babel subsystem (such as babel-loader) must consider option changes as cache invalidating scenarios. Today, it is in-fact the case that babel-loader operates as we expect, and with this change, caches are correctly invalidated.

We co-located the checksum with the template-compiler options, rather then a single global cache key to both simplify debugging and not prevent some more granular caching strategy from functioning optimally.

This does not explicitly include integration tests due to my believe that this is already covered sufficiently:

1) today’s tests already include this (running yarn test currently fails without this fix)
2) type checking ensure we now always pass the key in
3) this is ultimately a configuration change of a babel-loader feature (which itself should test this)

### DIFF
--- a/packages/compat/src/v1-addon.ts
+++ b/packages/compat/src/v1-addon.ts
@@ -26,6 +26,7 @@ import { isEmberAutoImportDynamic, isCompactReexports, isColocationPlugin } from
 import { ResolvedDep } from '@embroider/core/src/resolver';
 import TemplateCompilerBroccoliPlugin from './template-compiler-broccoli-plugin';
 import { fromPairs } from 'lodash';
+import { getEmberExports } from '@embroider/core/src/load-ember-template-compiler';
 
 const stockTreeNames = Object.freeze([
   'addon',
@@ -132,8 +133,11 @@ export default class V1Addon {
         // our macros don't run here in stage1
         options.plugins.ast = options.plugins.ast.filter((p: any) => !isEmbroiderMacrosPlugin(p));
         if (options.plugins.ast.length > 0) {
+          const { cacheKey: compilerChecksum } = getEmberExports(options.templateCompilerPath);
+
           return new NodeTemplateCompiler({
             compilerPath: options.templateCompilerPath,
+            compilerChecksum,
             EmberENV: {},
             plugins: options.plugins,
             resolver: this.templateResolver(),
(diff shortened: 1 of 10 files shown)


stefanpenner:
I indent to follow this up with (3) more changes:

1) include additional cache busting configuration changes for portable babel loader options (node_module version)
2) include additional cache busting configuration changes for each babel plugin (plugins version)
3) readme addition describing how this caching works, and when & how a power-user may need to bust the cache.

packages/compat/src/v1-addon.ts
```diff
@@ -132,8 +133,10 @@ export default class V1Addon {
         // our macros don't run here in stage1
         options.plugins.ast = options.plugins.ast.filter((p: any) => !isEmbroiderMacrosPlugin(p));
         if (options.plugins.ast.length > 0) {
+          const compilerChecksum = crypto.createHash('md5').update(readFileSync(options.templateCompilerPath, 'utf-8')).digest('hex');
```
rwjblue, line 136: Since we pass the path into this class in both `compat-app` and `v1-addon`, can we make `NodeTemplateCompiler` do the checksum in its constructor?

stefanpenner: Thanks for pointing that out, looking further, there is more we can do here to tidy this up. Especially given that we already do have a nearly-what-we-need subsystem for loading/caching & generating a checksum for the template compiler https://github.com/embroider-build/embroider/blob/46fd8277a0050769a534c7762437268f026a61ba/packages/core/src/template-compiler-node.ts#L20

packages/core/src/load-ember-template-compiler.ts
```diff
@@ -25,7 +25,7 @@ export function getEmberExports(templateCompilerPath: string): EmbersExports {
     if (
       currentStat.mode === entry.stat.mode &&
       currentStat.size === entry.stat.size &&
-      currentStat.mtime.getTime() === entry.stat.mtime.getTime()
+    currentStat.mtime.getTime() === entry.stat.mtime.getTime()
```
rwjblue, line 28: This seems odd spacing, can you fix?

rwjblue (approved):

PR #841, opened 2021-06-09
[Fix] Cache busting 

stefanpenner:
This includes 2 fixes:

1) Fix Placeholder cache solution when used with requireFile

Ensure we include the package’s version in the configuration, so that the babel-loader can detect module version changes in it’s cache key consideration. Without this, installing a new node_module may not result in babel-loaders cache expiration occurring.

2) Fix babel plugin version change cache busting.

We now ensure a given plugins node_module version is included (when available) in the babel configuration. Which ensures the cache is correctly evicted. This is accomplished by adding our own noop babel plugin into the plugin configuration, and having it contain the various plugin path / version pairs.


packages/core/src/app.ts
```diff
@@ -144,6 +145,36 @@ export function excludeDotFiles(files: string[]) {
   return files.filter(file => !file.startsWith('.') && !file.includes('/.'));
 }
 
+export const CACHE_BUSTING_PLUGIN = {
```
stefanpenner, line 148: this fixes (2)

packages/core/src/portable.ts
```diff
@@ -13,9 +15,23 @@ export interface PortableResult {
 
 export interface PortableHint {
   requireFile: string;
+  packageVersion: string | undefined;
   useMethod?: string;
 }
 
+const { findUpPackagePath } = resolvePackagePath;
+
+export function maybeNodeModuleVersion(path: string) {
```
stefanpenner, line 24: Fixes 1

test-packages/macro-sample-addon/package.json
```diff
@@ -49,7 +48,7 @@
     "ember-maybe-import-regenerator": "^0.1.6",
     "ember-qunit": "^4.4.1",
     "ember-resolver": "^5.0.1",
-    "ember-source": "~3.10.0",
```
stefanpenner, line 52: safe for future PR

packages/core/src/portable.ts
```diff
@@ -13,9 +15,23 @@ export interface PortableResult {
...
+  const packagePath = findUpPackagePath(path);
+
+  if (packagePath === null) {
+    // no package was found
+    return undefined; // should this bust the cache or ... ?
+  } else {
+    return readJSONSync(packagePath).version;
```
rwjblue, line 31: Why not use `require(packagePath)` (vs `readJSONSync`)?

packages/core/src/portable.ts
```diff
@@ -13,9 +15,23 @@ export interface PortableResult {
...
 }
 
+const { findUpPackagePath } = resolvePackagePath;
+
+export function maybeNodeModuleVersion(path: string) {
+  const packagePath = findUpPackagePath(path);
+
+  if (packagePath === null) {
+    // no package was found
+    return undefined; // should this bust the cache or ... ?
```
rwjblue, line 29: Let's error for now, I think this is very very unlikely anyways  but it seems pretty bad.

rwjblue (changes requested):

packages/core/src/portable.ts
stefanpenner: writing that + associated test changes now.

packages/core/src/portable.ts
stefanpenner: linting yelled at me

stefanpenner:
> Is it possible to write a test for the cache invalidation issue that this resolves?

We could add _another_ CI worker to run sequential ember-try scenarios... but it's unclear we can just continue to add those CI jobs... if we want to do that sure, otherwise I think this is adequate given our current priorities.

stefanpenner:
@ef4 / @rwjblue should we also include the current node version / OS in the cache key? ... If needed we can always expand later...

rwjblue:
> should we also include the current node version / OS in the cache key?

Probably. I think we should likely also include targets info, though it may already be included (since its probably part of options to `@babel/preset-env`'s preset).

stefanpenner:
> Probably. I think we should likely also include targets info, though it may already be included (since its probably part of options to @babel/preset-env's preset).

I believe these are in there already

stefanpenner:
@rwjblue with a fresh mind tomorrow I’ll think about approaches to do a more e2e test that is the right cost/benefit.

packages/core/src/app.ts
```diff
@@ -144,6 +145,39 @@ export function excludeDotFiles(files: string[]) {
...
+      if (Array.isArray(plugin) && typeof plugin[0] === 'string') {
+        absolutePathToPlugin = plugin[0] as string;
+      } else if (typeof plugin === 'string') {
+        absolutePathToPlugin = plugin;
+      } else {
+        throw new Error(`[Embroider] a babel plugin without an absolute path was from: ${plugin}`);
+      }
+
+      const pkgPath = resolvePackagePath.findUpPackagePath(absolutePathToPlugin);
+      plugins[absolutePathToPlugin] = pkgPath ? readJSONSync(pkgPath).version : undefined;
```
rwjblue, line 169: I thought we had decided to make the scenario where we can't find a `pkgPath` an error condition here?

packages/core/src/babel-plugin-cache-busting.ts
```diff
@@ -0,0 +1,11 @@
+export default function makePlugin(): any {
+  // Dear future @rwjblue,
```
rwjblue, line 2: ❤️

packages/core/src/app.ts
rwjblue: Hmm, looks like `maybeNodeModuleVersion` does that, can you update this to use `maybeNodeModuleVersion` to get the `version`?

packages/core/src/portable.ts
```diff
@@ -13,9 +14,22 @@ export interface PortableResult {
...
   useMethod?: string;
 }
 
+const { findUpPackagePath } = resolvePackagePath;
+
+export function maybeNodeModuleVersion(path: string) {
+  const packagePath = findUpPackagePath(path);
+
+  if (packagePath === null) {
+    throw new Error(`Could not find package.json for '${path}'`);
```
rwjblue, line 27:
```suggestion
    throw new Error(`[Embroider] Could not find package.json for '${path}'`);
```

Just for consistency with the same prefix as you had for the relative path babel plugin...

packages/core/src/app.ts
stefanpenner: forgot this scenario, fixing.

PR #842, opened 2021-06-10
[hygiene] re-roll lockfile

stefanpenner:


stefanpenner:
the 1 failure so far looks like an unrelated network failure in windows while volta attempts to fetch something: `Error: Response code 400 (Bad Request)`

PR #844, opened 2021-06-11
[hygiene] upgrade dependencies

stefanpenner:
To ensure we catch issues with dependencies sooner, I am beginning to upgrading our dependencies, and plan to address issues encountered during that process.


stefanpenner:
Upgrading `jest` introduces some `require` mock, which seems to break `require.cache` accesses used in some older versions of the template compiler code. Until we remove that from our tests, or fix jest, we may be stuck on an old version of jest.

package.json
```diff
@@ -40,16 +40,16 @@
     "**/qunit": "^2.14.1"
   },
   "devDependencies": {
-    "@types/jest": "^24.0.11",
+    "@types/jest": "^26.0.23",
```
rwjblue, line 43: If we can't update jest, shouldn't we also not update the types? Presumably they are adding types for features that we don't have?

package.json
```diff
@@ -40,16 +40,16 @@
...
     "cross-env": "^7.0.3",
     "eslint": "^7.14.0",
-    "eslint-config-prettier": "^6.11.0",
+    "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-import": "^2.22.0",
     "eslint-plugin-prettier": "^3.1.4",
-    "jest": "^24.5.0",
-    "prettier": "2.1.2",
+    "jest": "^24.9.0",
+    "prettier": "2.3.1",
```
rwjblue, line 52: Hmm, why do we pin? I'd expect `^` here and then leveraging the `yarn.lock` for the pinning.

packages/compat/package.json
```diff
@@ -24,59 +24,59 @@
...
-    "chalk": "^4.1.0",
-    "debug": "^3.1.0",
-    "fs-extra": "^7.0.0",
-    "fs-tree-diff": "^2.0.0",
-    "jsdom": "^16.4.0",
-    "lodash": "^4.17.10",
+    "broccoli-source": "^3.0.1",
+    "chalk": "^4.1.1",
+    "debug": "^4.3.2",
+    "fs-extra": "^10.0.0",
```
rwjblue, line 50: I don't think we can bump to `fs-extra@10` yet (until we drop support for Node 10). I think we can absolutely drop support, but it ought to be a dedicated PR.

package.json
stefanpenner: ya, pinning seems odd. I can make it drift

package.json
stefanpenner: good catch

packages/compat/package.json
stefanpenner: when should we do that?

PR #845, opened 2021-06-11
Adding addon template for future test migrations

thoov:
The next set of test migrations will need an addon template so this is purely to keep future PRs smaller


package.json
```diff
@@ -15,7 +15,8 @@
       "types/broccoli-concat",
       "types/ember-cli-htmlbars",
       "tests/scenarios",
-      "tests/app-template"
+      "tests/app-template",
+      "tests/addon-template"
```
rwjblue, line 19: Why it is in `tests/` instead of `test-packages`? Not a blocker or anything, just curious...

rwjblue, line 19: Should these be versioned (e.g. `app-template-3-24`)? It seems reasonable to have a template for each of our supported LTS versions (for example). That allows us to test against the ember-cli blueprint's of those supported LTS's.

Can you describe the expected workflow / process for updating these? Do we update them for every ember-cli release? Some other cadence? It would be great to start fleshing out our `CONTRIBUTING.md` with these kinds of details...

tests/addon-template/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "test": "ember test",
+    "test:all": "ember try:each"
+  },
+  "dependencies": {
+    "ember-cli-babel": "^7.7.3"
+  },
+  "devDependencies": {
+    "@ember/optional-features": "^0.7.0",
+    "broccoli-asset-rev": "^3.0.0",
+    "ember-cli": "~3.10.1",
```
rwjblue, line 29: Yikes, this seems quite old (and actually is unsupported). Why generate with ember-cli@3.10? Shouldn't we at least use one of our supported LTS versions?

rwjblue: I'd suggest using `npx ember-cli@3.24 addon addon-template --skip-npm --no-welcome` to regenerate.

rwjblue (changes requested):

package.json
thoov: As we refactor each test package into a scenario I am moving them over to keep them separate. I prefer the final state of just having 'tests' and that test-packages will be fully gone once all of them are converted.

thoov: > Can you describe the expected workflow / process for updating these?

We would run the following probably after every new LTS is released.

1.  Run ember-cli-update in the template directories to pull in the latest template changes.
2. Update 'tests/scenarios.ts' to link to make sure that our support matrix is correct.


> Should these be versioned

We can but we would have to restructure the test infra a bit if we want to. Right now we are just selectively installing the ember-cli and ember-source deps within this base template but if we feel like the template structure from one to the other is worth testing we can. If we go this route then we might want to rethink the above update workflow.

tests/addon-template/package.json
thoov: Oops I ran `ember addon addon-template` inside of this directory and seems like the global ember choose an older version. I have updated this to use a new version

package.json
rwjblue: 👍  - Sounds good. Mind throwing a small section into `CONTRIBUTING.md` with that "template update" plan?

tests/addon-template/package.json
rwjblue: 🎉

package.json
thoov: Added

PR #848, opened 2021-06-14
Update dependencies (packages/router)

stefanpenner:


packages/router/addon/index.js
```diff
@@ -1,9 +1,8 @@
 /*
-  This code is adapted from ember-engines/addon/-private/router-ext.js.
+  This cod- is adapted from ember-engines/addon/-private/router-ext.js.
 */
 import EmberRouter from '@ember/routing/router';
-import { registerWaiter, unregisterWaiter } from '@ember/test';
-import { DEBUG } from '@glimmer/env';
+import { waitForPromise } from '@ember/test-waiters';
```
stefanpenner, line 5: @rwjblue is it safe for us to use these new APIs?

packages/router/addon/index.js
```diff
@@ -61,56 +60,39 @@ if (macroCondition(getGlobalConfig()['@embroider/core']?.active)) {
     // overriding to provide our own handlerResolver.
     setupRouter() {
       let isSetup = this._super(...arguments);
+
+      // eslint-disable-next-line ember/no-private-routing-service
+      const routerMicrolib = this._routerMicrolib;
```
stefanpenner, line 65: I've left the original "private API" usage because:

1) I didn't want to change to much all at once
2) I'm not 100% sure yet if the new APIs work across the needed support versions (this is likely something a single follow up PR should focus on ... assuming is is possible)

packages/router/addon/index.js
```diff
@@ -22,22 +21,22 @@ if (macroCondition(getGlobalConfig()['@embroider/core']?.active)) {
...
-      return window._embroiderRouteBundles_.find(bundle => bundle.names.indexOf(routeName) !== -1);
+      return window._embroiderRouteBundles_.find(
+        (bundle) => bundle.names.indexOf(routeName) !== -1
+      );
     }
 
     return false;
   };
 
+  // eslint-disable-next-line ember/no-classic-classes
```
stefanpenner, line 38: I kept the classic class, because without it the provide `this._routerMicrolib` API was no longer present, and I would like to (if we decide to) do that upgrade independent of this one.

packages/router/addon/index.js
```diff
@@ -61,56 +60,39 @@ if (macroCondition(getGlobalConfig()['@embroider/core']?.active)) {
...
-            this._inFlightLazyRoutes--;
-            bundle.loaded = true;
-            return original(name);
-          },
-          err => {
-            this._inFlightLazyRoutes--;
-            throw err;
-          }
-        );
+        return waitForPromise(async () => {
```
stefanpenner, line 88: using the new test-waiters API (a nice improvement IMHO)

packages/router/addon/index.js
stefanpenner: I believe this is a breaking change, requiring users to include `@ember/test-waiters`. We now include the `@ember/test-waiters` as an dep of this add-on. I believe that is the correct approach... let me know if it is not.

rwjblue: > @rwjblue is it safe for us to use these new APIs?

It is safe if we add `@ember/test-waiters` to `@embroider/router`'s `dependencies`.

packages/router/addon/index.js
rwjblue: Ya, sounds good to me. Maybe make an issue to track that?

stefanpenner: Ya, I plan to continue working on that category of work until embroider code-base is as modern as it can be.

stefanpenner:
this depends on https://github.com/emberjs/ember-test-helpers/pull/1047 landing + releasing

stefanpenner:
pending: https://github.com/ef4/ember-auto-import/runs/2832950014

stefanpenner:
rebasing

ef4:
These packages need to get ported to run their tests through scenario-tester. Part of the point of it is to avoid the need for this kind of per-package upgrade work.

stefanpenner:
@ef4 ACK, I'll transition this work to that.

PR #852, opened 2021-06-15
Drop support for Node 10, 11, 13, and 15.

rwjblue:
This unifies the `engines` property across all of our packages (they didn't agree with each other, some were missing `engines` completely, etc) to support:

* Node 12.*
* Node 14.*
* Node 16 or higher

This matches the [currently supported releases](https://nodejs.org/en/about/releases/) by Node itself:

image

### DIFF
--- a/packages/addon-shim/package.json
+++ b/packages/addon-shim/package.json
@@ -29,7 +29,7 @@
     "@types/semver": "^7.3.6"
   },
   "engines": {
-    "node": "10.* || >= 12"
+    "node": "12.* || 14.* || >= 16"
   },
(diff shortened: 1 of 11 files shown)


thoov (approved):

stefanpenner:
:shipit:

stefanpenner (approved):

stefanpenner:
@ef4 et.al next release version should convey a "breaking change"

ef4:
It will go into the changelog with BREAKING automatically as long as it's tagged that way (and Rob already added the tag).

PR #853, opened 2021-06-15
Better error message when an asset cannot be found in entry file

thoov:
This is an error we have hit in all of our internal apps migrated thus far and wanted to improve the messaging for future travelers.

New error message:

Screen Shot 2021-06-15 at 12 45 57 PM

Old error message:

Screen Shot 2021-06-15 at 12 47 33 PM

Additionally, created the following wiki entry that gives more context for it: https://github.com/thoov/stitch/wiki/Could-not-find-asset-in-entry-file-error-help


stefanpenner (approved):
Nice improvement!

PR #863, opened 2021-06-18
Add back util types

### LINKED ISSUE #850: Missing util types
#633 added typings for `@embroider/util`, but this was later removed by #732. 
I wonder why? By accident maybe?

This now breaks upgrades for TS projects: `Cannot find module '@embroider/util' or its corresponding type declarations`.

simonihmig:
Fixes #850


ef4:
Thanks @simonihmig!

PR #866, opened 2021-06-21
[WIP] [Fix #862] add emberSourceRealModules flag

### LINKED ISSUE #862: Ember global not available to vendor code when using ember-source 3.27+
Issue found by @stefanpenner.

Starting with ember 3.27, Embroider can consume Ember as true ES modules. And this even supports the (deprecated) `window.Ember` as long as you are accessing it from an ES module. But if you try to access `window.Ember` from classic script context code (stuff emitted via treeForVendor and `app.imported` into vendor.js) it's not available.

We should make the true-ES-modules Ember into an optional flag until Ember 4.0, which completely removes the global.

stefanpenner:
* changes “real module” mode of ember-source to begin automatically at 4.0.0, rather then 3.27.0-beta.0
* adds emberSourceRealModules option, to opt into this mode prior to 4.0.0
* begins fixing ember-source@3.27.x issues

TODO:

- [x] sync up with @thoov re: test-matrix stuff
- [x] debug the expected ember 3.27.x issues
- [x] possible solution for ^
- [ ] ensure everything works
- [ ] write tests based on discussion with @thoov


packages/core/src/options.ts
```diff
@@ -57,6 +57,18 @@ export default interface Options {
...
+  // optimizing compiler, enabling tree-shaking and other optimizations of
+  // ember itself.
+  //
+  // Unfortunately due to backwards compatibility constraints consuming
+  // ember-source this way wont be the default until ember-source@4.0.0.
+  //
+  // This flag provideds adventurous ember-source users & addon-authors the
+  // ability to test & begin preperation prior to this feature becomeing the
+  // default.
+  emberSourceRealModules?: boolean,
```
stefanpenner, line 71: I'm not 100% sold on the flag name, but this is the first the came to mind. Anyone want to bike-shed this?

packages/core/src/options.ts
```diff
@@ -57,6 +57,18 @@ export default interface Options {
...
+  // optimizing compiler, enabling tree-shaking and other optimizations of
+  // ember itself.
+  //
+  // Unfortunately due to backwards compatibility constraints consuming
+  // ember-source this way wont be the default until ember-source@4.0.0.
+  //
+  // This flag provideds adventurous ember-source users & addon-authors the
+  // ability to test & begin preperation prior to this feature becomeing the
+  // default.
+  emberSourceRealModules?: boolean;
```
stefanpenner, line 71: I'm not 100% sold on the flag name, but this is the first the came to mind. Anyone want to bike-shed this?

thoov (approved):

packages/compat/src/compat-app.ts
```diff
@@ -356,9 +356,10 @@ class CompatAppAdapter implements AppAdapter<TreeNames> {
     }
 
     let emberSource = this.activeAddonChildren().find(a => a.name === 'ember-source')!;
-    let emberNeedsModulesPolyfill = semver.satisfies(emberSource.version, '<3.27.0-beta.0', {
-      includePrerelease: true,
-    });
+    let emberNeedsModulesPolyfill =
+      semver.satisfies(emberSource.version, '<4.0.0', {
+        includePrerelease: true,
+      }) && !this.options.emberSourceRealModules;
```
rwjblue, line 362:
```suggestion
    let emberNeedsModulesPolyfill = this.options.emberSourceRealModules || semver.satisfies(emberSource.version, '<4.0.0', {
      includePrerelease: true,
    });
```

packages/compat/src/compat-adapters/ember-source.ts
```diff
@@ -5,9 +5,25 @@ import AddToTree from '../add-to-tree';
...
+    addonInstance: any,
+    addonOptions: Required<Options>,
+    app: V1App,
+    packageCache: PackageCache,
+    orderIdx: number
+  ) {
+    super(addonInstance, addonOptions, app, packageCache, orderIdx);
+    this.useRealModules =
+      addonOptions.emberSourceRealModules ||
+      semver.satisfies(this.packageJSON.version, '>=4.0.0-beta.0', { includePrerelease: true });
```
rwjblue, line 25:
```suggestion
      semver.satisfies(this.packageJSON.version, '>=4.0.0-alpha.0', { includePrerelease: true });
```

packages/compat/src/compat-app.ts
```diff
@@ -356,9 +356,10 @@ class CompatAppAdapter implements AppAdapter<TreeNames> {
     }
 
     let emberSource = this.activeAddonChildren().find(a => a.name === 'ember-source')!;
-    let emberNeedsModulesPolyfill = semver.satisfies(emberSource.version, '<3.27.0-beta.0', {
-      includePrerelease: true,
-    });
+    let emberNeedsModulesPolyfill =
+      semver.satisfies(emberSource.version, '<4.0.0', {
```
rwjblue, line 360:
```suggestion
      semver.satisfies(emberSource.version, '<4.0.0-alpha.1', {
```

packages/compat/src/compat-app.ts
```diff
@@ -341,6 +341,17 @@ class CompatAppAdapter implements AppAdapter<TreeNames> {
     });
   }
 
+  @Memoize()
```
stefanpenner, line 344: currently unused

packages/core/src/app.ts
```diff
@@ -415,6 +416,17 @@ export class AppBuilder<TreeNames> {
       { absoluteRuntime: __dirname, useESModules: true, regenerator: false },
     ]);
 
+    {
```
stefanpenner, line 419: needs cleanup

packages/core/src/app.ts
```diff
@@ -1480,3 +1492,58 @@ function combinePackageJSON(...layers: object[]) {
   }
   return mergeWith({}, ...layers, custom);
 }
+
+function macroBabelPlugins() {
```
stefanpenner, line 1496: needs comment

packages/core/src/options.ts
```diff
@@ -57,6 +57,18 @@ export default interface Options {
...
+  // optimizing compiler, enabling tree-shaking and other optimizations of
+  // ember itself.
+  //
+  // Unfortunately due to backwards compatibility constraints consuming
+  // ember-source this way wont be the default until ember-source@4.0.0.
+  //
+  // This flag provideds adventurous ember-source users & addon-authors the
+  // ability to test & begin preperation prior to this feature becomeing the
+  // default.
+  emberSourceRealModules?: boolean | null;
```
stefanpenner, line 71: we likely no longer need `null` here

stefanpenner:
Test failures so far are just the expected deprecation warning. @rwjblue based on our chat, it would seem like we should allow those deprecations…. Although that does annoy me a little….

ef4:
I think if we use modules-api-polyfill it's going to cause deprecation spew on Ember 3.27+?

When code in a true ES module handled by webpack does:

```js
import Component from '@ember/component'`
```

modules-api-polyfill will rewrite that to `Ember.Component`, which is deprecated.

We could instead configure webpack's [externals](https://webpack.js.org/configuration/externals/) so that it knows to defer these imports to runtime and access them via runtime `require`, which would make us compatible with what the classic build does in this case.

stefanpenner:
> We could instead configure webpack's externals so that it knows to defer these imports to runtime and access them via runtime require, which would make us compatible with what the classic build does in this case.

That approach sounds better. I'll give it whirl

stefanpenner:
@ef4 It sounds like we may want to essentially add the data from https://github.com/ember-cli/ember-rfc176-data to externals for 2.27.0-alpha.0 -> 4.0.0-alpha.0 unless the user opts into the `emberSourceRealModules` WDYT

ef4:
@stefanpenner yes, but please try to use it via https://github.com/embroider-build/embroider/blob/7bab34a265e234ab45e8eaf40482f8dbce7aa5cc/packages/shared-internals/src/ember-standard-modules.ts which exists for this purpose.

ef4:
@stefanpenner I realized we probably already have enough support for externalizing ember dependencies in adjust-imports-plugin.

You may want to just try *not* doing anything to externalize the ember modules because this code probably already gets it right:

https://github.com/embroider-build/embroider/blob/7bab34a265e234ab45e8eaf40482f8dbce7aa5cc/packages/core/src/babel-plugin-adjust-imports.ts#L274-L301

stefanpenner:
> You may want to just try not doing anything to externalize the ember modules because this code probably already gets it right:

If memory serves this fails, but I will investigate further.

stefanpenner:
I will resume this and now that https://github.com/embroider-build/embroider/pull/881 has landed, I bet some of the issues will be resolved.

stefanpenner:
rebased (will continue to investigate as time permits)

stefanpenner:
quick rebase again

packages/compat/src/compat-adapters/ember-cli-fastboot.ts
```diff
@@ -141,6 +141,11 @@ class RewriteManifest extends Plugin {
       extraVendorFiles,
     };
 
+    json['ember-addon'] = {
```
stefanpenner, line 144: i don't remember if this was actually important now, i think it isn't and should likely be removed

packages/router/tests/dummy/config/ember-cli-update.json
```diff
@@ -0,0 +1,21 @@
+{
```
stefanpenner, line 1: likely just drop this file

packages/compat/src/addon-dependency-rules/ember-data.ts
```diff
@@ -4,6 +4,7 @@ let rules: PackageRules[] = [
   {
     package: '@ember-data/store',
     addonModules: {
+      //'-private/index.js': {
```
stefanpenner, line 7: do i need this? I don't even remember.

packages/core/src/app.ts
```diff
@@ -1476,3 +1493,58 @@ function combinePackageJSON(...layers: object[]) {
...
+  const isProduction = process.env.EMBER_ENV === 'production';
+  const isDebug = !isProduction;
+  // const ignore = {
+  //   '@ember/debug': ['assert', 'deprecate', 'warn'],
+  //   '@ember/application/deprecations': ['deprecate'],
+  // };
+
+  return [
+    // [
```
stefanpenner, line 1506: I don't remember this, i need to load this memory from cold storage..

packages/core/src/app.ts
```diff
@@ -1476,3 +1493,58 @@ function combinePackageJSON(...layers: object[]) {
   }
   return mergeWith({}, ...layers, custom);
 }
+
+function macroBabelPlugins() {
+  const isProduction = process.env.EMBER_ENV === 'production';
+  const isDebug = !isProduction;
+  // const ignore = {
```
stefanpenner, line 1500: I don't remember this, i need to load this memory from cold storage..

packages/core/src/babel-plugin-adjust-imports.ts
```diff
@@ -202,7 +203,17 @@ module.exports = m;
 
 function handleExternal(specifier: string, sourceFile: AdjustFile, opts: Options, isDynamic: boolean): string {
   let pkg = sourceFile.owningPackage();
+
+  if (specifier === 'ember') {
```
stefanpenner, line 207: remove

packages/core/src/babel-plugin-adjust-imports.ts
```diff
@@ -202,7 +203,17 @@ module.exports = m;
...
+    console.log('ember', JSON.stringify(pkg), pkg && JSON.stringify(pkg.packageJSON, null, 2), pkg && pkg.isV2Ember());
+  }
+
   if (!pkg || !pkg.isV2Ember()) {
+    if (specifier === 'ember') {
```
stefanpenner, line 212: remove block

packages/core/src/babel-plugin-adjust-imports.ts
```diff
@@ -235,6 +246,9 @@ function handleExternal(specifier: string, sourceFile: AdjustFile, opts: Options
 
     // self-imports are legal in the app tree, even for v2 packages
     if (packageName === pkg.name) {
+      if (specifier === 'ember') {
```
stefanpenner, line 249: remove block

packages/core/src/babel-plugin-adjust-imports.ts
```diff
@@ -331,6 +345,7 @@ function makeMissingModule(specifier: string, sourceFile: AdjustFile, opts: Opti
 }
 
 function makeExternal(specifier: string, sourceFile: AdjustFile, opts: Options): string {
+  // debugger;
```
stefanpenner, line 348: remove

packages/core/src/babel-plugin-adjust-imports.ts
```diff
@@ -401,6 +416,9 @@ export default function main(babel: unknown) {
             continue;
           }
 
+          if (source.value === 'ember') {
```
stefanpenner, line 419: remove

packages/core/src/babel-plugin-adjust-imports.ts
```diff
@@ -431,9 +449,15 @@ function rewriteTopLevelImport(
     }
   }
 
+  if (source.value === 'ember') {
+    // debugger;
+  }
   let specifier = adjustSpecifier(source.value, state.adjustFile, opts, false);
   if (specifier !== source.value) {
     source.value = specifier;
+  } else if (source.value === 'ember') {
```
stefanpenner, line 458: remove

packages/core/src/app.ts
```diff
@@ -1476,3 +1493,58 @@ function combinePackageJSON(...layers: object[]) {
...
+        },
+
+        debugTools: {
+          isDebug,
+          source: '@ember/debug',
+          assertPredicateIndex: 1,
+        },
+      },
+      '@embroider/core:@ember/debug stripping',
+    ],
```
stefanpenner, line 1532: Now that i re-enabled this block, I remember why I disabled it.

This block makes tests pass, but trips up against the globals warnings. Removing it makes some tests fail, especially when add-ons reference things like `@glimmer/tracking`.  There are several approaches we can take to address this, I hope to sync up with @ef4 tomorrow to discuss options.

ef4:
Closing in favor of https://github.com/embroider-build/embroider/pull/978

rwjblue:
(I think @ef4 meant to close this with his comment above)

PR #868, opened 2021-06-22
fix issue with thread-loader performance

krisselden:
avoid sending resolver config per file.

### DIFF
--- a/packages/babel-loader-8/index.js
+++ b/packages/babel-loader-8/index.js
@@ -1,0 +1,44 @@
+const { applyVariantToBabelConfig } = require('@embroider/core');
+module.exports = require('babel-loader').custom(babel => {
+  return {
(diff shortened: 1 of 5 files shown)


packages/babel-loader-8/index.js
```diff
@@ -0,0 +1,17 @@
+const { applyVariantToBabelConfig } = require('@embroider/core');
+module.exports = require('babel-loader').custom(babel => {
```
stefanpenner, line 2: A short description of the "fix" would be great. As it may not be obvious for future travelers.

stefanpenner (changes requested):

packages/babel-loader-8/index.js
krisselden: I added more to commit description

stefanpenner (approved):

rwjblue (approved):

PR #869, opened 2021-06-22
fix infinite loop

krisselden:


tests/scenarios/scenarios.ts
```diff
@@ -9,7 +9,7 @@ import { dirname, delimiter } from 'path';
   if (!voltaHome) return;
   let paths = process.env['PATH']!.split(delimiter);
   while (/\.volta/.test(paths[0])) {
-    paths.shift;
```
stefanpenner, line 12: someone was trying to write some ruby.

stefanpenner (approved):

tests/scenarios/scenarios.ts
ef4: I didn't notice the typo and copied it verbatim from https://github.com/volta-cli/volta/issues/702.

We also found and fixed this in ember-auto-import, but not before it got copied to here. 🙃

PR #873, opened 2021-06-24
[Discussion] Experiment with improving debugging of workers.

stefanpenner:
While debugging some stuff with parallel builds I've noticed several things, and have some ideas on how to possible improve the situation. 

1) we could make it easier to attach a debugger to one of the child-processes

One approach would be to automatically detect that the parent node process is in debug mode, and do the same with the children. This works alright, with the downside of needing more careful shutdown of the process.

Another approach would be to provide a separate environment variable to accomplish this.

2) when attempted to debug an issue that only appears only in parallel mode i noticed that JOBS=<N> doesn't correspond to the number of JOBS, rather it is any value but JOBS=1 will result in the number of available CPUs. This makes it both difficult to limit the number of JOBS based on available resources AND it was tricky to debug just 1 worker node scenarios.

One approach is to respect the number of JOBS if specified.

Another quirk is the inability to specify 1 single job, most of this time this likely doesn't matter, but it was a little odd.
Rather than JOBS=1 meaning use the primary process. We could instead have

* JOBS=false === use the primary process
* JOBS=<count> === explicitly set the worker count to <count>
* default experience is to let webpack decide.

### DIFF
--- a/packages/webpack/src/ember-webpack.ts
+++ b/packages/webpack/src/ember-webpack.ts
@@ -37,6 +37,7 @@ import crypto from 'crypto';
 import type { HbsLoaderConfig } from '@embroider/hbs-loader';
 import semverSatisfies from 'semver/functions/satisfies';
 import supportsColor from 'supports-color';
+import inspector from 'inspector';
 
 const debug = makeDebug('embroider:debug');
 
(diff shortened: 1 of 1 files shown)


stefanpenner:
I'm going to play with some other options as I debug stuff myself. This approach is kinda annoying...

PR #876, opened 2021-06-24
Convert `fastboot-addon` to new test infra

thoov:
Also removing the old test-packages/fastboot-addon code.

This should land after: #875


rwjblue:
Ready for a rebase now that #874 landed

PR #877, opened 2021-06-24
[BREAKING] Respect JOBS count if present

stefanpenner:
For web-pack:

JOBS=0 uses the current thread of execution
JOBS=1..n spawns the corresponding thread or sub-processes
No set JOBS results in a dynamically calculated number of jobs

Previously

JOBS=1 used the current thread of execution
All other variations would dynamically calculated number of jobs

### DIFF
--- a/packages/webpack/src/ember-webpack.ts
+++ b/packages/webpack/src/ember-webpack.ts
@@ -522,6 +522,7 @@ const Webpack: PackagerConstructor<Options> = class Webpack implements Packager 
 };
 
 const threadLoaderOptions = {
+  workers: 'JOBS' in process.env && Number(process.env.JOBS),
   // poolTimeout shuts down idle workers. The problem is, for
   // interactive rebuilds that means your startup cost for the
   // next rebuild is at least 600ms worse. So we insist on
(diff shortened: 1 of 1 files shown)


rwjblue (commented):
JOBS=0 and JOBS=1 should have the same behavior (in order to match what broccoli-babel-transpiler uses)

lifeart (approved):

stefanpenner:
> JOBS=0 and JOBS=1 should have the same behavior (in order to match what broccoli-babel-transpiler uses)


@rwjblue it's been a while now, but I believe we spoke about this after your comment, but maybe we should touch-base quick.


My thoughts, without this PR we cannot debug only 1 worker,  or run the build in a single job separate of the host, without running 2 or more workers. This negative impact of this PR when compared to the positive seems acceptable, but WDYT?

stefanpenner:
@ef4 pingaling

stefanpenner:
In an offline chat with @ef4 around the authoring of this PR he gave a +1. If that has changed, I will gladly revert + fix. But no need to let this linger.

Issue #826, opened 2021-05-28
Issue with @ember-data/debug/setup.js


thoov:
### Steps to reproduce:

1. `ember addon foobar --yarn` (Use ember-cli 3.26+)
2. `ember install ember-data`
3. `yarn add -D @embroider/core @embroider/compat @embroider/webpack webpack`
4. `EMBROIDER_TEST_SETUP_FORCE=true EMBROIDER_TEST_SETUP_OPTIONS=safe ember s`

### Error:

```
setup.js:20 Uncaught TypeError: Cannot read property 'prototype' of undefined
    at eval (setup.js:20)
    at Object.../../node_modules/@ember-data/debug/setup.js (chunk.d5c0c63cbd51f4014925.js:149)
    at __webpack_require__ (chunk.72811919dc18453a17c5.js:31)
    at eval (ember-data-data-adapter.js:5)
    at Object../initializers/ember-data-data-adapter.js (chunk.68a92cf93b54bca7a7d2.js:149)
    at __webpack_require__ (chunk.72811919dc18453a17c5.js:31)
    at Module.eval [as callback] (dummy.js:22)
    at Module.exports (vendor.js:115)
    at requireModule (vendor.js:36)
    at resolveInitializer (index.js:10)
```

cc @ef4 / @rwjblue

ef4:
(Reproduction steps are missing `yarn add -D webpack` since that is now required as a peer of @embroider/webpack, but that is not the source of the bug.)

thoov:
Updated the repo steps with that

rwjblue:
Fixed by https://github.com/embroider-build/embroider/pull/880

PR #880, opened 2021-06-28
Fix compatibility with ember-data@3.27+

ef4:
ember-data really needs to stop running its own complicated build-within-the-build and move these customized steps to prepublish (if they're truly needed at all, which I am skeptical about).

Todo:
 - [ ] check this against other recent ember-data versions to make sure we're not simlutaneously breaking them.


ef4:
This fixes https://github.com/embroider-build/embroider/issues/826

rwjblue:
> ember-data really needs to stop running its own complicated build-within-the-build and move these customized steps to prepublish

Ya, we are working on it (had a bunch of discussions over in the recent ember-data core calls).

rwjblue (approved):

thoov:
> check this against other recent ember-data versions to make sure we're not simlutaneously breaking them

Can you add ember-data to `scenarios.ts` such that we can get coverage back to 3.16?

PR #881, opened 2021-06-28
Fix race condition finding the owning package of a given file when using multiple workers

ef4:
`PackageCache`'s `ownerOfFile` can give the wrong result when dealing with in-repo nested package structures (like a dummy app, in-repo-addon, or runtime fastboot directory) such that there is no intervening `node_modules` folder between the parent and child packages.

It only happens if there's a request for the parent package before the files in the child package, which is why this was a sporadic race that only happened sometimes, when running with multiple parallel workers where the order is nondeterministic.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -56,7 +56,6 @@
     "js-string-escape": "^1.0.1",
     "jsdom": "^16.6.0",
     "lodash": "^4.17.21",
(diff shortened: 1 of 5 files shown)


lifeart (approved):

rwjblue (approved):

stefanpenner:
woohoo!

Issue #885, opened 2021-07-02
Overwriting addon macro config not working


mydea:
I have an addon with a macro config like this:

```js
module.exports = {
  name: require('./package').name,

    '@embroider/macros': {
      setOwnConfig: {
        mockUsername: 'jane@example.com',
      },
    },
  },
};
```

And an app using this addon with a config like this:

```js
 '@embroider/macros': {
      setConfig: {
        'ember-cognito-identity': {
          mockUsername: 'john@fabscale.com',
          other: 'test here',
        },
      },
    },
```

Now if I do this in my addon:

```js
console.log(getOwnConfig().mockUsername, getOwnConfig().other);
```

it will print out: "jane@example.com test here"

So it seems that any configuration from the addon itself takes precedence over configuration in the app, which I guess is not how this is supposed to work (at least that was my understanding)? If that is not how this is supposed to work (e.g. addon config should not be overwriteable), how would you go ahead providing default values for an addon?

PR #888, opened 2021-07-05
Expose sourceOfConfig to macro config mergers

mydea:
Currently, the merging strategy for `@embroider/macros` configs will just merge all given objects together. This means that the last config to be registered will "win", if a property is defined in more than one place.

Since the app's own config will be called first, this means that an app cannot overwrite configuration that is set in an addon.

This PR changes this by keeping a private `__priority` property on the config (which is stripped out at the end) by which we order the configs, if appropriate. The ordering works like this:

0. `setOwnConfig` will always have the lowest priority
1. `setConfig` called from an addon (e.g. setting config for another addon)
2. `setConfig` called from the host app

The following example shows the output hierarchy:

```js
// addon-1/index.js
setOwnConfig: {
  value1: 'default',
  value2: 'default',
  value3: 'default'
}

// addon-2/index.js
setConfig: {
  'addon-1': {
    value1: 'addon',
    value2: 'addon'
  }
}

// my-app/ember-cli-build.js
setConfig: {
  'addon-1': {
    value1: 'app'
  }
}

// Output from addon-1:
assert.deepEquals(
  getOwnConfig(), 
  { 
    value1: 'app', 
    value2: 'addon', 
    value3: 'default'
  }
);
```

I tried to build it in an as backwards-compatible way as possible. 

Closes https://github.com/embroider-build/embroider/issues/885

### DIFF
--- a/packages/macros/src/ember-addon-main.ts
+++ b/packages/macros/src/ember-addon-main.ts
@@ -14,24 +14,13 @@ export = {
     // if parent is an addon it has root. If it's an app it has project.root.
     let source = parent.root || parent.project.root;
 
-    // In order to allow to overwrite config from consumed addons, we keep a priority
-    // This is stripped before being returned via getConfig/getOwnConfig
-    // Own config has lowest priority - e.g. default values for an addon
-    // Addons can overwrite other addon's config (priority 1)
-    // The host app has highest priority (priority 2)
     if (ownOptions.setOwnConfig) {
-      MacrosConfig.for(appInstance).setOwnConfig(source, Object.assign({ __priority: 0 }, ownOptions.setOwnConfig));
+      MacrosConfig.for(appInstance).setOwnConfig(source, ownOptions.setOwnConfig);
     }
 
     if (ownOptions.setConfig) {
-      let parentIsHostApp = parent === appInstance;
-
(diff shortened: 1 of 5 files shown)


ef4:
Thanks for working on this. I know you were trying not to introduce API changes but I think it's worth having some small changes here so that custom mergers can see exactly where each config is coming from. That way, your change becomes just a change to the default merger.

 - add changes on top of yours that expose a new `sourceOfConfig` method to mergers
 - rebased on master

ef4:
I'm tagging this as breaking only because the type signature for config has changed from `unknown` to `object` (because it must be a valid `WeakMap` key).

mydea:
Great!
BTW while working on this, I noticed that it is currently not possible to actually _set_ a custom merger - is that on purpose? There is no reference to `useMerger` anywhere in the code, as far as I see.

ef4:
It's possible, it's just not documented or used in any examples yet.

```js
const { MacrosConfig } = require('@embroider/macros/node');
MacrosConfig.for(appInstance).useMerger(__filename, ...);
```

The `appInstance` is available to an addon's index.js as `this._findHost()` and to an app's ember-cli-build.js as the actual `app` local variable.

mydea:
Ah, very nice - I didn't know that! Great to know if I ever get into the position again that I need some custom handling :+1:

PR #893, opened 2021-07-08
Inline hbs reform

ef4:
This is part of a project to unify Ember's inline template compilation.

`@ef4/babel-plugin-htmlbars-inline-precompile` is a published copy of https://github.com/ember-cli/babel-plugin-htmlbars-inline-precompile/tree/streamline.  (It's easier to make typescript happy by using a published copy rather than pointing at git.)

The plan here is to make a new babel-plugin-htmlbars-inline-precompile that only support Ember 3.27+. Because that drops the need for modules-api-polyfill, much complexity can be removed and it becomes easier to compose babel plugins together, reducing the need to keep babel-plugin-htmlbars-inline-precompile a dumping ground of unrelated features (like ember-template-imports). 

While we're at it, I'd prefer to rename babel-plugin-htmlbars-inline-precompile to `@ember/template-compilation`, because that is its public API:

```js
import { precompileTemplate } from '@ember/template-compilation';
```

The older patterns like `import { hbs } from 'ember-cli-htmlbars'` are still available, but are understood to be sugar for `precompile`, because `precompile` is a superset of all the rest and it should be understood as the foundational primitive for all template compilation in ember. Even standalone `.hbs` files are conceptually sugar for `export default precompileTemplate(...)`.

The actual rename will presumably need an RFC, but for purposes of this discussion I'm calling the new version `@ember/template-compilation`.

Everybody else who wants to build on top of template compilation does so by composing with `@ember/template-compilation`.
 - embroider (in this branch) has a standalone babel plugin that is only responsible for inserting discovered template dependencies into JS, while delegating the actual template compilation. Soon, this will also be extended to insert the discovered dependencies directly into `scope`, instead of doing `window.define`. Embroider will be made to use this new pattern when ember 3.27+ is present, while keeping the old implementation on older embers for now.
 - I will have a branch of ember-cli-htmlbars coming that is updated to use the new `@ember/template-compilation` on Ember 3.27+. On older embers it will keep the present behavior.
 - I will have a branch of ember-template-imports coming that ports over the code I cut out of babel-plugin-htmlbars-inline-precompile.


ef4:
This is staying as draft until we can get a mainline package released instead of making people depend on `@ef4/babel-plugin-htmlbars-inline-precompile`.

Issue #891, opened 2021-07-05
New scenario tests running extraneous checks


ef4:
The app-template and addon-template are configured so that `yarn test` runs *many* things:

https://github.com/embroider-build/embroider/blob/81ca52f837bd6ff9e80bc3ac406677f05da524a4/tests/app-template/package.json#L23

And we are invoking this directly in various test scenarios:

https://github.com/embroider-build/embroider/blob/81ca52f837bd6ff9e80bc3ac406677f05da524a4/tests/scenarios/macro-test.ts#L42

I think the app-template and addon-template should probably change so `yarn test` is just `ember test`.

cc @thoov

PR #896, opened 2021-07-12
Only run ember test during scenario testing

thoov:
Fixes: #891


thoov:
cc: @ef4

stefanpenner:
@thoov is this because we already have a project-wide lint-stage run independent of this?

PR #897, opened 2021-07-13
Test Support: Get Packager from Dependencies

alexlafroscia:
As discussed in #759, we need some way to expand the tests to cover the behavior of packagers besides the existing Webpack one. We could do this by having the build-a-real-app tests load their packager dynamically based on locating a known packager in their dependencies, then use `scenario-tester` to configure scenarios with different packagers installed.

This PR sets up that infrastructure and configures the `app-template` tests to use it.


tests/scenarios/scenarios.ts
```diff
@@ -35,13 +35,21 @@ async function release(project: Project) {
...
-  });
+  return scenarios
+    .expand({
+      lts_3_16,
+      lts_3_20,
+      lts_3_24,
+      release,
+    })
+    .expand({
+      packager_webpack,
```
alexlafroscia, line 51: Point _maybe_ worth bike-shedding; using `packager_webpack` rather than just `webpack` means that the whole `packager_webpack` string is in the name of the CI job, etc. That's nice and specific but kind of verbose, in case it matters to anyone 🤷‍♂️

lifeart (approved):

thoov:
> Once the Vite packager is added to the mix, we'd need some way to avoid generating every possible job -- does Scenario 
> Tester have an API for ignoring particular combinations?

@alexlafroscia It's possible to filter the matrix before we hand it off to github: https://github.com/embroider-build/embroider/blob/d8edf75f1c973dccf0e5f53c538e7ed006961e9b/test-packages/support/suite-setup-util.ts#L83-L97

tests/scenarios/scenarios.ts
```diff
@@ -35,13 +35,21 @@ async function release(project: Project) {
...
-    release,
-  });
+  return scenarios
+    .expand({
+      lts_3_16,
+      lts_3_20,
+      lts_3_24,
+      release,
+    })
+    .expand({
```
ef4, line 50: I think this is overkill and we should just add vite to this list:

```js
export function supportMatrix(scenarios: Scenarios) {
  return scenarios.expand({
    lts_3_16,
    lts_3_20,
    lts_3_24,
    release,
    vite
  });
```

The vite scenario would use the default ember version. If we later find some particular combinations that are problematic and need regression coverage, we can explicitly add those combinations.

alexlafroscia: Oh okay, cool -- I can tweak it to work that way instead!

Should each of the "Ember Version" variants explicitly add the Embroider packager, and we'll remove the Webpack "variant" altogether?

PR #900, opened 2021-07-20
Only patch `ember-cli-deprecation-workflow` releases before `2.0.0`

alexlafroscia:
I ran into an error when trying out Embroider in my main Ember app.

The compat adapter for `ember-cli-deprecation-workflow` resolves an error with `ember-debug-handlers-polyfill` which is no longer present in version `2.0.0`. Because it's not installed, Node is unable to resolve the location of the package.

We should opt out of the patch if the installed version is `2.0.0` or greater.

### DIFF
--- a/packages/compat/src/compat-adapters/ember-cli-deprecation-workflow.ts
+++ b/packages/compat/src/compat-adapters/ember-cli-deprecation-workflow.ts
@@ -4,8 +4,15 @@ import { UnwatchedDir } from 'broccoli-source';
 import resolve from 'resolve';
 import { Memoize } from 'typescript-memoize';
 import buildFunnel from 'broccoli-funnel';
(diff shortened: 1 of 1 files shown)


alexlafroscia:
CI failure due to timeout seems to be present on `master` as well 🙍‍♂️

thoov:
@alexlafroscia you can also leverage `shouldApplyAdapter` (https://github.com/embroider-build/embroider/pull/772) which will completely ignore the adapter altogether.

alexlafroscia:
Ah, even better! I had looked around at the other compat adapters for something like this but didn't come across one. I'll solve it that way instead!

rwjblue (approved):

rwjblue (approved):

PR #904, opened 2021-07-23
Fix ember-test-selectors custom adapter for 6.x

mydea:
The 6.x release of ember-test-selectors dropped the attribute binding for classic components. This leads to a build error due to the current compat-adapter of embroider:

```
module not found: Error: 
Can't resolve '../node_modules/ember-test-selectors/utils/bind-data-test-attributes' 
in '$TMPDIR/embroider/d7920e/assets/my-app.js'
```

This PR makes the compat adapter only run for ember-test-selector versions below 6. Not sure if there is a "nicer" way to check this than to take the version of the addon pkg info, but I guess it works well enough for this case!

### DIFF
--- a/packages/compat/src/compat-adapters/ember-test-selectors.ts
+++ b/packages/compat/src/compat-adapters/ember-test-selectors.ts
@@ -1,12 +1,14 @@
 import V1Addon from '../v1-addon';
 import { forceIncludeModule } from '../compat-utils';
+import semver from 'semver';
 
 export default class extends V1Addon {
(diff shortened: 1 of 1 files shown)


thoov:
We can leverage `shouldApplyAdapter` (#772) which will completely ignore the adapter altogether. An example of this is: https://github.com/embroider-build/embroider/pull/900/files

mydea:
Ah, nice! I rewrote it to leverage that.

PR #905, opened 2021-07-23
Fix ember-test-selectors compat adapter for v6

josemarluedke:
Version 6 of ember-test-selectors removed the ability to bind attrs, therefore the file that was being included is not existing anymore.


thoov:
Dup of #904

Issue #908, opened 2021-07-27
PreprocessCss receiving different inputs that break ember-cli-eyeglass


thoov:
When upgrading an app that was on embroider v0.30 to v0.43 we ran into an issue with ember-cli-eyeglass. 

The root cause after debugging was that (https://github.com/linkedin/eyeglass/blob/master/packages/ember-cli-eyeglass/src/index.ts#L209)  was receiving different input trees.

Specifically in v0.30 it would receive a tree like:

```
├── app.scss
├── foo.scss
└── bar.scss
```

Where as in v0.43 it would receive a tree like:

```
├── app
│ └── styles
│     └── app.scss
│     └── foo.scss
└─    └── bar.scss
```

Backtracing this in Embroider pointed to [this line](https://github.com/embroider-build/embroider/blob/719700f6dce8fae1c57b16536da856e3eabee5b4/packages/compat/src/v1-app.ts#L479-L480) and the [following PR](https://github.com/embroider-build/embroider/pull/674).

Still need to evaluate what the expected tree should be in this case as reverting fixes the problem for our use case but also want to make sure https://github.com/embroider-build/embroider/issues/672 is addressed.

PR #910, opened 2021-07-28
Fix arguments to `preprocessCss` (to match classic build)

thoov:
In classic builds the second argument is '/app/styles' but we were passing 'app/styles'. While this difference might be small, in
ember-cli-eyeglass [it removes the leading slash](https://github.com/linkedin/eyeglass/blob/master/packages/ember-cli-eyeglass/src/index.ts#L210-L212) and this causes issues when the slash is missing. This PR aligns with what classic was doing along with adding a very simple test case.

Fixes: #908


rwjblue (approved):

ef4:
CI is now back to normal, the remaining failures here are real failures, I think.

tests/fixtures/preprocess-addon/index.js
```diff
@@ -0,0 +1,49 @@
...
+
+module.exports = {
+  name: require('./package').name,
+
+  setupPreprocessorRegistry(type, registry) {
+    registry.add('css', {
+      name: this.name,
+      ext: 'css',
+      toTree: (tree, inputPath, outputPath) => {
+        console.log('inputPath: ' + inputPath);
```
asakusuma, line 32: I assume you don't actually want to commit these console.logs in this file?

thoov: No I am trying to debug the windows path issue

thoov:
@rwjblue This should be good to go

PR #903, opened 2021-07-21
Upgrade ember-qunit to address ember-canary compatibility issue curre…

stefanpenner:
…ntly breaking CI

PR #911, opened 2021-07-28
Upgrade ember-qunit to address canary deprecations

ef4:
Based off https://github.com/embroider-build/embroider/pull/903, but avoiding latest conflicts with master and handling the breaking changes in all packages.


stefanpenner:
I'm currently working to address some issues related to this specifically https://github.com/emberjs/ember-string/issues/259 which is a "leaf" activity, preventing us from properly upgrading due to ember canary removal of some modules...

Issue #906, opened 2021-07-26
`@embroider/macros` uses previously cached macro condition


brendenpalmer:
We're seeing issues where a previously cached macro condition is being used as part of the build. For example, `ember-engines` uses `@embroider/macros` to have differing logic depending on `ember-source` version. It's possible for (seemingly) a previously cached macro condition to be used for an incompatible `ember-source` version. The macro condition that's evaluated as part of the build can be incorrect, and you can get into a state where the macro condition is evaluated correctly _without_ `broccoli-persistent-filter`'s caching enabled (`CI=true`); however, with `broccoli-persistent-filter`'s caching enabled (regular builds without `CI=true`) the macro condition is incorrect.

For a simple reproduction I've created a sample repo in OSS: https://github.com/brendenpalmer/embroider-macros-caching-repro.

**Steps to reproduce:**

1. Clone [repo](https://github.com/brendenpalmer/embroider-macros-caching-repro), `yarn install`
1. Run `ember b`
1. Open `dist/assets/vendor.js`; the evaluated macro (specifically the one used in `link-to-external`) from `ember-engines` is correct
1. Run `yarn add ember-source@3.24.0`
1. Run `ember b`
1. Open `dist/assets/vendor.js`; the evaluated macro (specifically the one used in `link-to-external`) from `ember-engines` is _incorrect_
1. Run `CI=true ember b`
1. Open `dist/assets/vendor.js`; notice that the evaluated macro (specifically the one used in `link-to-external`) is now different

thoov:
I was able to reproduce. One minor change to the above steps to reproduce:

~~4. Run yarn add ember-source@3.24.0~~
4. Run yarn add ember-source@3.24.1

Some of the dependencySatisfies checks in engines are checking for this version. I will look into this.

PR #913, opened 2021-07-29
Bust babel cache if dependencies change so dependencySatisfies is always valid

thoov:
dependencySatisfies can become invalid if babel caches the macros plugin + the dependency that is being checked for changes its package version. This adds a cache busting plugin that tracks a hash of the contents of the apps lock file so any changes to dependencies will bust the cache.

Fixes: #906


packages/macros/src/ember-addon-main.ts
```diff
@@ -70,6 +72,30 @@ export = {
...
+      let packagePath = join(appInstance.project.root, 'package.json');
+      let lockFileBuffer;
+
+      if (fs.existsSync(yarnLockPath)) {
+        lockFileBuffer = fs.readFileSync(yarnLockPath);
+      } else if (fs.existsSync(npmLockPath)) {
+        lockFileBuffer = fs.readFileSync(npmLockPath);
+      } else {
+        // no lock file found, using package.json as a fall back
+        lockFileBuffer = fs.readFileSync(packagePath);
```
stefanpenner, line 87: If we always include package.json, we can encourage our users to use `link:` (yarn) or `file:` (npm) to perform linking, as that would at-least make that level cache aware.

If we extended that further, and always invalided the cache for `link:` or `file:` if they are on npm, we could implicitly prevent deeply nested linked deps that themselves later get more links from causing cache corruption. TL;DR if this is the path we deem appropriate it would result in a single linked dependency disabling babel caching. This is pretty conservative... but may be appropriate. We should think through this

packages/macros/src/babel-plugin-cache-busting.ts
```diff
@@ -0,0 +1,3 @@
+export default function makePlugin(): any {
+  return {};
+}
```
rwjblue, line 3: This seems like exactly what was added in https://github.com/embroider-build/embroider/pull/841. 


Can we just use that plugin here (we might need to move it to `@embroider/utils`, but that seems fine)?

https://github.com/embroider-build/embroider/blob/7165f06bb6027e82c61fe0ab107e54f9f0c6ff8c/packages/core/src/babel-plugin-cache-busting.ts#L1-L11

packages/macros/src/ember-addon-main.ts
```diff
@@ -70,6 +72,30 @@ export = {
       let appInstance = this._findHost();
       let source = appOrAddonInstance.root || appOrAddonInstance.project.root;
       babelPlugins.unshift(MacrosConfig.for(appInstance).babelPluginConfig(source));
+
+      let yarnLockPath = join(appInstance.project.root, 'yarn.lock');
+      let npmLockPath = join(appInstance.project.root, 'package-lock.json');
+      let packagePath = join(appInstance.project.root, 'package.json');
+      let lockFileBuffer;
```
rwjblue, line 79: Mind adding `pnpm-lock.yaml` also? Some users are succesfully using `pnpm`, and it would be nice to support _basically_ all popular lockfile formats...

packages/macros/src/ember-addon-main.ts
rwjblue: Ya, this seems good to me.

rwjblue: FWIW, this also seems reasonable for the original purposes in #841. Can we move this logic over into `@embroider/utils` and use it from `@embroider/core` for #841's use case and also use it here for this `dependencySatisfies` invalidation?

rwjblue (changes requested):
This seems good overall, though I did leave a couple notes about sharing / reuse.

I'd also love to see a test here, the reproduction in "real world" was pretty straight forward (running `npm test` which runs multiple ember-try scenarios back to back in an addon ought to have triggered this).

packages/macros/src/ember-addon-main.ts
```diff
@@ -70,6 +72,30 @@ export = {
       let appInstance = this._findHost();
       let source = appOrAddonInstance.root || appOrAddonInstance.project.root;
       babelPlugins.unshift(MacrosConfig.for(appInstance).babelPluginConfig(source));
+
+      let yarnLockPath = join(appInstance.project.root, 'yarn.lock');
```
brendenpalmer, line 76: is it always guaranteed that `yarn.lock` exists at `project.root`? (eg, if a repo is using workspaces & the app/addon exists within a workspace)

rwjblue: You are correct, `yarn.lock` might not exist in `project.root` but I'm not sure what else we can do in that scenario. 🤔

brendenpalmer: Yeah, we could do a find up for `yarn.lock`/`package-lock.json`, but continue to fall back to `package.json` at `project.root`

packages/macros/src/ember-addon-main.ts
thoov: Both packages (macros and core) use `@embroider/shared-internals` which is probably the best home for this share plugin.

packages/macros/src/babel-plugin-cache-busting.ts
thoov: Will move to `@embroider/shared-internals`

thoov:
@rwjblue:

> I'd also love to see a test here, the reproduction in "real world" was pretty straight forward (running `npm test` which runs multiple ember-try scenarios back to back in an addon ought to have triggered this).

Added a test to cover this and verified that removing the plugin it fails the test

packages/macros/src/ember-addon-main.ts
thoov: Added!

packages/macros/src/ember-addon-main.ts
ef4: Yes, `@embroider/utils` is utilities for app & addons to help with embroider compatibility. `@embroider/shared-internals` is for this sort of thing.

packages/macros/src/ember-addon-main.ts
```diff
@@ -70,6 +72,37 @@ export = {
...
+
+      // @embroider/macros provides a macro called dependencySatisfies which checks if a given
+      // package name satisfies a given semver version range. Due to the way babel caches this can
+      // cause a problem where the macro plugin does not run (because it has been cached) but the version
+      // of the dependency being checked for changes (due to installing a different version). This will lead to
+      // the old evaluated state being used which might be invalid. This cache busting plugin keeps track of a
+      // hash representing the lock file of the app and if it ever changes forces babel to rerun its plugins.
+      // more information in issue #906
+      let cacheKey = crypto.createHash('sha256').update(lockFileBuffer).digest('hex');
+      babelPlugins.push([
```
ef4, line 101: I think this code should go into https://github.com/embroider-build/embroider/blob/f276e5d1e5158c0266a3f32cfac778f6b4638586/packages/macros/src/macros-config.ts#L284 instead of here.

packages/macros/src/ember-addon-main.ts
```diff
@@ -69,7 +69,7 @@ export = {
     if (!babelPlugins.some(isEmbroiderMacrosPlugin)) {
       let appInstance = this._findHost();
       let source = appOrAddonInstance.root || appOrAddonInstance.project.root;
-      babelPlugins.unshift(MacrosConfig.for(appInstance).babelPluginConfig(source));
+      babelOptions.plugins = babelPlugins.concat(MacrosConfig.for(appInstance).babelPluginConfig(source));
```
rwjblue, line 72:
```suggestion
      babelOptions.plugins = [
        ...MacrosConfig.for(appInstance).babelPluginConfig(source),
        ...babelOptions.plugins
      ];
```

rwjblue: Or an even smaller change:

```suggestion
      babelOptions.plugins.unshift(...MacrosConfig.for(appInstance).babelPluginConfig(source));
```

rwjblue (commented):
Overall 👍, only one minor suggestion RE: `babelPlugins.unshift` (made suggestion inline).

rwjblue (approved):

PR #915, opened 2021-07-30
Reduce memory pressure from compat layer by disabling Heimdall node gathering during OneShotPlugin

rwjblue:
This reduces the amount of memory that these one-shot's create by:

- Avoiding creating Heimdall nodes for each broccoli plugin
- For manually created heimdall nodes (e.g. ones made by individual broccoli plugins) remove any child nodes added during the one-shot process

---

Additionally, this makes broccoli instrumentation output quite a bit nicer for these `OneShot` instances (adding the addon name as an annotation). Now (with these changes) debugging _which_ OneShot is particularly slow, is much easier...

### DIFF
--- a/packages/compat/src/one-shot.ts
+++ b/packages/compat/src/one-shot.ts
@@ -2,27 +2,65 @@ import Plugin from 'broccoli-plugin';
 import { Node } from 'broccoli-node-api';
 import { Builder } from 'broccoli';
 import { copySync } from 'fs-extra';
+import heimdall from 'heimdalljs';
+
+class NerfHeimdallReparentingBuilder extends Builder {
+  /*
+    Replace the code used to track heimdall nodes: https://github.com/broccolijs/broccoli/blob/v3.5.2/lib/builder.ts#L463-L503
+
+    This reduces the amount of memory that these one-shot's create by:
+
+    - Avoiding creating Heimdall nodes for each broccoli plugin
+    - Disabling the "re-parenting" process done by Broccoli builder (which ends up creating **double** the heimdall nodes)
+  */
+  setupHeimdall() {}
+  buildHeimdallTree() {}
+}
 
 // Wraps a broccoli tree such that it (and everything it depends on) will only
 // build a single time.
 export default class OneShot extends Plugin {
-  private builder: Builder;
-  private didBuild = false;
+  private builder: NerfHeimdallReparentingBuilder | null;
 
-  constructor(originalTree: Node) {
+  constructor(originalTree: Node, private addonName: string) {
     // from broccoli's perspective, we don't depend on any input trees!
     super([], {
+      annotation: `@embroider/compat: ${addonName}`,
       persistentOutput: true,
     });
-    this.builder = new Builder(originalTree);
+
+    // create a nested builder in order to isolate the specific addon
+    this.builder = new NerfHeimdallReparentingBuilder(originalTree);
   }
+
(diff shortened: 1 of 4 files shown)


types/heimdalljs/index.d.ts
```diff
@@ -0,0 +1,3 @@
+declare module 'heimdalljs' {
+  export default any
+}
```
rwjblue, line 3: Yes, this is bad and I feel bad. We should fix the types publishing of heimdalljs@0.2.x if we actually want good types here.

I'm not sure I want to shave that particular yak at the moment...

packages/compat/src/one-shot.ts
```diff
@@ -2,27 +2,61 @@ import Plugin from 'broccoli-plugin';
...
+      this.didBuild = true;
+    } finally {
+      oneshotCookie.stop();
+
+      /*
+        Remove any of the current node's direct children, this ensures that we do not bloat the
+        current Broccoli builder's heimdall node graph (e.g. the one that is calling
+        OneShotPlugin; **not** the one that the OneShotPlugin internally creates).
+      */
+      oneshotHeimdallNode.remove();
```
krisselden, line 59: we should allow you to disable this based on addon name if you want to see detail

types/heimdalljs/index.d.ts
ef4: I mean, the existence of OneShot also seems bad and I feel bad. 😅

Do you have any sense for whether BROCOLLI_ENABLED_MEMOIZE is going to get widely enabled? It would let us drop OneShot.

We could also profile how much OneShot is really saving us on rebuilds. I think it mattered when I added it, but a lot has changed since then.

packages/compat/src/one-shot.ts
```diff
@@ -2,27 +2,61 @@ import Plugin from 'broccoli-plugin';
...
     }
-    await this.builder.build();
-    copySync(this.builder.outputPath, this.outputPath, { dereference: true });
-    await this.builder.cleanup();
-    this.didBuild = true;
+
+    // Make a heimdall node so that we know for sure, all nodes created during our
+    // inner builder can be remove easily
+    const oneshotCookie = heimdall.start({
+      name: `@embroider/compat: OneShot (${this.addonName})`,
```
krisselden, line 40: OneShot will already be the name of this broccoli-plugin this will look like `OneShot > @embroider/compat: OneShot  addonName`

types/heimdalljs/index.d.ts
rwjblue: > Do you have any sense for whether BROCOLLI_ENABLED_MEMOIZE is going to get widely enabled? It would let us drop OneShot.

The main problem is that we need to provide a reasonable escape hatch that isn't "restart the server". @stefanpenner and I have chatted about a few options, I think that the best one was to add a way to trigger a full rebuild (e.g. ignoring the memoization for that specific rebuild). The two ways we thought of doing that were something like `touch tmp/perform-full-rebuild` (just touch some file that we could monitor) or sending a signal to the running process (e.g. `SIGUSR1`) that would force it to full rebuild.

> We could also profile how much OneShot is really saving us on rebuilds. I think it mattered when I added it, but a lot has changed since then.

I suspect it still saves us quite a bit of time, and will until we start moving the needle on usage of `@embroider/addon-shim`.

krisselden (approved):

PR #918, opened 2021-08-03
Add `needsCache` and `persistentOutput` to internal broccoli-plugins.

rwjblue:


krisselden (approved):

ef4:
I mistakenly thought `needsCache: false` was the default.

PR #919, opened 2021-08-11
Ember internal deps experiment

mattmcmanus:


ef4:
I think this being closed was a github bug. I delete a branch here that was named the same, but the PR branch still exists.

mattmcmanus:
Reopened: https://github.com/embroider-build/embroider/pull/920

PR #922, opened 2021-08-17
allow `null`, `undefined`, `if` and `unless` for strict component arguments

### LINKED ISSUE #921: Passing optional component arguments through does not work
I am currently trying to write a PR for ember-power-select to work fully with embroider. However, I am running into this somewhat unexpected issue when trying to use the addon with embroider & staticComponents=true.

A component accepts an optional sub-component. I can realize this with @embroider/util like this:

```hbs
{{! addon/components/power-select.hbs }}

{{#if @afterOptionsComponent}}
  {{#let (component (ensure-safe-component @afterOptionsComponent)) as |AfterOptions|}}
    <AfterOptions
      @extra={{@extra}}
      @select={{publicAPI}}
    />
  {{/let}}
{{/if}}
```

Now if I have another component which should allow to (optionally) pass this through, like this:

```hbs
{{! addon/components/power-select-multiple.hbs }}
<PowerSelect
  @afterOptionsComponent={{@afterOptionsComponent}}
/>
```

It complains when running `ember b`:

```
Build Error (PackagerRunner) in ../../components/power-select-multiple.hbs

Module Error (from ember-power-select/node_modules/thread-loader/dist/cjs.js):
argument "afterOptionsComponent" to component "PowerSelect" is treated as a component, 
but the value you're passing is dynamic: @afterOptionsComponent in 
$TMPDIR/embroider/e23032/components/power-select-multiple.hbs/power-select-multiple.hbs
```

This happens with any argument really, also for other similar arguments. Interestingly, it even seems to happen when I remove any usage of `@afterOptionsComponent` in the addon (meaning it is not used anywhere except in `@afterOptionsComponent={{@afterOptionsComponent}}` in power-select-multiple). If I rename every occurrance to e.g. `@afterOptionsThing`, it _also_ works again (but complains about `@beforeOptionsComponent`). So not quite sure what is going on there..!

You can see a reproduction in my branch: https://github.com/mydea/ember-power-select/tree/fn/ensure-safe-components just check it out, run `npm install` and `ember b` and you'll see a bunch of errors.

mydea:
I am not 100% sure, but I _think_ that this should be correct and fixes https://github.com/embroider-build/embroider/issues/921.

With this PR, the following would be allowed with `strictComponents=true`:

```hbs
<PowerSelect
  @afterOptionsComponent={{if 
    @afterOptionsComponent 
    (ensure-safe-component @afterOptionsComponent) 
    null
  }}
  @afterOptionsComponent={{null}}
  @afterOptionsComponent={{undefined}}
  @afterOptionsComponent={{unless @thing (ensure-safe-component @afterOptionsComponent)}}
  @afterOptionsComponent={{ensure-safe-component @afterOptionsComponent}}
/>
```

Basically, it specifically allows `null`, `undefined`, `ensure-safe-component`, `if` and `unless`. When using `if` or `unless`, it will check if each path either returns null/undefined or a valid component. So the following will still fail:

```hbs
<PowerSelect
  @afterOptionsComponent={{if @check 'string'}}
  @afterOptionsComponent={{if @check (component @afterOptionsComponent)}}
/>
```

I added a bunch of tests hopefully covering this well.

### DIFF
--- a/packages/compat/src/resolver-transform.ts
+++ b/packages/compat/src/resolver-transform.ts
@@ -110,7 +110,7 @@ export function makeResolverTransform(resolver: Resolver) {
                   for (let name of argumentsAreComponents) {
                     let attr = node.attributes.find((attr: any) => attr.name === '@' + name);
                     if (attr) {
-                      handleComponentHelper(attr.value, resolver, filename, scopeStack, {
+                      handleOptionalComponentHelper(attr.value, resolver, filename, scopeStack, {
                         componentName: node.tag,
                         argumentName: name,
                       });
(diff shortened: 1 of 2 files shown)


ef4:
Thanks, this seems good but haven't had a chance to fully review these changes yet.

I do want to point out that this usage:

```hbs
  @afterOptionsComponent={{if 
    @afterOptionsComponent 
    (ensure-safe-component @afterOptionsComponent) 
    undefined
  }}
```

shouldn't be necessary because ensure-safe-component tolerates `null` and `undefined` by passing them through.

But in general, this is a nice extension to the rules system.

mydea:
> I do want to point out that this usage:
> 
> ```
>   @afterOptionsComponent={{if 
>     @afterOptionsComponent 
>     (ensure-safe-component @afterOptionsComponent) 
>     undefined
>   }}
> ```
> 
> shouldn't be necessary because ensure-safe-component tolerates `null` and `undefined` by passing them through.

Ah, true, the example is actually "wrong", I wanted to show it with `null` as fallback, which is used in power select (where it differentiates between `undefined` = default component, `null` = no component).

PR #923, opened 2021-08-18
Add documentation how to use safe components in tests

mydea:
This PR adds a section to the safe component helper usage docs on how to use it in tests. It took me a bit to figure it out, so I figured it might help others to have this in there!

### DIFF
--- a/REPLACING-COMPONENT-HELPER.md
+++ b/REPLACING-COMPONENT-HELPER.md
@@ -210,3 +210,49 @@ export default class extends Component {
 This code will cause every modules under the `./feed-items/` directory to be eagerly included in your build.
 
 To instead _lazily_ include them, refactor to use asynchronous `import()` instead of `importSync`. BUT CAUTION: using `import()` of your own app code is one of the few things that works _only_ under Embroider and not in classic builds, so don't do it until you have committed to Embroider.
+
+## When using one-off components in tests
+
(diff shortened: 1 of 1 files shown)


ef4:
Thanks, this looks good.

Just wanted to mention that this is an area that gets nicer as soon as we have a syntax for strict mode templates. For example, with GJS (which you can try already with https://github.com/ember-template-imports/ember-template-imports) your example becomes:

```js

test('my test', async function(assert) {
  this.testComponent = <template>Test content: {{@message}}</template>;
  await render(hbs`
    <MyComponent @display={{this.testComponent}} />
  `);
});

We don't need to show that in the docs yet as it's nonstandard. But that's where the design is heading.

File: REPLACING-COMPONENT-HELPER.md [hunk 1]
```diff
@@ -210,3 +210,49 @@ export default class extends Component {
...
+
+  setComponentTemplate(
+    hbs`Test content: {{@message}}`,
+    TestComponent
+  );
+
+  this.testComponent = TestComponent;
+
+  await render(hbs`
+    <MyComponent @display={{component this.testComponent}} />
```
ef4, line 255: I think this can drop the `component` helper like

```hbs
<MyComponent @display={{this.testComponent}} />
```

mydea: True, I updated it!

Issue #54, opened 2019-03-07
Unified parallelBabel API


ef4:
Both broccoli-babel-transpiler and ember-cli-htmlbars-inline-precompile have invented their own mutually-incompatible solutions for plugin serialization.

broccoli-babel-transpiler allows a `_parallelBabel` key on _any_ value within the plugin config, which will cause that value to get replaced based on its `requireFile`, `buildWith`, `params`, and/or `method` properties. Everything but `requireFile` is optional.

ember-cli-htmlbars-inline-precompile allows a `parallelBabel` key on only the top-level values within the plugin config, which will cause that value to get replaced based on its `requireFile`, `buildWith`, and `params`, properties, which are all mandatory.

We're now stuck with these, but before this situation gets any worse I want to refactor both of them and embroider to share one library.

NullVoxPopuli:
is this still an issue? we don't use ec-htmlbars-inline-precompile anymore

PR #924, opened 2021-08-18
Compat htmlbars ast plugin wrapper

eoneill:
This addresses two issues.

The first is that most registered `htmlbars-ast-plugin` will currently opt-out of caching becaue there is no [caching strategy](https://github.com/babel/broccoli-babel-transpiler#caching) (`baseDir`/`cacheKey`) on the raw plugin. In classic build, this is derived from the plugin wrapper. In embroider, that context is currently lost. The fix here is to mirror `baseDir`/`cacheKey` onto the plugin itself.

The second is that we only "unwrap" plugins at the App level, not at the Addon level. This resulted in our builds failing when `THROW_UNLESS_PARALLELIZABLE=1`. The fix here is to also unwrap the plugins in the `V1Addon`.

The two changes allow our app to build with `THROW_UNLESS_PARALLELIZABLE=1` and we no longer get a bunch of...
```
broccoli-babel-transpiler is opting out of caching due to a plugin that does not provide a caching strategy ...
```

NOTE: I changed `plugin.parallelBabel` to `plugin._parallelBabel` because that's what `broccoli-babel-transpiler` [is expecting](https://github.com/babel/broccoli-babel-transpiler#parallel-transpilation), but I'm not entirely sure this is correct (given the [previous commit](https://github.com/embroider-build/embroider/pull/754)). If we need both for compat, we could mirror this...
```js
plugin._parallelBabel = plugin.parallelBabel;
```

References:
- https://github.com/babel/broccoli-babel-transpiler#caching
- https://github.com/babel/broccoli-babel-transpiler#parallel-transpilation
- https://github.com/embroider-build/embroider/pull/754
- https://github.com/embroider-build/embroider/issues/54


stefanpenner:
What can we do to prevent regressing this again? Is there a test scenario to add, or unit test to write? Obviously we want to the regression prevention to be as light wait as possible

stefanpenner:
@eoneill the current test failures are unrelated to you work, I will be investigating those myself now as part of https://github.com/embroider-build/embroider/pull/925

ef4:
Yeah, I don't think we should be renaming the `parallelBabel` property. It's not being read by ember-cli-babel directly, it's being read by ember-cli-htmlbars [which does expect `parallelBabel` and not `_parallelBabel`](https://github.com/ember-cli/ember-cli-htmlbars/blob/0e3ef495412bb75096d1468d0771e97c0afb8d1b/lib/utils.js#L318).

Otherwise this looks good to me.

packages/compat/src/prepare-htmlbars-ast-plugins.ts
```diff
@@ -0,0 +1,28 @@
+import { join } from 'path';
+
+export default function prepHtmlbarsAstPluginsForUnwrap(registry: any): void {
```
stefanpenner, line 3: should this be unit tested to prevent future regressions?

stefanpenner (changes requested):

packages/compat/src/prepare-htmlbars-ast-plugins.ts
ef4: I would suggest setting `THROW_UNLESS_PARALLELIZABLE` during something like `stage1.test.ts`.

We already run with this flag, but only in the macros package's test suite.

stefanpenner:
Several of your tests that where failing due to the ember global being missing should now be green: https://github.com/embroider-build/embroider/pull/925

eoneill:
> Yeah, I don't think we should be renaming the `parallelBabel` property. It's not being read by ember-cli-babel directly, it's being read by ember-cli-htmlbars [which does expect `parallelBabel` and not `_parallelBabel`](https://github.com/ember-cli/ember-cli-htmlbars/blob/0e3ef495412bb75096d1468d0771e97c0afb8d1b/lib/utils.js#L318).
> 
> Otherwise this looks good to me.

If I don't set `_parallelBabel`, it fails with...
```
[Babel: foo-addon] was configured to `throwUnlessParallelizable` and was unable to parallelize a plugin. 
plugins:
1: name: unknown, location: unknown

Please see: https://github.com/babel/broccoli-babel-transpiler#parallel-transpilation for more details
    at new Babel (/workspace/node_modules/broccoli-babel-transpiler/index.js:90:11)
    at Class.transpileTree (/workspace/node_modules/ember-cli-babel/index.js:169:16)
    at Object.toTree (/workspace/node_modules/ember-cli-babel/index.js:179:30)
    at /workspace/node_modules/ember-cli-preprocess-registry/preprocessors.js:188:26
    at Array.forEach (<anonymous>)
    at processPlugins (/workspace/node_modules/ember-cli-preprocess-registry/preprocessors.js:186:11)
    at Function.module.exports.preprocessJs (/workspace/node_modules/ember-cli-preprocess-registry/preprocessors.js:179:10)
    at Class.preprocessJs (/workspace/node_modules/ember-cli/lib/models/addon.js:1297:25)
    at V1Addon.transpile (/workspace/node_modules/@embroider/compat/src/v1-addon.js:407:35)
    at V1Addon.treeForAddon (/workspace/node_modules/@embroider/compat/src/v1-addon.js:605:25)
    at V1Addon.buildTreeForAddon (/workspace/node_modules/@embroider/compat/src/v1-addon.js:630:25)
    at V1Addon.build (/workspace/node_modules/@embroider/compat/src/v1-addon.js:891:14)
    at V1Addon.<anonymous> (/workspace/node_modules/typescript-memoize/dist/memoize-decorator.js:100:52)
    at V1Addon.get v2Trees (/workspace/node_modules/@embroider/compat/src/v1-addon.js:491:30)
    at V1Addon.<anonymous> (/workspace/node_modules/typescript-memoize/dist/memoize-decorator.js:100:52)
    at /workspace/node_modules/@embroider/compat/src/v1-addon.js:478:64
    at V1Addon.throughTreeCache (/workspace/node_modules/@embroider/compat/src/v1-addon.js:521:20)
    at V1Addon.get v2Tree [as v2Tree] (/workspace/node_modules/@embroider/compat/src/v1-addon.js:463:21)
    at /workspace/node_modules/@embroider/compat/src/build-compat-addon.js:42:48
    at Array.map (<anonymous>)
    at buildCompatAddon (/workspace/node_modules/@embroider/compat/src/build-compat-addon.js:42:33)
    at Object.cachedBuildCompatAddon [as default] (/workspace/node_modules/@embroider/compat/src/build-compat-addon.js:13:16)
    at /workspace/node_modules/@embroider/compat/src/compat-addons.js:46:105
    at Array.map (<anonymous>)
    at CompatAddons.get tree [as tree] (/workspace/node_modules/@embroider/compat/src/compat-addons.js:46:63)
    at CompatApp.augment (/workspace/node_modules/@embroider/core/src/build-stage.js:57:64)
    at CompatApp.get tree (/workspace/node_modules/@embroider/core/src/build-stage.js:26:50)
    at CompatApp.<anonymous> (/workspace/node_modules/typescript-memoize/dist/memoize-decorator.js:100:52)
```

This is because `broccoli-babel-transpiler` walks the entire object and recurses. When it hits the `plugins.ast` (which is an `Array<Function>`, it fails because the plugin `Function`s do not have a `_parallelBabel` option on them. 
See [broccoli-babel-transpiler/lib/parallel-api.js](https://github.com/babel/broccoli-babel-transpiler/blob/f0e3b0a39f1413aaff3245cfdc50c0a472920a52/lib/parallel-api.js#L35-L61)

I think the reasonable fix forwards is...
```js
// set `parallelBabel` derived from wrapper
if (parallelBabel && !plugin.parallelBabel) {
  plugin.parallelBabel = { ... };
}
// mirror onto `_parallelBabel`
if (plugin.parallelBabel && !plugin._parallelBabel) {
  plugin._parallelBabel = plugin.parallelBabel;
}
```

ef4:
Ok, that makes sense.

packages/compat/src/prepare-htmlbars-ast-plugins.ts
eoneill: Setting `THROW_UNLESS_PARALLELIZABLE=1` in `stage1.test.ts` definitely causes the tests to fail. I was able to confirm that the patch here works. However, even with this patch, the tests continue to fail as the `inject-babel-helpers` plugin (from `ember-source`) is not parallelizable.

This was fixed in https://github.com/emberjs/ember.js/pull/19047, but not backported to `ember-source@3.17` which is currently used in a few test-packages (e.g. [here](https://github.com/embroider-build/embroider/blob/153a074d59ebacb6ff9b867c3376e61c0992946f/test-packages/support/package.json#L21) and [here](https://github.com/embroider-build/embroider/blob/a122d4d356c3ad185745c21829be78e13b59d1bf/test-packages/fastboot-addon/package.json#L47)).

Bumping `ember-source` resolves the issue, however, I don't think I can currently bump these instances of `ember-source` above `3.17` per https://github.com/embroider-build/embroider/issues/927 / https://github.com/embroider-build/embroider/pull/934 (@stefanpenner).

stefanpenner: Ok that makes sense. @eoneill how about:

1) we merge this
2) open an issue specifically to enable `THROW_UNLESS_PARALLELIZABLE=1` once ember-source is bumped across the project

stefanpenner (approved):

packages/compat/src/prepare-htmlbars-ast-plugins.ts
stefanpenner: @eoneill some of the ember-sources' are at the right version, can additional `THROW_UNLESS_PARALLELIZABLE` be turned on, or are we blocked from completely until ember-source is upgraded.

Also, what is the lowest version of ember-source that satisfies the requirement?

stefanpenner: > Bumping ember-source resolves the issue, however, I don't think I can currently bump these instances of ember-source above 3.17 per #927 / #934 (@stefanpenner)

@eoneill  #927 / #934 are not ember-source, they are for ember-cli, but the fix appears to be in ember-source -> https://github.com/emberjs/ember.js/pull/19047/commits/104265a63c8dd4e9f28972223be7ddfbb914381f

Looking at the ember-source commit, it suggests that if we get to ember-source@~3.22 the issue you describe is addressed.

```
rg 'ember-source":'
test-packages/support/package.json
21:    "ember-source": "3.17.0",

packages/router/package.json
63:    "ember-source": "~3.16.0",

test-packages/macro-sample-addon/package.json
54:    "ember-source": "~3.10.0",

test-packages/fastboot-addon/package.json
47:    "ember-source": "~3.17.0",

test-packages/sample-transforms/package.json
46:    "ember-source": "~3.10.0",

packages/util/package.json
64:    "ember-source": "~3.21.1",

tests/app-template/package.json
54:    "ember-source": "~3.26.1",

tests/addon-template/package.json
53:    "ember-source": "~3.26.1",
```

I can quickly try to upgrade those, and see if we can land this with the flag enabled

stefanpenner: Let's see if we can just quickly sneak in a conservative ember-source bump to help with https://github.com/embroider-build/embroider/pull/935

PR #928, opened 2021-08-20
Update custom package rules for ember-basic-dropdown

mydea:
Based on https://github.com/embroider-build/embroider/issues/921#issuecomment-901186594, we do not need the custom rules anymore, since ember-basic-dropdown has wrapped everything in `ensure-safe-component` now anyhow (https://github.com/cibernox/ember-basic-dropdown/blob/master/CHANGELOG.md)


ef4:
Thanks.

PR #930, opened 2021-08-23
add note on route splitting with pods in readme

mydea:
After spending some time trying to figure out why route splitting did not work, I finally found out that it was because I am using pod layout for my routes, but had no `podModulePrefix` defined - which does work normally as far as I can tell, but Embroider has some code checking for `podModulePrefix !== undefined` which is not triggered in this case.

I figured it might make sense to add this to the readme of @embroider/router, as others might also run into this issue!


ef4:
Thanks

PR #934, opened 2021-08-23
Bump test-packages/support’s to known failing version of ember-cli fo…

stefanpenner:
…r debugging


packages/compat/src/v1-addon.ts
```diff
@@ -729,6 +729,12 @@ export default class V1Addon {
   private buildAddonStyles(built: IntermediateBuild) {
     let addonStylesTree = this.addonStylesTree();
     if (addonStylesTree) {
+      let destDir = (addonStylesTree as any).destDir;
+
+      if (typeof destDir === 'string' && destDir.endsWith('__COMPILED_STYLES__')) {
```
stefanpenner, line 734: just a quick work-around to test CI

stefanpenner: Some some thoughts:

* chat with @rwjblue to verify my understanding of the `__COMPILED_STYLES__` change in ember-cli
* I don't believe we have a built-in relocating system for css, so if this path is the right approach we will need to implement something.
* One possible approach is: 1) if we detect this exact case, apply this exact fix ... otherwise 2) implement a broccoli funnel or plugin which undoes the nesting if needed.

ef4:
These tests haven't been ported to scenario-tester yet, which is why we didn't see this sooner.

stefanpenner:
@ef4 ya porting to scenario-tester seems critical. My prioritization for this PR though is: 1) land a fix 2) migrate tests `scenario-tester` as it seems like (1) basically blocks (2).

PR #947, opened 2021-08-26
tolerate missing rootURL on asset references

eoneill:
Currently, if `{{rootURL}}` is not present on the app js entrypoint in `index.html`, the build fails in stage 2 with:
```
could not find app javascript in index.html
```

The fix here (see `v1-app.ts`) is to be more forgiving when "finding" the app js entrypoint by making the `rootURL` prefix optional.
That is, `<script src="assets/app.js">` and `<script src="{{rootURL}}assets/app.js">` are both acceptable.

The primary argument for making `{{rootURL}}` optional is that prefixing the asset path with `{{rootURL}}` is not currently required in classic build, and this behavior is strictly being introduced in the `@embroider/compat` layer.

Additionally, `HTMLEntrypoint#relativeToApp` can currently result in inaccurate warnings for asset paths that contain the `rootURL` within the path. e.g. given, `rootURL = '/'`:
```
<link href="assets/foo/bar.css"/>
<script src="assets/fiz/buzz.js"></script>
```
... results in (note the first missing `/` in the paths):
```
warning: in index.html  <link rel="stylesheet" href="assetsfoo/bar.css"> does not exist on disk. If this is intentional, use a data-embroider-ignore attribute.
warning: in index.html  <script src="assetsfiz/buzz.js"> does not exist on disk. If this is intentional, use a data-embroider-ignore attribute.
```
The fix here (see `html-entrypoint.ts`) is to only strip the `rootURL` if the path `startsWith(rootURL)`. This ensures we surface accurate warnings.

## Alternate solutions
The above may be controversial as prefixing with `{{rootURL}}` is probably the _correct_ fix. However, we currently have an app that cannot successfully run with `{{rootURL}}` prefix (this is an issue of our own doing and not a compelling argument).

An alternate approach would be to look for a `data-*` attribute to find the entrypoint.

### DIFF
--- a/packages/compat/src/v1-app.ts
+++ b/packages/compat/src/v1-app.ts
@@ -652,12 +652,14 @@ export default class V1App {
 
   private withoutRootURL(src: string) {
     let rootURL = this.config.readConfig().rootURL;
+    // remove the leading slash for now
+    src = src.replace(/^\//, '');
+    // only remove the rootURL if the src startsWith it
     if (src.startsWith(rootURL)) {
-      src = '/' + src.slice(rootURL.length);
-    } else if (src.startsWith('/' + rootURL)) {
-      src = src.slice(rootURL.length);
+      src = src.replace(rootURL, '');
     }
-    return src;
+    // always add a leading slash
+    return '/' + src;
   }
 
   findAppScript(scripts: HTMLScriptElement[], entrypoint: string): HTMLScriptElement {
--- a/packages/compat/tests/stage2.test.ts
+++ b/packages/compat/tests/stage2.test.ts
@@ -764,4 +764,36 @@ i(\"../../node_modules/lazy-engine/lazy-engine.css\");
   }`);
     });
   });
+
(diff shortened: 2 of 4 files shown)


ef4:
After discussing, it sounds like the app in question can fix its own quirks in postprocessing.

I don't want to introduce this level of looseness into embroider. Even under classic builds, apps without `{{rootURL}}` have broken deep-linking to their nested child routes. It's just left as a runtime error as opposed to embroider's build-time error. While it's true that people patch over that problem with other custom infrastructure, that's not something embroider can count on.

I do think we should work to provide better error messages, docs, and/or audit tooling to point out this problem when people are trying to migrate to embroider.

stefanpenner:
@ef4 should we close this, or change this PR into said error messages?

stefanpenner:
re: https://github.com/embroider-build/embroider/pull/947#issuecomment-909530353 I believe we should close this. If I closed this in error, please feel to comment explaining why and we can gladly reopen.

Issue #894, opened 2021-07-11
Local helpers not resolved in tests


NullVoxPopuli:
```
Build Error (PackagerRunner) in tests/rendering/compile-hbs-test.ts

Module build failed (from ../../../../../home/me/Development/NullVoxPopuli/ember-play/node_modules/thread-loader/dist/cjs.js):
Thread Loader (Worker 0)
$TMPDIR/embroider/5e47be/tests/dummy/tests/rendering/compile-hbs-test.ts/compile-hbs-test.ts: Missing helper: this.compile in tests/rendering/compile-hbs-test.ts

    at CompatResolver.dependenciesOf (/home/me/Development/NullVoxPopuli/ember-play/node_modules/@embroider/compat/src/resolver.js:251:33)
    at NodeTemplateCompiler.precompile (/home/me/Development/NullVoxPopuli/ember-play/node_modules/@embroider/core/src/template-compiler-common.js:78:42)
    at handleTagged (/home/me/Development/NullVoxPopuli/ember-play/node_modules/@embroider/core/src/babel-plugin-inline-hbs.js:68:62)
    at PluginPass.TaggedTemplateExpression (/home/me/Development/NullVoxPopuli/ember-play/node_modules/@embroider/core/src/babel-plugin-inline-hbs.js:38:29)
    at PluginPass.<anonymous> (/home/me/Development/NullVoxPopuli/ember-play/node_modules/@embroider/core/src/portable-babel-launcher.js:37:29)
    at newFn (/home/me/Development/NullVoxPopuli/ember-play/node_modules/@babel/traverse/lib/visitors.js:171:21)
    at NodePath._call (/home/me/Development/NullVoxPopuli/ember-play/node_modules/@babel/traverse/lib/path/context.js:53:20)
    at NodePath.call (/home/me/Development/NullVoxPopuli/ember-play/node_modules/@babel/traverse/lib/path/context.js:40:17)
    at NodePath.visit (/home/me/Development/NullVoxPopuli/ember-play/node_modules/@babel/traverse/lib/path/context.js:90:31)
    at TraversalContext.visitQueue (/home/me/Development/NullVoxPopuli/ember-play/node_modules/@babel/traverse/lib/context.js:99:16)
```

```ts
  test('it works', async function (assert) {
    assert.expect(5);

    this.setProperties({
      compile: () => {
        let template = `
          {{#each (array 1 2) as |num|}}
            <output>{{num}}</output>
          {{/each}}
        `;

[...]

PR #953, opened 2021-08-30
fixes: Local helpers not resolved in tests #894

lifeart:
should fix https://github.com/embroider-build/embroider/issues/894
we suppose to skip instance scope-defined helpers, because it's declared in JS/TS file and could be resolved


ef4:
Nice! Thanks.

PR #957, opened 2021-09-01
[Experiment] Skip yarn —frozen-lockfile if we had a cache hit.

stefanpenner:
In theory this should be safe, as our cache key is based: `os + hashFile(**/yarn.lock)`.

This would reduce our install times by 40s -> 1.5minutes or up to about 100 GH-Action minutes or so.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,16 +24,13 @@ jobs:
       - uses: actions/cache@v2
         id: yarn-cache
         with:
-          path: '**/node_modules'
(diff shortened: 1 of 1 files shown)


stefanpenner:
Ah. Rather than caching the yarn cache dir, we would need to cache the various node_modules directories in our project. Which is something that most likely could be done

stefanpenner:
Bummer it appears to have some issues on windows. I bet I can fix those, most likely related to workspaces on windows. I’ll look at it further next time I have some spare cycles

Issue #968, opened 2021-09-09
Production builds produce incorrect browser runtimes for fastboot assets 


thoov:
## Steps to reproduce:

1.  `git clone git@github.com:thoov/fastboot-macro-issue.git`
2.  `cd fastboot-macro-issue && yarn`

Then running the following commands: 

- `CLASSIC=true ember s`
- `CLASSIC=true ember s -prod`
- `ember s`

And visiting http://ocalhost:4200 in the browser you will correctly see "browser". However if you run:

- `ember s -prod`

you will see "fastboot" printed indicating that the fastboot service is being used and not the app's service.

thoov:
Some background information:

When we encounter a fastboot asset (such as `fastboot/service/foo.js`) we create a ["switcher template"](https://github.com/embroider-build/embroider/blob/master/packages/core/src/app-differ.ts#L175-L189) for it. This uses a `macroCondition` to create a "switch" for fastboot and browser imports which works fine in dev mode. However, in production mode we eliminate the runtime aspect of macros (and elimiate the "false" branches) and thus this conditional removes the app service and only leaves the fastboot one.

thoov:
@ef4 Is this double build supposed to be working or is it not implemented?

ef4:
It's supposed to be working and tested. See `fastboot-app-test.ts`.

ef4:
That runs both in-browser tests and fastboot tests for both development and production.

ef4:
I think this was fixed by #974

PR #974, opened 2021-09-20
Issue with fastboot production assets and "switchable" template

thoov:
Failing test case for #968


ef4:
`applyVariantToBabelConfig` was mutating its input, which lived in the shared node require cache. So if the same file happened to get handed to the same worker thread for both fastboot and non-fastboot processing of the same file you'd get a polluted value of the babel config in the second one.

PR #975, opened 2021-09-20
unify matrix + linting jobs

stefanpenner:
Matrix generation is nearly instant, no need to spin up a whole job job
where 99.9% of the time is spent installing & building


stefanpenner:
Don't merge, CI is actually failing..

PR #976, opened 2021-09-21
Restructure workspaceDir logic

thoov:
Consolidate the logic for workspaceDir (the tmp location used for stage3) so that a proper `Building into ...` message can be displayed (it is currently displayed in the middle of other console.ui writes and can be lost / looks incorrect). This also standardizes the actual outputted "Building into ..." location and our internal workspaceDir. 

Building into now appears at the top of stdout:

Screen Shot 2021-09-21 at 4 57 06 PM

### DIFF
--- a/packages/compat/src/default-pipeline.ts
+++ b/packages/compat/src/default-pipeline.ts
@@ -25,8 +25,8 @@ export default function defaultPipeline<PackagerOptions>(
   packager?: PackagerConstructor<PackagerOptions>,
   options: PipelineOptions<PackagerOptions> = {}
 ): Node {
+  let outputPath: string;
   let addons;
-  let appDestDir: string;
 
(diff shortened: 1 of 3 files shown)


packages/compat/src/default-pipeline.ts
```diff
@@ -1,50 +1,57 @@
...
-        options.onOutputPath(result.outputPath);
-      } else {
-        console.log(`Building into ${result.outputPath}`);
-      }
-      outputPath = result.outputPath;
-    });
+
+    if (options && options.onOutputPath) {
+      options.onOutputPath(options.workspaceDir);
+    }
```
rwjblue, line 44: Let's remove this hook for now (instead of changing the semantics fundamental), if we need to bring it back we can rationalize + document the semantics.

rwjblue (approved):

PR #982, opened 2021-09-26
Try private getter reproduction

NullVoxPopuli:
Just seeing if tests pass with this little addition (which hopefully doesn't change any of the actual tests results).

I couldn't find any tests of real private properties or methods, so I arbitrarily adding something here


NullVoxPopuli:
I guess, I'm not sure if this even is a valid reproduction case 🤔

NullVoxPopuli:
private getters work fine now

PR #990, opened 2021-10-01
Invalidate @embroider/macro's babel cache when addon version's change without mutating lock file (e.g. linking)

thoov:
This ensures that using things like `yarn link`/`npm link` to test dependencies still invalidates our babel plugin's cache in most cases.

cc: @rwjblue

### DIFF
--- a/packages/macros/src/macros-config.ts
+++ b/packages/macros/src/macros-config.ts
@@ -41,6 +41,18 @@ type GlobalSharedState = WeakMap<
 // share the GlobalSharedState beneath.
 let localSharedState: WeakMap<any, MacrosConfig> = new WeakMap();
 
+// creates a string representing all addons and their versions
+// (foo@1.0.0|bar@2.0.0) to use as a cachekey
+function gatherAddonCacheKey(item: any, memo = new Set()) {
+  item.addons.forEach((addon: any) => {
+    let key = `${addon.pkg.name}@${addon.pkg.version}`;
+    memo.add(key);
+    gatherAddonCacheKey(addon, memo);
+  });
+
+  return [...memo].join('|');
+}
+
 export default class MacrosConfig {
   static for(key: any): MacrosConfig {
     let found = localSharedState.get(key);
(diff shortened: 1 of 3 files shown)


packages/macros/src/macros-config.ts
```diff
@@ -258,8 +270,9 @@ export default class MacrosConfig {
   // normal node_modules resolution can find their dependencies. In other words,
   // owningPackageRoot is needed when you use this inside classic ember-cli, and
   // it's not appropriate inside embroider.
-  babelPluginConfig(owningPackageRoot?: string): PluginItem[] {
+  babelPluginConfig(appOrAddonInstance?: any): PluginItem[] {
     let self = this;
+    let owningPackageRoot = appOrAddonInstance ? appOrAddonInstance.root || appOrAddonInstance.project.root : null;
```
rwjblue, line 275: Looks like this introduces a build error:

```
Error: packages/macros/src/macros-config.ts(275,108): error TS1005: ':' expected.
```

I think this might work:

```suggestion
    let owningPackageRoot = appOrAddonInstance?.root ?? null;
```

packages/macros/src/macros-config.ts
```diff
@@ -300,7 +313,13 @@ export default class MacrosConfig {
     // the old evaluated state being used which might be invalid. This cache busting plugin keeps track of a
     // hash representing the lock file of the app and if it ever changes forces babel to rerun its plugins.
     // more information in issue #906
-    let cacheKey = crypto.createHash('sha256').update(lockFileBuffer).digest('hex');
+    let hash = crypto.createHash('sha256');
+    hash = hash.update(lockFileBuffer);
+    if (appOrAddonInstance) {
+      hash = hash.update(gatherAddonCacheKey(appOrAddonInstance.project));
```
rwjblue, line 319:
```suggestion
      // ensure that the actual running addon names and versions are accounted
      // for in the cache key; this ensures that we still invalidate the cache
      // when linking another project (e.g. ember-source) which would normally
      // not cause the lockfile to change;
      hash = hash.update(gatherAddonCacheKey(appOrAddonInstance.project));
```

rwjblue (approved):

PR #995, opened 2021-10-02
Implement manifest.json output

NullVoxPopuli:
Todo: 

- [ ] tests
- [ ] prepend rootURL?

Questions:
 - should we add an option to embed this in the index.html?
 - I think the manifest should always be output, even though historically it's been production-only _AND_ opt-in. Thoughts?

### DIFF
--- a/packages/webpack/src/ember-webpack.ts
+++ b/packages/webpack/src/ember-webpack.ts
@@ -99,6 +99,7 @@ const Webpack: PackagerConstructor<Options> = class Webpack implements Packager 
   async build(): Promise<void> {
     let appInfo = this.examineApp();
     let webpack = this.getWebpack(appInfo);
+    debugger;
     let stats = this.summarizeStats(await this.runWebpack(webpack));
     await this.writeFiles(stats, appInfo);
   }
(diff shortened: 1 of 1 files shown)


ef4:
What are you actually trying to do with the asset map?

Because I am dubious about the value of asset maps. Historically they have caused lock-in to specific output formats which reduces our ability to optimize builds without breaking the format.

NullVoxPopuli:
These css imports are not swapped for their chunked versions:
https://github.com/NullVoxPopuli/limber/blob/main/frontend/app/components/shadowed.ts#L23

NullVoxPopuli:
So, my main use case is valid -- I need to be able to manually specify the CSS I want inside a shadow dom. 
The specifics of how are flexible of course, but, I've now tried to import my app css in the app.ts file, and it seemed to resolve, but the manifest from the just-webpack-plugins still does not include CSS :thinking:

ef4:
How much do you care about specifically getting the traditional app.css & vendor.css (which are not module-aware and don't go through webpack at all) vs getting styles that you can control in a more module-aware way?

Because if it's just about those two traditional output files, I would rather special-case those and create a way to express a dependency on them from within an app's code.

Whereas if you want to send your styles through webpack and avoid the legacy style pipeline, you can establish your own webpack entrypoint for them in order to take control over their output URL.

Either way, I don't think the solution is an asset manifest. Rather, it's a way to express a dependency on output URLs from within your code. Related to the things discussed in https://github.com/emberjs/rfcs/pull/763

NullVoxPopuli:
Yeah, I mean having things be more proper and easy with less hacks is def a direction I want to go in.
I'll try adding an entrypoint for my styles and avoid the legacy style pipeline -- this _sounds like_ a step towards getting tailwind JIT working, too. :D

PR #1005, opened 2021-10-15
Remove command documentation from `addon-shim` package

simonihmig:
As the command has been moved to `addon-dev`.

### DIFF
--- a/packages/addon-shim/README.md
+++ b/packages/addon-shim/README.md
@@ -55,16 +55,6 @@ module.exports = addonV1Shim(__dirname, {
 
 This option _only_ works in non-embroider builds. Under embroider, apps just won't import the parts of your addon they don't want.
 
-## addon-shim command
-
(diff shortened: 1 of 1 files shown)


ef4:
Good catch, thanks.

(And it looks like I need to figure out what dependency suddenly broke a lot of ci jobs, but that shouldn't block this docs pr.)

simonihmig:
> (And it looks like I need to figure out what dependency suddenly broke a lot of ci jobs, but that shouldn't block this docs pr.)

Yeah, this already cost me some hours today. It seems to be https://github.com/socketio/socket.io/issues/4121, which basically makes all ember tests that have no locked dependencies fail. We can temporarily fix this by pinning `socket.io` to `4.2.0` using yarn resolutions, or just wait for @chriskrycho's [fix](https://github.com/socketio/socket.io-client/pull/1503) to get merged and released.

PR #1006, opened 2021-10-16
ESBuild for minification

NullVoxPopuli:
Remaining TODOs
- [ ] figure out the sourcemaps URLs
- [x] add equiv of babelLoaderOptions for customizing the minification plugin (non vendor.js minification)
- [x] get tests passing


packages/webpack/src/ember-webpack.ts
```diff
@@ -238,15 +252,19 @@ const Webpack: PackagerConstructor<Options> = class Webpack implements Packager
 
     // loading these lazily here so they never load in non-production builds.
     // The node cache will ensures we only load them once.
-    const [Terser, srcURL] = await Promise.all([import('terser'), import('source-map-url')]);
+    const srcURL = await import('source-map-url');
+    // const [Terser, srcURL] = await Promise.all([import('terser'), import('source-map-url')]);
```
ef4, line 256: I think we can load esbuild here lazily just like we did with terser, so it never even loads in dev builds.

NullVoxPopuli: Do we want to manage the esbuild version in embroider (for consistent versions between the two minification bits)? If so, that requires the implementation key in the plugin, which requires a regular import 🤷

package.json
```diff
@@ -38,7 +38,8 @@
   "resolutions": {
     "**/browserslist": "^4.14.0",
     "**/fastboot": "^3.1.0",
-    "**/qunit": "^2.14.1"
+    "**/qunit": "^2.14.1",
+    "socket.io": "4.2.0"
```
NullVoxPopuli, line 42: i'll revert this change once socket.io releases a fix

NullVoxPopuli: ref: https://github.com/socketio/socket.io/issues/4121#issuecomment-943832512
ref: https://github.com/socketio/socket.io-client/pull/1503

packages/webpack/src/ember-webpack.ts
ef4: I was assuming we'd only configure the plugin in production builds anyway, so the `await import` and the appending of the minifier plugin could live in the same conditional block. But if there's a reason to keep the plugin configured all the time that's OK.

packages/webpack/package.json
```diff
@@ -30,17 +30,19 @@
...
+    "esbuild-loader": "^2.16.0",
     "fs-extra": "^9.1.0",
     "jsdom": "^16.6.0",
     "lodash": "^4.17.21",
     "mini-css-extract-plugin": "^1.6.0",
     "semver": "^7.3.5",
     "source-map-url": "^0.4.1",
     "style-loader": "^2.0.0",
+    "supports-color": "^8.1.0",
     "terser": "^5.7.0",
```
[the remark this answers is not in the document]
NullVoxPopuli: nope!

NullVoxPopuli:
do any ember-related packages of tests of what we expect for source map URLs? like.. what sorts of configurations can there be?

NullVoxPopuli:
no longer needed

PR #1010, opened 2021-10-23
template compilation improvements

ef4:
First, I'm refactoring so that we standardize all template compilation as inline template compilation, within Javascript, because that is the more fully-featured case. Specifically, a standalone HBS file can go through a trivial transformation to make it into a Javascript file:

```hbs
<Hello />
```

```js
export default hbs("<Hello />");
```

After which point we can share all the same standard machinery for compiling forward from there.

Second, I'm going to take advantage of template lexical scope in non-strict mode which we just successfully landed in ember-source 3.28.4. That will significantly streamline the build output when using `staticComponents`, etc, because we can pass the statically-resolved things directly into the template and completely sidestep the runtime resolver.

The first commit here is just testing out the basic idea, it needs further cleanup refactoring.


ef4:
Reminder to self: `template-compiler-broccoli-plugin.ts` can drop its stage3 support, it will only need to handle stage1.

ef4:
These changes can also address a cache-safety whole we currently have: the stage1 `TemplateCompilerBroccoliPlugin` uses broccoli-persistent-filter but only caches off:
 - hashDeps of `@embroider/compat`
 - the checksum of ember-template-compiler.js

It doesn't incorporate a checksum over the set of plugins.

ef4:
I'm going to land this long-running branch, it has a good refactor and all tests are passing.

The actual feature work on top to use lexical scope can be a followup.

PR #1012, opened 2021-11-01
Update sample-rollup.config.js

NullVoxPopuli:
Adds some things I encountered while setting this up: https://github.com/NullVoxPopuli/ember-resources/blob/main/ember-resources/rollup.config.js

### DIFF
--- a/packages/addon-dev/sample-rollup.config.js
+++ b/packages/addon-dev/sample-rollup.config.js
@@ -1,5 +1,8 @@
-import babel from '@rollup/plugin-babel';
+import { babel } from '@rollup/plugin-babel';
+
 import { Addon } from '@embroider/addon-dev/rollup';
+
+import packageJson from './package.json';
 
 const addon = new Addon({
   srcDir: 'src',
@@ -10,6 +13,9 @@ export default {
   // This provides defaults that work well alongside `publicEntrypoints` below.
   // You can augment this if you need to.
   output: addon.output(),
+
+  // This prevents peerDependencies from being included in your bundle
+  external: Object.keys(packageJson['peerDependencies']),
 
   plugins: [
     // These are the modules that users should be able to import from your
@@ -27,6 +33,7 @@ export default {
     babel({
       plugins: ['@embroider/addon-dev/template-colocation-plugin'],
       babelHelpers: 'bundled',
+      extensions: ['.js', '.ts'],
     }),
 
     // Follow the V2 Addon rules about dependencies. Your code can import from


packages/addon-dev/sample-rollup.config.js
```diff
@@ -1,15 +1,28 @@
...
   srcDir: 'src',
   destDir: 'dist',
 });
 
+const extensions = ['.js', '.ts'];
+
 export default {
+  // This tells rollup to follow the node-resolution algorithm for finding
+  // dependencies / and files within those dependencies
+  nodeResolve({ extensions }),
```
rwjblue, line 18: In general, I don't think we want folks to do this. This will make rollup embed your dependencies (not allow them to be resolved at application build time), which would be quite unfortunate. It would mean that the addon author in question would have to redeploy a new patch release for any security/patch releases in dependencies... 

More details below in the `externals` section.

packages/addon-dev/sample-rollup.config.js
```diff
@@ -1,15 +1,28 @@
...
+    
   // This provides defaults that work well alongside `publicEntrypoints` below.
   // You can augment this if you need to.
   output: addon.output(),
+  
+  // This prevents peerDependencies from being included in your bundle
+  external: Object.keys(packageJson['peerDependencies']),
```
rwjblue, line 25: This is great! We need to beef it up though, it should also include:

* The list of Ember's pseudo packages (e.g. `@ember/object` and friends). We should get that list from `@embroider/shared-internals` and expose it on the `Addon` instance we are using here.
* The list of dependencies (the actual resolution / bundling will be done by the final app packager; either ember-auto-import@2 in classic builds or @embroider/webpack in Embroider builds)

packages/addon-dev/sample-rollup.config.js
```diff
@@ -1,15 +1,28 @@
-import babel from '@rollup/plugin-babel';
+import { babel } from '@rollup/plugin-babel';
+import { nodeResolve } from '@rollup/plugin-node-resolve';
+
 import { Addon } from '@embroider/addon-dev/rollup';
 
+import packageJson from './package.json';
```
rwjblue, line 6: Hmm, what makes this work? By default you can't import from `.json` files in Node...

packages/addon-dev/sample-rollup.config.js
```diff
@@ -1,15 +1,28 @@
...
+
 const addon = new Addon({
   srcDir: 'src',
   destDir: 'dist',
 });
 
+const extensions = ['.js', '.ts'];
```
rwjblue, line 13: If we remove `nodeResolve`, I think we ought to be able move this to inline with `babel` with a nice little comment?

rwjblue (changes requested):

packages/addon-dev/sample-rollup.config.js
ef4: Rollup configs go through their own weird preprocessing. That's also why they're written as ES modules despite predating Node's native ES module support.

ef4:
I don't think we want `nodeResolve` in this default config. It will cause the dependencies to get copied into `dist`, which defeats yarn/npm deduplication and makes it impossible for consumers to benefit from semver-compatible releases of those dependencies.

Even in cases where rolling up a dependency is desired, it's only appropriate to do that for `devDependencies`, not `dependencies`, because otherwise consumers will end up with two copies.

So I think the real bug here is that we're letting rollup emit misleading warnings. The current output is correct, we just want to stop the warnings. I can whip up another little plugin to put in the default config that will achieve that.

ef4:
Here's how I would prefer to deal with the standard externals: https://github.com/embroider-build/embroider/pull/1015

NullVoxPopuli:
legit, I'll rebase and update this once that's merged

rwjblue:
Should we close this now that #1015 landed?

NullVoxPopuli:
I think the updates cover the dependency aspect of what I was trying to solve.

Should we still keep the `extensions` entry?

ef4:
I think this use of extensions is misleading. It makes it look like the config supports TypeScript, but it does not. For TS you still need to also add `@babel/plugin-transform-typescript` or `@rollup/plugin-typescript`. So I think we should leave it off.

We should probably be moving toward maintaining real blueprints rather than just this example file, in which case we could maintain a typescript blueprint that extends the base one, and that would be a good place to put all this.

NullVoxPopuli:
Sounds good!

PR #1015, opened 2021-11-04
Address misleading warnings from rollup about externals

ef4:
When building a v2 addon we deliberately *don't* want rollup to do anything with imports of our dependencies or the various ember-provided package names we might refer to. And that is what already happens, except rollup emits warnings about them.

This adds a plugin that handles these dependenices correctly, making them explicitly as external so there's no warning, while ensuring that if you stray from the allowed resolvable dependencies you will still see warnings.


NullVoxPopuli (approved):
noice!

PR #1017, opened 2021-11-11
Ensure components + helpers can work from `this` paths with `staticComponents = true` & `staticHelpers = true`

thoov:
Allow components from `{{#this.foobar}}baz{{/this.foobar}}` to be static. Needed for: adopted-ember-addons/ember-collapsible-panel#87

### DIFF
--- a/packages/compat/src/resolver-transform.ts
+++ b/packages/compat/src/resolver-transform.ts
@@ -64,9 +64,6 @@ export function makeResolverTransform(resolver: Resolver) {
           if (node.path.type !== 'PathExpression') {
             return;
           }
-          if (node.path.original.includes('liquidif')) {
-            debugger;
-          }
           if (node.path.this === true) {
             return;
           }
@@ -82,9 +79,6 @@ export function makeResolverTransform(resolver: Resolver) {
         MustacheStatement(node: ASTv1.MustacheStatement) {
           if (node.path.type !== 'PathExpression') {
             return;
-          }
-          if (node.path.original.includes('liquidif')) {
-            debugger;
           }
           if (scopeStack.inScope(node.path.parts[0])) {
             return;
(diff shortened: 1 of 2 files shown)


packages/compat/src/resolver-transform.ts
```diff
@@ -28,6 +28,9 @@ export function makeResolverTransform(resolver: Resolver) {
           if (scopeStack.inScope(node.path.parts[0])) {
             return;
           }
+          if (node.path.head && node.path.head.type === 'ThisHead') {
```
rwjblue, line 31: Can we do this?
```suggestion
          if (node.path.type === 'PathExpression' && node.path.this === true) {
```

packages/compat/tests/resolver.test.ts
```diff
@@ -476,6 +476,20 @@ describe('compat-resolver', function () {
     let findDependencies = configure({ staticHelpers: true });
     expect(findDependencies('templates/application.hbs', `{{(this.myHelper 42)}}`)).toEqual([]);
   });
+  test('class defined component not failing if there is no arguments', function () {
+    let findDependencies = configure({ staticComponents: true, staticHelpers: true });
+    expect(findDependencies('templates/application.hbs', `{{this.myComponent}}`)).toEqual([]);
+  });
```
rwjblue, line 482:
```suggestion
  test('helper defined in component not failing if there is no arguments', function () {
    let findDependencies = configure({ staticComponents: true, staticHelpers: true });
    expect(findDependencies('templates/application.hbs', `{{#if (this.myHelper)}}{{/if}}`)).toEqual([]);
  });
```

rwjblue: This one should already work (the subexpression hook already had the guard you are adding here)

PR #1021, opened 2021-11-15
Add `staticModifiers` option

### LINKED ISSUE #603: Add `staticModifiers` flag (to support shaking unused modifiers and route splitting)
`AppFiles` does not handle modifiers, they are all put under `otherAppFiles`. I believe they should be treated similar to helpers, so they get split into route bundles, right?

Windvis:
~~still need to add tests and actually test if it works :grimacing:~~ I added some tests. Let me know if I missed something :pray:.

Closes #603

### DIFF
--- a/packages/compat/src/resolver-transform.ts
+++ b/packages/compat/src/resolver-transform.ts
@@ -1,7 +1,7 @@
 import { default as Resolver, ComponentResolution, ComponentLocator } from './resolver';
 import type { ASTv1 } from '@glimmer/syntax';
 
-// This is the AST transform that resolves components and helpers at build time
+// This is the AST transform that resolves components, helpers and modifiers at build time
 // and puts them into `dependencies`.
 export function makeResolverTransform(resolver: Resolver) {
   function resolverTransform({ filename }: { filename: string }) {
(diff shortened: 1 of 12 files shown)


packages/compat/src/resolver-transform.ts
```diff
@@ -111,6 +111,16 @@ export function makeResolverTransform(resolver: Resolver) {
             }
           }
         },
+        ElementModifierStatement(node: ASTv1.ElementModifierStatement) {
+          if (node.path.type !== 'PathExpression') {
+            return;
+          }
+          if (node.path.this === true) {
```
lifeart, line 118: we need same for `node.path.data === true`  ({{@fooBar}})

Windvis: Thanks, didn't know about that!

Windvis: Added! I also forgot about "contextual" modifiers so I added that as well.

lifeart: @Windvis could we cover "modifier" helper usage?

like

```hbs
{{let (modifier "foo-bar" arg param=1) as |my-modifier|}}
   <div {{my-modifier}} ></div>
{{/let}}
```

lifeart (approved):

Windvis:
CI seems to fail due to #1024 which should be fixed by #1025

lifeart (approved):

PR #1025, opened 2021-11-18
[@embroider/router] update ember-auto-import to v2

### LINKED ISSUE #1024: `@embroider/router ember-release` job fails due to ember-auto-import v1
With the release of Ember 4 ember-auto-import v2 is now required. It seems that the `@embroider/router` addon still has [v1 listed as a devDependency ](https://github.com/embroider-build/embroider/blob/bec54e21af303b07a600c4c220ccb3032d5f4619/packages/router/package.json#L47)which means [the "release" jobs fail in CI](https://github.com/embroider-build/embroider/runs/4243114985?check_suite_focus=true).

I think we can just bump the ember-auto-import dependency in that package without requiring v2 in consuming apps since it's a devDependency?

Windvis:
Closes #1024

### DIFF
--- a/packages/router/package.json
+++ b/packages/router/package.json
@@ -44,7 +44,7 @@
     "babel-eslint": "^10.1.0",
     "broccoli-asset-rev": "^3.0.0",
     "cross-env": "^7.0.3",
-    "ember-auto-import": "^1.11.2",
+    "ember-auto-import": "^2.2.0",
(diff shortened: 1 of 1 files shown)


Windvis:
It seems bumping ember-auto-import to v2 was only a part of the work needed to support Ember v4 in tests. I created a new PR for that which also includes the ember-auto-import v2 version bump.

PR #1029, opened 2021-11-20
Don't resolve built-in components when used with the component helper

### LINKED ISSUE #815: `{{component "link-to"}}` + staticComponents error
While trying out Embroider (v0.40.0) and enabling the staticComponents compat flag, I noticed the following error:

`Missing component: link-to in node_modules/@appuniversum/ember-appuniversum/components/au-link.hbs`

Which points to this file: https://github.com/appuniversum/ember-appuniversum/blob/52566adf81dc1225bf2573eaa9a1aa5b2dd9b641/addon/components/au-link.hbs#L2

According to the [replacing the component helper docs](https://github.com/embroider-build/embroider/blob/master/REPLACING-COMPONENT-HELPER.md) this should work since it's a static string. From the Discord I understood that `link-to` is not _actually_ a component which is probably the reason for this error message.

I know there are workarounds to get it to work, but I just wanted to know if using `link-to` with the component helper like that _is supposed to_ work under Embroider and that this should be considered a bug.

Relevant discord discussion: https://discord.com/channels/480462759797063690/568935504288940056/841617811105120276

Windvis:
It seems that the code that handles the component helper doesn't ignore the built-in components like the other [component resolving code does](https://github.com/embroider-build/embroider/blob/bec54e21af303b07a600c4c220ccb3032d5f4619/packages/compat/src/resolver.ts#L667-L669).

Closes #815

### DIFF
--- a/packages/compat/src/resolver.ts
+++ b/packages/compat/src/resolver.ts
@@ -731,6 +731,11 @@ export default class CompatResolver implements Resolver {
         from
       );
     }
+
+    if (builtInComponents.includes(component.path)) {
+      return null;
+    }
+
     let found = this.tryComponent(component.path, from);
     if (found) {
       return this.add(found, from);
--- a/packages/compat/tests/resolver.test.ts
+++ b/packages/compat/tests/resolver.test.ts
@@ -518,6 +518,21 @@ describe('compat-resolver', function () {
       },
     ]);
   });
+  test('built-in components are ignored when used with the component helper', function () {
+    let findDependencies = configure({
+      staticComponents: true,
+    });
+    expect(
+      findDependencies(
+        'templates/application.hbs',
+        `
+      {{component "input"}}
+      {{component "link-to"}}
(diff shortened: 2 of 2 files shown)


ef4:
Nice work, thanks.

packages/compat/src/resolver.ts
```diff
@@ -731,6 +731,11 @@ export default class CompatResolver implements Resolver {
         from
       );
     }
+
+    if (builtInComponents.includes(component.path)) {
```
Windvis, line 735: Hmm, I just realized that early returning here will prevent people from "shadowing" built-in components by creating components with the same name in their project or addon. It doesn't seem like a good thing to do and I'm not sure if it's officially supported but it does seem to work (tested in Ember twiddle).

Is that something that needs to work under Embroider?

lifeart: as I understand, we could check it using 'tryComponent' method

PR #1031, opened 2021-11-24
Add `isTesting` and `isDevelopingApp` to readme

mydea:
These two macros were currently not documented in the readme.


ef4:
Thanks.

PR #1034, opened 2021-12-01
[DOC] Update macroCondition template helper example

achambers:
The `macroCondition` template example doesn't look correct. I think this fixes it.


ef4:
It's kinda weird but it actually matches the behavior. See: https://github.com/embroider-build/embroider/blob/1f400720511644b78d5dc4e336f3a7c7a18d312b/packages/macros/tests/glimmer/macro-condition.test.ts#L78-L90

I'm open to switching to the proposed form to make it easier to use, but somebody would need to take a stab at the implementation changes.

PR #1035, opened 2021-12-03
Optimize ember-cli-babel handling

ef4:
The ember-cli-babel addon introduces a large amount of useless work into builds. This compat adapter avoids a chunk of that work by making it clear to embroider that no runtime code is supposed to get emitted from it at all.


krisselden (approved):

NullVoxPopuli:
Excited for this!

Issue #1038, opened 2021-12-05
Ember 4.1+ || Error while processing route: ember window.Ember is undefined -- related: can't access lexical declaration '__WEBPACK_DEFAULT_EXPORT__' before initialization


NullVoxPopuli:
This is a preliminary issue, but cropped up -- not sure why yet.

There is only one occurrence of `window.Ember`:
![image](https://user-images.githubusercontent.com/199018/144767730-50917e1b-063f-488e-a1fb-3cda73f15551.png)
which comes from:
![image](https://user-images.githubusercontent.com/199018/144767758-5fdf543c-453b-46b0-91e4-ae57e1099ee6.png)
which is defined here: https://github.com/embroider-build/embroider/blob/fb866769e53b3ce543ce85ab3640be5be10239be/packages/util/addon/ember-private-api.js#L9-L19

Here is the code / error:
```
Error while processing route: index window.Ember is undefined ../node_modules/@embroider/util/ember-private-api.js@http://localhost:4200/assets/chunk.4c6a1d188e8b66e667de.js:32527:3
__webpack_require__@http://localhost:4200/assets/chunk.366fbe0dd169a8169677.js:23:42
../node_modules/@embroider/util/index.js@http://localhost:4200/assets/chunk.4c6a1d188e8b66e667de.js:32618:95
__webpack_require__@http://localhost:4200/assets/chunk.366fbe0dd169a8169677.js:23:42
```

and related error:
```
can't access lexical declaration '__WEBPACK_DEFAULT_EXPORT__' before initialization
```
![image](https://user-images.githubusercontent.com/199018/144767687-7e9a25e2-07e1-4daf-9e86-6d313234d9c7.png)

NullVoxPopuli:
if I revert back to Ember `4.0.1` the error goes away, so this seems to be an issue with `dependencySatisfies`?

NullVoxPopuli:
I confirm the fix by changing the macroCondition to just `true` in node_modules. but I am unable to get 4.1.0-beta or 4.2.0-alpha to match the dependencySatisfies

NullVoxPopuli:
I tried adding `>= 4.0.0` to the dependencySatisfies condition, but:
![image](https://user-images.githubusercontent.com/199018/144878965-9ef5b5c7-49e6-48c3-bdfa-7c8ebab338a3.png)
still false.

Windvis:
[...]

PR #1039, opened 2021-12-06
Reproduce issue https://github.com/embroider-build/embroider/issues/1038

NullVoxPopuli:
Things left to do:
 - the macroConditions tests are super thrown off by the project logic copied from dependencySatisfies tests.
 
 the tldr is that ember >= 4.1 doesn't work with @embroider/macros;


NullVoxPopuli:
hm, I'm now having this issue with ember 4.0.1 bah

Issue #1033, opened 2021-11-30
Embroider-native support for Sass


simonihmig:
I played a bit trying to make an Embroider app work with Sass natively, i.e. *without* using `ember-cli-sass`, as that seems to be the new way to integrate new tools to the build-pipelines (see https://github.com/embroider-build/embroider/issues/916#issuecomment-903175051). However that turned out to be not that trivial as I hoped. 

I tried the following three approaches, no one worked really perfectly. Wonder what the "right" way would be, and how to fix its issues?

For all approaches, I installed the related webpack plugins, and added their config to `ember-cli-build.js`. So this is `sass-loader` (and `sass` itself), `mini-css-extract-plugin` and `css-loader`. The latter two are already used by `@embroider/webpack`, but I made them also dependencies of the app itself. I chose to use `mini-css-extract-plugin` instead the usually proposed `style-loader` also in development, as I wanted the app to also work in FastBoot (which seems to not work with `style-loader`, as it uses `document` APIs at runtime). 

I have reproductions for all these approaches at https://github.com/simonihmig/embroider-sass-app in different branches.

## 1. Rename app.css to app.scss

The obvious and naive approach was to just rename `styles/app.css` to `styles/app.scss` and hope that it would just work. It didn't. 

It seems it doesn't even try to compile the sass. `app.scss` is found in `dist/assets/` in its original form. The stylesheet reference in index.html points to a now nonexisting file (`/assets/embroider-sass-app.css`)

[Reproduction](https://github.com/simonihmig/embroider-sass-app/tree/rename-app-scss)

## 2. Import app.scss from app.js

Similar to what the `sass-loader` itself suggests and what you mostly see in "webpack-native" projects AFAIK, I imported `app.scss` (moved from `/app/styles/app.scss` to `/app/app.scss`, as otherwise it would not find it somehow) from `app.js`.

This does work in general, the Sass is transformed, and you see the compiled CSS linked and used. However it is not quite ideal:
* you seem to need an empty `/app/styles/app.css`, which is a bit confusing. Otherwise the build crashes with `Could not find app css: "/assets/embroider-sass-app.css" in index.html. Found the following instead: - /assets/vendor.css`
* the `<link>` tag is added at the end of the index.html (alongside the scripts). For a "regular" SPA this is probably ok-ish, but for a SSR app (FastBoot) this causes a [FOUC](https://en.wikipedia.org/wiki/Flash_of_unstyled_content).

[...]

PR #1045, opened 2021-12-10
Append styles imported in JS to end of `document.head`

simonihmig:
This addresses the issue raised in #1033 ("2. Import app.scss from app.js") that link tags added through imports of CSS in JS get added next to the script entrypoint (at the body's end) rather than to the document's head, causing a FOUC in FastBoot rendered pages.

This was tested locally with my test app (https://github.com/simonihmig/embroider-sass-app/tree/import-app), so I am quite confident it does the right thing. Idk if we need automated tests for this, I haven't found any related test coverage, but maybe I missed something with the (new) test setup? We *could* add unit tests for the `Placeholder` class, but that seems to yield relatively little value given the rather trivial implementation. Better maybe would be some webpack-related e2e-like tests. To ensure the output in the generated index.html, especially the order of styles and scripts, one would probably need (jest) tests that parse the build output, rather than regular ember app tests?

### DIFF
--- a/packages/core/src/html-placeholder.ts
+++ b/packages/core/src/html-placeholder.ts
@@ -32,8 +32,12 @@ export default class Placeholder {
     }
   }
 
-  insert(node: Node) {
+  insert(node: Node): void {
     this.end.parentElement.insertBefore(node, this.end);
+  }
+
+  appendToHead(node: Node): void {
+    this.end.ownerDocument.head.appendChild(node);
   }
 
   isScript(): boolean {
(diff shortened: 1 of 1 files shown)


packages/core/src/html-placeholder.ts
```diff
@@ -70,12 +74,20 @@ export default class Placeholder {
...
-    this.insert(newTag);
-    this.insertNewline();
+
+    if (this.isScript()) {
+      // Add dynamic styles from scripts to the bottom of the head, and not to where the script was,
+      // to prevent FOUC when pre-rendering (FastBoot)
+      this.appendToHead(newTag);
+    } else {
+      // Keep the new style in the same place as the original one
+      this.insert(newTag);
```
simonihmig, line 84: This should ensure that styles already present in index.html will be kept where they are, to keep the existing order. Especially when dealing with external styles, that are not funneled through this. (e.g. vendor.css, then external  google-font-css, then app.css). All other dynamically added styles will be appended to `<head>`, in whatever order webpack gives us. 

This is all we can do here about ensuring the "right" order, is it?

simonihmig:
The one failing CI job is unrelated (some network error w/ yarn)

ef4:
Thanks

PR #1046, opened 2021-12-11
Disable publicAssetURL for FastBoot script render

harrywebdev:
When `publicAssetURL` is supplied and starts with `http`, generate
script tags for FastBoot without it.

### DIFF
--- a/packages/core/src/html-entrypoint.ts
+++ b/packages/core/src/html-entrypoint.ts
@@ -99,19 +99,28 @@ export class HTMLEntrypoint {
             if (!base) {
               // this bundle only exists in the fastboot variant
               let element = placeholder.start.ownerDocument.createElement('fastboot-script');
-              element.setAttribute('src', this.publicAssetURL + fastboot);
+              // HACK: actual URL in `publicAssetURL` breaks fastboot loading; only absolute path works
+              let src = this.publicAssetURL.match(/^http/) ? '/' : this.publicAssetURL;
+              element.setAttribute('src', src + fastboot);
               placeholder.insert(element);
               placeholder.insertNewline();
             } else if (!fastboot || base === fastboot) {
               // no specialized fastboot variant
               let src = this.publicAssetURL + base;
-              placeholder.insertURL(src);
+              // HACK: actually, URL in `publicAssetURL` breaks fastboot loading; only absolute path works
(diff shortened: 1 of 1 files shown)


ef4:
Thanks for taking a stab at this.

I wonder if you're hitting this because of a discrepancy between `publicAssetURL` and Ember's `rootURL`. They would normally align like:

```
rootURL === 'http://something';
publicAssetURL === 'http://something/assets';
```

And I think as long as they do we already handle that case.

harrywebdev:
Thanks for reply and pointing to this - actually, the URL is different; I was under the assumption that this serves a purpose of having assets on different URL than the app (that's how we used to use `fingerprint.prepend` in previous EmberApp config). In practice, we have `my-website.com` and we want to serve the assets from `our-cdn.com`.

Looking at the Ember docs, `rootURL` is only a relative path to the current domain, whereas `publicAssetURL` (as I thought) could include a full URL (and when that happens, I couldn't make FastBoot work). 

Do you propose any other solution? Is it not desired behaviour (what I've described)? Thanks for your help!

ef4:
Sorry, you're correct, the example I showed assumes only the case of serving assets and index.html together. Serving them separately is also supported.

I think the idea here is that we should never put `publicAssetURL` into the fastboot-specific URLs because those should always remain "local". Any time that `publicAssetURL !== `${rootURL}/assets`, we would need to keep `src` and `data-fastboot-src` separate as you do already in this PR.

harrywebdev:
Great! Is there anything else I should do with this PR so we could merge it in? There's one test failing but I cannot see any details or why is it happening.

The other thing you mentioned is this
```
if (publicAssetURL !== `${rootURL}/assets`) {
  // do as it is in the PR
}
```
I think you maybe meant `publicAssetURL !== rootURL`? There's also other files in root of `dist` that can be served via CDN (aka `publicAssetURL`).

So, shall I update the code to rather check for this instead of the current regex test for `http`?

ef4:
Yes, you're right, I was thinking that publicAssetURL pointed to the `dist/assets` subdir but it's really the path directly to dist/index.html.

So yes, let's use that test in this PR to decide when to insert `data-fastboot-src` and/or `<fastboot-script>` to keep Fastboot from trying to follow the public URLs.

kiwiupover (approved):

PR #1048, opened 2021-12-11
Fix imported CSS with FastBoot

simonihmig:
When importing CSS from a JS module in a FastBoot enabled app, Embroider would emit something like this into index.html:

```
<fastboot-script src="/assets/chunk.957ca35b25015b34de4d.js"></fastboot-script>
<fastboot-script src="/assets/chunk.96eb691090beaf949370.css"></fastboot-script>
```

This makes FastBoot try to parse and evaluate CSS as JavaScript, which obviously fails.

I tried to add a test case to `fastboot-app`, but this fails due to even more FastBoot/CSS issues (#1049)...

### DIFF
--- a/packages/core/src/html-entrypoint.ts
+++ b/packages/core/src/html-entrypoint.ts
@@ -154,10 +154,12 @@ function maybeInsertLazyBundles(
 ): boolean {
   if (!insertedLazy && placeholder.isScript()) {
     for (let bundle of lazyBundles) {
-      let element = placeholder.start.ownerDocument.createElement('fastboot-script');
-      element.setAttribute('src', publicAssetURL + bundle);
-      placeholder.insert(element);
-      placeholder.insertNewline();
+      if (bundle.endsWith('.js')) {
+        let element = placeholder.start.ownerDocument.createElement('fastboot-script');
+        element.setAttribute('src', publicAssetURL + bundle);
+        placeholder.insert(element);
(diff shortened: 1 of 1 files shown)


ef4:
I guess I can see how it's possible to introduce "fastboot-only CSS", since you can import css files into JS modules that are Fastboot-only.

And I guess this PR is implying that that CSS will never do anything, and it will be dropped?

simonihmig:
> I guess I can see how it's possible to introduce "fastboot-only CSS", since you can import css files into JS modules that are Fastboot-only.

Actually I am not importing it from a FastBoot-only module. It is a regular module (so browser + FastBoot), but one that is lazily loaded (from a splitted route). When rendering that route in FastBoot, those lazy chunks are added with `<fastboot-script>` for FastBoot to process them eagerly AFAIU, right? And that breaks for CSS.

You can see this in the reproduction I used for #1049, the [`import-app-fastboot`](https://github.com/simonihmig/embroider-sass-app/tree/import-app-fastboot) branch. I have the changes of this PR applied as a [patch](https://github.com/simonihmig/embroider-sass-app/blob/import-app-fastboot/patches/%40embroider%2Bcore%2B0.48.1.patch), but without it you should see the same error there. It is this [lazy route](https://github.com/simonihmig/embroider-sass-app/blob/import-app-fastboot/app/routes/lazy.js) that imports CSS and makes FastBoot fail.

PR #1052, opened 2021-12-13
restore tree caching via `cacheKeyForTree`

RuslanZavacky:
Similar idea to the https://github.com/babel/ember-cli-babel/pull/421, adding `cacheKeyForTree` helps to avoid calculate same trees multiple times when having addons inside addons.

### DIFF
--- a/packages/addon-shim/package.json
+++ b/packages/addon-shim/package.json
@@ -22,6 +22,7 @@
   },
   "dependencies": {
     "@embroider/shared-internals": "^0.48.1",
+    "calculate-cache-key-for-tree": "^2.0.0",
     "ember-auto-import": "^2.2.0",
     "semver": "^7.3.5"
   },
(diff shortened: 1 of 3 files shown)


NullVoxPopuli:
Related question for @ef4, is this still a responsibility of the build system in a post addonV2 world? Or does the packager/webpack handle all that?

With v2 addons, is the v1 shim a no-op? (Assuming ember auto import v2 or embroider is present?)

ef4:
> Related question for @ef4, is this still a responsibility of the build system in a post addonV2 world? Or does the packager/webpack handle all that?
> 
> With v2 addons, is the v1 shim a no-op? (Assuming ember auto import v2 or embroider is present?)

Even under embroider the shim is not yet quite a no-op. Not because anything in embroider needs it, but because `@embroider/compat` still bootstraps off ember-cli's `EmberApp` and `Addon` instances, and they need to see your addon as an Addon instance.

A v2-formatted *app* would be able to not care about the shims at all. But this isn't a thing that exists yet because it really wants to have all its addons in v2 format, and zero apps have that (given that ember itself isn't a v2 addon yet).

RuslanZavacky:
@ef4 updated key 👌

PR #1053, opened 2021-12-13
resolve failed macro condition in ember-private-api

NullVoxPopuli:
Goal to resolve: https://github.com/embroider-build/embroider/issues/1038

Only occurs when the host-app's ember-source is alpha or beta


NullVoxPopuli:
hm, nevermind, this isn't the fix

NullVoxPopuli:
@ef4 I did it! I just needed to use `*` for the version instead of specific versions

NullVoxPopuli:
I don't know how to make a test for this tho.

NullVoxPopuli:
hm, maybe the `*` breaks non-alpha/non-beta  `ember-source` :thinking: 

based on the failing tests, it seems that the `*` version is used _instead of_ the host app's version?

ef4:
There are two bad things that ember-cli does and we follow for compatibility that were causing trouble here.

One is that dummy apps and addons share a package.json, so adding the peerDep to the addon (which is correct) also adds it to the dummy app (which is unnecessary at best but here actually triggers the following bug).

Two is that we have special logic to cause apps to *not* discover addons in peerDeps, because that's what ember-cli does. So the dummy app stopped discovering the ember-source addon.

I had to adjust the logic so that a peerDep that's also a dep will get discovered.

Issue #1067, opened 2022-01-07
in a yarn-monorepo, `dependencySatisfies` is always true when referencing packages in the monorepo


NullVoxPopuli:
example dep tree:
```
app-a
├─ dep-base
├─ dep-optional-a
   └─ dep-base
app-b
├─ dep-base    
├─ dep-optional-b
   └─ dep-base
dep-base
└─ (no relavant dependencies)    
dep-optional-a
└─ dep-base
dep-optional-b
└─ dep-base  
```

in dep-base, I have:
```ts
let bag = [];

if (macroCondition(dependencySatisfies('dep-optional-a', '*'))) {
  let stuff = importSync('dep-optional-a/some-path');
  bag.push(...stuff.default);
}


if (macroCondition(dependencySatisfies('dep-optional-b', '*'))) {
  let stuff = importSync('dep-optional-b/some-path');
  bag.push(...stuff.default);
}

console.log( bag );
```

The intent is to, 
 - in `app-a`, have the macroCondition for dep-optional-a be true, and dep-optional-b be false
 - in `app-b`, have the macroCondition for dep-optional-b be true, and dep-optional-a be false
 
 However, what is happening is that all macroConditions are always true, as I get `Cannot find module "dep-optional-b" imported from (require)` (in app-a, for example)
 
 I would not expect that these conditions would be true, because, for example: app-a, dep-optional-b should not at all be in the dependency tree.

ef4:
We are following what NPM does here. But I agree that what NPM does is terrible.

I'm OK with adding strictness here. It's a pretty trivial code change. In these two places:

https://github.com/embroider-build/embroider/blob/85e67439f9b7b0f2a3aea721c17a56378921b0b8/packages/macros/src/glimmer/dependency-satisfies.ts#L31
https://github.com/embroider-build/embroider/blob/85e67439f9b7b0f2a3aea721c17a56378921b0b8/packages/macros/src/babel/dependency-satisfies.ts#L34

Before we do `packageCache.resolve`, we would check `us.dependencies` for the package name.

NullVoxPopuli:
I don't think we can close this because the core issue still exists -- the result is different though.
In the above snippet, macroCondition is passed `false` from `dependencySatisfies`.
I think you mentioned this elsewhere?, but we need to have the package looked up from the host app somehow?

ef4:
> but we need to have the package looked up from the host app somehow?

This isn't supported, on purpose.

What are you actually trying to do?

[...]

PR #1070, opened 2022-01-10
Do not try to resolve packages that are not declared in dependencies

NullVoxPopuli:
Resolves: https://github.com/embroider-build/embroider/issues/1067

### DIFF
--- a/packages/macros/src/babel/dependency-satisfies.ts
+++ b/packages/macros/src/babel/dependency-satisfies.ts
@@ -2,11 +2,8 @@ import type { NodePath } from '@babel/traverse';
 import type { types as t } from '@babel/core';
 import State, { sourceFile } from './state';
 import { satisfies } from 'semver';
-import { PackageCache } from '@embroider/shared-internals';
 import error from './error';
 import { assertArray } from './evaluate-json';
-
-const packageCache = PackageCache.shared('embroider-stage3');
 
 export default function dependencySatisfies(path: NodePath<t.CallExpression>, state: State): boolean {
   if (path.node.arguments.length !== 2) {
@@ -27,12 +24,12 @@ export default function dependencySatisfies(path: NodePath<t.CallExpression>, st
   }
   let sourceFileName = sourceFile(path, state);
   try {
-    let us = packageCache.ownerOfFile(sourceFileName);
+    let us = state.packageCache.ownerOfFile(sourceFileName);
     if (!us || us.dependencies.every(dep => dep.name !== packageName.value)) {
       return false;
     }
 
-    let version = packageCache.resolve(packageName.value, us).version;
+    let version = state.packageCache.resolve(packageName.value, us).version;
     return satisfies(version, range.value, {
       includePrerelease: true,
     });
--- a/packages/macros/src/babel/dependency-satisfies.ts
+++ b/packages/macros/src/babel/dependency-satisfies.ts
@@ -12,7 +12,7 @@ export default function dependencySatisfies(path: NodePath<t.CallExpression>, st
   if (path.node.arguments.length !== 2) {
     throw error(path, `dependencySatisfies takes exactly two arguments, you passed ${path.node.arguments.length}`);
   }
-  let [packageName, range] = path.node.arguments;
+  const [packageName, range] = path.node.arguments;
   if (packageName.type !== 'StringLiteral') {
     throw error(
       assertArray(path.get('arguments'))[0],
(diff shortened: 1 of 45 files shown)


packages/macros/src/babel/dependency-satisfies.ts
```diff
@@ -12,7 +12,7 @@ export default function dependencySatisfies(path: NodePath<t.CallExpression>, st
   if (path.node.arguments.length !== 2) {
     throw error(path, `dependencySatisfies takes exactly two arguments, you passed ${path.node.arguments.length}`);
   }
-  let [packageName, range] = path.node.arguments;
+  const [packageName, range] = path.node.arguments;
```
NullVoxPopuli, line 15: needed to be const due to the function usage in line 35

NullVoxPopuli:
can I make this draft after submission? :upside_down_face: 
I need to:
 - add a test
 - confirm this actually works in my real app

lifeart (approved):

lifeart:
@NullVoxPopuli I converted it to draft

NullVoxPopuli:
@lifeart thank you!

packages/macros/tests/babel/dependency-satisfies.test.ts
```diff
@@ -123,5 +123,25 @@ describe(`dependencySatisfies`, function () {
       );
       expect(runDefault(code)).toBe(true);
     });
+
+    test('monorepo resolutions resolve correctly', () => {
```
NullVoxPopuli, line 127: I _think_ this is pretty close to my setup

packages/macros/tests/babel/dependency-satisfies.test.ts
```diff
@@ -123,5 +123,25 @@ describe(`dependencySatisfies`, function () {
...
+        import { dependencySatisfies } from '@embroider/macros';
+
+        // specified in dependencies
+        export const util = dependencySatisfies('@embroider/util', '*');
+        // not specified as any kind of dep
+        export const webpack = dependencySatisfies('@embroider/webpack', '*');
+      `);
+
+      expect(code).toMatch(/util = true/);
+      expect(code).toMatch(/webpack = false/);
```
NullVoxPopuli, line 144: wip
```
    Expected pattern: /util = true/
    Received string:  "// specified in dependencies
    export const util = false; // not specified as any kind of dep·
    export const webpack = false;"
```

ef4:
Some of the existing babel tests were misleading you I think (`process.cwd` has nothing to do with the resolving for example). I cleaned them up.

There is a similar step still needed for the glimmer tests I think.

ef4:
The remaining test failures happen because the PackageCache at this layer doesn't distinguish which Package is the app vs any other library, and `Package.prototype.dependencies` is strict about only listing `devDependencies` if it knows the Package is the app. So at the moment dependencySatisfies will miss devDependencies of apps.

I'm refactoring PackageCache so that threading through the app root path is mandatory, so it can always give correct answers about the app's dependencies. Up until now we hadn't needed that here.

NullVoxPopuli:
Thanks for poking at this! This'll be a huge help!

ef4:
The two remaining failures are just sporadic, I confirmed they're passing locally.

(We have so many different jobs that even low-probability failures in CI are likely to accumulate and cause  a bad run.)

ef4:
This is ready to go. I hijacked your PR and made it a much better change. 😄

One thing we can consider doing next is making packageCache.resolve strict like this by default. There are definitely some cases in embroider/compat where we'd want to opt into non-strict resolve, but it would be a good default.

NullVoxPopuli:
Thank you!! <3

anehx:
Thanks for this fix, this resolves a lot of issues I had with packages using `dependencySatisfies` in a monorepo! :tada: However, since this makes `dependencySatisfies` much stricter ([needing at least a `peerDependency`](https://github.com/emberjs/ember-render-modifiers/pull/58)) we should make sure it's properly documented. Also, this should maybe even be marked as a breaking change? I'm not informed enough on why the `peerDependency` is required to add the documentation myself but I'd really appreciate if someone with that knowledge could add it (@NullVoxPopuli or @ef4 maybe? :wink: ).

Also, I'm wondering if a peer dependency should really be needed - what happens if a package wants to check whether a certain package is installed or not? A peer dependency for such a case would be wrong, no? This might be what @ef4 meant here: https://github.com/embroider-build/embroider/pull/1070#issuecomment-1011435283 ?

ef4:
> what happens if a package wants to check whether a certain package is installed or not? A peer dependency for such a case would be wrong, no? 

For that we use [optional peer dependencies](https://github.com/emberjs/rfcs/blob/master/text/0507-embroider-v2-package-format.md#optional-peer-dependencies).

PR #1071, opened 2022-01-11
Always unroll `each()` macro

### LINKED ISSUE #1063: `each()` macro doesn't unroll in development builds
TLDR: It should (and does in production), remove the mode check and look at arg info instead to ensure the unroll will work.

----------

In [discussion over on `ember-auto-import`](https://github.com/ef4/ember-auto-import/issues/477), it seems that the `each()` macro doesn't unroll for development (run-time mode) builds.

https://github.com/embroider-build/embroider/blob/57d58a7e3c0dbfece76c9ee2b5136622105d0a5a/packages/macros/src/babel/each.ts#L52-L54 

This will cause "(template) string literal" requirement errors when using the macro as [shown in the RFC example](https://github.com/emberjs/rfcs/blob/master/text/0507-embroider-v2-package-format.md#javascript-macro-each).

```js
// This example:
import { getOwnConfig, each, importSync } from "@ember/macros";
let plugins = [];
for (let plugin of each(getOwnConfig().registeredPlugins)) {
  plugins.push(importSync(plugin).default);
}

// could compile to this, given OwnConfig
// containing { registeredPlugins: ['@bigco/bar-chart', '@bigco/line-chart']}

let plugins = [];
plugins.push(importSync("@bigco/bar-chart").default);
plugins.push(importSync("@bigco/line-chart").default);
```

```cli
Build Error (broccoli-persistent-filter:Babel > [Babel: addon-name]) in addon-name/components/foobar.js

/path/to/addon-name/components/foobar.js: import() is only allowed to contain string literals or template string literals
  1 | // This example:
  2 | import { getOwnConfig, each, importSync } from "@ember/macros";
  3 | let plugins = [];
  4 | for (let plugin of each(getOwnConfig().registeredPlugins)) {
> 5 |   plugins.push(importSync(plugin).default);
    |                ^^^^^^^^^^^^^^^^^^
  6 | }
```

This happens because `ember-auto-import` imposes the [dynamic `import()` requirements](https://github.com/emberjs/rfcs/blob/73685c28378118bebb5e359b80e00b839a99f622/text/0507-embroider-v2-package-format.md#supported-subset-of-dynamic-import-syntax) for the `importSync()` macro here.

https://github.com/ef4/ember-auto-import/blob/d7031b95168c7f280d8be74d08a56e375b965104/packages/ember-auto-import/ts/analyzer-plugin.ts#L55-L57

```js
        } else if (callee.isIdentifier() && callee.referencesImport('@embroider/macros', 'importSync')) {
          state.imports.push(processImportCallExpression(path, false));
          state.handled.add(path.node);
        }
```

However, it _will_ compile successfully in production builds, when the `each()` is unrolled (as shown in the example).

https://github.com/embroider-build/embroider/blob/57d58a7e3c0dbfece76c9ee2b5136622105d0a5a/packages/macros/src/babel/each.ts#L56-L62

@ef4 says the best solution is to have the `each()` macro (and `macroCondition()`) [look at argument information instead of build mode when deciding their result](https://github.com/ef4/ember-auto-import/issues/477#issuecomment-1004405924). The argument information could be added to the `EvaluateResult`, and subsequently used by the macros when needed.

https://github.com/embroider-build/embroider/blob/6a2e220f08db8558af8366e9d186d2f5df86953f/packages/macros/src/babel/evaluate-json.ts#L103-L106

But I'm not really sure what argument information is needed and how to get that information. What I gather is that it should make sure the array argument is "confident" as well.?.

Panman82:
Currently the `each()` [example in the Embroider RFC](https://github.com/emberjs/rfcs/blob/master/text/0507-embroider-v2-package-format.md#javascript-macro-each) (shown below) doesn't always work because the loop doesn't get unrolled in dev builds, resulting in invalid calls to `importSync()` due to [(template) string literals](https://github.com/emberjs/rfcs/blob/master/text/0507-embroider-v2-package-format.md#supported-subset-of-dynamic-import-syntax) being [required](https://github.com/ef4/ember-auto-import/blob/d7031b95168c7f280d8be74d08a56e375b965104/packages/ember-auto-import/ts/analyzer-plugin.ts#L55-L57). In my digging, I don't see why the loop _can't_ always be unrolled so this PR removes the run-time vs build-time differences for this macro, allowing the RFC example to work for all builds.

```js
// This example:
import { getOwnConfig, each, importSync } from "@ember/macros";
let plugins = [];
for (let plugin of each(getOwnConfig().registeredPlugins)) {
  plugins.push(importSync(plugin).default);
}

// could compile to this, given OwnConfig
// containing { registeredPlugins: ['@bigco/bar-chart', '@bigco/line-chart']}

let plugins = [];
plugins.push(importSync("@bigco/bar-chart").default);
plugins.push(importSync("@bigco/line-chart").default);
```

Fixes #1063

### DIFF
--- a/packages/macros/README.md
+++ b/packages/macros/README.md
@@ -106,6 +106,26 @@ if (macroCondition(foo)) {
 }
 ```
 
+### each
+
+The `each` macro can be used to unroll `for ... of` loops, replacing entire loop with each iteration of the body. This can be useful when importing an array of files with `importSync` below. Doing so will ensure only the files needed are included in the build output, rather than lazy loading all possible file inclusions. This macro requires one argument, a statically analyzable array.
+
+```js
+import { each, getOwnConfig, importSync } from "@embroider/macros";
+
+for (let asset of each(getOwnConfig().importAssets)) {
+  importSync(asset);
+}
+
+// Compiles to the following, given getOwnConfig() returns;
+// { importAssets: ["pkg-name/assets/foo", "pkg-name/assets/bar"]}
+
+importSync("pkg-name/assets/foo");
+importSync("pkg-name/assets/bar");
+```
+
+Note that variable names in the loop body are simply replaced with the element value as-is. Ex: `importSync(``pkg-name/assets/${asset}``);` becomes `importSync(``pkg-name/assets/${"foo"}``);` and `importSync("pkg-name/assets/" + asset);` becomes `importSync("pkg-name/assets/" + "foo");`
+
 ### importSync
 
 The primary reason for Embroider's existence is to create statically analyzable builds. An under pinning of this
--- a/packages/macros/src/babel/each.ts
+++ b/packages/macros/src/babel/each.ts
@@ -45,21 +45,16 @@ export function insertEach(path: EachPath, state: State, context: typeof Babel) 
     throw error(args[0], `the argument to the each() macro must be statically known`);
   }
 
-  if (state.opts.mode === 'compile-time' && !Array.isArray(array.value)) {
+  if (!Array.isArray(array.value)) {
     throw error(args[0], `the argument to the each() macro must be an array`);
   }
 
-  if (state.opts.mode === 'run-time') {
-    let callee = path.get('right').get('callee');
-    state.neededRuntimeImports.set(callee.node.name, 'each');
-  } else {
-    for (let element of array.value) {
-      let literalElement = buildLiterals(element, context);
-      for (let target of nameRefs) {
-        target.replaceWith(literalElement);
-      }
-      path.insertBefore(cloneDeep(path.get('body').node, state));
+  for (let element of array.value) {
+    let literalElement = buildLiterals(element, context);
+    for (let target of nameRefs) {
+      target.replaceWith(literalElement);
     }
-    path.remove();
+    path.insertBefore(cloneDeep(path.get('body').node, state));
   }
+  path.remove();
 }
--- a/packages/macros/tests/babel/each.test.ts
+++ b/packages/macros/tests/babel/each.test.ts
@@ -13,12 +13,16 @@ describe('each', function () {
     createTests: allModes(function (transform, { applyMode }) {
       beforeEach(function () {
         macrosConfig = MacrosConfig.for({});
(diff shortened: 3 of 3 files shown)


Panman82:
FWIW, I don't see why the same couldn't be done for `macroCondition()`, which would also help remove invalid code in dev builds for the negative condition (mainly importing unavailable modules as [shown in the macros README](https://github.com/embroider-build/embroider/tree/master/packages/macros#dependencysatisfies)).

```js
import { macroCondition, dependencySatisfies, importSync } from '@embroider/macros';

if (macroCondition(dependencySatisfies('ember-qunit', '*'))) {
  return importSync('ember-qunit');
} else if (macroCondition(dependencySatisfies('ember-mocha', '*'))) {
  return importSync('ember-mocha');
}
```

ef4:
Thanks for working on this. We will definitely want to land at least the docs improvements here.

There is a larger change needed though to control which things happen at compile-time vs run-time. Instead of being a global mode, it needs to depend on the specific inputs to each macro.

For example, if the input to `each` is a macro that can change at runtime in development, it can't be unrolled. But if the input is a macro that can't change at runtime in development, then it can.

This also applies to `macroCondition` -- whether it does branch elimination should depend on the status of its input, not just a global mode. `macroCondition(dependencySatisfies(...))` can always do branch elimination, because you can't actually change your dependencies at runtime. Whereas `macroCondition(isTesting())` needs to keep both branches so that the result is different inside the test suite than outside.

Panman82:
> For example, if the input to each is a macro that can change at runtime in development, it can't be unrolled. But if the input is a macro that can't change at runtime in development, then it can.

Is that not what the evaluator `confident` check does already??

https://github.com/embroider-build/embroider/blob/278d466cfaaeec2e762c5c4a5b15d27580d229d9/packages/macros/src/babel/each.ts#L42-L46

ef4:
This would be an expansion of the concept of confidence from a boolean to a three-state system: "not confident", "confident but allowed to change at runtime", "confident and not allowed to change at runtime".

"not confident" is already an error as the input to these macros, and always should be.

"confident but allowed to change at runtime" would be a property of other macros like `isTesting()` and `getGlobalConfig()`

PR #1074, opened 2022-01-12
Fixes require babel filter config

krisselden:
Require the resolution of @embroider/core in the emitted config file.


ef4:
Thanks, looks good.

PR #1076, opened 2022-01-13
add non-es6-compat to importSync

ef4:
importSync should do the same thing that regular static import does when you use it against as non-es-module.


ef4:
Last failures here are some tests that need to be updated to expect this newer format.

But I'm thinking I'll take a stab at sharing this new compat code, rather than inlining it everywhere. It's tiny, but we use it a lot.

Issue #1066, opened 2022-01-07
@embroider/macros' dependencySatisfies does not understand canary versions


NullVoxPopuli:
related to: https://github.com/embroider-build/embroider/issues/1057

version is: `DEBUG: Ember : 4.2.0-alpha.1.canary+ef2ad15f`

my hunch is that the existing comparison code isn't expecting this many specifiers after alpha.1

ef4:
Please be more specific. 

We use `semver.satisfies` with the `includePrerelease: true` option, which definitely does understand versions like that:

```js
require('semver').satisfies('4.2.0-alpha.1.canary+ef2ad15f', '>=4.0', { includePrerelease: true })
// true
```

NullVoxPopuli:
I have the same issue with the latest beta. I'll spelunk soon

NullVoxPopuli:
nevermind this is resolved with 0.50.0

NullVoxPopuli:
j/k, I still had my patch script running

NullVoxPopuli:
I quick threw a (few) console.log in the node_modules/..../dependency-satisfies.js
(and this time I remember to have `rm -r /tmp/embroider/ && yarn start` as my run command) 🙃 
```js
    try {
        let us = packageCache.ownerOfFile(sourceFileName);
        console.log({ us: Boolean(us), name: packageName.value });
        if (!us) {
            return false;
        }
        let version = packageCache.resolve(packageName.value, us).version;
        console.log({ name: packageName.value, version: Boolean(version) });
        return (0, semver_1.satisfies)(version, range.value, {
            includePrerelease: true,
        });
    }
    catch (err) {
        console.log('catch', { name: packageName.value, range: range.value, error: err.code, sourceFileName });
        if (err.code !== 'MODULE_NOT_FOUND') {
            throw err;
        }
        return false;
    }

```

It seems the host `ember-source` version is ignored in all these catch scenarios? I'm still not clear on how dependencySatisfies is supposed to work

```
catch {
  name: 'ember-source',
  range: '>= 3.22.0-beta.1',
  error: 'MODULE_NOT_FOUND',
  sourceFileName: '/tmp/embroider/19ab55/node_modules/@ember/render-modifiers/modifiers/did-insert.js'
}
{ us: true, name: 'ember-source' }
catch {
  name: 'ember-source',
  range: '>= 3.22.0-beta.1',
  error: 'MODULE_NOT_FOUND',
[...]

PR #1080, opened 2022-01-16
Support a project having prerelease versions

NullVoxPopuli:
This is what was going on with https://github.com/embroider-build/embroider/issues/1066

Demonstration of `semver` behavior here: https://runkit.com/nullvoxpopuli/semver
Reproduction of behavior: https://github.com/NullVoxPopuli/limber/pull/375
(This PR is the last thing between me and _finally_ merging that https://github.com/NullVoxPopuli/limber/pull/375 🙃 )

tldr: `satisfies` does not return true if the version is a pre-release. the pre-release option to `satisfies` is only for the `range` parameter
 
 I've also been wondering if there is a better way to share code between the glimmer and babel versions of `dependencySatisfies` -- though, I don't know why the glimmer version exists, because I thought it was all build-time transforms anyway?

### DIFF
--- a/packages/macros/tests/babel/dependency-satisfies.test.ts
+++ b/packages/macros/tests/babel/dependency-satisfies.test.ts
@@ -147,19 +147,6 @@ describe(`dependencySatisfies`, function () {
         expect(runDefault(code)).toBe(true);
       });
 
-      test(`it considers a project's star-releases as allowed`, () => {
-        project.addDependency('foo', '*');
-        let code = transform(
-          `
-          import { dependencySatisfies } from '@embroider/macros';
-          export default function() {
-            return dependencySatisfies('foo', '>= 3.27.0-alpha.1 || >= 3.27.0-beta.1');
-          }
-        `
-        );
-        expect(runDefault(code)).toBe(true);
-      });
-
       test('monorepo resolutions resolve correctly', () => {
         project.addDependency('@embroider/util', '1.2.3');
         let code = transform(`
--- a/packages/macros/tests/babel/dependency-satisfies.test.ts
+++ b/packages/macros/tests/babel/dependency-satisfies.test.ts
@@ -147,6 +147,19 @@ describe(`dependencySatisfies`, function () {
         expect(runDefault(code)).toBe(true);
       });
 
+      test(`in monorepos, the * dep is commonly used to mean "latest"`, () => {
+        project.addDependency('foo', '*');
+        let code = transform(
+          `
+          import { dependencySatisfies } from '@embroider/macros';
+          export default function() {
+            return dependencySatisfies('foo', '>= 3.27.0-alpha.1 || >= 3.27.0-beta.1');
+          }
+        `
+        );
+        expect(runDefault(code)).toBe(true);
+      });
+
       test('monorepo resolutions resolve correctly', () => {
         project.addDependency('@embroider/util', '1.2.3');
         let code = transform(`
--- a/packages/macros/tests/babel/dependency-satisfies.test.ts
+++ b/packages/macros/tests/babel/dependency-satisfies.test.ts
@@ -134,13 +134,39 @@ describe(`dependencySatisfies`, function () {
         expect(runDefault(code)).toBe(true);
       });
 
+      test(`it considers a project's used pre-releases as allowed`, () => {
+        project.addDependency('foo', '4.2.0-beta.1');
+        let code = transform(
+          `
+          import { dependencySatisfies } from '@embroider/macros';
+          export default function() {
+            return dependencySatisfies('foo', '>= 3.27.0-alpha.1 || >= 3.27.0-beta.1');
+          }
+        `
+        );
+        expect(runDefault(code)).toBe(true);
+      });
+
+      test(`it considers a project's star-releases as allowed`, () => {
+        project.addDependency('foo', '*');
+        let code = transform(
+          `
+          import { dependencySatisfies } from '@embroider/macros';
+          export default function() {
+            return dependencySatisfies('foo', '>= 3.27.0-alpha.1 || >= 3.27.0-beta.1');
+          }
+        `
+        );
+        expect(runDefault(code)).toBe(true);
+      });
+
       test('monorepo resolutions resolve correctly', () => {
         project.addDependency('@embroider/util', '1.2.3');
         let code = transform(`
         import { dependencySatisfies } from '@embroider/macros';
 
         export default function() {
-          return { 
+          return {
             // specified in dependencies
             util: dependencySatisfies('@embroider/util', '*'),
 
--- a/packages/macros/tests/babel/dependency-satisfies.test.ts
+++ b/packages/macros/tests/babel/dependency-satisfies.test.ts
@@ -148,7 +148,13 @@ describe(`dependencySatisfies`, function () {
       });
 
       test(`in monorepos, the * dep is commonly used to mean "latest"`, () => {
-        project.addDependency('foo', '*');
+        project.addDependency('foo', '4.2.0-beta.1', project => {
+          project.pkg.dependencies = {
+            ...project.pkg.dependencies,
+            foo: '*',
+          };
+        });
+
         let code = transform(
           `
           import { dependencySatisfies } from '@embroider/macros';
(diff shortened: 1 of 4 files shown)


NullVoxPopuli:
@rwjblue I removed the test with `*` as the first-arg to `satisfies`, and I found an existing test that _does_ check for `*` on the _range_ arg (the second arg) -- so `*` should be covered

lifeart (approved):
@NullVoxPopuli could we have test case for “ pkg.version === undefined” ?

NullVoxPopuli:
@lifeart, iirc, I had a really hard time reproducing that in tests -- only came upon that scenario when linking to a real app

NullVoxPopuli:
@ef4 / @rwjblue I re-added the `*` test, because It's a valid scenario (regardless of the validity of the version) -- common in yarn workspaces to signify "this is the latest / linked version / the code that's in the monorpeo"

packages/macros/tests/babel/dependency-satisfies.test.ts
```diff
@@ -134,13 +134,39 @@ describe(`dependencySatisfies`, function () {
...
+          export default function() {
+            return dependencySatisfies('foo', '>= 3.27.0-alpha.1 || >= 3.27.0-beta.1');
+          }
+        `
+        );
+        expect(runDefault(code)).toBe(true);
+      });
+
+      test(`in monorepos, the * dep is commonly used to mean "latest"`, () => {
+        project.addDependency('foo', '*');
```
rwjblue, line 151: This isn't quite right, the second arg should be an actual version number

packages/macros/tests/babel/dependency-satisfies.test.ts
```diff
@@ -134,13 +134,39 @@ describe(`dependencySatisfies`, function () {
...
+        let code = transform(
+          `
+          import { dependencySatisfies } from '@embroider/macros';
+          export default function() {
+            return dependencySatisfies('foo', '>= 3.27.0-alpha.1 || >= 3.27.0-beta.1');
```
rwjblue, line 156: the star shoudl go here:

```suggestion
            return dependencySatisfies('foo', '*');
```

packages/macros/tests/babel/dependency-satisfies.test.ts
rwjblue: The `*` you are referring to (in the commit comment) is in a version range field (e.g. inside `dependencies` and `devDependencies`) **not** the top level `version` field of a `package.json` (which is what this second arg is all about).

ef4:
Seconding what @rwjblue said. 

`project.addDependency()` may be  misleading you because it uses the second argument for **two different things**. One is a version and one is a range. 

On new enough versions of fixturify-project (I don't recall if that's what we're using here), there's also an explicit `requestedRange` option to control the range separately from the version:

https://github.com/stefanpenner/node-fixturify-project/blob/e75e373aae4086406a088c8032cb203b46c6a28d/index.ts#L113

packages/macros/tests/babel/dependency-satisfies.test.ts
NullVoxPopuli: oh :thinking: the nomenclature here, `project.addDependency` is telling me that, _in the consuming project_, I'd have:

```
"dependencies": {
  "foo": "*"
}
```

so, for extra clarification, if that's _not_ the case, I'll go ahead and make your suggested change, no questions asked.

What is a "project"? 

Fwiw, _this is the piece of code I'm trying to "get around": https://github.com/embroider-build/embroider/blob/main/packages/util/addon/ember-private-api.js#L11
Where, I feel like if the consuming project has either `*` or `<whatever current canary is>`, this block should be true. (in the released version of embroider, that's not the case).

Is there some other way to test this scenario?

packages/macros/tests/babel/dependency-satisfies.test.ts
NullVoxPopuli: Not for what I'm wanting to test tho. 
Example: https://github.com/embroider-build/embroider/blob/main/packages/util/addon/ember-private-api.js#L11
(and then my consuming project has `*`)
(unless I totally misunderstand this test's APIs (project.addDependency, etc) -- possible!)

lifeart:
Looks like https://github.com/embroider-build/embroider/issues/1091 may be related

packages/macros/tests/babel/dependency-satisfies.test.ts
NullVoxPopuli: I think I added the dependency correctly?

NullVoxPopuli:
Gonna reslove the conversations, so we can start up new threads on the up-to-date code

packages/macros/tests/babel/dependency-satisfies.test.ts
rwjblue: Thanks for updating the tests!!

packages/macros/tests/babel/dependency-satisfies.test.ts
```diff
@@ -134,13 +134,45 @@ describe(`dependencySatisfies`, function () {
         expect(runDefault(code)).toBe(true);
       });
 
+      test(`it considers a project's used pre-releases as allowed`, () => {
```
NullVoxPopuli, line 137: these tests pass without my changes to the `src` directory :scream:

tests/scenarios/macro-test.ts
```diff
@@ -181,3 +181,46 @@ dummyAppScenarios
       });
     });
   });
+
+appReleaseScenario
```
NullVoxPopuli, line 185: hoping to re-create what I'm seeing in my actual app with this scenario test

NullVoxPopuli:
Trying to run tests locally using the javascript debug terminal. Getting:
```
❯ yarn test --filter "embroider/util"
Debugger attached.
yarn run v1.22.5
$ qunit --require ts-node/register *-test.ts --filter embroider/util
Debugger attached.
TAP version 13
```
and then in the debug console:
```
ARNING in ../../home/psego/Development/NullVoxPopuli/embroider-3/node_modules/ember-data-latest/node_modules/ember-inflector/index.js 3:0-48
export 'defaultRules' (reexported as 'defaultRules') was not found in './lib/system' (possible exports: Inflector, pluralize, singularize)
 @ ./assets/app-template.js 524:13-147
 ```
 
 does that seem familiar to anyone?

**Comment by ef4** [comment]
1. Clear your cache in `$TMPDIR/embroider` to rule that out.
2. Go to this spot in the actual node_modules directories and see what version of ember-source, if any, actually resolves from there. `node -e "console.log(require('ember-source/package.json').version)"`

**Comment by NullVoxPopuli** [comment]
1. no change (whew)
2.

```
❯ node -e "console.log(require('ember-source/package.json').version)"
Volta error: Could not determine path to project workspace: '../../package.json'

Please ensure that the file exists and is accessible.
Error details written to /home/nullvoxpopuli/.volta/log/volta-error-2022-02-08_11_44_14.688.log
# I deleted the volta entry in the package.json

limber/node_modules/@embroider/util on  upgrade-stuff-2 [$] 
❯ node -e "console.log(require('ember-source/package.json').version)"
4.1.0-alpha.7
```

This is the wrong version for my app _but is relevant to this PR (#1080)_, so that's good.

I changed the package in my monorepo using 4.1.0-alpha.7 to also use 4.2.0, and ran the node command again:
```
❯ node -e "console.log(require('ember-source/package.json').version)"
4.2.0
```

_progress_. However - this implies that every dependency in a monorepo should be the same version? is there a way around that?
:thinking: I wonder if this sort of think would be fixed by yarn 3 or pnpm.
yarn 1 is very buggy.

anywho, my app boots now. thanks!
So, this means I can continue to try to figure the ember-source pre-release scenario for this PR! :tada:

ef4:
> yarn 1 is very buggy.

It is, but in this case it's buggy because it's respecting Node's node_modules convention, which is the real root of the problem. You'd have the same problem with anything that doesn't replace node_modules. pnpm and yarn 3 are both ways to do that. But I doubt either works with embroider out of the box because `@embroider/compat`'s whole job is rewriting node_modules and it's not going to understand alternative layouts.

NullVoxPopuli:
I haven't see any problems with prerelease support as I've been messing around with vite + ember-source@alpha.

PR #1085, opened 2022-01-23
Adjust lts scenarios

ef4:
Adding an Ember 3.28 LTS to the test matrix. We used to test is as "release" but now that has moved on.

And in order to keep the suite the same size, I'm dropping the scenario for 3.20. We still test 3.16, our oldest supported LTS, which should give us plenty of coverage. Things that work in 3.16 and 3.24 are very unlikey to be broken in 3.20.


tests/scenarios/scenarios.ts
```diff
@@ -42,6 +33,12 @@ async function lts_3_24(project: Project) {
   project.linkDevDependency('@ember/string', { baseDir: __dirname });
 }
 
+async function lts_3_28(project: Project) {
+  project.linkDevDependency('ember-source', { baseDir: __dirname, resolveName: 'ember-source-3.28' });
+  project.linkDevDependency('ember-cli', { baseDir: __dirname, resolveName: 'ember-cli-3.28' });
+  project.linkDevDependency('ember-data', { baseDir: __dirname, resolveName: 'ember-data-3.38' });
```
Windvis, line 39: It seems like this typo is the cause of the CI faillures.
```suggestion
  project.linkDevDependency('ember-data', { baseDir: __dirname, resolveName: 'ember-data-3.28' });
```

ef4: Thanks, good catch, I had fired this off and didn't come back around to look at why it was failing yet.

tests/scenarios/package.json
```diff
@@ -30,25 +30,25 @@
     "broccoli-persistent-filter": "^3.1.2",
     "ember-bootstrap": "^4.6.3",
     "ember-cli-3.16": "npm:ember-cli@~3.16.0",
```
rwjblue, line 32: I'm curious, why keep 3.16 and drop 3.20?

ef4: Purely to keep the test suite smaller. We still support both, we're just testing the older one of the two and assuming it's unlikely that the newer one is broken if LTS releases on either side of it are working.

ef4:
Merged https://github.com/embroider-build/embroider/pull/1115 into here because I suspect it will unblock the failing tests here.

Issue #1086, opened 2022-01-23
built-in cached decorator causes "decorator is not a function" error.


NullVoxPopuli:
<details><summary>my component (in browser)</summary>

```js
let Game = (0,_externals_ember_component__WEBPACK_IMPORTED_MODULE_2__.setComponentTemplate)((0,_externals_ember_template_factory__WEBPACK_IMPORTED_MODULE_1__.createTemplateFactory)(
/*
  
    {{this.todaysWord}}



  
*/
{
  "id": "mu2ajeIA",
  "block": "[[[1,\"\\n    \"],[1,[30,0,[\"todaysWord\"]]],[1,\"\\n\\n\\n\\n  \"]],[],false,[]]",
  "moduleName": "wordle/components/game.ts",
  "isStrictMode": true
}), (_class = class Game extends _glimmer_component__WEBPACK_IMPORTED_MODULE_3__["default"] {
  // wordList = trackedFunction(this, async () => {
  //   return await import(/* webpackIgnore */ '/words.js');
  // });
  get todaysWord() {
    let {
      answers
    } = this.wordList.value;
    let index = toDate(this.args.day).getTime() % answers.length;
    return answers[index];
  }

  get state() {
    return initialStateFor(this.args.day);
  }

}, ((0,_home_lprestonsegoiii_Development_NullVoxPopuli_wordle_node_modules_babel_runtime_helpers_esm_applyDecoratedDescriptor_js__WEBPACK_IMPORTED_MODULE_0__["default"])(_class.prototype, "todaysWord", [_glimmer_tracking__WEBPACK_IMPORTED_MODULE_4__.cached], Object.getOwnPropertyDescriptor(_class.prototype, "todaysWord"), _class.prototype), (0,_home_lprestonsegoiii_Development_NullVoxPopuli_wordle_node_modules_babel_runtime_helpers_esm_applyDecoratedDescriptor_js__WEBPACK_IMPORTED_MODULE_0__["default"])(_class.prototype, "state", [_glimmer_tracking__WEBPACK_IMPORTED_MODULE_4__.cached], Object.getOwnPropertyDescriptor(_class.prototype, "state"), _class.prototype)), _class));
```

</details>

<details><summary>source</summary>

```tsx

import Component from '@glimmer/component';
import { service } from '@ember/service';
import { cached } from '@glimmer/tracking';
import { tracked } from 'ember-deep-tracked';
// import { trackedFunction } from 'ember-resources';

interface Args {
[...]

Issue #1020, opened 2021-11-14
Cannot import from `@glimmer/tracking/primitives/cache`


NullVoxPopuli:
Using 
`@embroider/test-setup` 0.47.2
`ember-source` ~3.28.0
`ember-cli-babel` ^7.26.6

And this code:
```js
import { getValue } from '@glimmer/tracking/primitives/cache';
```

I'm getting this error:
```
Module not found: Error: Can't resolve '@glimmer/tracking/primitives/cache' in '$TMPDIR/embroider/2340f5/node_modules/ember-resources/dist/index.js'
Did you mean 'cache.js'?
BREAKING CHANGE: The request '@glimmer/tracking/primitives/cache' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
```

build: https://github.com/NullVoxPopuli/ember-resources/runs/4205184397?check_suite_focus=true#step:5:145

NullVoxPopuli:
also occurs on ember-source ~4.0.0

simonihmig:
Worth noting that this (failing) import is in a **v2 addon** (`ember-resources`)! 👆

I guess this might be related to `@glimmer/tracking` being a "virtual package"? 

I looked and debugged a bit into https://github.com/embroider-build/embroider/blob/master/packages/core/src/babel-plugin-adjust-imports.ts, especially [this part](https://github.com/embroider-build/embroider/blob/master/packages/core/src/babel-plugin-adjust-imports.ts#L288-L307). Don't know if that's relevant, unfortunately I was not able to draw any conclusions...

Note that there are even more errors for different packages (see linked CI run above):
```
Module not found: Error: Can't resolve '../../../../externals/@ember/application' in '$TMPDIR/embroider/2340f5/node_modules/ember-resources/dist'
Did you mean 'application.js'?
```

NullVoxPopuli:
there is also this: https://github.com/embroider-build/embroider/blob/master/packages/compat/src/compat-adapters/%40glimmer/tracking.ts#L23

ef4:
No, v2 addons don't run through the compat adapters. They exist to help convert v1 to v2.

[...]

PR #1088, opened 2022-01-24
Support "cache" from @glimmer/tracking

NullVoxPopuli:
https://github.com/embroider-build/embroider/issues/1086#issuecomment-1019627550

### DIFF
--- a/packages/compat/src/compat-adapters/@glimmer/tracking.ts
+++ b/packages/compat/src/compat-adapters/@glimmer/tracking.ts
@@ -20,7 +20,12 @@ class RedirectToEmber extends Plugin {
       outputFileSync(join(this.outputPath, 'index.js'), `export { tracked } from '@ember/-internals/metal';`);
       outputFileSync(
         join(this.outputPath, 'primitives', 'cache.js'),
-        `export { cached, createCache, getValue, isConst } from "@ember/-internals/metal";`
+        // Prior to ember-source 4.1, cached didn't exist
+        // using this way of importing from metal, cached will be undefined if pre 4.1
+        `import * as metal from "@ember/-internals/metal";
+const { cached, createCache, getValue, isConst } = metal;
+export { cached, createCache, getValue, isConst };
+`
       );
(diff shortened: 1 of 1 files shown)


packages/compat/src/compat-adapters/@glimmer/tracking.ts
```diff
@@ -20,7 +20,7 @@ class RedirectToEmber extends Plugin {
       outputFileSync(join(this.outputPath, 'index.js'), `export { tracked } from '@ember/-internals/metal';`);
       outputFileSync(
         join(this.outputPath, 'primitives', 'cache.js'),
-        `export { createCache, getValue, isConst } from "@ember/-internals/metal";`
+        `export { cached, createCache, getValue, isConst } from "@ember/-internals/metal";`
```
NullVoxPopuli, line 23: @ef4 does this need to be in a macroCondition somehow? since pre ember-source-4.1, this cached export doesn't exist?

nightire:
Does this PR helps https://github.com/embroider-build/embroider/issues/1020 as well?

packages/compat/src/compat-adapters/@glimmer/tracking.ts
ef4: I think you can write this in a way that isn't a SyntaxError even when `cached` is not exported by ember metal:

```js
import * as metal from "@ember/-internals/metal";
const { cached, createCache, getValue, isConst } = metal;
export { cached, createCache, getValue, isConst };
```

It means our module always exports the name `cached`, but on older ember its value will be undefined.

NullVoxPopuli: will that work pre 3.27? I know _some_ modules weren't allowing `import *`
I mean, I guess I'll find out!

NullVoxPopuli:
@nightire I don't think so :( 
On my list of things to look at over this week tho

ef4:
Thanks, looking good.

Issue #1091, opened 2022-01-30
Cannot find peerDependency (miragejs) with importSync 


jrjohnson:
Getting the error `Could not find module miragejs imported from (require)` with miragejs in the `devDependencies` as well as `peerDependencies`. I created a reproduction and you can see the error at https://github.com/jrjohnson/test-miragejs-import/runs/4962685303?check_suite_focus=true#step:6:19 this is a new addon with miragejs installed and then used like:

```
if (macroCondition(dependencySatisfies('miragejs', '*'))) {
  const { Response } = importSync('miragejs');
  console.log(Response);
}
```

Not sure if I'm using this wrong or there is an issue with this specific import.

Windvis:
You need to install `ember-auto-import` as a dependency when you import npm packages in addon code. After doing that (and returning `Response` from the util,) the test doesn't seem to fail anymore.

jrjohnson:
Thank you so much @Windvis! Really really appreciate the help here. It seems maybe intuitive, but I would never have tried that assuming that `importSync` would bring it's own stuff and not rely on EAI. You've gotten me through a very frustrating situation and I can't thank you enough. I'm going to touch up the docs if I have a moment, but go ahead and close this issue.

PR #1096, opened 2022-02-01
Add section on package requirements to importSync documentation

jrjohnson:
Hopefully this will save someone else some time, especially the bit
about ember-auto-import being required.

Refs #1091

### DIFF
--- a/packages/macros/README.md
+++ b/packages/macros/README.md
@@ -121,6 +121,10 @@ let foo = importSync('foo');
 let foo = require('foo');
 ```
 
+#### hint
+
+When using `importSync` on non ember-addon packages both the package being imported from *and* `ember-auto-import` *must* be in the `dependencies` of your addons `package.json`.
+
 ### dependencySatisfies
 
 Tests whether a given dependency is present and satisfies the given semver range. Both arguments must be strings and the second argument will be passed into [semver's satisfies](https://github.com/npm/node-semver#usage) method.


lifeart (approved):

lifeart:
I think it's also could be build-time runtime assert, if importSync is found, we could check for deps

ef4:
This isn't quite correct.

If the package-being-imported is an ember addon, then you don't need ember-auto-import.

And if the package-doing-the-import is a v2-formatted addon, then you also don't need ember-auto-import, regardless of whether the package-being-imported is an ember addon.

The real rule here is that the argument to `importSync` should mean the exact same thing as the argument to plain old `import`. You can't import from NPM without ember-auto-import, so you can't importSync from NPM without ember-auto-import either. It's not that there's a special rule for importSync, it's that it follows the prevailing rules.

ef4:
So my suggestion for this PR is to word it more as a hint. Something like: "Remember that Ember apps can only import from non-ember-addon NPM packages if they have ember-auto-import, and that rule still applies to importSync".

jrjohnson:
Thanks for the feedback @ef4. I've updated the language a bit. I didn't use your exact verbiage because I'm not sure this is a "reminder" for most ember devs, it certainly wasn't for me as I don't really know how embroider and ember-auto-import are expected to interact.

ef4:
Looks good, thanks.

ef4:
And yes, I appreciate you making the wording more broadly approachable. It's helpful when people contribute to the docs who haven't been stewing so long in the code that they forget what parts are tricky. 😄

jrjohnson:
Thanks for stewing long enough to make something awesome!

PR #1098, opened 2022-02-01
Only enable the stats being used.

krisselden:
There are performance costs for all of stats.toJSON when doing small incremental build.

https://webpack.js.org/guides/build-performance/#statstojson-speed


ef4:
Thanks, this looks good.

Issue #1095, opened 2022-01-31
Provide a sample rollup config for typescript


NullVoxPopuli:
# Updated 

I think I got everything working _output_ wise.
I still have the issue where the consuming app is looking for dist/addon, but I can debug that.

Here is my current config which appears to generate things in a way `@embroider/addon-dev` is expecting:
 - automatic ember-addon.app-js updates
 - app-re-exports in `dist/_app_`
 
 **Notes**
  - babel config at `./config/babel.config.js`
  - rollup config at `./config/rollup.config.js`
  - build command in package.json: `    "build:js": "rollup -c ./config/rollup.config.js",`
  - builds type declarations _as well as the JS_ -- and these combined are faster than running `tsc` by itself (somehow)
 
 ```js
 // @ts-nocheck
import path from 'path';

import alias from '@rollup/plugin-alias';
import multiInput from 'rollup-plugin-multi-input';
import ts from 'rollup-plugin-ts';
import { defineConfig } from 'rollup';
import { nodeResolve } from '@rollup/plugin-node-resolve';

import { Addon } from '@embroider/addon-dev/rollup';

import babelConfig from './babel.config';

const addon = new Addon();

const extensions = ['.js', '.ts'];

const transpilation = [
  // Instruct rollup how to resolve ts and hbs imports
  // (importing a template-only component, for example)
  nodeResolve({ resolveOnly: ['./'], extensions: [...extensions, '.hbs'] }),

  // Allow top-level imports (what folks are used to from v1 addons)
  // During the build, anything referencing a top-level import will be
  // replaced with a relative import.
  // DANGER: it's somewhat easy to cause circular references with this tool
  alias({
    entries: [
      {
        find: '#types',
        replacement: path.resolve('src', '-private', 'types.ts'),
      },
      {
        find: '<my-addon>',
        replacement: path.resolve('src'),
      },
      {
        find: '<my-addon>/(.*)',
        replacement: path.resolve('src/$1'),
      },
    ],
  }),

  // This babel config should *not* apply presets or compile away ES modules.
  // It exists only to provide development niceties for you, like automatic
  // template colocation.
  // See `babel.config.json` for the actual Babel configuration!
  // babel({ babelHelpers: 'bundled', extensions }),
  ts({
    // can be changed to swc or other transpilers later
    // but we need the ember plugins converted first
    // (template compilation and co-location)
    transpiler: 'babel',
    browserslist: false,
    babelConfig,
    // setting this true greatly improves performance, but
[...]

Issue #1094, opened 2022-01-31
[addon-dev] Allow TS files in publicEntrypoints / appReexports


josemarluedke:
For addons authoring their code in TS (or soon gjs/gts), currently is not possible to use `@embroider/addon-dev`'s `publicEntrypoints` or `appReexports` methods. 

We cannot use them because the dist files will end up with the same extension as the input.

In glimmer-apollo, I ended up doing a simple replace of `.ts` with `.js`. 

https://github.com/josemarluedke/glimmer-apollo/blob/main/packages/glimmer-apollo/rollup.config.js#L7-L23

NullVoxPopuli:
It turns out this isn't needed - stay tuned!

NullVoxPopuli:
I think the solution in #1099 is solid~ish, but I need to figure out a peerDep issue -- https://github.com/embroider-build/embroider/issues/1062
I _think_ I'm getting two copies of `xstate` in the consuming app, and it's only revealing itself because of component-manager's reliance on object identity. :thinking:

rwjblue:
@josemarluedke - Can you put up a separate (smaller) PR from #1099 that ensures that the output path is `.js` similar to what you did [here](https://github.com/josemarluedke/glimmer-apollo/blob/6671a2191f2032cfe6e0cad018d9f66fbb10af75/packages/glimmer-apollo/rollup.config.js#L8-L23) (though with more robust extension replacement support probably)?

josemarluedke:
@rwjblue Can definitely do it. Should extension replacement support the following mapping? 

- `ts` -> `js`
- `gts` -> `js`
- `gjs` -> `js`
- `hbs` -> `js`

rwjblue:
yep! but I think we already have `hbs` -> `js` done right?

josemarluedke:
@rwjblue PR here: https://github.com/embroider-build/embroider/pull/1106

Issue #1062, opened 2021-12-22
What is the resolution for these types of errors: <addon> is trying to import <peerDep> from within its app tree. This is unsafe, because <addon> can't control which dependencies are resolvable from the app


NullVoxPopuli:
Note: this is within a v2 addon / normal npm package.

_I've reported / asked about this error before in Discord, and I don't remember what came of it, so I'm going to try some things and log my progress here in this issue_.

Generically (for me to copy-paste to the title):
```js
<addon> is trying to import <peerDep> from within its app tree. 
This is unsafe, because <addon> can't control which dependencies are resolvable from the app
```

Specifically, for me:
```
Thread Loader (Worker 0)
$TMPDIR/embroider/1e0772/instance-initializers/ember-statechart-component.js/ember-statechart-component.js: 
ember-statechart-component is trying to import xstate from within its app tree. 
This is unsafe, because ember-statechart-component can't control which dependencies are resolvable from the app
```

I know initializers are nefarious, and here is how I'm abusing them:
```ts
import { setComponentManager, setComponentTemplate } from '@ember/component';
import { hbs } from 'ember-cli-htmlbars';
import { StateNode } from 'xstate';
import ComponentManager from './-private/statechart-manager';

// Managers are managed globally, and not per app instance
setComponentManager((owner) => ComponentManager.create(owner), StateNode.prototype);
setComponentTemplate(hbs`{{yield this.state this.send}}`, StateNode.prototype);

export function initialize(): void {}

export default {
  initialize,
};
```

I know that the initializer is part of the app tree, and it imports `xstate`, as does `./-private/statechart-manager`, so there is no getting around that.

I'm going to try to eliminate the addon's app tree entirely by moving the contents of my initializer to a `setup` method in the addon tree.

NullVoxPopuli:
after trying that, it seems that I get multiple copies of xstate in my tree... which, when working with the componentManager api, ends up making everything broken with errors like this:
```
            Error: Attempted to load a component, but there wasn't a component manager associated with the definition. The definition was: StateNode
```

ef4:
Move the implementation into your addon instead, and in the app tree have only a reexport of the code from the addon.

This is the same thing the classic blueprint does if you ask it to generate an initializer.

NullVoxPopuli:
I gave that ago here: https://github.com/NullVoxPopuli/ember-statechart-component/pull/163
But now I _always_ have duplicate files (even in classic builds):
[...]

PR #1099, opened 2022-02-01
Add typecsript sample rollup config + exploration of different plugins' side-effects

NullVoxPopuli:
With `rollup-plugin-ts`
 - https://github.com/NullVoxPopuli/ember-addon-v2-typescript-demo
   - component, helper, instance initializer 
   - (all app-re-exports)
   - Issues
     - [x] `@glimmer/component` is undefined
       - this was caused by `"type": "module"` in the addon's package.json. We can't use `"type": "module"` quite yet
     - [x] https://github.com/embroider-build/embroider/issues/1121
   
 - https://github.com/NullVoxPopuli/ember-statechart-component/pull/244
   - gave up (for now) on this demo due to peerDep issues
     - https://github.com/embroider-build/embroider/issues/1062
     
 - https://github.com/NullVoxPopuli/ember-resources/pull/366
   - only utils, no app re-exports, compiled to single file
   - pending release of https://github.com/embroider-build/embroider/pull/1106
   - pending fix for: https://github.com/embroider-build/embroider/issues/1020
   - this approach convinced me to avoid the multi-input plugin
    
      <details><summary>because it does this:</summary>
    
      it has a bunch of weird imports.
      I understand it importing my own stuff so that it doesn't get optimized away (but these are side-effecting imports?) very weird
      
      ```js
      export { LifecycleResource } from './-private/resources/lifecycle.js';
      export { Resource } from './-private/resources/simple.js';
      export { useTask } from './-private/ember-concurrency.js';
      export { trackedFunction } from './-private/tracked-function.js';
      export { use } from './-private/use.js';
      export { useFunction } from './-private/use-function.js';
      export { useHelper } from './-private/use-helper.js';
      export { useResource } from './-private/use-resource.js';
      import '@glimmer/tracking/primitives/cache';
      import '@ember/application';
      import '@ember/destroyable';
      import '@ember/helper';
      import '@ember/debug';
      import './-private/resources/ember-concurrency-task.js';
      import '@ember/object';
      import './-private/utils.js';
      import './-private/resources/function-runner.js';
      import '@ember/runloop';
      import '@ember/test-waiters';
      //# sourceMappingURL=index.js.map
      ```
    
    
      </details>
    

## Resolves
 
 - https://github.com/embroider-build/embroider/issues/1095
 - https://github.com/embroider-build/embroider/issues/1094
 
 
## TODO

- [x] Figure out how to avoid specifying `input` at all
  - pending PR from @josemarluedke https://github.com/embroider-build/embroider/pull/1106
- [x] Figure out why `setComponentTemplate` appears to be invoked twice per template
  
   <details><summary>sample output from: 

    https://github.com/NullVoxPopuli/ember-addon-v2-typescript-demo

    </summary>

    ```js
    import { __decorate } from 'tslib';

    var TEMPLATE = setComponentTemplate(TEMPLATE, hbs`Hello there!

    <out>{{this.active}}</out>

    <button {{on 'click' this.flip}}>
      flip
    </button>
    `);

    class Demo extends Component {
      constructor(...args) {
        super(...args);
        this.active = false;

        this.flip = () => this.active = !this.active;
      }

    }

    __decorate([tracked], Demo.prototype, "active", void 0);

    export { Demo as default };
    //# sourceMappingURL=index.js.map
    ```

    </details>
    
    Potential solve:
    https://github.com/embroider-build/embroider/pull/1134

- [x] Ensure that app-js in package.json is updated
- [x] Ensure that template-only components can be imported
  - [x] Pending: https://github.com/embroider-build/embroider/pull/1126
- [x] Ensure that *at least* test-support is its own file
- [x] Answer the questions
  - [x] do we shipped compiled templates are no?
    - in most of the v2 addons out there that have components, they are still requiring the app to do the `hbs` to opcode conversion
      - discussion: https://github.com/embroider-build/embroider/pull/1107#discussion_r799062129
        tldr: the compiled format is tightly coupled to exact ember version, so it's unsafe to ship compiled templates.
         it _may_ make sense if you can _guarantee_ that addons and apps will always have the same exact ember version
         (maybe if you're doing a monorepo at work or something where everything is private)
  - [x] how should folks specify that they want sourcemaps?
    - Proposal: https://github.com/embroider-build/embroider/pull/1104

### DIFF
--- a/packages/addon-dev/sample-typescript-babel.config.cjs
+++ b/packages/addon-dev/sample-typescript-babel.config.cjs
@@ -21,6 +21,7 @@ module.exports = {
         legacy: true,
       },
     ],
+    [resolve('@babel/plugin-proposal-class-properties')],
     resolve('@embroider/addon-dev/template-colocation-plugin'),
   ],
 };
--- a/packages/addon-dev/sample-typescript-babel.config.cjs
+++ b/packages/addon-dev/sample-typescript-babel.config.cjs
@@ -1,0 +1,26 @@
+'use strict';
+
+const { resolve } = require;
+
+module.exports = {
+  presets: [resolve('@babel/preset-typescript')],
+  plugins: [
+    [
+      resolve('@babel/plugin-transform-typescript'),
+      {
+        allowDeclareFields: true,
+        onlyRemoveTypeImports: true,
+        // Default enums are IIFEs
+        optimizeConstEnums: true,
+      },
+    ],
+    [
+      resolve('@babel/plugin-proposal-decorators'),
+      {
+        // The stage 1 implementation
+        legacy: true,
+      },
+    ],
+    resolve('@embroider/addon-dev/template-colocation-plugin'),
+  ],
+};
--- a/packages/addon-dev/sample-typescript-rollup.config.js
+++ b/packages/addon-dev/sample-typescript-rollup.config.js
@@ -1,0 +1,113 @@
+/**
+  *  This file should live at ./config/rollup.config.js
+  *  and the babel config at ./config/babel.config.js (if the babel config is ESM)
+  *    _if the babel config is CJS, it can be at the addon root_
+  *
+  *  The command to build:
+  *
+  *    `rollup -c ./config/rollup.config.js`
+  */
+import path from 'path';
+
+import alias from '@rollup/plugin-alias';
+import multiInput from 'rollup-plugin-multi-input';
+import ts from 'rollup-plugin-ts';
+import { defineConfig } from 'rollup';
+import { nodeResolve } from '@rollup/plugin-node-resolve';
+import { Addon } from '@embroider/addon-dev/rollup';
+
+import babelConfig from './babel.config';
+import packageJson from '../package.json';
+
+const addon = new Addon({
+  srcDir: 'src',
+  destDir: 'dist',
+});
+
+const transpilation = [
+  // Instruct rollup how to resolve ts and hbs imports
+  // (importing a template-only component, for example)
+  nodeResolve({ resolveOnly: ['./'], extensions: ['.js', '.ts', '.hbs'] }),
+
+  // Allow top-level imports (what folks are used to from v1 addons)
+  // During the build, anything referencing a top-level import will be
+  // replaced with a relative import.
+  // DANGER: it's somewhat easy to cause circular references with this tool
+  alias({
+    entries: [
+      {
+        find: '#types',
+        replacement: path.resolve('src', '-private', 'types.ts'),
+      },
+      {
+        find: packageJson.name,
+        replacement: path.resolve('src'),
+      },
+      {
+        find: `${packageJson.name}/(.*)`,
+        replacement: path.resolve('src/$1'),
+      },
+    ],
+  }),
+
+  // This babel config should *not* apply presets or compile away ES modules.
+  // It exists only to provide development niceties for you, like automatic
+  // template colocation.
+  // See `babel.config.json` for the actual Babel configuration!
+  ts({
+    // can be changed to swc or other transpilers later
+    // but we need the ember plugins converted first
+    // (template compilation and co-location)
+    transpiler: 'babel',
+    browserslist: false,
+    babelConfig,
+    // setting this true greatly improves performance, but
+    // at the cost of safety (and no declarations output in your dist directory).
+    transpileOnly: false,
+    tsconfig: {
+      fileName: 'tsconfig.json',
+      hook: (config) => ({ ...config, declaration: true }),
+    },
+  }),
+
+  // Follow the V2 Addon rules about dependencies. Your code can import from
+  // `dependencies` and `peerDependencies` as well as standard Ember-provided
+  // package names.
+  addon.dependencies(),
+
+  // Ensure that standalone .hbs files are properly integrated as Javascript.
+  addon.hbs(),
+
+  // addons are allowed to contain imports of .css files, which we want rollup
+  // to leave alone and keep in the published output.
+  // addon.keepAssets(['**/*.css']),
+];
+
+// these should be JS, even though the authored format is TS
+const globallyAvailable = ['components/**/*.js', 'services/*.js'];
+
+export default defineConfig({
+  input: ['src/**/*{js,hbs,ts}'],
+  output: addon.output(),
+  plugins: [
+    // allows specifying glob inputs so we can transpile each module separately
+    multiInput(),
+
+    ...transpilation,
+
+    // These are the modules that users should be able to import from your
+    // addon. Anything not listed here may get optimized away.
+    addon.publicEntrypoints([...globallyAvailable]),
+
+    // These are the modules that should get reexported into the traditional
+    // "app" tree. Things in here should also be in publicEntrypoints above, but
+    // not everything in publicEntrypoints necessarily needs to go here.
+    //
+    // This generates an `_app_/` directory in your output directory
+    // and updates an 'ember-addon.app-js' entry in your package.json
+    addon.appReexports([...globallyAvailable]),
+
+    // Remove leftover build artifacts when starting a new build.
+    addon.clean(),
+  ],
+});
--- a/packages/addon-dev/sample-typescript-rollup.config.js
+++ b/packages/addon-dev/sample-typescript-rollup.config.js
@@ -1,11 +1,8 @@
 import path from 'path';
 
-import alias from '@rollup/plugin-alias';
 import ts from 'rollup-plugin-ts';
 import { defineConfig } from 'rollup';
 import { Addon } from '@embroider/addon-dev/rollup';
-
-import packageJson from './package.json';
 
 const addon = new Addon({
   srcDir: 'src',
@@ -31,27 +28,6 @@ export default defineConfig({
     // This generates an `_app_/` directory in your output directory
     // and updates an 'ember-addon.app-js' entry in your package.json
     addon.appReexports([...globallyAvailable]),
-
-    // Allow top-level imports (what folks are used to from v1 addons)
-    // During the build, anything referencing a top-level import will be
-    // replaced with a relative import.
-    // DANGER: it's somewhat easy to cause circular references with this tool
-    alias({
-      entries: [
-        {
-          find: '#types',
-          replacement: path.resolve('src', '-private', 'types.ts'),
-        },
-        {
-          find: packageJson.name,
-          replacement: path.resolve('src'),
-        },
-        {
-          find: `${packageJson.name}/(.*)`,
-          replacement: path.resolve('src/$1'),
-        },
-      ],
-    }),
 
     // This babel config should *not* apply presets or compile away ES modules.
     // It exists only to provide development niceties for you, like automatic
--- a/packages/addon-dev/sample-typescript-rollup.config.js
+++ b/packages/addon-dev/sample-typescript-rollup.config.js
@@ -10,94 +10,28 @@
 import path from 'path';
 
 import alias from '@rollup/plugin-alias';
-import multiInput from 'rollup-plugin-multi-input';
 import ts from 'rollup-plugin-ts';
 import { defineConfig } from 'rollup';
-import { nodeResolve } from '@rollup/plugin-node-resolve';
 import { Addon } from '@embroider/addon-dev/rollup';
 
-import babelConfig from './babel.config';
-import packageJson from '../package.json';
+import packageJson from './package.json';
 
 const addon = new Addon({
   srcDir: 'src',
   destDir: 'dist',
 });
 
-const transpilation = [
-  // Instruct rollup how to resolve ts and hbs imports
-  // (importing a template-only component, for example)
-  nodeResolve({ resolveOnly: ['./'], extensions: ['.js', '.ts', '.hbs'] }),
-
-  // Allow top-level imports (what folks are used to from v1 addons)
-  // During the build, anything referencing a top-level import will be
-  // replaced with a relative import.
-  // DANGER: it's somewhat easy to cause circular references with this tool
-  alias({
-    entries: [
-      {
-        find: '#types',
-        replacement: path.resolve('src', '-private', 'types.ts'),
-      },
-      {
-        find: packageJson.name,
-        replacement: path.resolve('src'),
-      },
-      {
-        find: `${packageJson.name}/(.*)`,
-        replacement: path.resolve('src/$1'),
-      },
-    ],
-  }),
-
-  // This babel config should *not* apply presets or compile away ES modules.
-  // It exists only to provide development niceties for you, like automatic
-  // template colocation.
-  // See `babel.config.json` for the actual Babel configuration!
-  ts({
-    // can be changed to swc or other transpilers later
-    // but we need the ember plugins converted first
-    // (template compilation and co-location)
-    transpiler: 'babel',
-    browserslist: false,
-    babelConfig,
-    // setting this true greatly improves performance, but
-    // at the cost of safety (and no declarations output in your dist directory).
-    transpileOnly: false,
-    tsconfig: {
-      fileName: 'tsconfig.json',
-      hook: (config) => ({ ...config, declaration: true }),
-    },
-  }),
-
-  // Follow the V2 Addon rules about dependencies. Your code can import from
-  // `dependencies` and `peerDependencies` as well as standard Ember-provided
-  // package names.
-  addon.dependencies(),
-
-  // Ensure that standalone .hbs files are properly integrated as Javascript.
-  addon.hbs(),
-
-  // addons are allowed to contain imports of .css files, which we want rollup
-  // to leave alone and keep in the published output.
-  // addon.keepAssets(['**/*.css']),
+const globallyAvailable = [
+  'components/**/*.{js,ts}', 'services/**/*.{js,ts}', 'helpers/**/*.{js,ts}',
+  'instance-initializers/**/*.{js,ts}'
 ];
 
-// these should be JS, even though the authored format is TS
-const globallyAvailable = ['components/**/*.js', 'services/*.js'];
-
 export default defineConfig({
-  input: ['src/**/*{js,hbs,ts}'],
   output: addon.output(),
   plugins: [
-    // allows specifying glob inputs so we can transpile each module separately
-    multiInput(),
-
-    ...transpilation,
-
     // These are the modules that users should be able to import from your
     // addon. Anything not listed here may get optimized away.
-    addon.publicEntrypoints([...globallyAvailable]),
+    addon.publicEntrypoints(['*.{js,ts}', ...globallyAvailable]),
 
     // These are the modules that should get reexported into the traditional
     // "app" tree. Things in here should also be in publicEntrypoints above, but
@@ -106,6 +40,62 @@ export default defineConfig({
     // This generates an `_app_/` directory in your output directory
     // and updates an 'ember-addon.app-js' entry in your package.json
     addon.appReexports([...globallyAvailable]),
+
+    // Allow top-level imports (what folks are used to from v1 addons)
+    // During the build, anything referencing a top-level import will be
+    // replaced with a relative import.
+    // DANGER: it's somewhat easy to cause circular references with this tool
+    alias({
+      entries: [
+        {
+          find: '#types',
+          replacement: path.resolve('src', '-private', 'types.ts'),
+        },
+        {
+          find: packageJson.name,
+          replacement: path.resolve('src'),
+        },
+        {
+          find: `${packageJson.name}/(.*)`,
+          replacement: path.resolve('src/$1'),
+        },
+      ],
+    }),
+
+    // This babel config should *not* apply presets or compile away ES modules.
+    // It exists only to provide development niceties for you, like automatic
+    // template colocation.
+    // See `babel.config.json` for the actual Babel configuration!
+    ts({
+      // can be changed to swc or other transpilers later
+      // but we need the ember plugins converted first
+      // (template compilation and co-location)
+      transpiler: 'babel',
+      browserslist: false,
+      // NOTE: babel config must be CJS if in the same directory as CWD
+      //       https://github.com/wessberg/rollup-plugin-ts/issues/167
+      //       otherwise ESM babel.config.js can be imported and set here
+      // babelConfig,
+      // setting this true greatly improves performance, but
+      // at the cost of safety (and no declarations output in your dist directory).
+      transpileOnly: false,
+      tsconfig: {
+        fileName: 'tsconfig.json',
+        hook: (config) => ({ ...config, declaration: true }),
+      },
+    }),
+
+    // Follow the V2 Addon rules about dependencies. Your code can import from
+    // `dependencies` and `peerDependencies` as well as standard Ember-provided
+    // package names.
+    addon.dependencies(),
+
+    // Ensure that standalone .hbs files are properly integrated as Javascript.
+    addon.hbs(),
+
+    // addons are allowed to contain imports of .css files, which we want rollup
+    // to leave alone and keep in the published output.
+    addon.keepAssets(['**/*.css']),
 
     // Remove leftover build artifacts when starting a new build.
     addon.clean(),
--- a/packages/addon-dev/sample-typescript-rollup.config.js
+++ b/packages/addon-dev/sample-typescript-rollup.config.js
@@ -1,12 +1,3 @@
-/**
-  *  This file should live at ./config/rollup.config.js
-  *  and the babel config at ./config/babel.config.js (if the babel config is ESM)
-  *    _if the babel config is CJS, it can be at the addon root_
-  *
-  *  The command to build:
-  *
-  *    `rollup -c ./config/rollup.config.js`
-  */
 import path from 'path';
 
 import alias from '@rollup/plugin-alias';
--- a/packages/addon-dev/sample-typescript-rollup.config.mjs
+++ b/packages/addon-dev/sample-typescript-rollup.config.mjs
@@ -27,12 +27,12 @@ export default defineConfig({
     // These are the modules that users should be able to import from your
     // addon. Anything not listed here may get optimized away.
     addon.publicEntrypoints([
-      'index.js',
-      'components/**/*.js',
-      'helpers/**/*.js',
-      'modifiers/**/*.js',
-      'services/**/*.js',
-      'test-support/**/*.js',
+      'index.ts',
+      'components/**/*.ts',
+      'helpers/**/*.ts',
+      'modifiers/**/*.ts',
+      'services/**/*.ts',
+      'test-support/**/*.ts',
     ]),
 
     // These are the modules that should get reexported into the traditional
--- a/packages/addon-dev/sample-typescript-rollup.config.mjs
+++ b/packages/addon-dev/sample-typescript-rollup.config.mjs
@@ -1,7 +1,6 @@
-import path from 'path';
-
 import ts from 'rollup-plugin-ts';
 import { defineConfig } from 'rollup';
+
 import { Addon } from '@embroider/addon-dev/rollup';
 
 const addon = new Addon({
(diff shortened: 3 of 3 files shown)


packages/addon-dev/sample-typescript-rollup.config.js
```diff
@@ -0,0 +1,113 @@
...
+
+// these should be JS, even though the authored format is TS
+const globallyAvailable = ['components/**/*.js', 'services/*.js'];
+
+export default defineConfig({
+  input: ['src/**/*{js,hbs,ts}'],
+  output: addon.output(),
+  plugins: [
+    // allows specifying glob inputs so we can transpile each module separately
+    multiInput(),
```
josemarluedke, line 94: Why did you need to include this plugin? Interestingly I had the same plugin in glimmer-apollo before moving to embroider addon-dev, which seem to solve the issue with `publicEntrypoints`.

packages/addon-dev/sample-typescript-rollup.config.js
```diff
@@ -0,0 +1,113 @@
...
+const transpilation = [
+  // Instruct rollup how to resolve ts and hbs imports
+  // (importing a template-only component, for example)
+  nodeResolve({ resolveOnly: ['./'], extensions: ['.js', '.ts', '.hbs'] }),
+
+  // Allow top-level imports (what folks are used to from v1 addons)
+  // During the build, anything referencing a top-level import will be
+  // replaced with a relative import.
+  // DANGER: it's somewhat easy to cause circular references with this tool
+  alias({
```
josemarluedke, line 36: In my opinion, having `alias` is not a requirement for typescript, so I think this example should be very similar to what the default example already has.

packages/addon-dev/sample-typescript-rollup.config.js
NullVoxPopuli: I couldn't _not_ need to specify inputs. Rollup wouldn't tell me why. But this enables each file to be transpiled separately so that the appReexports code here: https://github.com/embroider-build/embroider/blob/main/packages/addon-dev/src/rollup-app-reexports.ts#L21
Can find and re-export files correctly.

Previously, i'd combine all my components into a single file in dist, thinking it'd improve end-build time of the consuming app. But, fighting the tooling isn't worth it in this case.

If addon-dev had an api for specifying which file and which exports to use, that'd be cool. Like,
```js
// dist/_app_/components/my-component.js
export { MyComponent as default } from 'my-addon';
```

packages/addon-dev/sample-typescript-rollup.config.js
```diff
@@ -0,0 +1,113 @@
...
+  // It exists only to provide development niceties for you, like automatic
+  // template colocation.
+  // See `babel.config.json` for the actual Babel configuration!
+  ts({
+    // can be changed to swc or other transpilers later
+    // but we need the ember plugins converted first
+    // (template compilation and co-location)
+    transpiler: 'babel',
+    browserslist: false,
+    babelConfig,
```
rwjblue, line 63: Why do we have to specify this? Shouldn't it resolve the babel.config.js automatically?

packages/addon-dev/sample-typescript-rollup.config.js
rwjblue: @scalvert and I were just looking into this as well (trying to use addon-dev with a TS project), and are running into the "You must specify options.input" error you're referring to here. 

It is thrown here:

https://github.com/rollup/rollup/blob/e0b10605597ea2f7ee753e6f4e1ae152d19e6508/src/Graph.ts#L166-L179

packages/addon-dev/sample-typescript-rollup.config.js
NullVoxPopuli: it does, but hard errors if the babel config is ESM.
I opened an issue here: https://github.com/wessberg/rollup-plugin-ts/issues/167

packages/addon-dev/sample-typescript-rollup.config.js
NullVoxPopuli: I also found an issue with multi-input, where it adds a bunch of extra imports to the output files.
(I updated the PR description with an example).

Over at ember-resources, https://github.com/NullVoxPopuli/ember-statechart-component/pull/244
I've gradually been tweaking the rollup config until the output looks more correct.
Downside, is that, with everything in one file, consumers will require tree-shaking in order to get rid of the utilities they don't use.

ef4: I think all that might be required to avoid "You must specify options.input" is to make sure the `publicEntrypoints` plugin is renaming .ts inputs to .js outputs, so that there will be some .js chunks emitted to avoid that assertion.

rwjblue: It seems like in a non-TypeScript project things work without setting `input` through this pathway:

https://github.com/rollup/rollup/blob/e0b10605597ea2f7ee753e6f4e1ae152d19e6508/src/rollup/rollup.ts#L114-L116
https://github.com/rollup/rollup/blob/e0b10605597ea2f7ee753e6f4e1ae152d19e6508/src/utils/options/normalizeInputOptions.ts#L49
https://github.com/rollup/rollup/blob/e0b10605597ea2f7ee753e6f4e1ae152d19e6508/src/utils/options/normalizeInputOptions.ts#L156-L159

ef4: > Previously, i'd combine all my components into a single file in dist, thinking it'd improve end-build time of the consuming app. 

This is probably bad for final app optimization. While it's *possible* to tree shake inside modules, it's still a sketchy thing to do and if you ever want to be able to use native modules in the browser you shouldn't rely on it.

So shipping separate entrypoints per component is better, because then apps will only pull in the modules they actually use.

Rollup already does the right thing here by combining together modules that *definitely* get used together, while leaving apart modules that only *might* get used together.

NullVoxPopuli: > So shipping separate entrypoints per component is better, because then apps will only pull in the modules they actually use.

for sure -- is the primary motivation for trying to figure this out atm.
it's just "something" is making compiling correctly "impossible" when TS gets involved.
(still a lot of unknowns from my perspective)
 - I no longer think multi-input is a good idea
 - but I don't have other options lol (yet!)

ef4: glimmer-apollos already solved this problem of making publicEntrypoints work for typescript. We should upstream this: https://github.com/josemarluedke/glimmer-apollo/blob/6671a2191f2032cfe6e0cad018d9f66fbb10af75/packages/glimmer-apollo/rollup.config.js#L18

packages/addon-dev/sample-typescript-rollup.config.js
NullVoxPopuli: it turns out, because we can't use `"type": "module"` in our packages, the babel config needs to be cjs anyway

packages/addon-dev/sample-typescript-rollup.config.js
ef4: I don't think we'll need alias for things like self-imports as soon as typescript lands their own support for package.json exports: https://www.typescriptlang.org/docs/handbook/esm-node.html

NullVoxPopuli: oh, well if everyone is just going to support this out of the box:
https://nodejs.org/api/packages.html#packages_self_referencing_a_package_using_its_name

that's fantastic! no need for transform!

NullVoxPopuli:
> We have a monorepo with 70+ addons and having hbs pre-compiled may yield noticeable perf improvement (even some 30s on scale matters)

absolutely! I kinda want to try this exact same thing at work (with similar scale, too!)

the tldr is this:

in your babel config, add:
```js
const { precompile } = require('@glimmer/compiler'); // hope this is the same version used by your copy of ember-source 

const { resolve } = require;

module.exports = {
  plugins: [
    // other plugins
    // the important bit for compiling away the templates
    [
      resolve('babel-plugin-ember-template-compilation'),
      {
        precompile,
        enableLegacyModules: ['ember-cli-htmlbars'],
      },
    ],
  ],
}
```

NullVoxPopuli:
I just learned about `declarationMap` today.
brought up a question:
 - should our default configs, by default, demonstrate how to specify `sourcemaps: true` and `declarationMap: true`?

ef4:
I think using `declarationMap` implies that we would ship the original TS files within the published NPM packages?

If so, I think that's likely to cause trouble, because we don't want apps to start resolving the unbuilt TS files and trying to compile them, because if the typescript compiler version doesn't match perfectly they are likely to get type errors from down inside the addon. Shipping the `.js`, `.d.ts`, and `.js.map` is cleaner because apps can't accidentally try to compile your `.ts`.

NullVoxPopuli:
> I think using declarationMap implies that we would ship the original TS files within the published NPM packages?

I tested this in my big work monorepo (not using ember-resources, but in another package in the monorepo)

and without `declarationMap: true`:
 - doing go-to-definition in (neo)vim and webstorm took me directly to the declaration file instead of the source code
  - I had an occurrence of vscode going directly to where the function was defined in the v2 addon
  
and _with_ `declarationMap: true`,
 - doing go-to-definition in (neo)vim took me not to the function's source, but to the entrypoint file where it is exported -- which is an improvement to the type-declarations, because you can go-to-definition again to get to the function implementation.
 
 
 I tried this `yarn link`d to ember-resources as well, and the behavior is the same (going to the dts file instead the source (when available)
 
 I _don't_ think `declarationMap` is helpful when the source isn't available (so.. only when in monorepo or linked)  
 
 > Shipping the .js, .d.ts, and .js.map is cleaner because apps can't accidentally try to compile your .ts.
 
 100% agree. I added an additional type of test for my output files to help keep me honest as I change rollup configs.
 https://github.com/NullVoxPopuli/ember-resources/tree/main/testing/build
 
 atm, my output is:
 ```
 index.d.ts
index.d.ts.map
index.js
index.js.map
```

**Comment by simonihmig** [comment]
I tried to migrate `ember-stargate` (a v2 *Java*Script addon) to TS (WIP PR: https://github.com/kaliber5/ember-stargate/pull/491). It failed miserably, and today I think finally found the root cause...

I am basically using a similar setup like proposed here, using `rollup-plugin-ts` (which worked fine for another v2 addon). What is failing here though is that an injected service is always `undefined`. And the root cause seems to be how decorators are transpiled, especially `@inject`. (btw, the other addons mentioned here don't seem to use it.)

So the way `rollup-plugin-ts` works is to compile TS first, then Babel, AFAIU. And TS will always compile decorators away, there is no way around that (see e.g. https://github.com/microsoft/TypeScript/issues/16882 and @NullVoxPopuli's own https://github.com/wessberg/rollup-plugin-ts/issues/174). So what Babel then sees is regular JavaScript code without any decorators, but a class property for the injected service. With a babel config that only includes the minimally required `@embroider/addon-dev/template-colocation-plugin` and no other transforms, [this component](https://github.com/kaliber5/ember-stargate/blob/master/packages/ember-stargate/src/components/portal-target.js) - when running the rollup build - gets compiled to this:

<details>
  <summary>Compilation output</summary>

```js
import { _ as __decorate } from '../tslib.es6-4cf1a877.js';
import { action } from '@ember/object';
import { inject } from '@ember/service';
import { assert } from '@ember/debug';

var TEMPLATE = hbs("<div\n  {{did-insert this.register}}\n  {{will-destroy this.unregister}}\n  ...attributes\n>\n  {{yield this.count}}\n</div>");

class PortalTargetComponent extends Component {
  portalService;

  get count() {
    return this.portalService.getPortalCount(this.args.name);
  }

  register(element) {
    assert('PortalTargets needs a name', this.args.name);
    const options = {
      multiple: this.args.multiple,
      onChange: this.args.onChange
    };
    this.portalService.registerTarget(this.args.name, element, options);
  }

  unregister() {
    this.portalService.unregisterTarget(this.args.name);
  }

}

__decorate([inject('-portal')], PortalTargetComponent.prototype, "portalService", void 0);

__decorate([action], PortalTargetComponent.prototype, "register", null);

__decorate([action], PortalTargetComponent.prototype, "unregister", null);

setComponentTemplate(TEMPLATE, PortalTargetComponent);

export { PortalTargetComponent as default };
```
</details>

You can see there is still the `portalService` class property. But in the compiled *app*, this is also compiled away (I believe ember-auto-import does this for a non-Embroider app?). Same happens when you have `@babel/plugin-proposal-class-properties` in the addon's babel config. You then end up with this:

<details>
  <summary>Compilation output</summary>

```js
import { _ as _defineProperty } from '../defineProperty-1e8516b1.js';
import { _ as __decorate } from '../tslib.es6-4cf1a877.js';
import { action } from '@ember/object';
import { inject } from '@ember/service';
import { assert } from '@ember/debug';

var TEMPLATE = hbs("<div\n  {{did-insert this.register}}\n  {{will-destroy this.unregister}}\n  ...attributes\n>\n  {{yield this.count}}\n</div>");

class PortalTargetComponent extends Component {
  constructor(...args) {
    super(...args);

    _defineProperty(this, "portalService", void 0);
  }

  get count() {
    return this.portalService.getPortalCount(this.args.name);
  }

  register(element) {
    assert('PortalTargets needs a name', this.args.name);
    const options = {
      multiple: this.args.multiple,
      onChange: this.args.onChange
    };
    this.portalService.registerTarget(this.args.name, element, options);
  }

  unregister() {
    this.portalService.unregisterTarget(this.args.name);
  }

}

__decorate([inject('-portal')], PortalTargetComponent.prototype, "portalService", void 0);

__decorate([action], PortalTargetComponent.prototype, "register", null);

__decorate([action], PortalTargetComponent.prototype, "unregister", null);

setComponentTemplate(TEMPLATE, PortalTargetComponent);

export { PortalTargetComponent as default };
```
</details>

As you can see, we now have the class property removed in favor of defining it in the constructor, setting it effectively to `undefined`.

And this exactly causes the problem! `this.portalService` is now undefined, although the decorator was applied correctly. 
I can see that when debugging the point where the error is thrown: `portalService` on the component *instance* is undefined. However the `portalService` getter (added by the decorator) defined on the *prototype* returns the service instance just fine. So the issue is that this getter is shadowed by the constructor setting the property to undefined.

So basically the two step compilation of TS and then Babel is failing here AFAICT. 

An alternative would be reverting the approach to again only use the babel rollup plugin directly (with `@babel/plugin-transform-typescript`), but then we would need to run `tsc` additionally somehow to emit the declaration files. Basically what `ember-cli-typescript` is doing today. 

Any other ideas?

simonihmig:
BTW, @NullVoxPopuli I don't understand why you have [these two Babel plugins](https://github.com/NullVoxPopuli/ember-resources/blob/main/ember-resources/babel.config.cjs), as Babel would not see any TypeScript or decorators code when running as part of `rollup-plugin-ts`, as mentioned above. Or do I miss something?

SergeAstapov:
@simonihmig oh, I've been there with `ember-animated`. What I end up doing - removed `@babel/plugin-proposal-class-properties` from `babel.config.json`:

https://github.com/ember-animation/ember-animated/blob/master/addon/babel.config.json

TBH, I'm not sure we actually need to transpile away class properties at addon build time and I think app can do it based on targets

simonihmig:
Oh wow, that feedback was super fast, and more importantly it indeed fixed my issue! 🎉  Thanks a lot @SergeAstapov! 🙏

So for the record, the output is now like this:

<details>
  <summary>Compilation output</summary>

```js
import { _ as __decorate } from '../tslib.es6-4cf1a877.js';
import { action } from '@ember/object';
import { inject } from '@ember/service';
import { assert } from '@ember/debug';

var TEMPLATE = hbs("<div\n  {{did-insert this.register}}\n  {{will-destroy this.unregister}}\n  ...attributes\n>\n  {{yield this.count}}\n</div>");

class PortalTargetComponent extends Component {
  get count() {
    return this.portalService.getPortalCount(this.args.name);
  }

  register(element) {
    assert('PortalTargets needs a name', this.args.name);
    const options = {
      multiple: this.args.multiple,
      onChange: this.args.onChange
    };
    this.portalService.registerTarget(this.args.name, element, options);
  }

  unregister() {
    this.portalService.unregisterTarget(this.args.name);
  }

}

__decorate([inject('-portal')], PortalTargetComponent.prototype, "portalService", void 0);

__decorate([action], PortalTargetComponent.prototype, "register", null);

__decorate([action], PortalTargetComponent.prototype, "unregister", null);

setComponentTemplate(TEMPLATE, PortalTargetComponent);

export { PortalTargetComponent as default };
```
</details>

Which is almost the same, except for the removed class property! But this eliminates the problem I described above, that this gets transpiled further down the road in the app to assigning undefined in the constructor.

This still leaves me a bit puzzled:
* It seems only adding `@babel/preset-typescript` is enough. `@babel/plugin-proposal-decorators` was not needed for me. Decorators are still transpiled, but again by TS, not Babel
* Still I don't fully understand why we would need that plugin, given as mentioned before TypeScript types will already be compiled away when Babel kicks in, AFAIU
* That the (empty) class property is removed by `@babel/preset-typescript` feels more like an unexpected side effect, again given that the TypeScript compilation happened before Babel, doesn't it? 🤔 
* I did try adding `@babel/preset-typescript` before without success. But maybe I have just messed something up in the config setup before...

At least the [PR](https://github.com/kaliber5/ember-stargate/pull/491) is green now! 😍

simonihmig:
> TBH, I'm not sure we actually need to transpile away class properties at addon build time and I think app can do it based on targets

Yeah, class props on its own are probably ok. But in my case it seems it was a combination of an (untranspiled) class property plus the transpiled decorator code (for that same property) that blew things up...

NullVoxPopuli:
I've been having issues with converting [ember-headlessui](https://github.com/GavinJoyce/ember-headlessui/pull/144)
so, I (re)investigated doing transpilation "the hard way" here: https://github.com/NullVoxPopuli/ember-headlessui/pull/1/files (if any one is curious what the differences / considerations are)
(Un)fortunately, my issues with v2 addon conversion don't appear to have to do with how the addon is transpiled.

NullVoxPopuli:
> , I'm not sure we actually need to transpile away class properties at addon build time and I think app can do it based on targets

it _should_ be totally based on browsertargets., yeah agreed!

Here is a bare minimum demo of 
- _leaving in_ both private class properties and methods.
- transpiling away decorators

https://github.com/NullVoxPopuli/babel-transpilation-tests

NullVoxPopuli:
Progress on support for TS + components -- I have a working addon here:
 - https://github.com/NullVoxPopuli/ember-popperjs
 
 A couple things:
  - keywords "ember-addon" is required in the package, 
      - see: https://github.com/embroider-build/embroider/issues/1212
 - had to flip `hoistTransitiveImports` to `false`, because rollup was turning `ember-cli-htmlbars` into a side-effecting improt, which breaks the assumption we have where ember-cli-htmlbars is removed at build time because we only match on the `hbs` specifier.
  - https://github.com/NullVoxPopuli/ember-popperjs/blob/main/ember-popperjs/rollup.config.js#L22

NullVoxPopuli:
I think this is now ready for re-review.

with v3 of rollup-plugin-ts, we have babel doing all transpilation now, so things should be in a much better spot than before.

I must note, that it seems in typescript > 4.3, declarationMaps are no longer generated -- but it seems a problem with TS-upstream, rather than anything here

packages/addon-dev/sample-typescript-babel.config.cjs
```diff
@@ -0,0 +1,26 @@
...
+      resolve('@babel/plugin-transform-typescript'),
+      {
+        allowDeclareFields: true,
+        onlyRemoveTypeImports: true,
+        // Default enums are IIFEs
+        optimizeConstEnums: true,
+      },
+    ],
+    [
+      resolve('@babel/plugin-proposal-decorators'),
```
NullVoxPopuli, line 18: We don't need to specify the class-properties plugin.

By only specifying the decorators plugin, we leave properties "native" and untranspiled, since all browsers support class properties now.

packages/addon-dev/sample-typescript-babel.config.cjs
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+const { resolve } = require;
+
+module.exports = {
+  presets: [resolve('@babel/preset-typescript')],
```
NullVoxPopuli, line 6: Using resolve here, because I hate "plugin not found errors" in babel. Typos are much easier to catch earlier, and resolve helps out with that.

packages/addon-dev/sample-typescript-rollup.config.mjs
```diff
@@ -0,0 +1,87 @@
...
+      'helpers/**/*.ts',
+      'modifiers/**/*.ts',
+      'services/**/*.ts',
+      'test-support/**/*.ts',
+    ]),
+
+    // These are the modules that should get reexported into the traditional
+    // "app" tree. Things in here should also be in publicEntrypoints above, but
+    // not everything in publicEntrypoints necessarily needs to go here.
+    addon.appReexports([
```
NullVoxPopuli, line 41: as an aside, is there anything we can do with `appReexports` and `publicEntrypoints` to either not need file extensions, or have some sort of assertions on over-specification of extensions? The plugins are very forgiving here, and I'm not sure what's "proper".

packages/addon-dev/sample-typescript-rollup.config.mjs
```diff
@@ -0,0 +1,87 @@
...
+      'helpers/**/*.{js,ts}',
+      'modifiers/**/*.{js,ts}',
+      'services/**/*.{js,ts}',
+      'test-support/**/*.{js,ts}',
+    ]),
+    // This babel config should *not* apply presets or compile away ES modules.
+    // It exists only to provide development niceties for you, like automatic
+    // template colocation.
+    // See `babel.config.json` for the actual Babel configuration!
+    ts({
```
NullVoxPopuli, line 53: do we also want to provide a sample tsconfig?

I've been using:
```ts
{
  "compilerOptions": {
    // Path resolution
    "baseUrl": "./src",
    "moduleResolution": "node",

    // types are built by rollup-plugin-ts
    "declarationDir": "./dist",
    "emitDeclarationOnly": true,
    "declaration": true,

    // Build settings
    "noEmitOnError": false,
    "module": "ESNext",
    "target": "ESNext",

    // Features
    "experimentalDecorators": true,
    "allowJs": false,
    "allowSyntheticDefaultImports": false,

    // Strictness / Correctness
    "strict": true,
    "noImplicitAny": true,
    "noImplicitThis": true,
    "alwaysStrict": true,
    "strictNullChecks": true,
    "strictPropertyInitialization": true,
    "noFallthroughCasesInSwitch": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "noImplicitReturns": true,
    "paths": {

      // No paths, no absolute imports, only node_modules imports allowed
      // But fallback for type-overrides and such
      "*": ["types/*"]
    }
  },
  "include": [
    "src/**/*",
    "types/*"
  ]
}
```

packages/addon-dev/sample-typescript-babel.config.cjs
ef4: This code is fine, but your explanation for why it's fine is incorrect:

> since all browsers support class properties now.

Browser support has nothing to do with it. That is a concern for apps, not addons.

The only reason we need to transpile the decorators here is that acorn doesn't parse them, so rollup blows up. As soon as that is fixed, we should drop decorator transpilation, and that won't depend on whether any browsers have implemented decorators.

ef4:
Ideally, I would prefer if we can keep publicEntrypoints and appReexports arguments in terms of the publicly exposed filenames, which are always JS not TS. This is already how we treat components that are authored as standalone hbs files (in publicEntrypoints they appear as JS, not HBS, because the exact authoring format is an internal implementation detail).

NullVoxPopuli:
Yeah, I have a ts demo that I started long ago, and the output `dist` folder looks good. the app has been failing for unrelated reasons though -- trying to resolve those before re-posting said demo app.

NullVoxPopuli:
ok, so I gave up on https://github.com/NullVoxPopuli/ember-addon-v2-typescript-demo/
and decided to try on a project I already have working: https://github.com/NullVoxPopuli/ember-popperjs

the rollup build _fails_ when publicEntrypoints points only at JS (and all files on disk are TS).

ef4:
Yes, so let's fix the publicEntrypoints and/or appReexports plugins so they accept the .js extensions, and then keep the .js extensions in these examples.

NullVoxPopuli:
> Yes, so let's fix the publicEntrypoints and/or appReexports plugins so they accept the .js extensions, and then keep the .js extensions in these examples.

Since we use the publicEntrypoints to match the file system, the only way to make this work would be to scan *all* files and assume that if a file matches without the specified extension in publicEntrypoints that _that match_ (and all additional matching files?) are what we want to pull in to rollup -- brings up the question -- what happens when we encounter things rollup doesn't support? like, maybe  `src/components/demo/index.css` would be matched, given: `publicEntrypoints(['**/*.js'])`

ef4:
This is consistent with [our handling of component hbs files](https://github.com/embroider-build/embroider/blob/7f4ae9078b6234c5130e05f91ef656952cabf2be/packages/addon-dev/src/rollup-public-entrypoints.ts#L34-L37).

It's not that we match *anything*, it's that we know a short list of extension mappings caused by the build. Right now we only predict the hbs -> js transformation, but we also [have a list of extensions that normalize to js](https://github.com/embroider-build/embroider/blob/7f4ae9078b6234c5130e05f91ef656952cabf2be/packages/addon-dev/src/rollup-public-entrypoints.ts#L9) that we could also predict.

NullVoxPopuli:
PR for that here: https://github.com/embroider-build/embroider/pull/1223

packages/addon-dev/sample-typescript-rollup.config.mjs
```diff
@@ -0,0 +1,87 @@
...
+      // can be changed to swc or other transpilers later
+      // but we need the ember plugins converted first
+      // (template compilation and co-location)
+      transpiler: 'babel',
+      browserslist: ['last 2 firefox versions', 'last 2 chrome versions'],
```
simonihmig, line 58: Why do you have this? Browser support should be an app concern.

packages/addon-dev/sample-typescript-rollup.config.mjs
```diff
@@ -0,0 +1,87 @@
...
+      tsconfig: {
+        fileName: 'tsconfig.json',
```
simonihmig, line 60: Can we omit this, as this is just declaring what is already the default?

packages/addon-dev/sample-typescript-rollup.config.mjs
```diff
@@ -0,0 +1,87 @@
...
+        hook: (config) => ({
+          ...config,
+          declaration: true,
+          declarationMap: true,
+          // See: https://devblogs.microsoft.com/typescript/announcing-typescript-4-5/#beta-delta
+          // Allows us to use `exports` to define types per export
+          // However, we can't use that feature until the minimum supported TS is 4.7+
+          declarationDir: './dist',
+        }),
```
simonihmig, line 69: Do we need this hook thing? Most of the options are already set in your `tsconfig.json` example, except for `declarationMap`, which we also could set there!?

packages/addon-dev/sample-typescript-babel.config.cjs
ef4: Why do we have both `@babel/preset-typescript` and `@babel/plugin-transform-typescript`? I think it should be one or the other. The only thing the preset does is install that plugin.

packages/addon-dev/sample-typescript-rollup.config.mjs
ef4: Yes, the docs confirm it: https://github.com/wessberg/rollup-plugin-ts#babelpreset-env-behavior-and-how-to-opt-out

NullVoxPopuli: setting browserslist to false added poplyfills for ownKeys and objectSpread in my addon.

packages/addon-dev/sample-typescript-babel.config.cjs
NullVoxPopuli: plugin-transform-typescript is required when doing `resolve` in the babel.config.js (by eslint + node:recommended)
if I just use strings for the plugin names, the issue is side-stepped.

packages/addon-dev/sample-typescript-rollup.config.mjs
NullVoxPopuli: just kidding, it's required when hooks are present

packages/addon-dev/sample-typescript-rollup.config.mjs
simonihmig: @ef4 @NullVoxPopuli  As discussed yesterday, I looked into this today. Our suspicion, that `rollup-plugin-ts` does some unexpected magic, seems to be right unfortunately. I filed this issue: https://github.com/wessberg/rollup-plugin-ts/issues/189

NullVoxPopuli:
Gonna close this, because I think we've figured out everything we need to figure out w/r/t TS support in v2 addons (and there are demos of these in the wild).

Also with the upcoming `--typescript` support landing in both ember-cli and the `addon-blueprint`, the `addon-blueprint` will be the canonical (and tested) place to look for how to do TS

PR #1101, opened 2022-02-03
Add test for v2 addon importing from v1 addon's fake module

NullVoxPopuli:
This added test is to try to recreate some troubles I've been experiencing with 
 - https://github.com/NullVoxPopuli/ember-resources/pull/279
 - which has related issue: https://github.com/embroider-build/embroider/issues/1020
 
 The end goal is to reproduce the problem in embroider's test suite, so I can debug what's going on a little easier.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,6 +16,8 @@ Tests can be run per-package, or combined all together.
 
 Per-package test suites are defined as the `yarn test` command within each `packages/*` or `test-packages/*`.
 
+Scenario tests can be ran in `tests/scenarios` with `yarn test --filter "scenario name"`.
+
 The combined suite can be run via the top-level `yarn test` command. It's defined by combining
 
 - all the `yarn test:*` commands within each `packages/*` or `test-packages/*`. These are typically complete Ember apps that need to get built and tested in the browser.
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "scenario-tester/fixturify-project": "4.1.0"
   },
   "devDependencies": {
+    "@babel/helper-regex": "^7.10.5",
     "@types/jest": "^24.0.0",
     "@typescript-eslint/eslint-plugin": "^4.1.1",
     "@typescript-eslint/parser": "^4.1.1",
--- a/tests/scenarios/v2-addon-test.ts
+++ b/tests/scenarios/v2-addon-test.ts
@@ -1,9 +1,399 @@
-import { appScenarios, baseV2Addon } from './scenarios';
-import { PreparedApp } from 'scenario-tester';
+import { appScenarios, baseAddon, baseV2Addon } from './scenarios';
+import { PreparedApp, Project } from 'scenario-tester';
 import QUnit from 'qunit';
 import merge from 'lodash/merge';
 
 const { module: Qmodule, test } = QUnit;
+
+function buildV2Addon() {
+  let addon = new Project('my-v2-addon', {
+    files: {
+      'addon-main.js': `
+        const { addonV1Shim } = require('@embroider/addon-shim');
+        module.exports = addonV1Shim(__dirname);
+      `,
+      'index.js': `
+        import plainDep from 'plain-dep';
+        import { innerV1Addon } from 'inner-v1-addon';
+        import { innerV2Addon } from 'inner-v2-addon';
+        export function usePlainDep() {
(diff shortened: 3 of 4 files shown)


ef4:
Either way this will be good to keep in the test suite.

ef4:
I think your original bug may have been around `@glimmer/component`, which is definitely different from the other "fake" ember packages because a real package *does* exist, but the implementation inside it is really only for standalone glimmerX and shouldn't actually be used, because Ember contains the real implementation.

NullVoxPopuli:
> I think your original bug may have been around @glimmer/component

yeah, I have a couple things I'm trying to reproduce:

from here: https://github.com/NullVoxPopuli/ember-addon-v2-typescript-demo
`@glimmer/component` appears to be undefined.

from here: https://github.com/NullVoxPopuli/ember-resources/pull/366
 - this issue: https://github.com/embroider-build/embroider/issues/1020

josemarluedke:
I was playing with [ember-addon-v2-typescript-demo](https://github.com/NullVoxPopuli/ember-addon-v2-typescript-demo) and figured out that `@glimmer/component` is returning an object with default instead of the component class. 

See:
Screen Shot 2022-02-03 at 2 21 13 PM

Screen Shot 2022-02-03 at 2 21 08 PM

When logging `Component`, we can see it as an object:

Screen Shot 2022-02-03 at 2 21 26 PM

If I change the dist files to `class FlatDemo extends Component.default`, that works.

ef4:
The problem in that demo repo is:

https://github.com/NullVoxPopuli/ember-addon-v2-typescript-demo/blob/125700fc9aa77faec4d9a49c514a4fa90e477996/my-addon/package.json#L10

When webpack sees that it gets Very Strict with your module, including disabling things like interoperability between `import` and `require`. 

But we need the interoperability, because in a classic build (which is what that repo is using), `@glimmer/component` has long since been transpiled from an ES module to AMD by the time our v2 addon gets it. Without the interop, we don't know to grab the `.default` property off the value that comes back from `require()`.

Under Embroider things are different, because webpack *will* see `@glimmer/component` as a real ES module. But the build fails in that case for other reasons. I'm not really sure if we can turn on this level of Webpack strictness until much more of Ember itself ships as ES modules. We liberally use `require()` in the compiled output to paper over the gap between Ember's runtime loader and ES module loading.

ef4:
I think you can use [Rule.type](https://webpack.js.org/configuration/module/#ruletype) to override webpack's behavior here and go back from `javascript/esm` to `javascript/auto`.

NullVoxPopuli:
ah, interesting -- I hadn't heard of rule.type before.

Changing my config in the test app to this has everything working with @josemarluedke's addon-dev changes. :tada: 

<details><summary>updated emebr-cli-build.js</summary>

```js
'use strict';

  let app = new EmberApp(defaults, {
    autoImport: {
      watchDependencies: [Object.keys(require('./package').dependencies)],
      webpack: {
        devtool: 'inline-source-map',
        module: {
          rules: [{ test: /\.(js|ts)$/, type: 'javascript/auto' }],
        },
      },
    },
  });

  const { maybeEmbroider } = require('@embroider/test-setup');

  return maybeEmbroider(app, {
    packagerOptions: {
      webpackConfig: {
        devtool: 'inline-source-map',
        module: {
          rules: [{ test: /\.(js|ts)$/, type: 'javascript/auto' }],
        },
      },
    },
  });
};
```

</details>


Some followup questions though -- and maybe "elsewhere" is more appropriate for this sort of discussion, but idk:
 
> Without the interop, we don't know to grab the .default property off the value that comes back from require().
   
   does the problem go away if we convert `@glimmer/component` to a v2 addon? (I know that's not easy and has a lot of nuance right now, since all the implementations of stuff are actually in glimmer -- but just conceptually -- is it the whole ecosystem that needs this sort of interop? or just the `@glimmer/*` stuff?
   
> I'm not really sure if we can turn on this level of Webpack strictness until much more of Ember itself ships as ES modules. We liberally use require() in the compiled output to paper over the gap between Ember's runtime loader and ES module loading.

  does a PR to flip to the defaults to `javascript/auto` in both auto-import and embroider make sense? (at least until more native ES Modules are shipped? (also, does this mean it's more than just `@glimmer/*` that has some special interop? (I suppose the template-compiler?)))

NullVoxPopuli:
sufficient testing exists now on main

PR #1102, opened 2022-02-03
Optimize addonCacheKey computation

SergeAstapov:
Since adoption of `ember-auto-import` v2 and ecosystem migration to `@embroider/macros@1.0.0` we seen significant performance degradation in our not that small app - it went from 4 mins to 20 mins (but with way to inefficient dependencies graph and too many cross-dependencies, which is separate problem).

Debugging and profiling leads to `gatherAddonCacheKey` function being biggest contributor.

**First issue**
Every time `@embroider/macros` addon is included (and it's included way more than few times due to `ember-auto-import` and many other addons depend on it) the whole tree travels happens over and over.

**Second issue**
there is `return [...memo].join('|')` line which runs for each end every recursion level but the result is used only from top-level return when invoked with `project` as an argument.

Changes in this PR propose to traverse addons three only once and build `cacheKey` string only once per project.

I wasn't sure if `project` *always* is the same reference or there are use cases when we may have more than one `project` during build time hence used `WeakMap`.

If we know for sure `project` *always* is the same reference - than we can replace `WeakMap` with simple string variable that would be returned once we invoke `gatherAddonCacheKey` second and subsequent times.

### DIFF
--- a/packages/macros/src/macros-config.ts
+++ b/packages/macros/src/macros-config.ts
@@ -39,16 +39,34 @@ type GlobalSharedState = WeakMap<
 // share the GlobalSharedState beneath.
 let localSharedState: WeakMap<any, MacrosConfig> = new WeakMap();
 
-// creates a string representing all addons and their versions
-// (foo@1.0.0|bar@2.0.0) to use as a cachekey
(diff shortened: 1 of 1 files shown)


ef4:
Thanks, very nice.

Issue #1018, opened 2021-11-12
Contextual modifiers and helpers not supported?


simonihmig:
Ember 3.27 introduced the `(modifier)` and `(helper)` helper counterparts to `(component)`. Using `(modifier)` in an addon to yield a contextual modifier works in non-Embroider and the "safe" scenario, but fails in "optimized" mode with `Missing helper: modifier`.

It seems this new feature is not properly supported yet? I guess we need similar special treatment for these as we have for the [component helper](https://github.com/embroider-build/embroider/blob/dcd4eaaf851368e3393367c474e9eccdfbae13e9/packages/compat/src/resolver-transform.ts#L249)?

ef4:
Yes, we need to make them work almost exactly the same as the component helper, just resolving to slightly different paths.

Along with this we also still need a staticModifiers mode. None of it is very different from staticHelpers and staticComponents.

simonihmig:
> Yes, we need to make them work almost exactly the same as the component helper, just resolving to slightly different paths.

Thanks for confirming!

> Along with this we also still need a staticModifiers mode. None of it is very different from staticHelpers and staticComponents.

Yup, I raised this already here: https://github.com/embroider-build/embroider/issues/603 🙂

Maybe I can find some time for this, but no promises...

rahulk94:
Looks like this has been added but missed from being added as a config option in the README.

Windvis:
AFAIK this hasn't been implemented yet? The PR which is referenced here only added the `staticModifiers` option.

simonihmig:
I think you are right, so I reopened this issue!

rahulk94:
Woops my bad 🤦🏾‍♂️  sorry team

Windvis:
@ef4 Forgive my ignorance but do we also need to add `ensureSafeHelper` and `ensureSafeModifier` utils to fully support these helpers?

ef4:
I think "no", because I'm pretty sure the newly added helper-helper and modifier-helper never accepted dynamic values. They only accept string literals.

So it's not possible to write a component that accepts a helper as an argument "the wrong way", by accepting the string name of the helper, rather than the helper itself.

Windvis:
As of the 1.4.0 release both these helpers should now be fully supported. :tada:

PR #1103, opened 2022-02-03
Update the ember-template-compiler to v4.1.0

Windvis:
I started looking into https://github.com/embroider-build/embroider/issues/1018 again and I noticed the tests failed because the template compiler doesn't support the new helpers yet.

I just plucked this out of the ember-source package of a 4.1.0 Ember app. Let me know if there is a better way to update it :smile:.

### DIFF
--- a/packages/compat/tests/resolver.test.ts
+++ b/packages/compat/tests/resolver.test.ts
@@ -874,7 +874,7 @@ describe('compat-resolver', function () {
       findDependencies(
         'templates/application.hbs',
         `
-        {{outlet "foo"}}
+        {{outlet}}
         {{yield bar}}
         {{#with (hash submit=(action doit)) as |thing| }}
         {{/with}}
--- a/test-packages/support/vendor/README.md
+++ b/test-packages/support/vendor/README.md
@@ -1,3 +1,3 @@
-This is vendored from ember 3.25.0.
+This is vendored from ember 4.1.0.
 
 I did it this way because if I try to depend directly on ember-source, I end up with a version of fs-tree-diff that has bad types in it that messes up my build.
--- a/test-packages/support/vendor/ember-template-compiler.js
+++ b/test-packages/support/vendor/ember-template-compiler.js
@@ -1,23 +1,44 @@
 (function() {
 /*!
  * @overview  Ember - JavaScript Application Framework
- * @copyright Copyright 2011-2020 Tilde Inc. and contributors
+ * @copyright Copyright 2011-2021 Tilde Inc. and contributors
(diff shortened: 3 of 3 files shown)


packages/macros/src/glimmer/macro-maybe-attrs.ts
```diff
@@ -20,11 +20,11 @@ export function maybeAttrs(elementNode: any, node: any, builders: any) {
 
   if (result.value) {
     for (let bareAttr of bareAttrs) {
-      elementNode.attributes.push(builders.attr(bareAttr.original, builders.text('')));
+      elementNode.attributes.push(builders.attr(bareAttr.original, builders.text(''), bareAttr.loc));
```
Windvis, line 23: Some `macroMaybeAttrs` tests were failing due to some changes in the ember-template-compiler. 

[This PR](https://github.com/glimmerjs/glimmer-vm/pull/1266) added sorting logic to the printer which [sorts attributes](https://github.com/charlespierce/glimmer-vm/blob/b963b0955214166109a6560ea2c685a6a2b8ff22/packages/%40glimmer/syntax/lib/generation/printer.ts#L222) based on their `loc` data. Since the `macroMaybeAttrs` transform didn't attach `loc` information to the newly created attributes they were being moved to the front of the element due to this sorting. In particular, [this condition](https://github.com/charlespierce/glimmer-vm/blob/b963b0955214166109a6560ea2c685a6a2b8ff22/packages/%40glimmer/syntax/lib/generation/util.ts#L71) is hit because the attributes without `loc` data seem to default to a `0` value for `column`.

I reused the `loc` from the original source so that the printer maintains the correct order and the tests pass again.

Let me know if that wasn't the correct fix. :smile:

Windvis:
The failing check seems unrelated to this change?

`connect ECONNREFUSED 52.73.87.228:443`

ef4:
Yes, that CI failure is something in Github's caching layer as used by the Volta action and it seems to happen about 1% of the time, which means in our suite of 111 jobs it happens once almost every run. 🙃

ef4:
Thanks for working on this. 

This part of the test suite is lacking a way to test multiple ember versions. Can you verify (at least manually) that these changes to macro-maybe-attrs are compatible with the 3.25-era template-compiler that this PR is replacing?

Ultimately we need to finish moving all the node tests into the tests/scenarios so that they can be run under multiple ember versions.

Windvis:
@ef4 The tests still pass with the template compiler of ember 3.25. I can split the `lec` commit in a separate branch and PR it on its own so that it can be verified through CI?

ef4:
Thanks, nice work.

PR #1104, opened 2022-02-03
[Addon Dev] allow all of rollup's OutputOptions to be passed to addon.output()

NullVoxPopuli:
Update!

This PR adds pass-through support for all of rollup's output options:

```js
output: addon.output({ sourcemap: true });
```

Intellisense confirmed working, too:
image

--------------------
(old PR description)

I found myself wanting to provide sourcemaps for libraries.
`addon.output()` does not allow for that, but you could do:
```js
output: {
  ...addon.output(),
  sourcemap: true,
}
```

but, that's extra code on the consuming side, when it could just be 1 extra line of config in the `new Addon` part of the rollup config file.

There are _a lot_ of output options tho: https://rollupjs.org/guide/en/#outputoptions-object.
:shrug: 

Maybe instead, we can take any object as an argument to `.output()` and merge that in the rollup config?


packages/addon-dev/src/rollup.ts
```diff
@@ -62,7 +90,12 @@ export class Addon {
   // their bundle names (the bundle names get generated by `publicEntrypoints`
   // above).
   output() {
```
rwjblue, line 92: Why not just take options here?

NullVoxPopuli: I _just_ edited the PR to ask that same question. Happy to change if that's preferred.

To answer "why not":  I didn't think of it initially :upside_down_face:

NullVoxPopuli: > I just edited the PR to ask that same question

I'm really upset at github for continuing to revert my PR descriptions :cry: 
image

NullVoxPopuli: oh it's back now. idk what's going on with github :(

NullVoxPopuli:
I removed `sourcemap` from the constructor, kept the docs tho, and added _all_ output options to the output function. Looks kinda fun (intellisense test):

image

kinda weird they don't have any description or anything on these, but that's work we PR to rollup to help everyone out

packages/addon-dev/src/rollup.ts
NullVoxPopuli: advantage of passing options to output function was too high! (changed to that approach)

ef4:
This PR comes down the different between:

```js
{
  // current
  output: { ...addon.output(),  sourceMap: true }
}
```

and

```js
{
  // proposed
  output: addon.output({ sourceMap: true })
}
```

They're extremely close, but the proposed option forces you to read the documentation for `addon.output()` to understand what's going to happen, whereas the current option is Just Javascript™️ and it's transparent what's happening. So I'm not a fan.

NullVoxPopuli:
aight, sounds good! 🥳

PR #1105, opened 2022-02-03
Cancel workflows when they become outdated

Windvis:
I noticed that multiple workflows were added to the queue when I pushed new commits and the previous ones weren't canceled.
This change ensures that workflows are canceled when new changes come in which means resources aren't wasted on outdated workflows.

ember-cli added the same config a while a go: https://github.com/ember-cli/ember-cli/pull/9726

More information: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,10 @@ on:
   pull_request:
     branches: [main]
 
+concurrency:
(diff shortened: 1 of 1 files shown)


ef4:
Thanks, this looks good.

PR #1107, opened 2022-02-03
Creating v2 addon guide

ef4:
This is the start of a guide on porting addons to v2 format.

### DIFF
--- a/PORTING-ADDONS-TO-V2.md
+++ b/PORTING-ADDONS-TO-V2.md
@@ -1,0 +1,260 @@
+# Guide: Porting an Addon to V2
+
+This is a guide for addon authors who want to publish their addon in **v2 format**.
+
+> The actual V2 Format RFC only cares what format you **publish** to NPM. It doesn't necessarily care about your **authoring** format or toolchain. But in this guide we are picking good defaults, and we hope to polish this experience until it's ready to become a new RFC as the default new Ember Addon authoring blueprint.
+
+## What Addons should and should not be converted to V2?
+
+The best candidates to convert to V2 are addons that provide only runtime features, like components, helpers, modifiers, and services. That kind of addon should definitely port to V2.
+
+In contrast, addons that are primarily an extension to the build system (like `ember-cli-sass` or `ember-cli-typescript`) are not good candidates to be V2 addons, at least at present. As V1 addons, they will continue to work through `@embroider/compat` even for apps with Embroider.
+
+If your addon is a mix of both build-time and run-time features, consider trying to replace the build-time features with `@embroider/macros`. This would let you drop all your custom build-time code and port to V2. Alternatively, if you really need build customizations, you can provide users instructions and utilities (like a webpack rule or plugin) to add those customizations to their Embroider build. We do _not_ let V2 addons automatically manipulate the app's build pipeline. Thar be dragons.
+
+## Monorepo Organization
+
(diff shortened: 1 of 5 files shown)


packages/addon-dev/sample-babel.config.json
```diff
@@ -0,0 +1,7 @@
+{
+  "plugins": [
+    "@embroider/addon-dev/template-colocation-plugin",
+    ["@babel/plugin-proposal-decorators", { "legacy": true }],
+    "@babel/plugin-proposal-class-properties"
+  ]
+}
```
NullVoxPopuli, line 7: why or why shouldn't we also ship compiled templates?

(for example: https://github.com/NullVoxPopuli/ember-addon-v2-typescript-demo/blob/main/my-addon/babel.config.cjs#L34-L39 )

```js
const { precompile } = require('@glimmer/compiler');

// ...

      resolve('babel-plugin-ember-template-compilation'),
      {
        precompile,
        enableLegacyModules: ['ember-cli-htmlbars'],
      },
    ],
```    

by not compiling away the templates, apps have more work to do, yea?

NullVoxPopuli (approved):
Nice! this is very detailed!

PORTING-ADDONS-TO-V2.md
```diff
@@ -0,0 +1,264 @@
...
+
+1. Make a new top-level .gitignore:
+
+   ```
+   # you definitely want this:
+   node_modules
+
+   # and you can put in anything else that tends to accumulate in your environment:
+   yarn-error.log
+   .DS_Store
```
SergeAstapov, line 79: we may mention `.github` as it's often overlooked

ef4: Is there stuff in `.github` that should be gitignored?

packages/addon-dev/sample-babel.config.json
ef4: Because the compiled format is deeply tied to the exact ember version that's going to consume it.

ef4:
Please try it and give more feedback, but I'm merging so we have a starting point for more PRs.

PORTING-ADDONS-TO-V2.md
SergeAstapov: ah sorry, this is `.gitignore` and I read `.npmignore` 🙈

packages/addon-dev/sample-babel.config.json
NullVoxPopuli: gotchya, so this is probably only useful if you can guarantee that addons have the same ember version as apps :thinking: 

could be tricky

ef4: It would be possible to make the wire format (the thing that you'd have if you use babel-plugin-ember-template-compilation) a stable public API, but somebody with knowhow would need to prioritize a little design work.

angelayanpan:
> | Optimized Embroider Safe |   v1   |

is this the same as Embroider Optimized?

ef4:
Yes, this is the same thing tested by the embroider-safe scenario provide by `@embroider/test-setup`.

angelayanpan:
> Once you achieve Optimized Embroider Safe, you should enable the `embroider-optimized`

Maybe this makes it sound like there is a difference between `Optimized Embroider Safe` ? and embroider-optimized? I was confused when I read "Optimized Embroider Safe" multiple times before `embroider-optimized` was mentioned.

PR #1109, opened 2022-02-05
fastboot chunk preloading fix

ef4:
Ember Fastboot is deliberately designed to be able to run the same build output that was produced for the browser in Node, without an entire second build for this second environment.

In order to fit in with that architecture, Embroider preloads all lazy chunks in fastboot, so that we never hit webpack's runtime loader (which will expect to be running in the browser and break in fastboot).

Previously, the way we identified the lazy chunks was simply by listing all the chunks and seeing which ones were not used as entry chunks for any entrypoint. But this was incorrect.

A chunk could be used *lazily* by one entrypoint (like the app) while being used *eagerly* by another (like the test suite). In this case we would miss preloading it and get a runtime error.

This changes the chunk categorization strategy to track the lazy chunks per-entrypoint.

### DIFF
--- a/packages/webpack/src/ember-webpack.ts
+++ b/packages/webpack/src/ember-webpack.ts
@@ -396,14 +396,14 @@ const Webpack: PackagerConstructor<Options> = class Webpack implements Packager 
   private summarizeStats(multiStats: webpack.MultiStats): BundleSummary {
     let output: BundleSummary = {
       entrypoints: new Map(),
-      lazyBundles: new Set(),
+      lazyBundles: new Map(),
       variants: this.variants,
     };
     for (let [variantIndex, variant] of this.variants.entries()) {
-      let { entrypoints, assets } = multiStats.stats[variantIndex].toJson({
+      let { entrypoints, chunks } = multiStats.stats[variantIndex].toJson({
         all: false,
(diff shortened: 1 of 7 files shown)


packages/webpack/src/ember-webpack.ts
```diff
@@ -427,20 +426,16 @@ const Webpack: PackagerConstructor<Options> = class Webpack implements Packager
...
-            output.lazyBundles.add('assets/' + asset.name);
-          }
+        if (variant.runtime !== 'browser') {
+          // in the browser we don't need to worry about lazy assets (they will be
+          // handled automatically by webpack as needed), but in any other runtime
+          // we need the ability to preload them
+          output.lazyBundles.set(
+            id,
+            chunks
+              .filter(chunk => chunk.runtime?.includes(id) && !entrypointAssets?.find(a => a.name === chunk.files?.[0]))
```
ef4, line 436: I can't find any obvious docs about `chunk.runtime` but it appears to mean what we need it to mean: which entrypoints can use this chunk at runtime?

ef4:
@krisselden I think this might be the fix to the problem you were having with fastboot dying on `document is undefined`.

PR #1115, opened 2022-02-08
Use broccoli memoization by default

### LINKED ISSUE #1112: ember-cli shares broccoli trees across builders
After @krisselden fixed cleanup of the OneShot builders in https://github.com/embroider-build/embroider/pull/1084, it exposed the fact that we were leaking broccoli trees across the boundary between different broccoli builders via cacheKeyForTree. I fixed that in https://github.com/embroider-build/embroider/pull/1100.

But now I have also discovered that ember-cli itself stores a tree in module scope:

https://github.com/ember-cli/ember-cli/blob/a7c0d3dd932fe78369fd538bb2cb00c7329bd0a6/lib/broccoli/merge-trees.js#L6

This makes ember-cli unsafe to use when more than one broccoli builder can exist in the same process, as we do with OneShot.

This is the source of the test failure in #1085 in [lts_3_28-engines ubuntu](https://github.com/embroider-build/embroider/runs/5081404027?check_suite_focus=true). I don't fully understand the sequence of events that causes the bug to trigger under that particular set of dependencies, but you can make it go away by doing any of these things:
 - patch out the [shared EMPTY_MERGE_TREE](https://github.com/ember-cli/ember-cli/blob/a7c0d3dd932fe78369fd538bb2cb00c7329bd0a6/lib/broccoli/merge-trees.js#L14)
 - or patch out [the builder cleanup in OneShot](https://github.com/embroider-build/embroider/blob/9bc26f2e950ca0fb2e1d915d70da085a7583bbea/packages/compat/src/one-shot.ts#L43)
 - or set BROCCOLI_ENABLED_MEMOIZE=true which disables OneShot

So it's pretty clear that we're sharing the tree between the main builder and the OneShot builders, and that's blowing up.

ef4:
This closes #1112 by removing our OneShot optimization hack in favor of turning on broccoli memoization by default.

Broccoli memoization should absolutely be broccoli's default. It's not yet, but we can turn the flag on here unless users have explicitly turned it off.


ef4:
If we land this, we *could* revert https://github.com/embroider-build/embroider/pull/1100. While I remain not a fan of cacheKeyForTree, it would become safe again for us to use it with OneShot gone, and our previous implementation worked harder to be compatible with the old caching semantics.

PR #1119, opened 2022-02-08
Reuse the `loc` of the macroMaybeAttrs hashes for the created attributes

Windvis:
The newer template compiler sorts attributes based on their `loc` location when printing. Since previously no `loc` info was added to the newly created attributes, they were being moved to the front of the element. By reusing their original `loc` data we can ensure that the order from the source file is maintained.

This commit was included in #1103 since it's a requirement for the tests to pass under the newer compiler version.


ef4:
Thanks, looking good.

PR #1120, opened 2022-02-08
Add support for the `helper` helper

Windvis:


packages/compat/src/resolver-transform.ts
```diff
@@ -350,3 +358,51 @@ function handleComponentHelper(
...
+  // let locator: ComponentLocator;
+  // switch (param.type) {
+  //   case 'StringLiteral':
+  //     locator = { type: 'literal', path: param.value };
+  //     break;
+  //   case 'PathExpression':
+  //     locator = { type: 'path', path: param.original };
+  //     break;
+  //   case 'MustacheStatement':
+  //     if (param.hash.pairs.length === 0 && param.params.length === 0) {
```
lifeart, line 377: it will be great to resolve super simple dynamic cases, like

```hbs
{{helper (if this.a "foo" "bar")}}
```

```hbs
{{helper (if this.localHelper this.localHelper "fallback-helper")}}
```

Windvis: @lifeart I think it was a deliberate choice not to support that for these helpers. https://github.com/embroider-build/embroider/issues/1018#issuecomment-999716009

Maybe I just misunderstood, but the RFC doesn't mention that either I think.

ef4: Yes, the dynamic case can be dynamic *outside* of the `(helper )` call:

```hbs
{{if this.localHelper this.localHelper (helper "fallback-helper") }}
```

But also, since helpers are first-class values that can be passed around IMO it's nicer to implement your fallback-helper in the corresponding JS file anyway, so that it's not polluting the global namespace:

```hbs
{{if this.localHelper this.localHelper this.fallbackHelper }}
```

```js
import { helper } from '@ember/component/helper';
export default class extends Component {
  fallbackHelper = helper(function() {...});
}
```

lifeart: https://github.com/ember-template-lint/ember-template-lint/issues/2375

ef4: I confirmed that ember already hard-enforces the main thing we care about here:

>  Assertion Failed: Passing a dynamic string to the `(helper)` keyword is disallowed. (You specified `(helper s)` and `s` evaluated into "my-helper".) This ensures we can statically analyze the template and determine which helpers are used. If the helper name is always the same, use a string literal instead, i.e. `(helper "my-helper")`. Otherwise, import the helpers into JavaScript and pass them to the helper keyword. See https://github.com/emberjs/rfcs/blob/master/text/0496-handlebars-strict-mode.md#4-no-dynamic-resolution for details. ('sample2/templates/application.hbs' @ L3:C9) 

The last bit of that message is actually out of date for Ember 3.25+ because you don't need to even pass them to the helper keyword, you can just use them as-is.

ef4: Fired off a PR to update those out-of-date instructions in Ember https://github.com/emberjs/ember.js/pull/19950

ef4:
Thanks @Windvis, I was able to push this over the line.

Windvis:
@ef4 Sorry about that. I haven't had the time nor energy to do much open source work the past weeks. :disappointed: 

Thanks for finishing it up!

packages/compat/src/resolver-transform.ts
```diff
@@ -320,3 +328,16 @@ function handleComponentHelper(
...
+function handleDynamicHelper(param: ASTv1.Node, resolver: Resolver, moduleName: string): void {
+  switch (param.type) {
+    case 'StringLiteral':
+      resolver.resolveDynamicHelper({ type: 'literal', path: param.value }, moduleName, param.loc);
+      break;
+    case 'TextNode':
+      resolver.resolveDynamicHelper({ type: 'literal', path: param.chars }, moduleName, param.loc);
+      break;
+    default:
+      resolver.resolveDynamicHelper({ type: 'other' }, moduleName, param.loc);
```
Windvis, line 341: @ef4 Do we actually want to handle this case? The contextual helpers RFCs mentions that passing a helper reference into the helper should be supported so that extra arguments can be added: https://emberjs.github.io/rfcs/0432-contextual-helpers.html 

I left it out in the draft PR since [Ember itself verifies that the helper isn't dynamic](https://github.com/emberjs/ember.js/blob/bfaed5a9a944bf31c99f5772da4f7681b7004555/packages/ember-template-compiler/lib/plugins/transform-resolutions.ts#L34). I think Embroider only needs to do anything if users pass in a helper name?

I think the code snippet in the RFC would throw an error with the current implementation?
```hbs
{{#let (helper "join-words" separator=",") as |join|}}
  {{#let (helper join "foo") as |foo|}}
    {{#let (helper foo "bar") as |foo-bar|}}
      {{foo-bar "baz"}}
    {{/let}}
  {{/let}}
{{/let}}
```

packages/compat/src/resolver-transform.ts
```diff
@@ -320,3 +328,16 @@ function handleComponentHelper(
 
   resolver.resolveComponentHelper(locator, moduleName, param.loc, impliedBecause);
 }
+
+function handleDynamicHelper(param: ASTv1.Node, resolver: Resolver, moduleName: string): void {
+  switch (param.type) {
+    case 'StringLiteral':
+      resolver.resolveDynamicHelper({ type: 'literal', path: param.value }, moduleName, param.loc);
+      break;
+    case 'TextNode':
```
Windvis, line 337: Silly question, but how is it possible to pass a `TextNode` to the helper keyword? There is probably a legit use-case I'm missing, but I can't seem to create that scenario in AST-explorer :smile:.

packages/compat/src/resolver-transform.ts
ef4: Yes, this also generates an error, it's just done from the resolver instead of directly here because the resolver keeps track of all the errors.

packages/compat/src/resolver-transform.ts
ef4: Maybe it's not? I was copying the earlier work on the component helper and that does have handling for TextNode, and I don't recall if that's because we found a case or because some typescript types implied it was possible.

packages/compat/src/resolver-transform.ts
Windvis: But I think it _shouldn't_ throw an error for that code snippet? Passing in a helper reference should work, according to the RFC, but with the currently merged implementation it will throw an error I think?

Unless I misunderstood the code, I think at the moment, errors will be thrown under Embroider while it works as expected in a classic build.

PR #1125, opened 2022-02-11
rollup/babel: template-only component support

NullVoxPopuli:
right now, I don't remember how this is supposed to work :(

the existing test harness is _very_ focused on existing javascript files.

as is the actual babel plugin.

questions:
 - do we even need this? if we're to emit _extra_ files it seems like the template-colocation plugin is causing extra work for the other plugins.
 - if the answer is to emit those extra files, how would we do it from the scope of the babel plugin? seems like it could only be possible if the plugin wasn't _per-file_


NullVoxPopuli:
closed in favor of https://github.com/embroider-build/embroider/pull/1126

Issue #1121, opened 2022-02-09
V2 Addons: cannot import template-only components


NullVoxPopuli:
Reproduction: https://github.com/NullVoxPopuli/ember-addon-v2-template-only-import

This was originally discovered here: https://github.com/embroider-build/embroider/pull/1099
but I figured I'd make a more focused issue (cause closing things and checking boxes off is fun)

Atm, this is using the default rollup/babel config shown in the addon-dev package.
This is the error:
```
$ concurrently "npm:build:*"
[build:*js] 
[build:*js] > my-addon@0.0.0 build:js
[build:*js] > rollup -c ./config/rollup.config.js
[build:*js] 
[build:*js] 
[build:*js]  → dist...
[build:*js] [!] Error: Could not resolve './button' from src/components/demo/index.js
[build:*js] Error: Could not resolve './button' from src/components/demo/index.js
[build:*js]     at error (<repo>/node_modules/rollup/dist/shared/rollup.js:160:30)
[build:*js]     at ModuleLoader.handleResolveId (<repo>/node_modules/rollup/dist/shared/rollup.js:22394:24)
[build:*js]     at <repo>/node_modules/rollup/dist/shared/rollup.js:22357:26
[build:*js] 
[build:*js] npm run build:js exited with code 1
error Command failed with exit code 1.}
```

my _hunch_ is that rollup is trying to resolve everything before the `addon.hbs()` plugin has ran?
Is there a way to force plugin order?

"For fun", I added:
```
    nodeResolve({ extensions: ['.gjs', '.gts', '.js', '.ts', '.hbs'] }),
```
to the plugins list, but that gives this error:
```
[!] Error: 'hbs' is not exported by ../node_modules/ember-cli-htmlbars/lib/index.js, imported by src/components/demo/button.hbs
```
which kinda tells me that the `externals` bit from `addon.dependencies()` somehow isn't being used?

:thinking:

NullVoxPopuli:
When I remove `hbs` from publicEntrypoints and remove `nodeResolve`
```
❯ yarn build
yarn run v1.22.17
$ concurrently "npm:build:*"
[build:*js] 
[build:*js] > my-addon@0.0.0 build:js
[build:*js] > rollup -c ./config/rollup.config.js
[build:*js] 
[build:*js] 
[build:*js]  → dist...
[build:*js] [!] Error: You must supply options.input to rollup
[build:*js] Error: You must supply options.input to rollup
[...]

PR #1126, opened 2022-02-11
Resolve template-only components

NullVoxPopuli:
Accompanying repo for testing: https://github.com/NullVoxPopuli/ember-addon-v2-template-only-import

TODO
- [x] importing template-only component
- [x] template-only component included in app re-exeports

Resolves: https://github.com/embroider-build/embroider/issues/1121


packages/addon-dev/src/rollup-addon-dependencies.ts
```diff
@@ -33,7 +35,27 @@ export default function emberExternals(): Plugin {
...
+        *
+        * ./button
+        * ../button
+        *
+        */
+      if (isRelative && importer && !path.extname(source)) {
+        let dir = path.dirname(importer);
+        let target = path.resolve(dir, source);
+
+        let result = glob.sync(`${target}.*`);
```
lifeart, line 52: Sounds as expensive fs work :(

NullVoxPopuli: yeah, idk how to resolve template-only imports otherwise :sweat_smile:

ef4: But there's only one possible legal file extension for a template-only component. `.hbs`. You don't need a wildcard.

ef4:
Think of the `resolveId` as providing an extra fallback. 

In pseudocode, you need to do something like this where resolveId handles relative paths:

```js
resolveId(id) {
  // ...

  // for relative paths inside our own package, first try normal resolving.
  // This is calling https://www.rollupjs.org/guide/en/#thisresolve
  let result = this.resolve(source);
  if (result) {
    return result;
  }

  // only if that doesn't work, try to rewrite to .hbs
  result = this.resolve(source.replace(/\.js$/, '.hbs'));
  if (result) {
    return result;
  }
```

ef4:
Co-located templates (meaning app/components/*.hbs) are never available to the resolver. Compare with a classic app where the behavior is the same.

They always become an internal detail of the component JS module.

ef4:
> Or do you mean that, at the time the resolver becomes aware of anything, it's already been converted to JS?

Yes, that. If you have:

app/components/thing.js
app/templates/components/thing.hbs

Those are two separate entries in the runtime resolver that ember puts together after finding both at runtime. But if you have

app/components/thing.js
app/components/thing.hbs

Those become a single module from the perspective of the runtime resolver. And if you have

app/components/thing.hbs

It also becomes a single JS modules from ember's perspective.

> imported hbs files are now inlined -- which, imo, is a great thing to do, but it's not a safe thing to do as these components previously were still available to the globals resolver.

This is misunderstanding how things work. We already let rollup inline lots of things, including templates. That has nothing to do with what will be accessible to the resolver. That's controlled entirely by the appReexports, which in turn depend on the publicEntrypoints. `publicEntrypoints` is how you're telling rollup which things can't be inlined because they need to be addressable from the outside.

NullVoxPopuli:
> app/components/thing.js
> app/templates/components/thing.hbs

I forgot this option existed :sweat_smile: 

> It also becomes a single JS modules from ember's perspective.

or (with v2 addons), _will soon be_ (what I'm trying to figure out with this PR as well)

> hat's controlled entirely by the appReexports, which in turn depend on the publicEntrypoints. 

but it's all files based, yeah? like, you can't say "inline everything into one file, and then generate app re-exports for that", like,
```js
// dist/_app_/components/foo.js
export { Foo as default } from 'my-addon';

// dist/_app_/components/bar.js
export { Bar as default } from 'my-addon';
```

NullVoxPopuli:
Speaking of, I think I just did it :partying_face: 

https://github.com/NullVoxPopuli/ember-addon-v2-template-only-import
now has:
 - template-only component converted to JS
 - template-only component imported by co-located component
 - template-only component now in app re-exports

packages/addon-dev/src/rollup-public-entrypoints.ts
```diff
@@ -12,10 +13,28 @@ export default function publicEntrypoints(args: {
...
   return {
     name: 'addon-modules',
-    buildStart() {
-      for (let name of walkSync(args.srcDir, {
-        globs: args.include,
-      })) {
+    async buildStart() {
+      // Include *.hbs to add template-only files to publicEntrypoints
+      let matches = walkSync(args.srcDir, {
+        globs: [...args.include, 'components/**/*.hbs'],
```
ef4, line 19: This seems to be forcing every template-only component in `components/**/*.hbs` to be exposed a public entrypoint? We wouldn't want that.

I don't think we should be adding anything to the user's pattern. If the users wants to match template-only components they can give us a pattern that includes .hbs. We can make that the default in the example rollup config.

I would also prefer not to give the path "components/" any special behavior. We can treat every `.hbs` file as a component. Apps have other scenarios they need to worry about like route templates, but we don't ,and even if some addon author needs that they can customize their config.

packages/addon-dev/src/rollup-hbs-plugin.ts
```diff
@@ -19,5 +23,39 @@ export default function rollupHbsPlugin(): Plugin {
         id: id + '.js',
       };
     },
+
+    // template-only components may be imported
+    async resolveId(source, importer, options) {
+      if (!options.isEntry) {
```
ef4, line 29: I think a template-only component could be an entrypoint, which might make options.isEntry true. Can you test that case -- an addon that wants to use a template-only component in publicEntrypoints and appReexports?

packages/addon-dev/src/rollup-hbs-plugin.ts
```diff
@@ -19,5 +23,39 @@ export default function rollupHbsPlugin(): Plugin {
...
+        let isLocal = source.startsWith('.');
+        let hasExt = path.extname(source);
+
+        // imports with extensions will resolve automatically
+        if (!isLocal || hasExt) return;
```
ef4, line 34: I don't think `hasExt` is a good way to discriminate here. it's legal to say both `./components/foo` and `./component/foo.js`, and both should get the same handling (if the js doesn't exist and the hbs does, we synthesize the template-only component JS.)

ef4 (changes requested):
Where does this emit the call to `templateOnlyComponent()`? That is critical to generate correct template-only components.

The idea is that if somebody tries to load a nonexistent JS file beside an HBS file, we fill in the contents of the JS file with something like:

```js
import templateOnlyComponent from "@ember/component/template-only";
export default templateOnlyComponent();
```

That is probably sufficient as long as the colocation plugin also has a chance to run and handle that. Otherwise we would also need to synthesize the import of the template and the `setComponentTemplate` here.

ef4:
The linked demo app seems to misunderstand the intended semantics of hbs imports in the v2 addon spec. This:

https://github.com/NullVoxPopuli/ember-addon-v2-template-only-import/blob/2f31e515f66437702b58b287c2e3184aa9495ff6/my-addon/src/components/demo/index.js#L5

should not produce a component, it should be just the template. We need this both for compatibility and because we use it internally to bind together a component's JS and HBS. That is, if you author both foo.js and foo.hbs, the colocation inserts an `import TEMPLATE from './foo.hbs'; setComponentTemplate(TEMPLATE, whatever);` into `foo.js`. So an import ending in an explicit .hbs extension needs to mean "just the template", not a component with that template.

ef4:
I will add that having the demo repo to go with this is a very good idea and once we get it to demonstrates all the features we want it can move directly into this repo as a test suite.

packages/addon-dev/src/rollup-hbs-plugin.ts
NullVoxPopuli: yeah, I'll add something to the demo repo! thanks!

packages/addon-dev/src/rollup-hbs-plugin.ts
NullVoxPopuli: true true

packages/addon-dev/src/rollup-public-entrypoints.ts
NullVoxPopuli: removed!

packages/addon-dev/src/rollup-hbs-plugin.ts
```diff
@@ -1,23 +1,90 @@
...
+    },
   };
 }
+
+async function isTemplateOnly(hbsPath: string) {
+  let jsPath = hbsPath.replace(/\.hbs$/, '.js');
+  let tsPath = hbsPath.replace(/\.hbs$/, '.ts');
+
+  let hasJs = await pathExists(jsPath);
+  let hasTs = await pathExists(tsPath);
```
lifeart, line 64: looks like it could be paralleled
```js
let [hasJs, hasTs] = await Promise.all([pathExists(jsPath), pathExists(tsPath)]);
```

lifeart:
There is a possible ways to test resolveId - https://github.com/rollup/plugins/blob/master/packages/virtual/test/test.js#L7

packages/addon-dev/src/rollup-hbs-plugin.ts
```diff
@@ -1,23 +1,90 @@
...
+
+    // template-only components may be imported
+    async resolveId(source, importer, options) {
+      let resolution = await this.resolve(source, importer, {
+        ...options,
+        skipSelf: true,
+      });
+
+      let ext = path.extname(source);
+
```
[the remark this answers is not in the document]
NullVoxPopuli: oh cool, I didn't know about this API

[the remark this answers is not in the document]
NullVoxPopuli: dirname is on the `importer` tho, so I don't think this is useful in my case?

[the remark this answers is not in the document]
NullVoxPopuli: (source doesn't have the full path)

NullVoxPopuli:
> There is a possible ways to test resolveId

I was thinking testing a full build would be useful because the hbs plugin needs resolveId and resolve to work in concert with one another

packages/addon-dev/src/rollup-hbs-plugin.ts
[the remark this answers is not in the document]
lifeart: ah, sorry for confusion

tests/scenarios/v2-addon-dev-test.ts
```diff
@@ -0,0 +1,209 @@
+import path from 'path';
```
NullVoxPopuli, line 1: first addon-dev test! :tada:

tests/scenarios/v2-addon-dev-test.ts
```diff
@@ -0,0 +1,209 @@
...
+    Qmodule(scenario.name, function (hooks) {
+      let app: PreparedApp;
+      hooks.before(async () => {
+        app = await scenario.prepare();
+      });
+
+      Qmodule('The addon', function (hooks) {
+        let addonDir = '';
+        hooks.before(async () => {
+          debugger;
```
NullVoxPopuli, line 186: This is temporary, because I can't trust my work computer atm with the registry errors I'm seeing from the yarn install.

If there is a different way I should install deps with a fixturify project, lemme know

ef4: To add a dependency

1. Add it to tests/scenarios/package.json so that it will be visible to tests/scenarios/v2-addon-dev-test.ts. You can use aliasing to make multiple versions available as necessary.
2. Add it to a specific `project` using `project.linkDevDependency` or `project.linkDependency`.

tests/scenarios/v2-addon-dev-test.ts
```diff
@@ -0,0 +1,209 @@
...
+        './addon-main.js': './config/addon-main.cjs',
+      },
+      files: ['dist', 'config/addon-main.cjs'],
+      'ember-addon': {
+        type: 'addon',
+        version: 2,
+        main: './config/addon-main.cjs',
+        'app-js': {},
+      },
+      devDependencies: {
```
ef4, line 25: This is an unreliable way to add dependencies, you don't need to edit package.json manually, use `project.linkDevDependency` instead.

tests/scenarios/v2-addon-dev-test.ts
```diff
@@ -0,0 +1,209 @@
...
+        rollup: '^2.67.0',
+      },
+    };
+
+    let addon = new Project({
+      name: 'v2-addon',
+      files: {
+        'package.json': JSON.stringify(packageJSON, null, 2),
+        config: {
+          'addon-main.cjs': `
```
ef4, line 41: The PR I merged last night added a `baseV2Addon` utility. That would let you delete code like this that is provided by the base v2 addon template.

See for example https://github.com/embroider-build/embroider/blob/2a44b737e120f4159aa4731efd9a6484cd81fdb5/tests/scenarios/fastboot-app-test.ts#L31-L64

tests/scenarios/v2-addon-dev-test.ts
```diff
@@ -0,0 +1,209 @@
...
+      },
+    });
+
+    addon.linkDependency('@embroider/addon-shim', { baseDir: __dirname });
+    addon.linkDependency('@embroider/addon-dev', { baseDir: __dirname });
+
+    await addon.write();
+    let addonDir = addon.baseDir;
+
+    await execa('yarn', ['install'], { cwd: addonDir });
```
ef4, line 140: Most definitely do not do this. The entire reason scenario-tester exists is to avoid this. This is why we have `project.linkDependency` and `Project.fromDir(whatever, { linkDeps: true })`.

I realize there's no teaching materials about this, but as a general point of advice: to learn how a system like this works, read any other test. All of them need dependencies too, and none of them call yarn install to get them.

ef4 (changes requested):

tests/scenarios/v2-addon-dev-test.ts
NullVoxPopuli: it this part, I'm just wanting the addon's location on disk so I can assert that the package.json changed successfully.
I only have access to `scenario` and `app` at this location (of the relevant objects anyway), so once I get through your other suggestions, I'll see if there is a way to get that info

NullVoxPopuli: resolved by storing `addon` in module-scope for later access in the tests

tests/scenarios/v2-addon-dev-test.ts
NullVoxPopuli: updated

tests/scenarios/v2-addon-dev-test.ts
NullVoxPopuli: thanks! this really cut down the boilerplate

tests/scenarios/v2-addon-dev-test.ts
NullVoxPopuli: I learned that the node_modules are nested within the test app's node_modules, and it meant that `yarn` also couldn't be used to start rollup -- had to reach in to the package and run the bin directly

tests/scenarios/v2-addon-dev-test.ts
```diff
@@ -0,0 +1,185 @@
...
+      },
+    });
+  })
+  .forEachScenario(scenario => {
+    Qmodule(scenario.name, function (hooks) {
+      let app: PreparedApp;
+      hooks.before(async assert => {
+        app = await scenario.prepare();
+
+        await execa(path.join(addon.baseDir, 'node_modules/rollup/dist/bin/rollup'), ['-c', './rollup.config.mjs'], {
```
ef4, line 154: I think `addon.baseDir` here is probably messing you up. `forEachScenario` is going to run many times and each one gets a different set of input files, probably in entirely different temp directories. Trying to share the `addon` across all of them isn't likely to work.

This test needs a new feature that no test has needed yet: it wants to run a command in a directory that is not the app root, but instead the addon root. Normally tests can use `app.execute` and that isolates them from needing to worry about most of this. 

I think we should extend PreparedApp so you can conveniently execute in dependencies:

```js
await app.inDependendency('v2-addon').execute('yarn rollup -c ./rollup.config.mjs')
```

Short of having that feature, you can adjust the PreparedApp's dir instead:

```js
app.dir = dirname(require.resolve('v2-addon/package.json', { paths: [app.dir] }))
await app.execute('yarn rollup -c ./rollup.config.mjs');
```

NullVoxPopuli: why set app.dir? that feels incorrect because the app is the test app? 
I'm confused

NullVoxPopuli: I don't think the app.dir and execute trick actually builds the files. my package.json test now fails with that code.

NullVoxPopuli: Opened: https://github.com/ef4/scenario-tester/issues/5

packages/addon-dev/src/rollup-hbs-plugin.ts
```diff
@@ -1,23 +1,92 @@
...
+}
+
+function getTemplateOnly(hbsPath: string) {
+  let input = readFileSync(hbsPath, 'utf8');
+  let code =
+    `import { hbs } from 'ember-cli-htmlbars';\n` +
+    `import templateOnly from '@ember/component/template-only';\n` +
+    `import { setComponentTemplate } from '@ember/component';\n` +
+    `export default setComponentTemplate(\n` +
+    `hbs${backtick}${input.trim()}${backtick}, templateOnly());`;
```
rwjblue, line 80: Hmmm, I don't think we should `.trim()` here

packages/addon-dev/src/rollup-hbs-plugin.ts
```diff
@@ -1,23 +1,92 @@
...
+
+  return code;
+}
+
+function getTemplateFactory(hbsPath: string) {
+  let input = readFileSync(hbsPath, 'utf8');
+  let code =
+    `import { hbs } from 'ember-cli-htmlbars';\n` +
+    `export default hbs${backtick}${input.trim()}${backtick};`;
```
rwjblue, line 89: ✂️  `.trim()`

tests/scenarios/package.json
```diff
@@ -4,6 +4,7 @@
   "dependencies": {
     "@embroider/shared-internals": "1.5.0",
     "@types/qunit": "^2.11.1",
+    "execa": "^5.0.0",
```
rwjblue, line 7: Should this be a devDep?

packages/addon-dev/src/rollup-hbs-plugin.ts
NullVoxPopuli: removed, thanks! <3

packages/addon-dev/sample-rollup.config.js
```diff
@@ -1,4 +1,4 @@
-import babel from '@rollup/plugin-babel';
+import { babel } from '@rollup/plugin-babel';
```
rwjblue, line 1: I think these are the same?

https://github.com/rollup/plugins/blob/5363f55aa1933b6c650832b08d6a54cb9ea64539/packages/babel/src/index.js#L258-L260

NullVoxPopuli: I forgot about this -- I had made the change because the default export was undefined -- and I "just wanted to focus on the important bits" of what I was working on and not deal with sillyness. But yeah, I forgot to check if this was actually needed when I got the tests passing locally.

Checking now tho (locally):
 - changing the import in v2-addon-dev-test back to use the default import
 -  default import reveals: `TypeError: babel is not a function` -- which is suspicious, because your link points to the same value for both named and default exports....
    - my next hunch is that the embroider-monorepoe's typescript setup is doing dirty things to default imports/exports due to the compile target not being ESM

packages/addon-dev/src/rollup-public-entrypoints.ts
```diff
@@ -12,10 +13,12 @@ export default function publicEntrypoints(args: {
...
-    buildStart() {
-      for (let name of walkSync(args.srcDir, {
-        globs: args.include,
-      })) {
+    async buildStart() {
+      let matches = walkSync(args.srcDir, {
+        globs: [...args.include],
+      });
+
+      for (let name of matches) {
```
rwjblue, line 21: I think we need to make this aware of colocation. Right now, I _think_ we'll end up calling `this.emitFile` for both the `src/components/foo.js` and `src/components/foo.hbs`.

tests/scenarios/v2-addon-dev-test.ts
```diff
@@ -0,0 +1,227 @@
+import path from 'path';
```
NullVoxPopuli, line 1: Be sure to test this scenario with peer deps: https://github.com/embroider-build/embroider/issues/1179#issuecomment-1090462587

NullVoxPopuli:
This is rebased, and my little demo repo works.

example output:
```js
import { _ as _applyDecoratedDescriptor, a as _defineProperty, b as _initializerDefineProperty } from '../../_rollupPluginBabelHelpers-2eca8644.js';
import templateOnly from '@ember/component/template-only';
import Out from './out.js';

var TEMPLATE = hbs("Hello there!\n\n<this.Out>{{this.active}}</this.Out>\n\n<this.Button @onClick={{this.flip}} />\n<this.Button2 @onClick={{this.flip}} />\n");

var BlahButton = setComponentTemplate(hbs`<button {{on 'click' @onClick}}>
  flip
</button>

`, templateOnly());

var _class, _descriptor;
let Demo = (_class = class Demo extends Component {
  constructor(...args) {
    super(...args);

    _defineProperty(this, "Button", BlahButton);

    _defineProperty(this, "Button2", BlahButton);

    _defineProperty(this, "Out", Out);

    _initializerDefineProperty(this, "active", _descriptor, this);

    _defineProperty(this, "flip", () => this.active = !this.active);
  }

}, (_descriptor = _applyDecoratedDescriptor(_class.prototype, "active", [tracked], {
  configurable: true,
  enumerable: true,
  writable: true,
  initializer: function () {
    return false;
  }
})), _class);

export { Demo as default };

```

The important thing I was trying to make sure happened:
 - imports of the same template-only module are de-duped
 - template-only modules acquire setComponentTemplate + templateOnly()
 
 So, success :D

packages/addon-dev/src/rollup-public-entrypoints.ts
NullVoxPopuli: emit files is a no go cause it puts files in the output, and isn't compat with what we want (at least in my debugging) -- I added a bunch of comments around the current implementation

NullVoxPopuli:
I've left the execa usage, because:

 - I don't know how to 
    - get an addon project from a PreparedApp (so that I can call execute)
   
I don't think this sort of problem should be a blocker, because seems like we need to add behaviors to scenario-tester
 - add some way of getting a linked project from a PreparedApp
   ([only APIs right now](https://github.com/ef4/scenario-tester/blob/main/index.ts#L170) are `dir` and `execute`)

packages/addon-dev/sample-rollup.config.js
NullVoxPopuli: reverted -- unrelated to this work anyway. If we need to update the sample, that should be separate PR

tests/scenarios/package.json
NullVoxPopuli: yes! thanks

NullVoxPopuli: just kidding, this is required to be a dep, due to lints

packages/addon-dev/src/rollup-hbs-plugin.ts
```diff
@@ -1,22 +1,64 @@
...
         ...options,
       });
 
-      const id = resolution?.id;
+      let id = resolution?.id;
 
-      if (!filter(id)) return null;
+      // if id is undefined, it's possible we're importing a file that that rollup
+      // doesn't natively support such as a template-only component that the author
+      // doesn't want to be available on the globals resolver
```
ef4, line 24: This comment is misleading. The globals resolver isn't really relevant to this. You control global resolving by deciding whether or not to put a module into appReexports. But this code runs for your modules whether or not they're in appReexports.

tests/scenarios/v2-addon-dev-test.ts
```diff
@@ -0,0 +1,214 @@
...
+          destDir: 'dist',
+        });
+
+        export default {
+          output: addon.output(),
+
+          plugins: [
+            addon.publicEntrypoints([
+              '**/*.js',
+              'components/demo/out.hbs',
```
ef4, line 60: Similar to the meaning in imports, we need a way to talk about template files, so people need to be allowed to put explicit `.hbs` here like this. But that can't mean the same thing as the template-only component. The template-only component is a JS file -- even if that JS file is synthesized by the build, rather than typed by the author.

ef4 (changes requested):

ef4 (approved):
@NullVoxPopuli thanks for pushing this forward, and especially for getting good tests in place. I was able to sort out the implementation.

tests/scenarios/v2-addon-dev-test.ts
```diff
@@ -225,3 +207,8 @@ appScenarios
       });
     });
   });
+
+// https://github.com/ef4/scenario-tester/issues/5
+function inDependency(app: PreparedApp, dependencyName: string): PreparedApp {
```
NullVoxPopuli, line 208: Ah, this makes sense, now that i see it 🤦

packages/addon-dev/src/rollup-hbs-plugin.ts
```diff
@@ -17,111 +19,109 @@ export default function rollupHbsPlugin(): Plugin {
...
+      ...options,
+    }
+  );
+  if (!templateResolution) {
+    return null;
+  }
+  // we're trying to resolve a JS module but only the corresponding HBS
+  // file exists. Synthesize the template-only component JS.
+  return {
+    id: templateResolution.id.replace(/\.hbs$/, '.js'),
```
NullVoxPopuli, line 100: This is way simpler than what i was doing 🤔

tests/scenarios/v2-addon-dev-test.ts
```diff
@@ -56,13 +56,12 @@ appScenarios
 
           plugins: [
             addon.publicEntrypoints([
-              '**/*.js',
-              'components/demo/out.hbs',
+              'components/**/*.js',
```
NullVoxPopuli, line 59: Should we throw on non-js extensions in here and is app-re-exports?

packages/addon-dev/src/rollup-public-entrypoints.ts
```diff
@@ -15,15 +19,37 @@ export default function publicEntrypoints(args: {
...
+          // match the user's patterns.
+          let normalizedName = normalizeFileExt(name);
+          let id = join(args.srcDir, normalizedName);
+          if (
+            args.include.some((pattern) =>
+              minimatch(normalizedName, pattern)
+            ) &&
+            !pathExistsSync(id)
+          ) {
+            this.emitFile({
```
NullVoxPopuli, line 46: Ah,. This was probably what i was missing (and maybe all the changes in this file).


Because i was so focused on the other plugin, i didn't remember how in interacts with this plugin.

PR #1134, opened 2022-02-19
Solve ts related double setComponentTemplate

NullVoxPopuli:
Depends on: https://github.com/embroider-build/embroider/pull/1126
(Only because my reproduction repos cover both TS and template-only)

Reproduction repo: https://github.com/NullVoxPopuli/ember-addon-v2-typescript-demo/

This PR aims to solve the problem discovered here: https://github.com/embroider-build/embroider/pull/1099
Where _prior_ to the changes in this PR, component output for TS files (using rollup-plugin-ts) included _two_ `setComponentTemplate` calls, such as: 
```js
import { __decorate } from 'tslib';

var TEMPLATE = setComponentTemplate(TEMPLATE, hbs`Hello there!

<out>{{this.active}}</out>

<button {{on 'click' this.flip}}>
  flip
</button>
`);

class Demo extends Component {
  constructor(...args) {
    super(...args);
    this.active = false;

    this.flip = () => this.active = !this.active;
  }

}

__decorate([tracked], Demo.prototype, "active", void 0);

export { Demo as default };
//# sourceMappingURL=index.js.map
```

The line that has:
```js
var TEMPLATE = setComponentTemplate(TEMPLATE, hbs`Hello there!
```
erroneously has setComponentTemplate doing a self-reference, which wouldn't cause a an error during app boot, but _would_ cause additional boot time for apps (the less to do during boot the bett)


packages/addon-dev/src/rollup-hbs-plugin.ts
```diff
@@ -1,23 +1,90 @@
...
+    `import { hbs } from 'ember-cli-htmlbars';\n` +
+    `import templateOnly from '@ember/component/template-only';\n` +
+    `import { setComponentTemplate } from '@ember/component';\n` +
+    `export default setComponentTemplate(\n` +
+    `hbs${backtick}${input.trim()}${backtick}, templateOnly());`;
+
+  return code;
+}
+
+function getTemplateFactory(hbsPath: string) {
```
SergeAstapov, line 83: @NullVoxPopuli am I missing something or looks like this does not actually solve the problem with double `setComponentTemplate` call?

NullVoxPopuli: Correct. The diff is goofy becauso the upsteam PR that handles template only components hass't been merged

NullVoxPopuli:
It looks like the latest changes to how co-located templates and template-only components are handled fixed this! yay!

There is another issue I ran in to at runtime about `ember-cli-htmlbars` not being found -- so I'm going to start with other PR with a failing test for that (if I can get a reproduction figured out)

Issue #1132, opened 2022-02-18
`Build Error (PackagerRunner) in node_modules\@glimmer\tracking\index.js`


mkszepp:
Ember: 4.1.0
Embroider 1.2.0
OS: Win 10

Sometime after file changes the build fails.

The only fix is to stop and restart ember serve.
This bug cames randomly after x file changes

We use `ember-cached-decorator-polyfill` because `@cached` doesn't work with the latest embroider version

Maybe related: https://github.com/embroider-build/embroider/issues/1086#issuecomment-1044028028

```
[@embroider/webpack]assets by status 11.5 MiB [cached] 6 assets
Entrypoint assets/my-app.js [big] 9.96 MiB = chunk.a014068d97d15b716a73.js 6.45 MiB chunk.738f12f02191cd08114e.js 3.5 MiB chunk.a38afc09352e8ed55d0b.js 13.1 KiB
Entrypoint assets/test.js [big] 10.9 MiB = chunk.a014068d97d15b716a73.js 6.45 MiB chunk.2522794c3224ecc4d25b.js 341 KiB chunk.738f12f02191cd08114e.js 3.5 MiB chunk.e198f43b9a52d72b6f1e.js 636 KiB
cached modules 7.8 MiB (javascript) 14.4 KiB (runtime) [cached] 1831 modules
javascript modules 78 bytes
  ./node_modules/@glimmer/tracking/index.js 39 bytes [built] [1 error]
  ./node_modules/@glimmer/tracking/primitives/cache.js 39 bytes [built] [1 error]

ERROR in ./node_modules/@glimmer/tracking/index.js
Module build failed (from ../../../../../../../Projects/rg/my-app/node_modules/thread-loader/dist/cjs.js):
Thread Loader (Worker 0)
ENOENT: no such file or directory, open '$TMPDIR\embroider\a1864e\node_modules\@glimmer\tracking\index.js'
    at PoolWorker.fromErrorObj (C:\Projects\rg\my-app\node_modules\thread-loader\dist\WorkerPool.js:346:12)    at C:\Projects\rg\my-app\node_modules\thread-loader\dist\WorkerPool.js:219:29
    at mapSeries (C:\Projects\rg\my-app\node_modules\neo-async\async.js:3625:14)
    at PoolWorker.onWorkerMessage (C:\Projects\rg\my-app\node_modules\thread-loader\dist\WorkerPool.js:173:34)
[...]

PR #1135, opened 2022-02-19
Fix @cached decorator export from fake @glimmer/tracking module

NullVoxPopuli:
Discovered during: https://github.com/embroider-build/embroider/issues/1132#issuecomment-1045280240

(I put `@cached` in the wrong spot :( )

### DIFF
--- a/packages/compat/src/compat-adapters/@glimmer/tracking.ts
+++ b/packages/compat/src/compat-adapters/@glimmer/tracking.ts
@@ -17,15 +17,17 @@ class RedirectToEmber extends Plugin {
   build() {
     if (!this.didBuild) {
       copyFileSync(join(this.inputPaths[0], 'package.json'), join(this.outputPath, 'package.json'));
-      outputFileSync(join(this.outputPath, 'index.js'), `export { tracked } from '@ember/-internals/metal';`);
       outputFileSync(
-        join(this.outputPath, 'primitives', 'cache.js'),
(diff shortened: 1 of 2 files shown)


lifeart (approved):

lifeart:
@NullVoxPopuli looks like we could use https://github.com/embroider-build/embroider/blob/main/packages/compat/tests/route-split.test.ts as reference for test, in my understanding we could have minimal ember app with cached decorator usage and it's should pass build

NullVoxPopuli:
:upside_down_face: 
image

NullVoxPopuli:
Test success:
image

tests/scenarios/scenarios.ts
```diff
@@ -39,7 +39,7 @@ async function lts_3_28(project: Project) {
   project.linkDevDependency('ember-data', { baseDir: __dirname, resolveName: 'ember-data' });
 }
 
-async function release(project: Project) {
+export async function release(project: Project) {
```
ef4, line 42: Instead of this, you can say `appScenarios.only('release')`

ef4 (approved):

tests/scenarios/scenarios.ts
NullVoxPopuli: ooo that's handy, because I have other tests I'm working on that need just the release bit. thanks!

NullVoxPopuli: Updated!

PR #1137, opened 2022-02-23
Update `Status` section in main README

bertdeblock:
Maybe this entire section could use a bit of an update, but I thought this part could already go as we are at v1.2


SergeAstapov (approved):

ef4:
Thanks, this makes sense.

PR #1140, opened 2022-02-24
Fastboot lazy css support

ef4:
This fixes a crash that can occur if an app tries to lazily load CSS in fastboot.

This change works by making all the lazy CSS eager. This is strictly better than being broken. Further work can follow to make the lazy chunks actually lazy in fastboot, such that we detect during rendering which ones were wanted and only insert link tags for those ones.

Apps that don't use fastboot should be unaffected.


ef4:
This is working as intended, but the test failures are places where we already had tests for lazy CSS in engines that have now been forced to be eager. 

Next step it to investigate how those were ever working -- it's possible they weren't and the tests simply didn't hit the bad cases.

ef4:
I confirmed that the lazy engines CSS case was actually broken all along in fastboot, and the test wasn't running properly in CI before.

So I'm adjusting the tests to guard that case and only assert that lazy engine CSS is lazy in non-fastboot apps.

PR #1145, opened 2022-02-26
ensure addon app-js trees are merged in the correct order

eoneill:
This resolves a compatibility issue in which the order that `addon.app-js` trees are merged is not consistent with classic builds. The fix here is to sort the addons by the `order-index` (if present).

I have yet to be able to reproduce a test case of the issue, but can confirm this resolves the issue discovered in our app.


ef4:
Nice.

This fix might want to move to a slightly different place because there's already code that was *attempting* to do this correctly:

https://github.com/embroider-build/embroider/blob/e9bbbadf5721ba754ec06af93985c3785593ffcb/packages/core/src/app.ts#L596-L607

https://github.com/embroider-build/embroider/blob/e9bbbadf5721ba754ec06af93985c3785593ffcb/packages/compat/src/compat-app.ts#L186-L199

Issue #1049, opened 2021-12-11
How to lazy load CSS with FastBoot


simonihmig:
Somewhat related to #1033, I played with importing CSS (Sass) from JS. This seems to work quite fine with a classic Ember SPA, but not so much when FastBoot comes into play. 

First thing I had to do was to use `mini-css-extract-plugin` instead of `style-loader`, even in development mode, as `style-loader` would immediately break FastBoot by using not supported DOM APIs. 

<details>
<summary>Stack trace </summary>

```
ReferenceError: document is not defined

at insertStyleElement (node_modules/style-loader/dist/runtime/injectStylesIntoStyleTag.js?:93:15)
at addStyle (node_modules/style-loader/dist/runtime/injectStylesIntoStyleTag.js?:208:13)
at modulesToDom (node_modules/style-loader/dist/runtime/injectStylesIntoStyleTag.js?:81:18)
at module.exports (node_modules/style-loader/dist/runtime/injectStylesIntoStyleTag.js?:239:25)
at eval (webpack:/app-template/components/lazy-css.css?:16:156)
...
```
</details>

Now with using only `mini-css-extract-plugin`, and with #1048 also applied, loading the index route of a FastBoot powered app that imports CSS from `app.js` does work correctly! I.e. the imported CSS has a link in the prerendered HTML (with #1045 applied even in the correct place). However when I have for example lazy loaded routes, and those import CSS (in `route.js`, or some component not previously used), then `mini-css-extract-plugin` will then (correctly) try to dynamically inject the `<link>` for that CSS chunk [here](https://github.com/webpack-contrib/mini-css-extract-plugin/blob/master/src/index.js#L666-L685), however that fails when this runs in the FastBoot VM.

<details>
<summary>Stack trace </summary>

```
ReferenceError: document is not defined
    at findStylesheet (/var/folders/q3/yzggf_l965zbzps_bk8xr37c0000gn/T/broccoli-34347pIJ0N0jVFeLE/out-155-packager_runner_embroider_webpack/assets/chunk.ab6b31dc622375f2f6d5.js:255:36)
    at /var/folders/q3/yzggf_l965zbzps_bk8xr37c0000gn/T/broccoli-34347pIJ0N0jVFeLE/out-155-packager_runner_embroider_webpack/assets/chunk.ab6b31dc622375f2f6d5.js:272:17
    at new Promise (<anonymous>)
[...]

PR #1146, opened 2022-02-28
Use MiniCssExtractPlugin for FastBoot builds

simonihmig:
> First: In apps with fastboot, `@embroider/webpack` should automatically use MiniCssExtractPlugin even in development (as opposed to style-loader, which is fine in development for apps that don't have fastboot). ember-auto-import [already has this policy](https://github.com/ef4/ember-auto-import/blob/67a8607369030b5b811a97d2e025f5e7c7fb7c4b/packages/ember-auto-import/ts/webpack.ts#L224) and embroider should do the same. This would be a great well-scoped PR.

https://github.com/embroider-build/embroider/issues/1049#issuecomment-1034079882

### DIFF
--- a/packages/compat/src/audit.ts
+++ b/packages/compat/src/audit.ts
@@ -257,7 +257,7 @@ export class Audit {
     resolver.enableAuditMode();
 
     let compile = applyVariantToTemplateCompiler(
-      { name: 'default', runtime: 'all', optimizeForProduction: false },
+      { name: 'default', runtime: 'all', optimizeForProduction: false, hasFastBoot: false },
       templateCompiler.compile
     );
     return { compile, resolver };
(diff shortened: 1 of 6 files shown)


ef4:
Thanks!

simonihmig:
Thanks for tweaking the PR with [this commit](https://github.com/embroider-build/embroider/pull/1146/commits/de5adccaa2de915d71d58179651f6cc6770c9eff), reverting my choice to add an additional `hasFastBoot` flag to `Variant`. I was initially hesitant to add this, but I thought we would need to disambiguate a `browser` build with and without FastBoot. But you are right, given how the combinations of `runtime` and `optimizeForProduction` are created, that flag was actually redundant. Good catch then, I should have looked into it more deeply!

ef4:
Well none of this is well documented of you to discover the original intent, so it's very understandable!

But yes, my intention for "runtime" is that it should cover this case.

PR #1147, opened 2022-02-28
ensure addon app-js trees are merged in the correct order

ef4:
Building on #1145, with the sorting moved to a slightly more general spot.


ef4:
Superseded by work in https://github.com/embroider-build/embroider/pull/1145

PR #1149, opened 2022-03-03
Update externals stubs atomically

ef4:
Since babel can run in many workers, and since we write the external stubs as a side-effect of discovering that they're needed from within babel, it's possible that a stub is written at the exact moment another worker is reading it. So we need to make sure our writes are atomic.

### DIFF
--- a/packages/core/src/babel-plugin-adjust-imports.ts
+++ b/packages/core/src/babel-plugin-adjust-imports.ts
@@ -4,11 +4,12 @@ import type { NodePath } from '@babel/traverse';
 import type * as Babel from '@babel/core';
 import type { types as t } from '@babel/core';
 import { PackageCache, Package, V2Package, explicitRelative } from '@embroider/shared-internals';
-import { outputFileSync } from 'fs-extra';
 import { Memoize } from 'typescript-memoize';
 import { compile } from './js-handlebars';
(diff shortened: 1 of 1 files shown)


ef4:
Ugh, yes, windows doesn't have atomic rename. 😭

packages/core/src/babel-plugin-adjust-imports.ts
```diff
@@ -353,6 +354,24 @@ function makeExternal(specifier: string, sourceFile: AdjustFile, opts: Options):
...
+    return;
+  }
+  let suffix = randomBytes(8).toString('hex');
+  outputFileSync(path + suffix, content);
+  try {
+    renameSync(path + suffix, path);
+  } catch (err: any) {
+    // windows throws EPERM for concurrent access. For us it's not an error
+    // condition because the other thread is writing the exact same value we
+    // would have.
```
lifeart, line 368: Thinking about async func here with few retry attempts. I think it’s important to do job, instead of fail

ef4: If we hit EPERM here it's only because another process is already doing the exact same job we were trying to do, so we can move on and let that process take care of it.

As far as I can tell, that's the only way to get EPERM here. Since this is inside our own temp dir, you shouldn't ever have unrelated permission failures in the usual Unix EPERM sense.

PR #1150, opened 2022-03-05
Make the helper keyword handling less strict

Windvis:
The current implementation throws an error if you try to curry helpers using the helper keyword. This reduces the strictness of the resolver since Ember itself catches the cases were users provide unsupported dynamic values to the helper keyword.

This PR is a continuation of the discussion in the previous PR: https://github.com/embroider-build/embroider/pull/1120#discussion_r816763874

### DIFF
--- a/packages/compat/src/resolver-transform.ts
+++ b/packages/compat/src/resolver-transform.ts
@@ -329,15 +329,12 @@ function handleComponentHelper(
   resolver.resolveComponentHelper(locator, moduleName, param.loc, impliedBecause);
 }
 
-function handleDynamicHelper(param: ASTv1.Node, resolver: Resolver, moduleName: string): void {
-  switch (param.type) {
-    case 'StringLiteral':
-      resolver.resolveDynamicHelper({ type: 'literal', path: param.value }, moduleName, param.loc);
-      break;
-    case 'TextNode':
-      resolver.resolveDynamicHelper({ type: 'literal', path: param.chars }, moduleName, param.loc);
-      break;
-    default:
-      resolver.resolveDynamicHelper({ type: 'other' }, moduleName, param.loc);
+function handleDynamicHelper(param: ASTv1.Expression, resolver: Resolver, moduleName: string): void {
+  // We only need to handle StringLiterals since Ember already throws an error if unsupported values
+  // are passed to the helper keyword.
+  // If a helper reference is passed in we don't need to do anything since it's either the result of a previous
+  // helper keyword invocation, or a helper reference that was imported somewhere.
+  if (param.type === 'StringLiteral') {
+    resolver.resolveDynamicHelper({ type: 'literal', path: param.value }, moduleName, param.loc);
   }
 }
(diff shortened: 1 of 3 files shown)


packages/compat/tests/resolver.test.ts
```diff
@@ -519,7 +519,7 @@ describe('compat-resolver', function () {
       },
     ]);
   });
-  test('string literal passed to `helper` helper in content position', function () {
+  test('string literal passed to "helper" keyword in content position', function () {
```
Windvis, line 522: I started renaming "helper" to "keyword" since that's the word they use in the error in ember-source:

> Uncaught Error: Assertion Failed: Passing a dynamic string to the `(helper)` keyword is disallowed. (You specified `(helper (concat "hel" "lo"))` and `(concat "hel" "lo")` evaluated into "hello".) This ensures we can statically analyze the template and determine which helpers are used. If the helper name is always the same, use a string literal instead, i.e. `(helper "hello")`. Otherwise, import the helpers into JavaScript and pass them to the helper keyword. See https://github.com/emberjs/rfcs/blob/master/text/0496-handlebars-strict-mode.md#4-no-dynamic-resolution for details. ('contextual-built-ins/templates/application.hbs' @ L39:C0)

packages/compat/tests/resolver.test.ts
```diff
@@ -681,14 +681,35 @@ describe('compat-resolver', function () {
...
     ]);
   });
+  test('helper currying using the "helper" keyword', function () {
+    let findDependencies = configure({ staticHelpers: true });
+    givenFile('helpers/hello-world.js');
+    expect(
+      findDependencies(
+        'templates/application.hbs',
+        `
+        {{#let (helper "hello-world" name="World") as |hello|}}
```
Windvis, line 691: This is based on the code snippet in the [RFC](https://github.com/emberjs/rfcs/blob/master/text/0432-contextual-helpers.md).

packages/compat/src/resolver-transform.ts
```diff
@@ -329,15 +329,12 @@ function handleComponentHelper(
...
-  switch (param.type) {
-    case 'StringLiteral':
-      resolver.resolveDynamicHelper({ type: 'literal', path: param.value }, moduleName, param.loc);
-      break;
-    case 'TextNode':
-      resolver.resolveDynamicHelper({ type: 'literal', path: param.chars }, moduleName, param.loc);
-      break;
-    default:
-      resolver.resolveDynamicHelper({ type: 'other' }, moduleName, param.loc);
+function handleDynamicHelper(param: ASTv1.Expression, resolver: Resolver, moduleName: string): void {
```
Windvis, line 332: I'm new to TypeScript, but I think `Expression` is the correct type here since `node.params` seems to be typed as `Expression[]`? `Expression` doesn't include the `TextNode` type, so that doesn't need to be covered anymore.

packages/compat/tests/resolver.test.ts
ef4: Yes, I agree, it seems clearer to say "keyword"

PR #1151, opened 2022-03-06
Add support for the modifier keyword

Windvis:
The modifier counterpart of #1120

Once this PR and #1150 are merged, #1018 should be fully resolved.


ef4:
Nice!

PR #1155, opened 2022-03-07
Turn ensureSafeComponent into a no-op for class values on Ember 3.25+

### LINKED ISSUE #1061: ensureSafeComponent can be a no-op for classes on newer embers
Starting with the ember-source 3.25 beta series, it became safe to invoke component classes directly with angle bracket invocation.

We should update to ensureSafeComponent so it leaves classes alone on those versions, instead of converting them to curried component definitions. 

There's no bug here, it's just unnecessary work we can optimize away.

Windvis:
Closes #1061


rwjblue (approved):

ef4:
Very nice!

PR #1156, opened 2022-03-08
fix exception in macros babel plugin

ef4:
This fixes a crash in the macros babel plugin that could be triggered by a particular combination of SpreadElement and OptionalMemberExpression.

### DIFF
--- a/packages/macros/src/babel/evaluate-json.ts
+++ b/packages/macros/src/babel/evaluate-json.ts
@@ -222,6 +222,9 @@ export class Evaluator {
 
     if (path.isObjectExpression()) {
       let props = assertArray(path.get('properties')).map(p => {
+        if (p.isSpreadElement()) {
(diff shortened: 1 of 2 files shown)


ef4:
For searchability, the failure looks like:

```
TypeError: /path/to/some.js: Cannot read property 'type' of undefined
      201 |     // Here we are glossing over the lack of a real OptionalMemberExpression type
      202 |     // in our @babel/traverse typings.
    > 203 |     if (path.node.type === 'OptionalMemberExpression') {
          |                   ^
      204 |       return this.evaluateMember(path as NodePath<t.OptionalMemberExpression>, true);
      205 |     }
      206 |

      at Evaluator.realEvaluate (src/babel/evaluate-json.ts:203:19)
      at Evaluator.evaluate (src/babel/evaluate-json.ts:187:23)
```

PR #1161, opened 2022-03-11
Remove outdated core and compat options

bertdeblock:
I couldn't find these options in the corresponding interfaces or throughout
the rest of the codebase, so I assumed these might be leftovers?


packages/core/src/options.ts
```diff
@@ -112,10 +112,7 @@ export function optionsWithDefaults(options?: Options): Required<Options> {
     staticHelpers: false,
     staticModifiers: false,
     staticComponents: false,
-    packageRules: [],
```
bertdeblock, line 115: Not 100% sure about this one, but it isn't defined in the interface above and it's also [a compat option](https://github.com/embroider-build/embroider/blob/main/packages/compat/src/options.ts#L85).

ef4:
Thanks, I hadn't even noticed that these were lingering.

For those following along at home:
 - `optionalComponents` is covered instead by `safeToIgnore` inside `packageRules`.
 - `packageRules` is still fully-supported, it's just that it's an option to `@embroider/compat`, not `@embroider/core`. You're almost certainly using compat anyway.
 - `splitControllers` and `splitRouteClasses` are subsumed by `splitAtRoutes`, which necessary must handle controllers, route classes, and route templates together.

PR #1164, opened 2022-03-14
Update `master` to `main` in links

bertdeblock:
Avoids the `Branch master was renamed to main.` info alert when clicking on any of these links.


ef4:
Thanks

PR #1171, opened 2022-03-29
FIX: rollup-hbs-plugin add resolveId hook

wondersloth:
- In the `load()` hook, `id` isn't a valid property in the returned object.
- This assumed that a resolution was created for the intermediate
  file `*.hbs.js` file.
- This created an issue where a babel plugin could not transform
  `*.hbs.js` files.
- Added `resolveId()` hook to create a resolution for `*.hbs` to `*.hbs.js` file.

### DIFF
--- a/packages/addon-dev/src/rollup-hbs-plugin.ts
+++ b/packages/addon-dev/src/rollup-hbs-plugin.ts
@@ -4,8 +4,7 @@ import { readFileSync } from 'fs';
 const backtick = '`';
 
 export default function rollupHbsPlugin(): Plugin {
-  const templateFilter = createFilter('**/*.hbs');
-  const moduleFilter = createFilter('**/*.hbs.js');
+  const filter = createFilter('**/*.hbs');
 
   return {
     name: 'rollup-hbs-plugin',
@@ -17,7 +16,7 @@ export default function rollupHbsPlugin(): Plugin {
 
       const id = resolution?.id;
 
-      if (!templateFilter(id)) return null;
+      if (!filter(id)) return null;
 
       // This creates an `*.hbs.js` that we will populate in `load()` hook.
       return {
@@ -31,8 +30,6 @@ export default function rollupHbsPlugin(): Plugin {
       };
     },
     load(id: string) {
-      if (!moduleFilter(id)) return null;
-
       const meta = this.getModuleInfo(id)?.meta;
       const originalId = meta?.['rollup-hbs-plugin']?.originalId;
 
--- a/packages/addon-dev/src/rollup-hbs-plugin.ts
+++ b/packages/addon-dev/src/rollup-hbs-plugin.ts
@@ -4,19 +4,48 @@ import { readFileSync } from 'fs';
 const backtick = '`';
 
 export default function rollupHbsPlugin(): Plugin {
-  const filter = createFilter('**/*.hbs');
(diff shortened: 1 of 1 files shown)


packages/addon-dev/src/rollup-hbs-plugin.ts
```diff
@@ -4,19 +4,48 @@ import { readFileSync } from 'fs';
...
     name: 'rollup-hbs-plugin',
+    async resolveId(source: string, importer: string | undefined, options) {
+      const resolution = await this.resolve(source, importer, {
+        skipSelf: true,
+        ...options,
+      });
+
+      const id = resolution?.id;
+
+      if (!templateFilter(id)) return null;
```
rwjblue, line 20: I think we can move this up to be first (no need to `this.resolve` if we are just going to return null)

packages/addon-dev/src/rollup-hbs-plugin.ts
```diff
@@ -4,19 +4,48 @@ import { readFileSync } from 'fs';
...
-      if (!filter(id)) return;
-      let input = readFileSync(id, 'utf8');
+      if (!moduleFilter(id)) return null;
+
+      const meta = this.getModuleInfo(id)?.meta;
+      const originalId = meta?.['rollup-hbs-plugin']?.originalId;
+
+      if (!originalId) {
+        return;
+      }
```
rwjblue, line 41: I think we can remove the `moduleFilter` conditional, because the thing we _actually_ care about is that the meta contains the original id path info (and we return early anyways).

```suggestion
      const meta = this.getModuleInfo(id)?.meta;
      const originalId = meta?.['rollup-hbs-plugin']?.originalId;

      if (!originalId) {
        return;
      }
```

rwjblue (commented):
Thanks for working on this @wondersloth!

@ef4 / @NullVoxPopuli - The issue that @wondersloth is fixing here is the root cause of why `.hbs` files themselves aren't going through babel transpilation (`rollup-plugin-babel` only looks at `.js` files, `load` isn't allowed to return a new `id`, so babel was just skipping these `.hbs` files completely).

packages/addon-dev/src/rollup-hbs-plugin.ts
rwjblue, line 20:
```suggestion
      if (!templateFilter(id)) return null;

      const resolution = await this.resolve(source, importer, {
        skipSelf: true,
        ...options,
      });

      const id = resolution?.id;
```

wondersloth: We can't move it up because we don't have an `id` yet.

wondersloth: Dropping this because we don't have `id` at that point.

packages/addon-dev/src/rollup-hbs-plugin.ts
wondersloth: Updated my PR.

packages/addon-dev/src/rollup-hbs-plugin.ts
rwjblue: Ah! Thank you!

rwjblue (approved):

ef4 (approved):

PR #1177, opened 2022-04-04
Write files plugin

krisselden:
this moves the post build stuff into webpack, this is also a step toward running webpack out of process (this stuff can be separated more but this PR is trying to make the diff smaller).

Since this moves one post process into 2 compilers, this adds some microtask concurrency, this makes sure the has() guards synchronously set() after has() fails and it makes both sides get to the html entrypoint check, then only has the variant 0 emit the html entrypoint.


krisselden:
Seems to be something wrong with the fastboot variant stuff.

rwjblue (approved):

PR #1187, opened 2022-04-10
Add test for app depending on an addon depending on an optional peerDependency

NullVoxPopuli:
This test was meant to reproduce the issue I'm running in to here:
 - https://github.com/NullVoxPopuli/ember-resources/runs/5961868348?check_suite_focus=true#step:7:459

but everything is fine and dandy.

Could this be pnpm (the ember-resources repo) vs yarn (embroider) issue? idk

Locally, on the ember-resources PR, I did the require.resolve trick to make sure that the package is resolvable to node:
```
❯ node -e "console.log(require.resolve('ember-async-data'))"
<repo>/node_modules/.pnpm/ember-async-data@0.6.0/node_modules/ember-async-data/index.js
```

### DIFF
--- a/tests/scenarios/v2-addon-peer-deps-test.ts
+++ b/tests/scenarios/v2-addon-peer-deps-test.ts
@@ -1,0 +1,67 @@
+import { appScenarios, baseV2Addon, baseAddon } from './scenarios';
+import { PreparedApp } from 'scenario-tester';
+import QUnit from 'qunit';
+import merge from 'lodash/merge';
+
+const { module: Qmodule, test } = QUnit;
+
+appScenarios
+  .map('v2-addon-with-optional-peer-dep-on-v1-addon', project => {
+    let v1Addon = baseAddon();
+
+    v1Addon.pkg.name = 'v1-addon';
+    merge(v1Addon.files, {
+      addon: {
+        'index.js': `export class TrackedAsyncData {}`,
+      },
+    });
+
+    let v2Addon = baseV2Addon();
+
+    merge(v2Addon.files, {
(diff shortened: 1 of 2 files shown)


tests/scenarios/v2-addon-peer-deps-test.ts
```diff
@@ -0,0 +1,67 @@
...
+    merge(v1Addon.files, {
+      addon: {
+        'index.js': `export class TrackedAsyncData {}`,
+      },
+    });
+
+    let v2Addon = baseV2Addon();
+
+    merge(v2Addon.files, {
+      'index.js': `export { TrackedAsyncData } from 'v1-addon';`,
```
rwjblue, line 22: FWIW, this doesn't really seem like an optional peer dependency. 

I doubt it matters for the test though, I don't think we have any custom/specific logic around a peer dependency being optional or not (though it's possible that webpack does?)...

NullVoxPopuli: >  doesn't really seem like an optional peer dependency.

but, if you don't import `index.js`, you don't need `v1-addon`? (or .... how do optional peer deps work?)

ef4: An example of using an optional peer dep would be

```js

try {
  await import('thing')
} catch (e) {
  // never mind, go on without it
}
```

Or alternatively, using dependencySatisfies from embroider macros to do a similar thing.

NullVoxPopuli: ohhh! then I think I need to update this test!
(and my ember-resources stuff)

NullVoxPopuli:
this test is not needed here

PR #1188, opened 2022-04-11
v2 guide: move `.md` files to published package

SergeAstapov:


PORTING-ADDONS-TO-V2.md
```diff
@@ -43,11 +43,25 @@ The steps:
 
    - yes, this means you will have an `addon/addon` directory. This looks silly, but it will go away when we finish porting the addon to v2.
 
+1. Move CHANGELOG.md, LICENSE.md, and README.md files into `_addon`
+
+   In doing so, the three files will be picked up by npm pack and npm publish respectively,
+   so that the npmjs.com listing shows the proper README.md instead of an error message about missing readme,
+   and so that these files are also available in the users `node_modules`.
+
+1. Adds symlinks to `.md` files in the root to the `addon` folder
```
rwjblue, line 52: Hmm. Doesn't this break the GitHub experience though? IIRC, GitHub doesn't render a `README.md` if it's a symlink (and also clicking on a symlink in GitHub doesn't take you to the file, it shows you a text file of what that symlink points to).

SergeAstapov: > Hmm. Doesn't this break the GitHub experience though? IIRC, GitHub doesn't render a `README.md` if it's a symlink (and also clicking on a symlink in GitHub doesn't take you to the file, it shows you a text file of what that symlink points to).

@rwjblue it's "no" and "yes":
* GitHub does properly render `README.md` on the home page of repo even if `README.md` is a symlink to `addon/README.md`
* but if you open/navigate to the file in GitHub, it just renders the symlink path.

e.g. you can check https://github.com/ember-animation/ember-animated-tools and https://github.com/ember-animation/ember-animated-tools/blob/master/README.md

So `README.md` does not seem to be a real problem.

Not sure people often open `LICENSE.md` so I think it's also not real problem.

The most problematic is `CHANGELOG.md` as I think people often may want to see it.
So it's probably a question of what we'd like to achieve openable `CHANGELOG.md` in the root of the repo or `CHANGELOG.md` included in .tgz and available in `node_modules`

ef4:
Since we have a build step before publishing anyway, we could presumably copy the files from the root into the built package at that time?

SergeAstapov:
> Since we have a build step before publishing anyway, we could presumably copy the files from the root into the built package at that time?

Good point!
I think we should also remove copied files after publishing to ensure it's single source of truth and copied files do not get accidentally committed.

ef4:
Putting them in gitignore should be enough to avoid that.

SergeAstapov:
Thanks for ideas @ef4 @rwjblue @NullVoxPopuli ! I've updated with suggestion Ed made above.

PORTING-ADDONS-TO-V2.md
```diff
@@ -74,6 +74,10 @@ The steps:
    # you definitely want this:
    node_modules
 
+   # you don't want *.md copies
+   /addon/README.md
+   /addon/LICENSE.md
```
rwjblue, line 79: Would you mind sending a PR to the blueprint for these also? https://github.com/embroider-build/addon-blueprint/blob/main/files/gitignore

PORTING-ADDONS-TO-V2.md
```diff
@@ -249,7 +253,7 @@ Now that we've separated the test-app and docs app concerns from the addon, we c
     ],
     "scripts": {
       "build": "rollup --config",
-      "prepublishOnly": "rollup --config",
+      "prepublishOnly": "cp ../README.md ../LICENSE.md . && rollup --config",
```
rwjblue, line 256: Let's suggest `npm-run-all` instead (most apps/addons already have it, it's in the default ember-cli blueprint)

simonihmig: We could also integrate copying into the rollup build itself, e.g. with something like https://github.com/vladshcherbin/rollup-plugin-copy. To not split concerns between npm script and rollup.

At least I was thinking about this for the blueprint (https://github.com/embroider-build/addon-blueprint/issues/8)

SergeAstapov: @simonihmig the only downside is that it will do so even during regular build step. Can't say it's big deal/an issue, just a side effect of such approach

SergeAstapov: had another idea:

use `before:init` hook for the `release-it` like so:

```json
"release-it": {
  "hooks": {
    "before:init": "cp README.md LICENSE.md addon/"
  }
}
```

that way, release process concerns are specific to the release tooling.
Since we copy files form monorepo root folder to the addon folder, I think ideally addon should not care about anything outside of the addon folder itself.

made PR https://github.com/ember-animation/ember-animated/pull/466/files so you may check it out how this looks like in practice.

lifeart: In general, it seems quite tricky to maintain all addons (at scale) release process in package.json file.
Thinking about single source of truth (may be release.js at the start), and later, could be extracted as an ember ecosystem package.

```js
{
  release: "npx ember-addon-release@2 --release"
}
```

Issue #1157, opened 2022-03-08
publicAssetURL has wrong semantics


ef4:
The `publicAssetURL` option to `@embroider/webpack` is supposed to be the full path to where your *assets* -- not your whole **app** -- will get published. It should not assume the path will always include the string `/assets/`. 

ember-auto-import gets this correct -- when you want `/assets` in the URL you need to pass it explicitly.

But embroider gets it wrong. It automatically appends `/assets`. This means you don't really have full control over the URL.

We should implement the fix behind a new option to avoid breaking existing apps.

These spots would stop appending "assets":

https://github.com/embroider-build/embroider/blob/5bd545493ccd751fd7bf69560b9fde7b2d2320bd/packages/webpack/src/ember-webpack.ts#L193

https://github.com/embroider-build/embroider/blob/5bd545493ccd751fd7bf69560b9fde7b2d2320bd/packages/webpack/src/ember-webpack.ts#L423

Corresponding places that are talking about the *filesystem* would need to append "assets" instead, whereas places that are talking about URLs will now get the right results by just appending publicAssetURL.

22a:
I plan on picking this up, hope to have a PR ready for review today/tomorrow.

ef4:
Looking at the related PR and the test failures made me realize I was wrong in the original description above.

I don't think we *can* redefine publicAssetURL to mean only the assets subdir, because "assets" is really only a convention and in principle you don't even have to have such a directory. You can emit assets elsewhere, including at the root directory of the application. 

So I think the actual bug is that we don't currently let you emit your webpack chunks anywhere other than the assets subdir. For example:

1. the build produces `assets/chunk.123.js`
2. you set publicAssetURL to `https://cdn/`
3. we combine them into `https://cdn/assets/chunk.123.js`

We had to include `/assets/` in the url to the CDN because your build could have also emitted `whatever.js` which would need to go at `https://cdn/whatever.js`. 

If you really want `https://cdn/chunk.123.js`, we should let you change step 1 above:

1. the build produces `chunk.123.js`
2. you set publicAssetURL to `https://cdn/`
3. we combine them into `https://cdn/chunk.123.js`

PR #1194, opened 2022-04-18
Align webpack's outputPath with the whole app

ef4:
We hard-code webpack's output path to be the "assets" subdirectory of the built app (which appears as "./dist/assets" after you run a build).

This is needlessly confusing and limiting, because webpack's view of the world doesn't match what the app author actually writes. Mostly this doesn't matter when people are using all the defaults, but if they wanted to customize (for example) the location of the generated chunks they can't escape the assets subdir.

This PR fixes this problem by setting webpack's `output.path` to the root of the (output) ember app (meaning the "./dist" subdir in a typical build), and making corresponding adjustments to `output.filename` and `output.chunkFileName` so that we're still producing the same files by default, but they're now addressed relative to the whole app.

This is fix for https://github.com/embroider-build/embroider/issues/1157.

### DIFF
--- a/packages/webpack/src/ember-webpack.ts
+++ b/packages/webpack/src/ember-webpack.ts
@@ -184,10 +184,10 @@ const Webpack: PackagerConstructor<Options> = class Webpack implements Packager 
         ],
       },
       output: {
-        path: join(this.outputPath, 'assets'),
-        filename: `chunk.[chunkhash].js`,
-        chunkFilename: `chunk.[chunkhash].js`,
-        publicPath: publicAssetURL + 'assets/',
+        path: join(this.outputPath),
+        filename: `assets/chunk.[chunkhash].js`,
+        chunkFilename: `assets/chunk.[chunkhash].js`,
+        publicPath: publicAssetURL,
       },
       optimization: {
         splitChunks: {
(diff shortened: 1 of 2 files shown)


packages/webpack/src/ember-webpack.ts
```diff
@@ -184,10 +184,10 @@ const Webpack: PackagerConstructor<Options> = class Webpack implements Packager
         ],
       },
       output: {
-        path: join(this.outputPath, 'assets'),
-        filename: `chunk.[chunkhash].js`,
-        chunkFilename: `chunk.[chunkhash].js`,
-        publicPath: publicAssetURL + 'assets/',
+        path: join(this.outputPath),
+        filename: `assets/chunk.[chunkhash].js`,
```
NullVoxPopuli, line 188: do we have existing tests to assert these files are created in this folder?

ef4: Tests like[these](https://github.com/embroider-build/embroider/blob/9271aa877d70e4c1bf214fc4e34321ffeeb8d1ae/packages/compat/tests/stage2.test.ts#L597-L611) are directly inspecting the output files.

And the various tests/scenarios are ensuring that these files actually execute.

NullVoxPopuli (approved):

packages/webpack/src/options.ts
```diff
@@ -11,13 +11,41 @@ export interface BabelLoaderOptions {
...
+  // @embroider/webpack.
   webpackConfig: Configuration;
 
   // the base public URL for your assets in production. Use this when you want
   // to serve all your assets from a different origin (like a CDN) than your
   // actual index.html will be served on.
   //
   // This should be a URL ending in "/".
+  //
+  // For example:
```
NullVoxPopuli, line 25: thanks for adding additional docs + examples!

SergeAstapov (approved):

PR #1199, opened 2022-04-29
Add babel plugin for preprocessing templates with ast transforms

wondersloth:
In v2 addons there is no way to apply template transforms to components. This is common in addons that use `ember-cli-preprocessor-registry` and register transforms for modifying `templates`.

- [x] Needs some documentation somewhere on how to use this in the `sample-babel.config.json` or in a `sample-babel.config.js` file.

### DIFF
--- a/PORTING-ADDONS-TO-V2.md
+++ b/PORTING-ADDONS-TO-V2.md
@@ -61,10 +61,7 @@ The steps:
    ```json
    {
      "private": true,
-     "workspaces": [
-       "addon",
-       "test-app"
-     ]
+     "workspaces": ["addon", "test-app"]
    }
    ```
 
@@ -219,6 +216,7 @@ Now that we've separated the test-app and docs app concerns from the addon, we c
    `yarn add --dev @embroider/addon-dev rollup @rollup/plugin-babel @babel/core @babel/plugin-proposal-class-properties @babel/plugin-proposal-decorators`
 
 6. Grab the [example babel config](https://github.com/embroider-build/embroider/blob/main/packages/addon-dev/sample-babel.config.json) and save it as `addon/babel.config.json`
+   - If you addon requires template transforms in order to publish to a shareable format. Apply transforms using the `@embroider/addon-dev/template-transform-plugin`. View how to use this in the [example babel.config.js](https://github.com/embroider-build/embroider/blob/main/packages/addon-dev/sample-babel.config.js)
 7. Grab the [example rollup config](https://github.com/embroider-build/embroider/blob/main/packages/addon-dev/sample-rollup.config.js) and save it as `addon/rollup.config.js`.
 8. Identify your **app reexports**. This is the list of modules from your addon that get reexported by files in the `addon/app` directory.
 9. Delete the `addon/app` directory. You aren't going to need it anymore.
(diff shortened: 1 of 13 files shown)


ef4:
Hi @wondersloth, thanks for tackling this.

This plugin does the same thing as [babel-plugin-stage1-inline-hbs](https://github.com/embroider-build/embroider/blob/93148c45d9b2d4101853ecd948659928a9e9e1b5/packages/core/src/babel-plugin-stage1-inline-hbs.ts). So perhaps instead you could 

 - move that plugin into `@embroider/shared-internals` so it would be available both to `@embroider/core` and to `@embroider/addon-dev`.
 - reexport it from `@embroider/addon-dev` (there is already [an example of another plugin that works this way](https://github.com/embroider-build/embroider/blob/93148c45d9b2d4101853ecd948659928a9e9e1b5/packages/addon-dev/src/template-colocation-plugin.ts)) so it's convenient for addon authors
 - document how to use it from the sample babel config.

ef4:
Oh, also be aware that there are two separate files related to that plugin:
 - babel-plugin-stage1-inline-hbs.ts is the actual implementation, written in a way that isn't tied to any node APIs so it's portable to other environments
 - babel-plugin-stage1-inline-hbs-node.ts is a wrapper around that for convenient usage from Node (it can load the ember template compiler from the filesystem, for example).

wondersloth:
@ef4 

I want to clarify what you're suggestions are here:

### move `babel-plugin-stage1-inline-hbs.ts` into `@embroider/shared-internals`

If I move this file into `shared-internals`:
- I need to hoist a `TemplateCompiler` type/interface.
- I need to have a compiler in `addon-dev` so that I can properly `make()` the `babel-plugin-state1-inline-hbs` plugin.

At this point I need to hoist more code from `@embroider/core` to `@embroider/shared-internals` to make that work across packages.  FIles like `template-compiler-common.ts`, `resolver.ts`, `babel-pluign-adjust-imports.ts`

### Refactor the pieces of  `template-transform-plugin.ts` and `babel-plugin-stage1-inline-hbs.ts` to `shared-internals`

I suppose for this it's doable, just requires abstracting out the visitors?

wondersloth:
Discussed concerns in Embroider Office Hours with @ef4 

Moving `babel-plugin-stage1-inline-hbs.ts` to `shared-internals` will require many files to be hoisted to shared, which will cause a couple issues.
- It will gut `@embroider/core` of a lot of features.
- It will bloat `@embroider/shared-internals` which is consumed by `ember-auto-imports`.

Recommended path
- [ ] Refactor template-transform-plugin to depend on  `babel-plugin-stage1-inline-hbs.ts` via `@embroider/core`.
- [ ] In `@embroider/add-dev` drop `shared-internals` as a dep.

NullVoxPopuli:
Can this be rebased? looks like a lot of stuff from `main` is (for some reason) showing up in this PR's diff

wondersloth:
@ef4 is there any additional work needed before merge?

PORTING-ADDONS-TO-V2.md
```diff
@@ -219,6 +216,7 @@ Now that we've separated the test-app and docs app concerns from the addon, we c
    `yarn add --dev @embroider/addon-dev rollup @rollup/plugin-babel @babel/core @babel/plugin-proposal-class-properties @babel/plugin-proposal-decorators`
 
 6. Grab the [example babel config](https://github.com/embroider-build/embroider/blob/main/packages/addon-dev/sample-babel.config.json) and save it as `addon/babel.config.json`
+   - If you addon requires template transforms in order to publish to a shareable format. Apply transforms using the `@embroider/addon-dev/template-transform-plugin`. View how to use this in the [example babel.config.js](https://github.com/embroider-build/embroider/blob/main/packages/addon-dev/sample-babel.config.js)
```
vladucu, line 219: _nit_ the example file link here is 404'ing

wondersloth: The file, `sample-babel.config.js`, doesn't exist in `main` yet so it will 404.  FWIW's if a relative link is used it may work.

Kaceykaso:
@ef4 @wondersloth any chance we can get this merged today or tomorrow? Looking to announce internally ;)
Thanks a bunch!

PR #1202, opened 2022-05-16
Accept optional `mapFilename` config for `rollup-app-reexports`

dfreeman:
I brought this up in Discord the other day and didn't get any immediate vehement "no way" responses, so I'm going ahead and proposing it here.

Tl;dr: in `@embroider/addon-dev`'s tooling for generating app reexports, I'd like to be able to define a mapping between the source and reexport paths for an entity instead of assuming they're exactly the same.

## Background

For addons that expose many resolvable entities (components, helpers, services, etc), it can be helpful to somehow namespace those entities to make it clear to consumers where they're coming from when they see a component invocation or service injection.

Short of no-longer-the-future [Batman namespacing](https://github.com/rwjblue/ember-holy-futuristic-template-namespacing-batman), a lighter-weight way to do that is by grouping those entities together under a directory so that everything your addon exposes is clearly prefixed with an indicator of where it's from. Concretely, rather than directly having `components/foo` in my `cool-addon` package, I might have `components/cool-addon/foo`, and consumers would write `<CoolAddon::Foo />` in their templates.

It's a bit cumbersome to have that extra directory in the canonical path for the implementation, though: it's an additional layer of nesting to navigate through in editors and code browsers, and you end up writing somewhat redundant import paths like `cool-addon/components/cool-addon/foo`. The latter point also becomes a bit more painful as [first-class component templates](https://github.com/emberjs/rfcs/pull/779) pick up steam and consumers begin needing to care about the import paths of the components, helpers and modifiers they use.

To mitigate this in a v1 addon, I can put my reexport in `app/components/cool-addon/foo.js` while leaving my implementation in `addon/components/foo.hbs`, but since reexports are managed as part of build tooling for v2 addons, I don't have that flexibility.

## This Change

This change adds an optional `appReexports` function that devs can pass to `appReexports` to remap the names of files emitted as part of their `app-js` content.

```js
  plugins: [
    // ...
    addon.appReexports([
      'components/**/*.js',
      'services/**/*.js',
    ], {
      mapFilename: (name) => name.replace(/^(services|components)\//, '$1/cool-addon/'),
    }),
    // ...
  ],
```

By configuring my v2 `cool-addon` like the snippet above, I can provide a reasonably ergonomic form factor for consuming my `Foo` component unambiguously in a loose-mode template (`<CoolAddon::Foo />`) or a strict-mode one (`import Foo from 'cool-addon/components/foo'`).

### DIFF
--- a/packages/addon-dev/src/rollup-app-reexports.ts
+++ b/packages/addon-dev/src/rollup-app-reexports.ts
@@ -6,22 +6,25 @@ export default function appReexports(opts: {
   from: string;
   to: string;
   include: string[];
+  mapFilename?: (filename: string) => string;
 }): Plugin {
   return {
     name: 'app-reexports',
     generateBundle(_, bundle) {
       let pkg = readJsonSync('package.json');
       let appJS: Record<string, string> = {};
-      for (let filename of Object.keys(bundle)) {
+      for (let addonFilename of Object.keys(bundle)) {
+        let appFilename = opts.mapFilename?.(addonFilename) ?? addonFilename;
+
         if (
-          opts.include.some((glob) => minimatch(filename, glob)) &&
-          !minimatch(filename, '**/*.d.ts')
+          opts.include.some((glob) => minimatch(addonFilename, glob)) &&
+          !minimatch(addonFilename, '**/*.d.ts')
         ) {
-          appJS[`./${filename}`] = `./dist/_app_/${filename}`;
+          appJS[`./${appFilename}`] = `./dist/_app_/${appFilename}`;
           this.emitFile({
             type: 'asset',
-            fileName: `_app_/${filename}`,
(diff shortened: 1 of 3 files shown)


ef4:
> but since reexports are managed as part of build tooling for v2 addons, I don't have that flexibility.

Plugins like appReexports appear explicitly in rollup.config.js so you can replace or remove them. Nothing stops a v2 addon author from manually managing an `app` directory and the corresponding app-js entries in package.json.

That said, this change seems fine. The main nudge I want to give people with this plugin is away from putting arbitrary code in the app tree (limiting to only reexports). If people choose to drop the plugin, they lose that nudge. This option lets more people stick with the plugin.

jamescdavis:
Can this be documented somewhere? Maybe in https://github.com/embroider-build/embroider/blob/main/ADDON-AUTHOR-GUIDE.md?

PR #1205, opened 2022-05-19
Fix preprocessors tree by wrapping with moduleName

wondersloth:
This PR addresses the following issues:
- https://github.com/embroider-build/embroider/issues/1206
- https://github.com/salsify/ember-css-modules/issues/278

Replicated the issue here:
https://github.com/wondersloth/demo-embroider-preprocessors-bug

In a classic build the tree passed to the [preprocessor is scoped with the moduleName.](https://github.com/ember-cli/ember-cli/blob/177f52340dcaa59cf27402ed77fda5e55860fd35/lib/models/addon.js#L1195)

In embroider, th app or addon tree is not scoped with the name/moduleName

screenshot

This creates backcompat issues with addons like `ember-css-modules` that uses the file path to generate hash for some values. [Initial attempt fixing from an addon](https://github.com/salsify/ember-css-modules/pull/279).

- [X] Validated fix via `yarn link @embroider/compat` in `demo-embroider-preprocessors-bug`
- [x] Fix `v1-app.ts` to scope/wrap with moduleName when passing to preprocessors.
- [x] Fix tests so that preprocessors apply transform  to `my-addon`.

### DIFF
--- a/packages/compat/src/v1-addon.ts
+++ b/packages/compat/src/v1-addon.ts
@@ -488,14 +488,22 @@ export default class V1Addon {
 
   // applies preprocessors to JS and HBS
   private transpile(tree: Node) {
+    tree = buildFunnel(tree, { destDir: this.moduleName });
+
     tree = this.addonInstance.preprocessJs(tree, '/', this.moduleName, {
       registry: this.addonInstance.registry,
     });
+
     if (this.addonInstance.shouldCompileTemplates() && this.addonInstance.registry.load('template')?.length > 0) {
       tree = this.app.preprocessRegistry.preprocessTemplates(tree, {
         registry: this.addonInstance.registry,
       });
     }
+
+    tree = buildFunnel(tree, {
+      srcDir: this.moduleName,
+    });
+
     return tree;
   }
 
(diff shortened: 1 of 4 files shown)


packages/compat/src/v1-addon.ts
```diff
@@ -488,15 +492,35 @@ export default class V1Addon {
 
   // applies preprocessors to JS and HBS
   private transpile(tree: Node) {
+    tree = debugTree(tree, { label: `${this.moduleName}-1-before`, force: true });
+
+    let hasAdjustedPath = false;
+
+    if (!this.moduleName.startsWith('@')) {
```
wondersloth, line 499: Remove the `@`

packages/addon-dev/.gitignore
```diff
@@ -4,6 +4,9 @@
 /src/**/*.js
 /src/**/*.d.ts
 /src/**/*.map
+/tests/**/*.js
+/tests/**/*.d.ts
+/tests/**/*.map
```
wondersloth, line 10: Remove all conflated changes from `packages/addon-dev`

wondersloth:
@ef4 I've got my tests for the preprocessors to pass. 

Next would be to validate this more broadly.

wondersloth:
In my integration testing with the example apps in ember-css-modules, I found another bug. I need to move the unwrapping of the moduleName from the files, because the template ast transforms already get the moduleName in the path.

PR #1213, opened 2022-06-03
Prevent redundant toTree wrapping for macros

raycohen:
In an app with _many_ addons, we are seeing this wrapping get applied a significant number of times, eventually leading to  "Maximum call stack size exceeded" errors when toTree is finally executed.

When I copy the stack trace out of visual code just before the "Maximum call stack size exceeded" error, there are 7364 lines in the stack that match `node_modules/@embroider/macros/src/ember-addon-main.js:56`

I think this wrapper only needs to be applied once per app instance. Is there something else I should be doing to prevent `@embroider/macros` `included` method from getting hit so much?

Screen Shot 2022-06-03 at 12 15 34 PM

Screen Shot 2022-06-03 at 12 19 33 PM

### DIFF
--- a/packages/macros/src/ember-addon-main.ts
+++ b/packages/macros/src/ember-addon-main.ts
@@ -48,16 +48,23 @@ export = {
 
     appInstance.import('vendor/embroider-macros-test-support.js', { type: 'test' });
 
-    // When we're used inside the traditional ember-cli build pipeline without
-    // Embroider, we unfortunately need to hook into here uncleanly because we
-    // need to delineate the point in time after which writing macro config is
-    // forbidden and consuming it becomes allowed. There's no existing hook with
-    // that timing.
     const originalToTree = appInstance.toTree;
-    appInstance.toTree = function (...args) {
-      macrosConfig.finalize();
-      return originalToTree.apply(appInstance, args);
-    };
+
+    // as macrosConfig is 1:1 with the appInstance, if
+    // the toTree method is already wrapped, we don't need
+    // to wrap it again
+    if (!originalToTree.__hasMacrosConfigWrapper) {
+      // When we're used inside the traditional ember-cli build pipeline without
+      // Embroider, we unfortunately need to hook into here uncleanly because we
+      // need to delineate the point in time after which writing macro config is
+      // forbidden and consuming it becomes allowed. There's no existing hook with
+      // that timing.
+      appInstance.toTree = function (...args) {
+        macrosConfig.finalize();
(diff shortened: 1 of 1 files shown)


lifeart (approved):

packages/macros/src/ember-addon-main.ts
```diff
@@ -48,16 +50,23 @@ export = {
...
     const originalToTree = appInstance.toTree;
-    appInstance.toTree = function (...args) {
-      macrosConfig.finalize();
-      return originalToTree.apply(appInstance, args);
-    };
+
+    // as macrosConfig is 1:1 with the appInstance, if
+    // the toTree method is already wrapped, we don't need
+    // to wrap it again
+    if (!wrappedTrees.has(originalToTree)) {
```
simonihmig, line 58: Does it have to be a `Set`? Couldn't it be just a simple boolean flag? I.e. we will only be called with a single `appInstance`, right?

raycohen: That sounds right... I wasn't sure if this could somehow get hit with multiple apps in the same process. Maybe during tests?

raycohen: Updated without using a Set. It works for me locally this way.

simonihmig (approved):
LGTM!

> Is there something else I should be doing to prevent @embroider/macros included method from getting hit so much?

No, I don't think so. You are getting called (not literally *you* of course :wink:), by EmberCLI, so there is nothing to change that on this side. However I am quite puzzled that `include()` is called *that much*! (>7000x 😱)

ef4:
I've seen some pretty crazy things like that in big apps. ember-cli creates separate addon instances per consumer.

If you have an addon that uses embroider/macros, and that addon is used by 10 other addons, and those addons are each used by 10 other addons, and *those* addons are used by 10 other addons, embroider/macros will get instantiated 1000 times.

raycohen:
> ember-cli creates separate addon instances per consumer.

Yeah that seems to be the situation I'm seeing. I've got ~175 addons recognized by the project, but the recursive `included` calls from 

https://github.com/ember-cli/ember-cli/blob/1519391eb562c446123e6b63fa400ed7b9e6b77a/lib/models/addon.js#L769

blow things up

rwjblue:
This will still happen once per `@embroider/macros` version, right (because the module scoped boolean or set would be per-version)? Unless there is some sort of highlander logic going on (which I don't recall having here),

ef4:
I think that's correct, but that's probably not enough to blow up the stack limit.

PR #1214, opened 2022-06-03
fix: reset HTMLEntrypoint state after build (fix #1211)

### LINKED ISSUE #1211: live reload broken (duplicate scripts like vendor.js appear)
Reproduction: https://github.com/jacobq/embroider-live-reload-bug-repro

Enabling embroider in minimal (ember-cli v4.2.0) app breaks live reload feature. When `ember s` first gets run things are OK. However, changing a file (triggering update/reload) causes something weird to happen such that multiple `vendor.js` scripts appear (one gets added after each change/reload).

Compare link for v1.6.0 vs v1.7.1 (known good / bad):
https://github.com/embroider-build/embroider/compare/v1.6.0...v1.7.1

jacobq:
I'm not sure if this is "the right solution", but it did seem the most obvious one to me as an outsider, and it does appear to work (fixes #1211). I don't love that it exposes a method (`HTMLEntrypoint::initialize`) that should probably be private somehow (e.g. calling it after construction but before being "done" may lead to weird/inconsistent states), but I needed a way for the Webpack plugin to reset the HTMLEntrypoint to its pristine/initial state so that it could be reused for subsequent calls to `build` that used the cached `webpack` instance (and perhaps a stern warning in a doc comment would be sufficient here...). Open to feedback / alternatives, of course, just wanted to help get this unblocked.

@krisselden I think you made the original PR that refactored the file output to a Webpack plugin. Do you have thoughts on this?

### DIFF
--- a/packages/core/src/html-entrypoint.ts
+++ b/packages/core/src/html-entrypoint.ts
@@ -8,7 +8,7 @@ import Placeholder from './html-placeholder';
 import { Variant } from './packager';
 
 export class HTMLEntrypoint {
-  private dom: JSDOM;
+  private dom!: JSDOM;
   private placeholders: Map<string, Placeholder[]> = new Map();
   modules: string[] = [];
   scripts: string[] = [];
@@ -20,6 +20,10 @@ export class HTMLEntrypoint {
     private publicAssetURL: string,
     public filename: string
   ) {
+    this.initialize();
+  }
+
+  initialize() {
     this.dom = new JSDOM(readFileSync(join(this.pathToVanillaApp, this.filename), 'utf8'));
 
     for (let tag of this.handledStyles()) {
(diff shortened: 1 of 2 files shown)


jacobq:
@ef4 Would you mind approving CI run to make sure I didn't accidentally break any tests? (My local testing only consisted of linking my bug repro app and confirming that it worked correctly with this change)

ef4:
Thanks for taking a look at this.

It's intentional that the HTMLEntrypoint preserves state across builds. The idea is that you don't bother re-parsing an HTML file that hasn't changed. It's only the contents of the [placeholders](https://github.com/embroider-build/embroider/blob/aec11b7204e1da7fbc85f8b6f2995cd0f59a81e5/packages/core/src/html-entrypoint.ts#L12) that need to get recreated each build.

I think what's missing is a call to `placeholder.clear()` right about [here](https://github.com/embroider-build/embroider/blob/aec11b7204e1da7fbc85f8b6f2995cd0f59a81e5/packages/core/src/html-entrypoint.ts#L93).

ef4:
Superseded by #1219.

PR #1215, opened 2022-06-03
Strip extension from reexport of `rollup-app-reexports`

simonihmig:
This makes `rollup-app-reexports` strip file extensions (`.js`) from reexports, aligning with (the blueprint of) v1 addons, and allowing better interoperability with `package.json#exports` with path mappings including extensions, as discussed in https://github.com/typed-ember/glint/issues/338.


ef4:
Thanks @simonihmig

PR #1216, opened 2022-06-04
Migrate from yarn to pnpm

NullVoxPopuli:
The number of bugs in yarn@v1 is too high!
(and pnpm is really fast, has good (any) support, which is good)


ef4:
We rely on Volta, including for some tests that run older ember-cli versions using older node versions. 

I'm hoping Volta will have pnpm support soon.

NullVoxPopuli:
Yeah, Volta not having pnpm support is a major downer for me, too. No way to ensure multiple folks have the same version.
At work, across nearly 200 projects now, that'll be really important.

mastastealth:
[Almost there](https://github.com/volta-cli/volta/pull/1273). :eyes:

Issue #1217, opened 2022-06-06
Make keepAssets also edit the packageJson like appReexports does


NullVoxPopuli:
Was pairing with a co-worker today with some WASM + WebWorker stuff in a v2 addon, and noticed that keepAssets doesn't modify the package.json. 

I think, by default, could could do something easy, like:

```js
addon.keepAssets('public/**/*.{wasm,js}')
```
gets built in
```
dist/
  public/
     my-wasm.wasm
     my-worker.js
```
and the package.json entry would be:
```js
"ember-addon": {
   "public-assets": {
      "dist/public/my-wasm.wasm": "/<the-addon-name>/my-wasm.wasm",     
      "dist/public/my-worker.js": "/<the-addon-name>/my-worker.js",    
  }
}
```

and then like with appReexports, it can take a function for changing how the output is generated, like:

```js
addon.keepAssets('public/**/*.{wasm,js}', (srcName) => {
  switch (srcName) {
    case 'public/my-wasm.wasm': return '/my-secret-path/my-wasm-the-best.wasm';
    case 'public/my-worker.js': return '/my-secret-worker/the-best-worker.js';
  }
})
```
which would output
```js
"ember-addon": {
   "public-assets": {
      "dist/public/my-wasm.wasm": "/my-secret-path/my-wasm-the-best.wasm",     
      "dist/public/my-worker.js": "/my-secret-path/the-best-worker.js",    
  }
}
```

NullVoxPopuli:
After talking with @ef4 about this in https://github.com/embroider-build/embroider/pull/1218, this feature isn't needed :partying_face:

PR #1218, opened 2022-06-07
Implement automated keepAssets package.json augmentation

NullVoxPopuli:
Resolves: https://github.com/embroider-build/embroider/issues/1217

This *could* be a breaking change as behavior was "do nothing" before, unless the lack of behavior is considered a bug.

Todo: add test

### DIFF
--- a/packages/addon-dev/src/rollup-keep-assets.ts
+++ b/packages/addon-dev/src/rollup-keep-assets.ts
@@ -1,15 +1,20 @@
 import walkSync from 'walk-sync';
 import type { Plugin } from 'rollup';
 import { readFileSync } from 'fs';
+import { readJsonSync, writeJsonSync } from 'fs-extra';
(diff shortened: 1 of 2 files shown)


ef4:
@NullVoxPopuli and I discussed this during office hours, the conclusion was: not everything that goes into keepAssets should go into the public-assets list in package.json. public-assets is a very niche feature, mostly for auto-upgraded v1 addons. Most of the time v2 addons who are using keepAssets are doing it so that their code can refer to assets in a way that is legible to the final stage packager (ie webpack), and that is better than forcing a file into the consuming app's actual public dist.

CSS for example is in keepAssets by default, but not because all the addons CSS files should be shipped separately and unchanged in the final app's dist. Rather it's so they can be consumed by the app's bundler, which compiled references to them into the correct URL.

PR #1223, opened 2022-06-28
Stabilize extensions for addon-dev's publicEntrypoints.

NullVoxPopuli:
Per some discussion in https://github.com/embroider-build/embroider/pull/1099
(and some common questions in discord),
we want to use `.js` for all file paths in the rollup config.

This PR transparently adds _every_ extension (we only support 5) to each glob passed to `addon.publicEntrypoints` so that we can only specify `.js`, regardless of what the filetypes are on disk.

I'm currently getting stuck on this error:
```
(!) The emitted file "components/demo/index.js" overwrites a previously emitted file of the same name.
```
which you can reproduce via:
```
cd tests/scenarios
yarn test:output --scenario v2-addon-dev-typescript
cd output
cd node_modules/v2-addon
node node_modules/rollup/dist/bin/rollup -c ./rollup.config.mjs
```

### DIFF
--- a/packages/addon-dev/src/rollup-public-entrypoints.ts
+++ b/packages/addon-dev/src/rollup-public-entrypoints.ts
@@ -25,7 +25,7 @@ export default function publicEntrypoints(args: {
 
         // anything that doesn't match the users patterns, and wasn't a template-only
         // component needs to be emitted "as-is" so that other plugins may handle it.
-        let isTO = isTemplateOnly(args.srcDir, name);
+        let isTO = isTemplateOnly(matches, name);
 
         let isHbs = path.extname(name) === '.hbs';
 
@@ -72,7 +72,7 @@ export default function publicEntrypoints(args: {
   };
 }
 
-function isTemplateOnly(srcDir: string, filePath: string) {
+function isTemplateOnly(matches: string[], filePath: string) {
   let isHbs = path.extname(filePath) === '.hbs';
 
   if (!isHbs) return false;
@@ -82,10 +82,9 @@ function isTemplateOnly(srcDir: string, filePath: string) {
     path.basename(filePath).replace(/hbs$/, '*')
   );
 
-  let relatedFiles = walkSync(srcDir, {
-    globs: [correspondingFileGlob],
-  });
-
+  let relatedFiles = matches.filter((match) =>
+    minimatch(match, correspondingFileGlob)
+  );
   let isTO = relatedFiles.filter((x) => x !== filePath).length === 0;
 
   return isTO;
--- a/packages/addon-dev/src/rollup-public-entrypoints.ts
+++ b/packages/addon-dev/src/rollup-public-entrypoints.ts
@@ -1,7 +1,6 @@
 import walkSync from 'walk-sync';
-import { join } from 'path';
+import path from 'path';
 import minimatch from 'minimatch';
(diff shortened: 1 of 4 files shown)


NullVoxPopuli:
where I'm failing is that I seem to somehow be removing ts files from the build entirely. they aren't transpiled by rollup-plugin-ts and the resulting ts+hbs co-located component ends up as a template-only component in `dist`

packages/addon-dev/src/rollup-public-entrypoints.ts
```diff
@@ -30,19 +28,34 @@ export default function publicEntrypoints(args: {
...
-              minimatch(normalizedName, pattern)
-            ) &&
-            !pathExistsSync(id)
-          ) {
+          let normalizedMatch = args.include.some((pattern) =>
+            minimatch(normalizedName, pattern)
+          );
+          let normalizedExists = pathExistsSync(id);
+
+          let matchesDeferred = [...args.include, '**/*.ts'].some(pattern => minimatch(name, pattern) )
```
NullVoxPopuli, line 44: This is what I'm least certain about :-\

NullVoxPopuli: if I _only_ do: 
```ts
            this.emitFile({
              type: 'chunk',
              id: join(args.srcDir, name),
              fileName: normalizedName,
            });
```
then we get: 
```
(!) The emitted file "components/demo/index.js" overwrites a previously emitted file of the same name.
```
and the resulting component ends up being a template-only component -- ie, the class is lost

ef4: I would guess the TS plugin is already emitting that JS file. But it probably shouldn't be. Perhaps it has options to control that?

For example, it might be trying to emit a JS file for every TS file, but that would be wrong. We only want to emit public entrypoints, and let other modules get rolled up.

NullVoxPopuli: > Perhaps it has options to control that?

plugins can't interact with other plugins tho? or can they?

maybe I don't understand?

ef4: I mean the options you're passing directly to the ts plugin in the example config.

NullVoxPopuli: there are no options for passing entryPoints to the plugin? https://github.com/wessberg/rollup-plugin-ts

nightire: @NullVoxPopuli Hi, we use rollup-plugin-ts in a recent project and found it ditches the default types of generics in the output.

For example, for input like:

```ts
export interface Foo<T extends Bar | undefined = undefined> {
  ...
}
```

the output is:

```ts
export interface Foo <T extends Bar | undefined> {
  ...
}
```

So we end up to use rollup-plugin-typescript2 + rollup-plugin-dts instead. Just for your information, because this project is not an embroider project, we can not be sure if rollup-plugin-ts will be a problem here.

NullVoxPopuli: I think something else may have been at play in your project because rollup-plugin-ts very faithfully generates type definitions (I just confirmed just now by looking at the dists' d.ts files with default type args).

PR #1225, opened 2022-07-01
docs: Add lazy loaded engines EmbroiderRouter details to README

richgt:
Lazy Loaded engines require `@embroider/router`. Adding to README.


ef4:
Thanks!

PR #1228, opened 2022-07-03
ember-addon keyword required?

NullVoxPopuli:
I don't know if this is required or not -- but it's bitten me a few times.

We also have the `ember-addon` object (soon to be just `ember`), which also specifies addonness -- which I can include in this PR -- but wanted to verify about the keyword first


ef4:
By requiring the keyword we ensure that every ember addon is searchable as such in the NPM registry. The separate ember-addon or ember top-level property is nonstandard as far as public registries are concerned.

PR #1229, opened 2022-07-04
Run the `clean` plugin of `addon-dev` as late as possible

simonihmig:
This attempts to fix the problem described in https://github.com/embroider-build/addon-blueprint/issues/32.

While it is not the perfect solution like a truly atomic build update, it seems "good enough". Previously the cleanup would happen at the earliest point in time, at [`buildStart`](https://github.com/vladshcherbin/rollup-plugin-delete/blob/master/src/index.js#L5), making the time window large enough for Ember CLI to see the transient build output in an inconsistent state. Now it happens at the *latest* possible time, at [`generateBundle`](https://rollupjs.org/guide/en/#generatebundle) right before files are written, making the time window small enough to not cause any problems in practice. (successfully tested locally)

### DIFF
--- a/packages/addon-dev/src/rollup.ts
+++ b/packages/addon-dev/src/rollup.ts
@@ -47,7 +47,7 @@ export class Addon {
   // By default rollup does not clear the output directory between builds. This
   // does that.
   clean() {
-    return clean({ targets: `${this.#destDir}/*` });
+    return clean({ targets: `${this.#destDir}/*`, hook: 'generateBundle' });
   }
 
   // V2 Addons are allowed to contain imports of .css files. This tells rollup


ef4:
Thanks. 

I wonder if, now that we have https://github.com/embroider-build/addon-blueprint/blob/d64ff367581eb95e5c92ea4d9799ddcfb44ef437/files/__addonLocation__/rollup.config.js we could just link people to there instead of maintaining this example file here too, and worrying about keeping them in sync.

ef4:
Oh, I'm sorry I misunderstood. I was imagining this was a change to https://github.com/embroider-build/embroider/blob/7bbd49c55a4b0f3bf4a1605f5307c412c089cfa0/packages/addon-dev/sample-rollup.config.js (which we could indeed remove from this repo in favor of the blueprint repo).

Issue #889, opened 2021-07-05
Bug: No TypeScript support (without ember-cli-typescript)


NullVoxPopuli:
My situation:
 - ember 3.27
 - embroider 0.42.x
 - no ember-cli-typescript

The closest issue I found was this closed one: https://github.com/embroider-build/embroider/issues/691

What I'm observing:
 When I make the app, router, or test-helper files TS, embroider can no longer locate them (ModuleNotFound errors).

With my tests files in TS, they are not detected at all, and my embroider part of CI passes with only having run the Ember.onerror check and nothing else.

Using ember-cli-typescript in this case is not an option, because: https://github.com/typed-ember/ember-cli-typescript/issues/1439
But, on work projects we also don't use ember-cli-typescript, because we use composite TS projects, and I experienced the same behavior in an app. So, this TS support issue is not isolated to just add-ons.

ef4:
You don't explain here how you're setting up typescript. If you're using a standalone typescript compiler, that should work fine because everything would be JS before embroider does anything.

If you're configuring typescript support directly in webpack, then yes, probably `@embroider/compat` doesn't know to keep typescript files around as its rewriting everything to get it ready for webpack. See https://github.com/embroider-build/embroider/blob/81ca52f837bd6ff9e80bc3ac406677f05da524a4/packages/compat/src/compat-app.ts#L240

Can you try patching that place so it always unshifts `.ts` and see if that fixes your problem?

Also, you can probably workaround by creating an in-repo-addon that does nothing but is named `ember-cli-typescript`.

NullVoxPopuli:
> You don't explain here how you're setting up typescript.

I thought the ember-cli-babel config would have been sufficient. If not, then I guess these problems make sense.

I'll try your suggestions soon

NullVoxPopuli:
- adding an in-repo-addon to my addon called `ember-cli-typescript`
  - detects TS files
  - but maybe doesn't transpile them? 
```
    $TMPDIR/embroider/80e0e8/tests/dummy/tests/helpers/await.ts/await.ts: Unexpected reserved word 'interface'. (7:0)

   5 | import { hbs } from 'ember-cli-htmlbars';
   6 |
>  7 | interface Args {
     | ^
```
  https://github.com/NullVoxPopuli/ember-repl/pull/27/checks?check_run_id=2990856820#step:5:385
  
  
  - always unshifting `.ts` in compat-app:
     - same as above
     
[...]

PR #1236, opened 2022-07-27
Support TypeScript without ember-cli-typescript

NullVoxPopuli:
Repro'd and then patched here: https://github.com/NullVoxPopuli/ember-demo-typescript-without-ec-ts/commit/dd5baa622e2c564ec150a9f0b4f028ac24f8c9bf

With glint + `<template>`, ember-cli-typescript's type-checking worker is incorrect as tsserver can't, by itself, resolve .gts imports. 

Since glint projects fully manage type checking within glint (and gts transformation via ember-template-imports), ember-cli-typescript should be removed as a requirement for projects using typescript so that we can reduce terminal noise in the build.

Later, we may bundle glint in some webpack (or preferrably, [unplugin](https://github.com/unjs/unplugin)) plugin such that we properly type check ember projects with glint / template-imports (maybe similar to [`fork-ts-checker--webpack-plugin`](https://www.npmjs.com/package/fork-ts-checker-webpack-plugin) )

anywho, this PR is mostly tests:
 - additional app template, based off of the ember-cli-typescript blueprint
   but also
   - some simplifications of the tsconfig.json
   - convert app.js, router.js, and test-helper.js to ts to have a minimal set of "ts stuff" for every test/scenario
 
Resolves: https://github.com/embroider-build/embroider/issues/889


NullVoxPopuli:
I probably broke something :sweat_smile: 

```
Build Error (PackagerRunner) in assets/test.js

Module build failed (from ../../../../../..<repo>/tests/scenarios/output/node_modules/@embroider/webpack/node_modules/thread-loader/dist/cjs.js):
Thread Loader (Worker 1)
[BABEL] $TMPDIR/embroider/d9b711/tests/scenarios/output/assets/test.js/test.js: module.call is not a function (While processing: "<repo>/packages/core/src/portable-babel-launcher.js")

    at babelLauncher (<repo>/packages/core/src/portable-babel-launcher.js:18:25)}
```

packages/compat/src/compat-app.ts
```diff
@@ -233,23 +233,11 @@ class CompatAppAdapter implements AppAdapter<TreeNames> {
...
-    // resolvable and there's no other "loader" or anything to apply.
-    if (this.activeAddonChildren().find(pkg => pkg.name === 'ember-cli-typescript')) {
-      extensions.unshift('.ts');
-    }
-
-    return extensions;
+    // For TS, we defer to ember-cli-babel, and the setting for
+    // "enableTypescriptTransform" can be set with and without
+    // ember-cli-typescript
+    return ['.wasm', '.mjs', '.js', '.json', '.hbs', '.ts'];
```
bwbuchanan, line 240: This introduces a bug that causes TypeScript component files to be ignored when `staticComponents: true`

should be: `return ['.ts', '.wasm', '.mjs', '.js', '.json', '.hbs'];`

NullVoxPopuli: how does that happen / how does moving ts to the front resolve it?

bwbuchanan: I'm not at all familiar with the internals of embroider and had to diagnose this by trial-and-error.

The previous version of the code unshifts '`.ts'` onto the array (placing it at the front). This change moves it to the end of the array.

When I inspected $TMPDIR/embroider/.../components, I found three files for each of my TypeScript components: .hbs, .js, and .ts.

.hbs and .ts were as expected. .js contains a templateOnlyComponent stub.

I can only deduce that in the process of assembling the app, embroider looks for the component implementation file in the order of the extensions in this array, so moving .ts to the end of the array means that it finds the empty stub implementation and ignores the actual implementation in the .ts file.

I have no idea why the stub .js implementations are being generated if a .ts backing class definition exists. Perhaps this is a different bug, or perhaps intended behavior?

The result is a broken component with no backing class.

Doesn't happen if staticComponents = false in the embroider config.

PR #1239, opened 2022-08-03
add unique-id to builtInHelpers list

jakesjews:
This is preventing staticHelpers from working when the unique-id helper is used


SergeAstapov (approved):

ef4:
Thanks!

PR #1242, opened 2022-08-10
Streamline stage1 template compilation

ef4:
Working to integrate https://github.com/emberjs/babel-plugin-ember-template-compilation/pull/9

This should significantly simplify our template compilation pipeline.

The first commit just shims the new version into the place of the old one, but this PR should ultimately be able to delete a bunch of code by taking advantage of the new features.


ef4:
Note to self: the next step here is that I broke component template colocation in addons that have custom AST transforms. We can't just unilaterally convert the hbs to js without also doing the colocation step all at once.

PR #1245, opened 2022-08-12
Do not use absolute path in vendor files

stevcooo:


ef4:
A few tests need to be updated to match the new output this produces. I suspect it's as simple as changing from `test()` to `runtimeTest()` for the ones that are asserting about the config being in the output.

`test()` will run under both modes, `runtimeTime()` runs only when in run-time mode, and there's also `buildTimeTest()` for the opposite.

ef4:
Thanks!

PR #1246, opened 2022-08-13
Dropping support for older ember's

ef4:
Ember's oldest supported LTS no longer needs the old modules api polyfill and has first-class components. This allows us to drop a whole bunch of complexity.

This PR also includes a refactor of the test suite to get everything onto scenario-tester, so that everything respects the same support matrix. As a practical matter, this means some tests that were co-located within their packages will move into tests/scenarios, because it's somewhat annoying to make it work otherwise. As a benefit, this will improve windows test coverage because all the newer-style tests run on both windows and unix and the older ones did not.


ef4:
This breaking release will apply to the core and compat packages. We don't need breaking releases to macros, util, etc.

ef4:
There is more to do here but this has now completed the refactor of the test suite, which resulted in fixing a bunch of windows bugs, and this branch is huge, so I'm merging to continue in other PRs.

PR #1252, opened 2022-09-01
extend existing EmberENV for ember-inspector

### LINKED ISSUE #1251: production build breaks ember-inspector component tab
relevant issue:  https://github.com/emberjs/ember-inspector/issues/2125

here is definition how it should work:
https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/environment/lib/env.ts#L129

relevant code in inspector
https://github.com/emberjs/ember-inspector/blob/main/skeletons/web-extension/boot.js#L13
https://github.com/emberjs/ember-inspector/blob/main/ember_debug/libs/capture-render-tree.js#L10

embroider just overwrites EmberEnv:
https://github.com/embroider-build/embroider/blob/main/packages/core/src/app.ts#L293

patricklx:
fixes #1251

### DIFF
--- a/packages/core/src/app.ts
+++ b/packages/core/src/app.ts
@@ -290,7 +290,7 @@ export class AppBuilder<TreeNames> {
       result.unshift({
         kind: 'in-memory',
         relativePath: '_ember_env_.js',
-        source: `window.EmberENV=${JSON.stringify(emberENV, null, 2)};`,
+        source: `window.EmberENV=Object.assign(window.EmberENV || {}, ${JSON.stringify(emberENV, null, 2)});`,
       });
 
       result.push({
--- a/packages/core/src/app.ts
+++ b/packages/core/src/app.ts
@@ -290,7 +290,7 @@ export class AppBuilder<TreeNames> {
       result.unshift({
         kind: 'in-memory',
         relativePath: '_ember_env_.js',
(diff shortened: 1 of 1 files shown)


patricklx:
if object.assign is not well enough supported for you then this could be used
```
function(e, t) {
    for (var r in t)
        e[r] = t[r]
    return e
}(window.EmberENV || {}, {...});
```

void-mAlex:
> Thanks for working on this, this looks like a good fix.
> 
> But yes, we do still support Ember 3.x versions and those support IE11, so let's not add Object.assign here.
> 
> You can use object spread (because if the app needs IE11 support their babel settings will transpile this):
> 
> ```js
> window.EmberENV={ ...(window.EmberENV || {)}, ...${JSON.stringify(emberENV, null, 2)} }
> ```

@ef4 I have that exact ie11 use case and I can confirm it doesn't get transpiled.
looks to be the single instance in all the chunks/vendor that is not fixed.

not sure if it is a side effect due to staticEmberSource or not

ef4:
That's pretty weird, can you check if this exact same code would transpile correctly if you put it into any other spot in your app code?

ef4:
Oh, I see, I think this is code that gets directly appended to vendor.js. Yeah, that would skip babel. We should strategize what's the cleanest fix.

Issue #1249, opened 2022-08-30
"Unsafe dynamic component" error when using local template-only component with `{{component ...}}` helper


dknutsen:
### Scenario
1. I started playing around with Embroider and strict mode templates via ember-template-imports today. 

2. I set up a new project with the --embroider flag using latest ember-cli, embroider, etc

3. I enabled all the things:
```js
  return require('@embroider/compat').compatBuild(app, Webpack, {
    staticAddonTestSupportTrees: true,
    staticAddonTrees: true,
    staticHelpers: true,
    staticModifiers: true,
    staticComponents: true,
    splitAtRoutes: [],
```

4. I installed ember-template-imports

5. to play around with contextual components I created a nonsense sidebar component like this:
```js
import { array, component } from '@ember/helper';

const join = (segments) => segments.join('.');

const SidebarItem = <template>
  <li id={{join (array @prefix @id)}}>{{@title}}</li>
</template>

const SidebarCategory = <template>
  <li>{{@title}}</li>
  {{yield (component SidebarItem prefix=@id)}}
</template>

export default <template>
  <SidebarCategory @title="test" @id="test" as |Item|>
    <Item @title="a" @id="a" />
    <Item @title="b" @id="b" />
  </SidebarCategory>
</template>
```

### Result
When `staticComponents: true` is enabled this throws the following error:
```
$TMPDIR/embroider/a09361/components/sidebar.js/sidebar.js: Unsafe dynamic component: SidebarItem in components/sidebar.js
```

Obviously if I turn `staticComponents` to `false` it works fine.

I would expect this to work since SidebarItem is a component class and not actually a "dynamic component" per se. @NullVoxPopuli seemed to think this would work as well. 

But maybe that's a bad assumption? Is there a better way to partially apply arguments for a contextual component pattern? Or is this just a bug/gap?

ef4:
I think this is just a bug and the static components analyzer doesn't know that it should completely ignore templates that are in handlebars strict mode.

bwbuchanan:
I think this was addressed by https://github.com/embroider-build/embroider/commit/f3e225764f5b7acd16e95d3c68c2985fb60463bd, which I just discovered when trying to rebase my own attempt to fix this. 🤦‍♂️

PR #1255, opened 2022-09-10
Resolve https://github.com/embroider-build/embroider/issues/1249

NullVoxPopuli:
From #1249, 

> I think this is just a bug and the static components analyzer doesn't know that it should completely ignore templates that are in handlebars strict mode.

in gjs/gts, we know that everything is strict mode all the time, so we can skip over those files.


ef4:
Thanks for working on this.

I think this isn't exactly the right solution, because as soon as we add `<template>` support to `@embroider/addon-dev`, we will expect addons to publish their strict mode templates *not* in GTS/GJS files, but already compiled to JS that contains:

```js
import { precompileTemplate } from "@ember/template-compilation";
precompileTemplate("<YourCodeHere />", { strict: true, scope: () => ({ YourCodeHere })  })
```

Also, users are allowed to type that by hand too, and it shouldn't blow up. While I expect this low-level format to mostly be a compile target and not an authoring format, we shouldn't blow up if somebody uses it literally. 

I also expect there will always be some contexts where you'll want to use the longhand form to get around the need for a custom syntax preprocessor, like if you wanted to demo something quickly in a hosted JS playground that doesn't understand GJS.

So I think our implementation should look for `strict: true` rather than using the filename. Making this easy to do is something I have already been refactoring toward. My goal is to land the features from babel-plugin-ember-template-compilation 2.0-alpha and use them to redo our staticComponents analyzer to be much simpler.

NullVoxPopuli:
vite supports gjs natively 🎉

PR #1271, opened 2022-10-28
drop vendored template compiler

ef4:


ef4:
(Still being used by some tests that aren't worth refactoring at the moment.

PR #1273, opened 2022-10-29
Don't generate .js component stubs for .ts components

### LINKED ISSUE #1261: @embroider/compat v1.9.0 breaks TypeScript components when staticComponents: true
@ef4 @NullVoxPopuli 

See [my comment on #1236](https://github.com/embroider-build/embroider/pull/1236/files#r990693797)

#1236 changes the order of `resolvableExtensions` to put `.ts` at the end, instead of the beginning, of the array.

The stub template-only component .js file is then being resolved, instead of the actual component implementation in the .ts file, and the component breaks at runtime because its backing class is empty.

(Why are these template-only component stub files being generated in the first place, if an implementation file exists? Is this also a bug somewhere?)

bwbuchanan:
Fixes #1261

### DIFF
--- a/packages/compat/src/synthesize-template-only-components.ts
+++ b/packages/compat/src/synthesize-template-only-components.ts
@@ -9,10 +9,7 @@ export default templateOnlyComponent();`;
 
 const templateExtension = '.hbs';
 
-// we don't need to worry about all resolvable extensions in here (like .ts)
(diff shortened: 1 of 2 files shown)


ef4:
Thanks

xomaczar:
can we cut a bug fix release?

PR #1276, opened 2022-11-07
Streamline template compilation, part 2

ef4:
#1242 refactored the stage1 (addon compatibility) template compilation to use the new capabilities of babel-plugin-ember-template-compilation 2.0-alpha.

This PR will do the same for the stage3 (final app build) template compilation.

TODOs and areas to cleanup: 
 - [x] `app.ts` babel config changes
 - [x] `load-ember-template-compiler.ts`
 - [x] `babel-plugin-inline-hbs-deps*`
 - [x] `template-compiler-node.ts`
 - [x] `template-compiler-common.ts`
 - [x] `patch-template-compiler.ts`
 - [x] `ember-template-compiler-types.ts`
 - [x] `core/src/resolver.ts`
 - [x] `compat/src/resolver.ts`
 - [x]  release babel-plugin-ember-template-compilation


ef4:
Next steps here: 
 - refactor the resolver tests, probably using code-equality-assertions since we're eliminating the need for the explicit `dependenciesOf` API
 - switch most cases to use lexical scope. The only case that can't is a component with an un-associated template.
   - need to confirm that `{{component}}` helper is OK with taking lexically scope component classes as input (for the cases where it's being used for currying and can't just be replaced outright)

PR #1282, opened 2022-11-14
Addon-shim was missing dependencies

NullVoxPopuli:
This came up when I moved broccoli-funnel to `nohoist` in a yarn project -- got an error about broccoli-funnel not being found

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -5756,7 +5756,26 @@ browserslist-generator@^1.0.66:
     semver "^7.3.7"
     ua-parser-js "^1.0.2"
 
-browserslist@4.20.2, browserslist@^3.2.6, browserslist@^4.0.0, browserslist@^4.14.0, browserslist@^4.14.5, browserslist@^4.20.4, browserslist@^4.21.3, browserslist@^4.21.4:
+browserslist@4.20.2:
+  version "4.20.2"
+  resolved "https://registry.yarnpkg.com/browserslist/-/browserslist-4.20.2.tgz#567b41508757ecd904dab4d1c646c612cd3d4f88"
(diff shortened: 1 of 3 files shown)


lifeart (approved):

yarn.lock
```diff
@@ -5756,7 +5756,26 @@ browserslist-generator@^1.0.66:
...
+
+browserslist@^3.2.6:
+  version "3.2.8"
+  resolved "https://registry.yarnpkg.com/browserslist/-/browserslist-3.2.8.tgz#b0005361d6471f0f5952797a76fc985f1f978fc6"
+  integrity sha512-WHVocJYavUwVgVViC0ORikPHQquXwVh939TaelZ4WDqpWgTX/FsGhl/+P4qBUAGcRvtOgDgC+xftNWWp2RUTAQ==
+  dependencies:
+    caniuse-lite "^1.0.30000844"
+    electron-to-chromium "^1.3.47"
+
+browserslist@^4.0.0, browserslist@^4.14.5, browserslist@^4.20.4, browserslist@^4.21.3, browserslist@^4.21.4:
```
lifeart, line 5778: looks like `yarn-deduplicate` needed here for 3 browserlists

ef4:
Does broccoli-node-api break if you don't change it? Because it's types-only and I don't think our own types refer to them, so it seems like it could stay a devDependency.

As for the other one, we should also figure out why we don't have a lint rule to catch this.

yarn.lock
NullVoxPopuli: thanks!

NullVoxPopuli:
yeah, eslint-plugin-node barely supports anything these days (except old cjs)

PR #1281, opened 2022-11-13
enable ember-engine tests for 4.x versions as support has landed

void-mAlex:
https://github.com/ember-engines/ember-engines/pull/798

PR #1283, opened 2022-11-15
add some steps to show how to work with tests scenarios

void-mAlex:
https://github.com/embroider-build/embroider/pull/1281#issuecomment-1315361349

happy to tweak any words as needed


SergeAstapov (approved):
This is great! Thanks you @void-mAlex!

we indeed need @ef4 to confirm these are correct and complete steps

PR #1284, opened 2022-11-15
Update lints w/ eslint-plugin-n

NullVoxPopuli:
`n/no-extraneous-import` is pretty good, and catches the error that #1282 fixes

Should this PR also replace all occurrences of eslint-plugin-node?


NullVoxPopuli:
too many conflicts

PR #1291, opened 2022-11-22
Porting @embroider/router to v2 addon format

ef4:
Since we're about to release 2.0 anyway (due to the peerDep on `@embroider/core` which is going 2.0), seems like a good time to make the switch.


packages/router/dist/index.d.ts
```diff
@@ -0,0 +1,4 @@
+import EmberRouter from '@ember/routing/router';
```
SergeAstapov, line 1: should we gitignore `packages/router/dist/` folder?

ef4: good catch. I used the v2 addon blueprint but then lifted the actual addon folder out of the generated monorepo, and that left the default gitignore behind.

PR #1301, opened 2022-12-02
Make `ensureSafeComponent` usable with Glint

chriskrycho:
Introduce a peer (and dev) dependency on `@glint/template` for the type of `ensureSafeComponent` to always return a `ComponentLike`, which is what its implementation does from the point of view of a consumer using Glint. This is an optional peer dependency, but should be installed by anyone using TS who wants to use Glint types.

Note: This would be a breaking change if our TS support story were stable; in this case it is *part of* making our support story stable.


ef4:
Thanks, this seems good to me.

PR #1306, opened 2022-12-10
Fix HTML entrypoint for module scripts

### LINKED ISSUE #1305: HTML entrypoint with script refering to package not working
According to https://github.com/simonihmig/ember-responsive-image/issues/412#issuecomment-1176574070, our `index.html` serving as webpack's entrypoint should allow referring to npm packages as module-type scripts, as in

```html
<script src="@ember-responsive-image/loader/blurhash-script" type="module"></script>
```

However this throws with `Error: Can't resolve './@ember-responsive-imageloader/blurhash-script' in '$TMPDIR/embroider/197651/test-app'`. Note two things here:
* the `./` prefix is obviously wrong 
👉 [Source](https://github.com/embroider-build/embroider/blob/main/packages/webpack/src/ember-webpack.ts#L191)
* the first found dash is removed (`@ember-responsive-imageloader` should be `@ember-responsive-image/loader`). 
👉 [Source](https://github.com/embroider-build/embroider/blob/main/packages/core/src/html-entrypoint.ts#L54)

*Preparing a PR...*

simonihmig:
Attempt to fix #1305

To Do:

- [ ] Add a test

### DIFF
--- a/packages/core/src/html-entrypoint.ts
+++ b/packages/core/src/html-entrypoint.ts
@@ -51,7 +51,7 @@ export class HTMLEntrypoint {
   }
 
   private relativeToApp(rootRelativeURL: string) {
-    return rootRelativeURL.replace(this.rootURL, '');
+    return rootRelativeURL.replace(new RegExp(`^${this.rootURL}`), './');
   }
 
   private handledScripts() {
--- a/packages/webpack/src/ember-webpack.ts
+++ b/packages/webpack/src/ember-webpack.ts
@@ -188,7 +188,7 @@ const Webpack: PackagerConstructor<Options> = class Webpack implements Packager 
     let entry: { [name: string]: string } = {};
     for (let entrypoint of entrypoints) {
       for (let moduleName of entrypoint.modules) {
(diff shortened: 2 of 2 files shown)


packages/core/src/html-entrypoint.ts
```diff
@@ -51,7 +51,7 @@ export class HTMLEntrypoint {
   }
 
   private relativeToApp(rootRelativeURL: string) {
-    return rootRelativeURL.replace(this.rootURL, '');
+    return rootRelativeURL.replace(new RegExp(`^${this.rootURL}`), './');
```
simonihmig, line 54: This introduces a change, in that passing `/assets/app.js` would now return `./assets/app.js` instead of `assets/app.js`. But this was necessary to make the `resolve()` call in the other file work, otherwise `assets/app.js` would be treated like a package and not resolve correctly. 

CI seems happy, but I wonder if that could cause any unwanted side effects, that neither CI nor me understand?

simonihmig:
Closing this, will need to be replaced with an experimental flag to opt into "input semantics" of index.html, as discussed recently in the office hours...

PR #1308, opened 2022-12-16
Resolver transform fixes

ef4:
~Two~ Several bug fixes:

 - don't bother trying to do global resolution in strict templates. They don't need it. They are natively doing what embroider works to make traditional templates do.
 - respect existing lexically scoped values passed to templates. Even non-strict templates can have things in lexical scope and those are not subject to global resolving.
 - fix the `acceptsComponentArguments` packageRule in a case where there are multiple invocations of the same component in the same template where no renaming was required to avoid collisions with the javascript scope.


ef4:
This is actually merged but GitHub seems confused.

PR #1309, opened 2022-12-16
Split resolution decisions out of babel-plugin-adjust-specifiers

ef4:
(Currently building off #1308, when that lands I can rebase for the more focused diff.)

This is a step toward moving resolution decisions out of babel entirely.


ef4:
@void-mAlex this is relevant to areas we've been talking about.

PR #1311, opened 2022-12-19
starting to move resolution into an unplugin

ef4:


packages/webpack/src/ember-webpack.ts
```diff
@@ -19,10 +19,21 @@ import {
...
-import { readFileSync, outputFileSync, copySync, realpathSync, Stats, statSync, readJsonSync } from 'fs-extra';
+import {
+  readFileSync,
+  outputFileSync,
+  copySync,
+  realpathSync,
+  Stats,
+  statSync,
+  readJsonSync,
+  readJSONSync,
```
void-mAlex, line 35: turns out `fs-extra` actually exports both of these as aliases of each other, but it's confusing to see them both imported

ef4: Oh yeah, this is just autocomplete being silly.

ef4:
Closing in favor of #1331

PR #1313, opened 2022-12-20
Update for upstream type changes

ef4:
`@types/ember__test-helpers` has been published with empty contents because you're supposed to migrate to the natively-provided types. This necessitates a series of upgrades.

The failure (which is pretty confusing, so I'm including it for future searchers) when you don't do this upgrade is:

```
error TS2688: Cannot find type definition file for 'ember__test-helpers'.
  The file is in the program because:
    Entry point for implicit type library 'ember__test-helpers'
 ```


ef4:
Upstream change was https://github.com/DefinitelyTyped/DefinitelyTyped/pull/63669

ef4:
@chriskrycho this solved my type problems that we discussed earlier. Thanks.

PR #1321, opened 2023-01-09
Upgrade resolver tests

void-mAlex:
@ef4 here are a couple of them to make sure I'm doing this correctly, the rest will follow


ef4:
Looking good, thanks.

void-mAlex:
@ef4 have gone through all skipped tests in the resolver.tests.ts file
went through 70+ tests
12 are still skipped as they don't do the expected outcome from the test and cause failures

there are several commits that look like they have the expected outcome (from the test perspective) but I have doubts about intended output of the transform step - I've marked such commits with an * at the beginning so they are easy to spot

looking for input on next steps / criteria to land this

PR #1323, opened 2023-01-10
Add static asset loader for fonts and images.

NullVoxPopuli:
People don't like configuring webpack (this is known, across all ecosystems, not just ember).

Today, we support importing CSS in JS, like :

```js
import 'some-path/to-a.css'
```
for example, which mostly does the right thing, and adds the CSS file to your module graph (doesn't make the CSS available in JS for use class-use in JS, etc)

However, this file also has fonts in it.

ember-auto-import and embroider don't currently have a font loader.

In this PR, I've added a config while working with @spuxx1701, which I believe is "just from around the internet".

It brings up some questions:
 - are there different ways we want to handle image formats?
    - low-quality-image-placeholders?
    - size optimizations?
 - are there different ways we want to handle SVG?
    - `<use>`   
    - as components?
 - for fonts, do similar questions about technique exist that we maybe can't solve for everyone?
   - if not, should this PR _only_ focus on fonts, since images and SVGs can be handled at least 5 ways each?

### DIFF
--- a/packages/webpack/src/ember-webpack.ts
+++ b/packages/webpack/src/ember-webpack.ts
@@ -214,6 +214,17 @@ const Webpack: PackagerConstructor<Options> = class Webpack implements Packager 
       module: {
         rules: [
           {
+            test: /\.(jpg|jpeg|png|woff|woff2|eot|ttf|svg)$/,
+            use: [
+              {
+                loader: 'url-loader',
+                options: {
+                  limit: 8192,
+                },
+              },
+            ],
+          },          
+          {
             test: /\.hbs$/,
             use: nonNullArray([
               maybeThreadLoader(babel.isParallelSafe, this.extraThreadLoaderOptions),


SergeAstapov:
2 cents for 1. and 2.:
we shouldn't do anything fancy in Embroider itself and simply make files available "as is".
any optimisations, different strategies should be implemented in packages like ember-svg-jar or/and ember-responsive-image

NullVoxPopuli:
ember-svg-jar and ember-responsive-image would need to provide either webpack plugins are unplugins -- which still need webpack / etc configuration -- they cannot do anything automatically -- but they could document how to do this configuration for both ember-auto-import and embroider.
But! you are right -- these things needed by these addons would likely conflict with this PR's changes as-is. 

So, we should _only_ do fonts?

SergeAstapov:
maybe we could do images but e.g. in a way that is easily "swappable" with ember-svg-jar or/and ember-responsive-image ?

NullVoxPopuli:
that would be ideal for config-resistant folks, I think -- but I'm not sure how it'd work, technically

simonihmig:
FWIW, the next-gen v2 version of ember-responsive-image will ship with custom webpack loaders, but it will also suggest users to add a custom module rule like [this](https://github.com/simonihmig/ember-responsive-image/pull/442/files#diff-79bd52d36498e1639f62dec1be8cdcc1782f4f489cbf0b5d9bbf73c8da9137ffR37-R40) that will _only_ activate its loaders when a certain query param is in the module specifier, so users can opt into the responsive-way of importing images while still keeping any other image imports the same as they are (the "virtual" module that these loaders emit are vastly different and incompatible, so using responsive loader for _all_ image imports would break things).

So I think the default for imports of _any_ static assets should be what the RFC suggests, to return the URL of that asset. If people want more, they can opt into that by adding custom loaders.

packages/webpack/src/ember-webpack.ts
```diff
@@ -213,6 +213,17 @@ const Webpack: PackagerConstructor<Options> = class Webpack implements Packager
       node: false,
       module: {
         rules: [
+          {
+            test: /\.(jpg|jpeg|png|woff|woff2|eot|ttf|svg)$/,
+            use: [
+              {
+                loader: 'url-loader',
```
simonihmig, line 220: This one is deprecated and replaced by [`asset` loader](https://v4.webpack.js.org/guides/asset-modules/) for webpack 5!

NullVoxPopuli:
Closing, as vite handles this use case very nicely, and with 0 config.

PR #1327, opened 2023-01-14
Fix Nested V2 Addons

### LINKED ISSUE #1325: Cannot read properties of undefined (reading 'registerV2Addon')
I'm hunting a NPE for quite some time now - might be related to #1108 

```
Cannot read properties of undefined (reading 'registerV2Addon')
```

I _may_ have found a fix, but cannot be sure about this 100% yet - but enough to share here already.

## The Situation

I'm upgrading [hokulea](https://github.com/hokulea/hokulea) to [`addons-v2`](https://github.com/hokulea/hokulea/tree/addons-v2). And with these addons v2 I see the error message above. Here is the monorepo and the dependency tree of relevant addons:

```
@hokulea/ember-actions-test-app
|- @hokulea/ember-actions
   |- ember-modifier
   `- @hokulea/ember-foundation
      `- ember-modifer
```

These are the noticable v2 addons, that will go through `@embroider/addon-shim`.

## The Problem

Starting `ember serve` in `@hokulea/ember-actions-test-app` shows the error, that happens when an addon v2 depends on an addon v2 depends on an addon v2 - Two levels is the "threshold" here.

Reason for this in this line:

https://github.com/embroider-build/embroider/blob/aa97453277272f533c97db174f6d4b1851e77f9d/packages/addon-shim/src/index.ts?rgh-link-date=2022-02-04T12%3A53%3A26Z#L77

which calls this method:

https://github.com/embroider-build/embroider/blob/aa97453277272f533c97db174f6d4b1851e77f9d/packages/addon-shim/src/index.ts?rgh-link-date=2022-02-04T12%3A53%3A26Z#L161-L163

Now `autoImportInstance` is set in the code block linked above. Here is the trace in prose:

1. this instance is only available to its direct children `@hokulea/ember-actions` (in the tree above). 
2. That is `@hokulea/ember-actions` has the chance to register its direct v2 addons (`ember-modifer` + `@hokulea/ember-foundation`). 
3. But at the addon instance of `@hokulea/ember-foundation` this `autoImportInstance` is `undefined`
4. `@hokulea/ember-foundation` throws the error above

Given `autoImportInstance` is a local variable not "deeply" available, which makes sense and the reason for two levels being the threshold

1. first level: instantiation
2. second level: use
3. third level+: `undefined` 💣 

## The Suggested Fix

The fix is mentioned in the comment:

https://github.com/embroider-build/embroider/blob/aa97453277272f533c97db174f6d4b1851e77f9d/packages/addon-shim/src/index.ts?rgh-link-date=2022-02-04T12%3A53%3A26Z#L75-L76

but not applied, it is not thrown "higher and higher". The fix would target `registerV2Addon()` method:

```ts
    registerV2Addon(name: string, root: string): void {
      if (autoImportInstance) {
        autoImportInstance.registerV2Addon(name, root);
      } else if (this.parent && this.parent.registerV2Addon) {
        this.parent.registerV2Addon(name, root);
      }
    },
```

Why I'm saying this is only a _suggested_ fix: I'm not sure, if this piece of code is in the intention of the comment or if that means a different implementation (given I have basically no real knowledge about embroider).

What I can say, this makes my build _passes_ this step - I have errors later on, so I cannot be sure as well.

This is ready enough for me for now to share. If my suggested fix is enough, then cool. If not a hint would be nice, so the next time I'm working on this, I can continue digging into it.

gossi:
Fix #1325 

please refer to linked issue for detailed explanation. I didn't add any tests though, there is no such scenario in the repo. Please let me know how a good test would fit within the existing suite.

### DIFF
--- a/packages/addon-shim/src/index.ts
+++ b/packages/addon-shim/src/index.ts
@@ -158,8 +158,12 @@ export function addonV1Shim(directory: string, options: ShimOptions = {}) {
       return isInside(directory, appInstance.project.root);
     },
 
-    registerV2Addon(name: string, root: string): void {
-      autoImportInstance!.registerV2Addon(name, root);
+    registerV2Addon(this: AddonInstance, name: string, root: string): void {
+      if (autoImportInstance) {
+        autoImportInstance.registerV2Addon(name, root);
+      } else {
+        (this.parent as EAI2Instance).registerV2Addon(name, root);
+      }
     },
   };
 }


packages/addon-shim/src/index.ts
```diff
@@ -158,8 +158,12 @@ export function addonV1Shim(directory: string, options: ShimOptions = {}) {
       return isInside(directory, appInstance.project.root);
     },
 
-    registerV2Addon(name: string, root: string): void {
-      autoImportInstance!.registerV2Addon(name, root);
+    registerV2Addon(this: AddonInstance, name: string, root: string): void {
+      if (autoImportInstance) {
+        autoImportInstance.registerV2Addon(name, root);
+      } else {
+        (this.parent as EAI2Instance).registerV2Addon(name, root);
```
NullVoxPopuli, line 165: How do we know it's safe to call this on the parent?

gossi: That's the biggest question mark I have, quote from the issue above:

> Why I'm saying this is only a suggested fix: I'm not sure, if this piece of code is in the intention of the comment or if that means a different implementation (given I have basically no real knowledge about embroider).

which basically targets this line.

But following the analysis I put in the issue, there is either a `autoImportInstance` or `this.parent` with `registerV2Addon` available.

gossi: Potentially, you can safeguard this. But I'd say, it would be better to throw an error over silently do nothing and not registering the addon at all, without a dev knowing about it ?!?

gossi:
@ef4 is this ready to move forward? It is improving and at the same time doesn't make it worse?

packages/addon-shim/src/index.ts
ef4: It's safe once refactored inside the `if (this.parent.pkg['ember-addon']?.version ?? 1) < 2)` check.

ef4:
The test coverage for this is in ember-auto-import, not embroider, because all of this registering v2 addons with ember-auto-import stuff is inert under embroider, it only matters for non-embroider apps where ember-auto-import is doing the work.

Opened https://github.com/ef4/ember-auto-import/pull/571 to cover this.

ef4:
Thanks @gossi

wagenet:
Has this made it into a release yet?

njradford:
Also encountering this issue fwiw

PR #1328, opened 2023-01-16
test-setup: use caret version modifier for Embroider dependencies

### LINKED ISSUE #1268: Missing helper: unique-id with embroider-optimized
In a fresh `ember-cli@4.8.0` addon adding `unique-id` to `application.hbs`
```hbs
{{#let (unique-id) as |templateId|}}
  <h2 id="{{templateId}}-title">Welcome to Ember</h2>
{{/let}}
```
results in `Missing helper: unique-id in templates/application.hbs` when running `ember try:one embroider-optimized`.

simonihmig:
When all monorepo packages were released in-sync, this ensured `test-setup` is referring always to the latest. But with the release process decoupled now, `test-setup` would effectively pull in an older pinned version of Embroider. This adds the caret modifier to allow pulling the latest version (on the same major).

Fixes #1268.

### DIFF
--- a/packages/test-setup/src/index.ts
+++ b/packages/test-setup/src/index.ts
@@ -5,7 +5,7 @@ import type { Webpack } from '@embroider/webpack';
 type EmberWebpackOptions = typeof Webpack extends PackagerConstructor<infer Options> ? Options : never;
 
 // eslint-disable-next-line @typescript-eslint/no-require-imports
-const currentEmbroiderVersion = require('../package.json').version;
+const currentEmbroiderVersion = `^${require('../package.json').version}`;
 
 /*
   Use this instead of `app.toTree()` in your ember-cli-build.js:


simonihmig:
Tbh, I haven't tested this, given that there are no tests and no straightforward way to test this manually (is there?). But the change seems small enough to be confident it is ok, hopefully?

simonihmig:
Yeah, but you can already pass customisation options to specify the exact versions of dependencies that you want, as I did [here](https://github.com/CrowdStrike/ember-headless-form/pull/14/commits/ed1f6c1a56acfaf9695a59c693b7d7d5f0db89e8) to workaround that issue for now.

The new arguments (`'latest'`) you proposed here would be a breaking change, not sure if that is worth it?

I think we should primarily decide on what the best default behaviour would be: latest or same major, not? 
If people want to override that, then they already can...

> which are all co-released, afaict

Well, but that's just how it turned out so far, that's not set in stone, right? I don't think we should rely on that.

NullVoxPopuli:
> The new arguments ('latest') you proposed here would be a breaking change, not sure if that is worth it?

nah, no breaking change, just an overload.

> I think we should primarily decide on what the best default behaviour would be: latest or same major, not?

latest makes sense as a default to me

>  If people want to override that, then they already can...

yeah, but it's quite a few extra lines to do so

> Well, but that's just how it turned out so far, that's not set in stone, right? I don't think we should rely on that.

yeah -- no clue if it'll stay that way, and that would be the defining limitation of what I proposed

NullVoxPopuli (approved):
better than pinning, imo!

though, this could be a breaking change

PR #1331, opened 2023-01-19
Move resolving into dedicated plugins

ef4:
Supersedes #1311.

While this refactor is landing a webpack plugin, I have already tested that the same shared module-resolver API is fully working under rollup and vite.


ef4:
Goals as we land this:
 - [ ] move all app tree merging into the module-resolver rules, so no physical rewriting of the app is necessary
 - [ ] redirect resolution to and from auto-upgraded addons so we don't need to physically recreate the node_modules tree (this would simplify the existing case where v2 addons depend on emberVirtualPeerDeps)
 - [ ] refactor `class CompatResolver` into separate steps: the first would be a one-to-one transformation of the template that always emits standardized imports, the second would be additional rules in the module-resolver that satisfies those imports based on what's actually available.

ef4:
The unchecked "goals" list above is things to do in the next PR.

PR #1334, opened 2023-01-24
Allow `ember-cli-babel` >= v8 for apps and addons

bertdeblock:


packages/router/package.json
```diff
@@ -70,7 +70,7 @@
     "prettier": "^2.5.1",
     "rollup": "^2.67.0",
     "rollup-plugin-copy": "^3.4.0",
-    "rollup-plugin-ts": "^3.0.2",
+    "rollup-plugin-ts": "~3.1.1",
```
bertdeblock, line 73: Temp to see if CI will pass

ef4:
Since the point of this check is to make sure the app is using babel 7, when we encounter ember-cli-babel 8 I think we need to actually check the app's version of `@babel/core` itself.

```js
if (app has ember-cli-babel 7) {
  we're definitely using @babel/core 7
} else if (app has ember-cli-babel > 7) {
  we need to check what version of @babel/core is in the app
}
```

ef4:
Released in `@embroider/compat` 2.2.1.

Issue #714, opened 2021-03-08
Generated output with `staticHelpers: true` seems invalid


mydea:
Forked this out from https://github.com/stefanpenner/ember-strict-resolver/pull/45

When building an app with strict helpers, I noticed that e.g. `require('helper:page-title')` would not return the normal module with a default property, but directly return the default export itself (!?). Not sure what exactly is going on, but after some digging I noticed this small difference in output:

## With `staticHelpers: true`

```js
__webpack_require__.r(__webpack_exports__);
/* harmony export (binding) */
__webpack_require__.d(__webpack_exports__, 'default ', function () {
  return PageTitle;
});
```

## With `staticHelpers: false`

```js
__webpack_require__.r(__webpack_exports__);
/* harmony export (binding) */
__webpack_require__.d(__webpack_exports__, 'default', function () {
  return PageTitle;
});
```

Notice the space after default in the first example.

rwjblue:
That `'default '` string seems wrong to me, but I'm not sure exactly internally where we handle emitting those modules.

ef4:
That certainly looks weird, but I will point out that using `require` to access helpers when `staticHelpers` is on is highly likely to break.

Right now if you try it after a template has evaluated that uses the given helper, it will work. If you try it before any template that uses the helper evalutes, it will not.

In upcoming releases we will take advantage of ember's template lexical scope and then it will never work -- the helper will never be registered with the runtime AMD loader.

rwjblue:
@ef4 - This is the resolver resolving the helper, is there some override for `ember-resolver` in these cases?

ef4:
No, it does go through the resolver internally. I just mean people shouldn't rely on looking in requirejs.entries themselves.

rwjblue:
> I just mean people shouldn't rely on looking in requirejs.entries themselves.

Yes, totally agreed!

mydea:
Yes, to clarify, that has nothing to do with manually using `require()`, it has the same output with just a regular app using staticHelpers, even with the default ember-resolver. Only that ember-resolver seems to handle this by also allowing the module returned from `require()` to be used directly (instead of `module['default']`), which makes this work, although I am still not sure why that would even work that way...

PR #1340, opened 2023-01-26
Static helpers and static modifiers transform fixes

void-mAlex:
@ef4 I've hopefully correctly fixed some of the code that we previously discovered was misbehaving
this should I hope correctly scope helpers and modifiers when their respective options are turned on
I've also added 4 tests for bare mustache invocation of modifiers as I wasn't able to find other relevant tests to cover that see 31e7e2584cd3a7cb49091e1670023a130045adf1

quickly scanning through open issues this one may seem relevant https://github.com/embroider-build/embroider/issues/714

### DIFF
--- a/packages/compat/src/resolver-transform.ts
+++ b/packages/compat/src/resolver-transform.ts
@@ -5,6 +5,7 @@ import {
   ResolutionFail,
   Resolution,
   ResolvedDep,
+  HelperResolution,
 } from './resolver';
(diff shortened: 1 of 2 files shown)


void-mAlex:
I've ran both failing scenarios locally with no failures, could we rerun ci?

ef4 (approved):
Nice work, this looks good.

Issue #1341, opened 2023-01-27
Error importing component classes (TypeScript)


esbanarango:
### 🐞 Bug description
Importing a component class ends up being a different object than the class itself. Preventing extending it or interacting with it.

_**This works fine without using TypeScript**_

### 👨‍🔬 Example
Trying to stub some properties or methods of a _Component_ class.

```ts
// This is no longer a valid/normal Class as Embroider has converted into a different (owner) object.
import TestComponent from 'owner-embroider-issue/components/test-component';

module('Integration | Component | test-component', function (hooks) {
  setupRenderingTest(hooks);

  test('...', async function (assert) {
    this.owner.register(
      'component:test-component',
      class extends TestComponent {
        // Propertier and methods I want to override
      }
    );

    await render(hbs`<TestComponent />`);
  });
});
```
When doing: `import TestComponent from 'owner-embroider-issue/components/test-component';`, `TestComponent` is no longer a Class, but a different "object", so we can not "extend" it anymore.

### 😕 Actual Behavior
<img width="1570" alt="215009088-21f31d51-9678-4d68-9d91-cbd6a75d890b_png__2532×1606__🔊" src="https://user-images.githubusercontent.com/1156865/215010662-36bea68d-1a77-4aa4-a8d6-0eb1533e30f0.png">
<img width="946" alt="✖_OwnerEmbroiderIssue_Tests_🔊" src="https://user-images.githubusercontent.com/1156865/215009456-97fd6307-01f2-40e9-9c66-a06fa7ba2aed.png">

### 🤔 Expected Behavior
This is the output without using Typescript
<img width="1244" alt="Cursor_and_OwnerEmbroiderIssue_Tests_🔊" src="https://user-images.githubusercontent.com/1156865/215010353-fb1fc433-1283-4a35-b0b8-a0d6fc48283d.png">

### 🌍 Environment
```json
"@embroider/compat": "^2.0.2",
"@embroider/core": "^2.0.2",
"@embroider/webpack": "^2.0.2",
"ember-source": "~4.10.0",
"ember-cli": "~4.10.0",
"ember-cli-typescript": "^5.2.1",
"typescript": "^4.9.4",
"webpack": "^5.75.0"
```

### 🔀 App to reproduce the issue

https://github.com/esbanarango/owner-embroider-issue/tree/typescript

dmuneras:
Got the same issue :(

ef4:
[...]

Issue #1336, opened 2023-01-25
Errors with embroider and ember-table / ensure-safe-component / TypeScript


rwwagner90:
I am hitting some very specific error cases where when I try to pass a `cellComponent` to ember-table, which is wrapped in `ensure-safe-component`, I get `Uncaught Error: Attempted to load a component, but there wasn't a component manager associated with the definition. The definition was: factory`.

If I step through the error, it hits a couple lines inside ember-table where it is registering a container.

This error only occurs if the component I am trying to use as a `cellComponent` has a TypeScript backing class file. If I remove the TS file and use JS instead or have no backing class, the error goes away.

I made a reproduction repo here for debugging:

https://github.com/rwwagner90/embroider-bug

NullVoxPopuli:
Something I noticed while debugging this:
I turned all the static flags off, got this error:
```

Error occurred:

- While rendering:
  -top-level
    application
      table-wrapper
        ember-table
          ember-tbody
            vertical-collection
              ember-table-private/row-wrapper
                ember-tr
                  ember-td

Uncaught Error: Attempted to load a component, but there wasn't a component manager associated with the definition. The definition was: factory
```

With the static flags on, the error is the same, except the template stack is terrible:
```


Error occurred:

- While rendering:
  -top-level
    application
      TableWrapperComponent
        Class
          Class
            Class
              Class
                Class
                  Class
```

NullVoxPopuli:
I removed ensure-safe-component (with static flags all off, so testing more compatibility than strictness):

and the error persists:
```
Uncaught Error: Attempted to load a component, but there wasn't a component manager associated with the definition. The definition was: factory
```
```hbs
      <r.cell class='h-10' as |cell column row cellMeta|>
        {{#if column.cellComponent}}
          {{log column.cellComponent}}
          {{#let (component column.cellComponent) as |Wrapper|}}
            {{log Wrapper}}
            <Wrapper>
              {{cell}}
            </Wrapper>
          {{/let}}
        {{else}}
          {{cell}}
        {{/if}}
      </r.cell>
```

NullVoxPopuli:
<details><summary>The compiled file with backing TS class</summary>

```js
import { setComponentTemplate } from '@ember/component';
import { precompileTemplate } from '@ember/template-compilation';
import Component from '@glimmer/component';

[...]

PR #1342, opened 2023-01-27
restore original ts extension priority

ef4:
I don't really know what https://github.com/embroider-build/embroider/commit/7882a2da97ce31f720dc7eabb8bef7d149a5b18b was trying to achieve and it definitely looks weird and it's the cause of #1336 and probably #1341.

Maybe CI will tell me why this was done in the first place...


ef4:
I think I see now. It was bad to have TS before JS (because then libraries that ship both would try to recompile their TS). But it's also bad to have TS after HBS (because when a component has both, it's supposed to be the TS that represents the component).

esbanarango:
🙏

esbanarango:
Thank you 🤝

PR #1343, opened 2023-01-28
Fix `@babel/core` version check to support `ember-cli-babel` v8

bertdeblock:


packages/compat/src/v1-app.ts
```diff
@@ -287,14 +287,14 @@ export default class V1App {
...
-      let babelMajor = babelAddonMajor;
+      let babelMajor: number | undefined = babelAddonMajor;
       if (babelAddonMajor >= 8) {
         // `ember-cli-babel` v8 breaks lockstep with Babel, because it now
         // defines `@babel/core` as a peer dependency, so we need to check the
         // project's version of `@babel/core`:
         let babelVersion = this.app.project.pkg.devDependencies?.['@babel/core'];
         if (babelVersion) {
-          babelMajor = Number(babelVersion.split('.')[0]);
+          babelMajor = semver.coerce(babelVersion)?.major;
```
bertdeblock, line 297: I stupidly only tested this with a pinned version of `@babel/core`. This was failing when using a range (`^` or `~`).

PR #1344, opened 2023-01-29
Delete files that are now in the blueprint

NullVoxPopuli:
folks should be using the addon-blueprint here: https://github.com/embroider-build/addon-blueprint


SergeAstapov:
Should we update docs referencing those deleted files?

NullVoxPopuli:
That's a good call out :thinking: 

I'm now wondering if the docs should be updated to use the blueprint :thinking:

PR #1346, opened 2023-01-30
Bugfix: inconsistent handling of webpack virtual modules

ef4:
This should match the pattern we use when emitting virtual modules.

It works *sometimes* as it was, but not reliably. This change seems to make the tests pass every time.


ef4:
I think we're having sporadically bad interactions between webpack-virtual-modules and thread-loader. I *had* reproduced the problem without thread-loader present so I thought I ruled that out, but apparently not.

PR #1348, opened 2023-02-01
Move docs into a docs folder

mansona:
Considering that we're likely going to be writing a lot more documentation over the coming months, I thought it was about time that we move all the random docs hanging around into a nice and neat docs folder 👍


void-mAlex:
in my own opinion this would make then way less accessible to people, I've gotten to click through other .md docs because I just opened the repo and they were there to click on
there is great value to open find function in the browser and type in `addon` and find the document related to that from the main page because it's just there, I personally feel there would be more to lose by moving them than there is to gain from organizational purposes
I hear what you're saying about more documentation, but I thought I would offer this other perspective as well

NullVoxPopuli:
there are also a bunch of places linking to the docs where they are now. -- since we don't have the ability to redirect on GitHub, we'd break all those links.

README.md
```diff
@@ -27,7 +27,7 @@ This repo implements a new three-stage build system for Ember apps:
    and 3 improves our ability to innovate and experiment with taking the best
    parts of wider JS ecosystem tooling.
 
-You can read more about the motivation and key ideas in the [intro to the SPEC](SPEC.md).
+You can read more about the motivation and key ideas in the [intro to the SPEC](docs/spec.md).
```
void-mAlex, line 30: also referenced in https://github.com/embroider-build/embroider/blob/3a232cc04a68f86d0acfa5e7edfa764b990efdbf/packages/macros/README.md?plain=1#L9

void-mAlex (changes requested):

void-mAlex (approved):

ef4 (approved):

Issue #1338, opened 2023-01-26
Module not found: Error: Can't resolve '@ember/test-helpers'


mansona:
This issue has hit me for a little while, and after our last Embroider meeting, I am finally at the point where I feel confident enough to open the issue about this 😂 still quite a while until I understand how to fix the problem though 😭 

I'm going to say it from the start because I think it's relevant to the problem: We are using `pnpm` in a monorepo when I'm seeing this problem. 

The error that I've been facing is the following: 

```
Build Error (PackagerRunner) in ../../addons/addon-name/test-support/accordion.js

Module not found: Error: Can't resolve '@ember/test-helpers' in '$TMPDIR/embroider/62cd22/addons/addon-name/test-support/accordion.js'
```

**Note:** I  have replace the real addon name with `<addon-name>` as it's an internal addon 👍 

When discussing this in the #embroider-dev channel on discord the first thing that people suggested was for me to "make sure that I have the same version of `@ember/test-helpers` everywhere" so I run `pnpm why @ember/test-helpers` and get the following output: 

```
devDependencies:
@ember/test-helpers 2.9.3
ember-a11y-testing 5.0.0
└── @ember/test-helpers 2.9.3 peer
ember-qunit 6.0.0
└── @ember/test-helpers 2.9.3 peer
```

so that's totally fine.

When looking at the folder that error says it can't resolve the dependency `$TMPDIR/embroider/62cd22/addons/addon-name/test-support/` and following the node resolution rules I find that the place it would try to import from is `$TMPDIR/embroider/62cd22/addons/addon-name/node_modules/@ember/test-helpers` and running `ls` in that folder I see that it's mostly empty: 

```
➜  test-helpers ls
package.json
```

and the contents of the package.json are incredibly sparse i.e. just the name field: 

```
➜  test-helpers cat package.json
{"name":"@ember/test-helpers"}
```

After the embroider meeting it's clear that this empty package is being exported by the `build-compat-addon` file in the `@embroider/compat` package: https://github.com/embroider-build/embroider/blob/main/packages/compat/src/build-compat-addon.ts#L37 

Now that I know where the problem was coming from I dropped a debugger in there to find out **why** the issue was happening.

The key to this problem is related to the `v1Cache` and the line `let oldPackages = v1Cache.getAddons(originalPackage.root);`

[...]

PR #1350, opened 2023-02-01
add a comment to the empty package key generator

mansona:
We discussed doing this to side-step the issue described in https://github.com/embroider-build/embroider/issues/1338 at least giving people a breadcrumb to follow if they come across the same issue

I have included a link to the file that I'm adding in #1349 so this is blocked on that PR getting merged. I don't know if they need to be 2 separate PRs but it felt natural so I thought I would do it that way 🤷


packages/compat/src/empty-package-tree.ts
```diff
@@ -15,6 +15,7 @@ export default class extends Plugin {
   build() {
     if (!this.built) {
       writeJSONSync(join(this.outputPath, 'package.json'), {
+        '//': 'This empty package was created by embroider. If you are having issues resolving this package and you have followed your dependency tree to this file you could be experiencing an issue with your dependencies. See https://github.com/embroider-build/embroider/blob/main/docs/pnpm-injected.md for more information or open an issue in the embroider repo with a recreation of your problem.',
```
NullVoxPopuli, line 18: do folks using yarn or npm have a path to resolve here?

or is it:
- if you're using `yarn@1`. Don't. Use `yarn@3`
- if you're using `npm@9`... ?
- if you're using `pnpm@7`, use `dependenciesMeta.*.injected`

ef4: Both the PR and this comment are over-focused on package managers here. This comment should not be about package managers at all. There are multiple reasons somebody could have an empty package, and this should only point to a document about that without assuming one of the reasons or assuming what the fix is.

mansona: So my thoughts are that this comment should be pretty generic and that the thing that it links to should have all the caveats and differences there.

Maybe it's important to make the markdown file not have pnpm in the name then? maybe `docs/injected-dependencies.md`? what do you think?

ef4: It's not about injected dependencies either. See https://github.com/embroider-build/embroider/pull/1349#issuecomment-1412542979

PR #1354, opened 2023-02-09
Add glint helper types for more macros

vlascik:
Adds more glint types for the macros.

Not entirely sure about the accuracy of these though, so probably should be checked a bit more closely before merging.

### DIFF
--- a/packages/macros/src/index.ts
+++ b/packages/macros/src/index.ts
@@ -35,12 +35,12 @@ export function importSync(specifier: string): unknown {
   throw new Oops(specifier);
 }
 
-export function getConfig<T>(packageName: string, ...keys: string[]): T {
-  throw new Oops(packageName, keys);
+export function getConfig<T>(packageName: string): T {
+  throw new Oops(packageName);
 }
 
-export function getOwnConfig<T>(...keys: string[]): T {
-  throw new Oops(keys);
+export function getOwnConfig<T>(): T {
+  throw new Oops();
 }
 
 export function getGlobalConfig<T>(): T {
--- a/packages/macros/src/index.ts
+++ b/packages/macros/src/index.ts
@@ -59,10 +59,6 @@ export function failBuild(message: string): void {
   throw new Oops(message);
 }
 
-export function maybeAttrs<T>(predicate: boolean, ...bareAttrs: unknown[]): T {
-  throw new Oops(predicate, bareAttrs);
-}
-
 export function moduleExists(packageName: string): boolean {
   throw new Oops(packageName);
 }
--- a/packages/macros/src/index.ts
+++ b/packages/macros/src/index.ts
@@ -35,12 +35,12 @@ export function importSync(specifier: string): unknown {
   throw new Oops(specifier);
 }
 
-export function getConfig<T>(packageName: string): T {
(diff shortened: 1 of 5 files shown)


packages/macros/src/index.ts
```diff
@@ -35,12 +35,12 @@ export function importSync(specifier: string): unknown {
   throw new Oops(specifier);
 }
 
-export function getConfig<T>(packageName: string): T {
-  throw new Oops(packageName);
+export function getConfig<T>(packageName: string, ...keys: string[]): T {
```
ef4, line 38: These are the types for the Javascript macros, not the template macros. The Javascript `getConfig` doesn't take any `keys` argument (because it works with regular property access and optional chaining instead).

So probably this change should go in `class macroGetConfig extends Helper` instead, which would change to declare its own type and not rely on this type, since they're supposed to be different.

packages/util/index.d.ts
```diff
@@ -15,3 +16,43 @@ export class EnsureSafeComponentHelper<
...
+
+export interface EmbroiderMacrosRegistry {
+  'ensure-safe-component': typeof EnsureSafeComponentHelper;
+  macroGetOwnConfig: typeof macroGetOwnConfig;
+  macroGetConfig: typeof macroGetConfig;
+  macroCondition: typeof macroCondition;
+  macroDependencySatisfies: typeof macroDependencySatisfies;
+  macroMaybeAttrs: typeof macroMaybeAttrs;
+  macroFailBuild: typeof macroFailBuild;
+}
```
ef4, line 58: These types look good for the template macros but they should go in `@embroider/macros` not `@embroider/util`.

ef4 (changes requested):

packages/macros/src/index.ts
```diff
@@ -59,6 +59,10 @@ export function failBuild(message: string): void {
   throw new Oops(message);
 }
 
+export function maybeAttrs<T>(predicate: boolean, ...bareAttrs: unknown[]): T {
```
vlascik, line 62: Not sure if this function belongs here and was just omitted by mistake?

packages/macros/src/index.ts
vlascik: removed these parameters and use them only in template types

packages/util/index.d.ts
vlascik: ok, moved these to packages/macros/src/template-registry.ts, not sure if that's the right spot, but it didn't seem like index.ts would be a proper fit.

Also had to rename all helpers definitions to PascalCase because of an ESLint rule, don't think it matters, registry names them in camelCase as they appear in templates.

packages/macros/src/index.ts
ef4: I think this is incorrect. This file is the Javascript macros, not the Handlebars macros.

vlascik: So `maybeAttrs` is just a handlebars macro and it should be removed from this file completely, is that correct?

vlascik: Anyway, I removed maybeAttrs from the JS macros file, so I guess the PR is ready for the review

ef4 (approved):

PR #1357, opened 2023-02-14
App tree resolving

ef4:
In-progress work to replace on-disk app tree merging with in-resolver app tree merging.

At this step I have disabled (but not cleanly removed) the old behavior so I can work through the issues that are revealed in the tests.


ef4:
A lot of the tests failures are just tests that need to be refactored so they don't assume the on-disk file merging will be present. They can all use the newer setupAuditTest infrastructure instead.

ef4:
Closed in favor of #1373

PR #1365, opened 2023-02-24
Prevent HTML-escaping of module specifiers

simonihmig:
Switching a pnpm-based repo to its newer `lockfileVersion: '6.0'` caused `node_modules` to have folders containing an `=` equal sign (for patched packages), which breaks [the build](https://github.com/CrowdStrike/ember-headless-form/actions/runs/4261156528/jobs/7415828786#logs) as that equal sign gets wrongly HTML-escaped to `&#x3D;`, causing module not found errors. 

This PR basically replicates the same [fix as for ember-auto-import](https://github.com/ef4/ember-auto-import/pull/562). 

I was able to confirm - by patching the local file in `node_modules` with the same changes as in this PR - that it fixes the build issue I had.


packages/core/src/js-handlebars.ts
```diff
@@ -11,4 +11,7 @@ handlebars.registerHelper('json-stringify', function (input: any, indent?: numbe
 handlebars.registerHelper('eq', function (a: any, b: any) {
   return a === b;
 });
-export const compile = handlebars.compile;
+
+export function compile(input: string) {
```
simonihmig, line 15: Technically, this changes the signature of this function, as you _could_ previously pass custom options. But this was nowhere the case AFAICT. And it _seems_ better to me to not expose this, but own the config internally here.

I can change that of course in case of disagreement!

simonihmig:
Failing test scenarios seem to be related to https://github.com/embroider-build/embroider/issues/1366

ef4:
Previously I had done this in each place it came up by using triple curlies. I like your solution better, but we should remove all the triple curlies too so it's not confusing.

ef4:
Thanks for working on this, I think it's ready to go.

Issue #1366, opened 2023-02-24
CI failing due to network/resources issues


simonihmig:
Is this known already?

CI jobs, even scheduled ones on `main`, are failing in `actions/setup`, e.g.
* Windows: [Could not download node@14.19.3 - A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond. (os error 10060)](https://github.com/embroider-build/embroider/actions/runs/4260139922/jobs/7413046743#step:3:28)
* Linux: [522 error downloading node.js](https://github.com/embroider-build/embroider/actions/runs/4260139922/jobs/7413045429#step:3:28)
* Linux: [Could not unpack Node v14.19.3 - Resource temporarily unavailable (os error 11)](https://github.com/embroider-build/embroider/actions/runs/4260139922/jobs/7413046743#step:3:28)

PR #1368, opened 2023-02-25
Add support for keeping public assets and ember-addon.public-assets meta in sync

phndiaye:
While migrating an addon to the v2 format, I noticed that we need to declare the files from the `public` directory in the `ember-addon.public-assets` meta in the package.json, which is easy if there isn't much of them, but can get out of hand quickly (I have a couple addons providing a lot of assets to the parent app)

This adds support for keeping an addon's package.json `ember-addon.public-assets` meta in sync with the `public` directory by default.

By passing `opts.exclude` (glob pattern array), you can still ignore some of the files if needed.

```javascript
export default {
  output: addon.output(),

  plugins: [
    addon.publicEntrypoints([]),

   // Other plugins declared in the samples

    addon.publicAssets('assets'), // Or addon.publicAssets('assets', { include: ['**/*.svg'], exclude: ['**/*.png'] }) to skip all PNG files and include SVG ones
  ]
};

```

### DIFF
--- a/packages/addon-dev/src/rollup-public-reexports.ts
+++ b/packages/addon-dev/src/rollup-public-reexports.ts
@@ -7,13 +7,10 @@ export default function publicAssets(opts: { exclude: string[] }): Plugin {
     name: 'public-assets-bundler',
     generateBundle() {
       let pkg = readJsonSync('package.json');
-      let filenames;
-
-      filenames = walkSync('public', {
+      const filenames = walkSync('public', {
         directories: false,
-        ignore: opts.exclude || [],
+        ignore: opts?.exclude || [],
       });
-
       const publicAssets: Record<string, string> = filenames.reduce(
         (acc: Record<string, string>, v): Record<string, string> => {
           acc['./public/' + v] = ['/', pkg.name, '/', v].join('');
--- a/packages/addon-dev/src/rollup-public-reexports.ts
+++ b/packages/addon-dev/src/rollup-public-reexports.ts
@@ -1,0 +1,32 @@
+import { readJsonSync, writeJsonSync } from 'fs-extra';
+import walkSync from 'walk-sync';
+import type { Plugin } from 'rollup';
+
+export default function publicAssets(opts: { exclude: string[] }): Plugin {
+  return {
+    name: 'public-assets-bundler',
+    generateBundle() {
+      let pkg = readJsonSync('package.json');
+      let filenames;
+
+      filenames = walkSync('public', {
+        directories: false,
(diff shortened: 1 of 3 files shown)


packages/addon-dev/src/rollup-public-reexports.ts
```diff
@@ -0,0 +1,29 @@
+import { readJsonSync, writeJsonSync } from 'fs-extra';
+import walkSync from 'walk-sync';
+import type { Plugin } from 'rollup';
+
+export default function publicAssets(opts: { exclude: string[] }): Plugin {
+  return {
+    name: 'public-assets-bundler',
+    generateBundle() {
+      let pkg = readJsonSync('package.json');
+      const filenames = walkSync('public', {
```
phndiaye, line 10: To my knowledge, the `public` folder is not something that's configurable. Otherwise, this could be part of the options (w/ a default value)

simonihmig (commented):
I think that's a good feature to have!

Although the `public` folder is IMHO a bit of a remnant of the old (push-based) world. We might instead recommend people to the pull-based world in the future, when we have a default setup for importing static assets (related: https://github.com/embroider-build/embroider/pull/1323, https://github.com/emberjs/rfcs/pull/763). But as we have that feature currently, we can also make it more ergonomic for sure...

ef4:
This seems good and I agree with Simon's feedback. 

The one thing I'll point out is that we hope to not make addons use this for most cases. It's actually pretty problematic if addons are putting things in here that the addon plans to access at runtime, because the connection between the consumption site and the asset is not statically visible. It means that people run into trouble when they deploy assets to a CDN or adjust fingerprinting rules, etc, because the arbitrary runtime code needs to know about those changes.

We want to instead allow code to explicitly ask for an asset's URL. In embroider apps today people often use the native webpack 5 asset handling to import the URL of an asset:

```js
import iconURL as "./icon.png";
```

And that works well but it's not really a standard and we'd like to come up with something less implementation-specific. I'm hoping the import assertions spec proposal actually moves forward again at TC39, because that would create a web-standards-compliant way to do this.

ef4:
Even for shared branding assets I think we'd want people to use "the new way". Meaning, users of your addon can either import the assets directly out of the addon, or the addon can provide components that directly import the assets.

But yeah, since there's no standardized "new way" right now I'm find with moving forward this PR. I can be a document option that's not enabled by default.

phndiaye:
Looks like the CI is failing w/ this issue https://github.com/embroider-build/embroider/issues/1366 :/

PR #1369, opened 2023-02-28
Add semverRange <=4.11.0 for ember-data

### LINKED ISSUE #1322: can't locate @ember-data/record-data/-private referred to in module rules
I'm running tests in ember-try `embroider-safe` and we're getting the following error: 

```
Build Error (WaitForTrees)

can't locate @ember-data/record-data/-private referred to in module rules:{
  "dependsOnModules": [
    "@ember-data/record-data/-private"
  ]
}
```

I've done some debugging and I've found the error that is being thrown in the `resolver.js`

```
'Cannot find module '@ember-data/record-data/-private' from '/private/var/folders/r5/bw5frvnn5k794hk62w3vp7km0000gn/T/embroider/2daa8e/node_modules/.pnpm/@ember-data+store@4.8.4_k6kp2ds5o4a3bn46aja3did5ee/node_modules/@ember-data/store''
```

I had a look at the file on disk and it looks like it's right, there was no sibling to `@ember-data/store` for `record-data` and normal node resolution would be expected to fail.

mkszepp:
fix #1322

Starting with ember-data `v4.11.1` there is not anymore necessary to add package rules (https://github.com/emberjs/data/pull/8427)

### DIFF
--- a/packages/compat/src/addon-dependency-rules/ember-data.ts
+++ b/packages/compat/src/addon-dependency-rules/ember-data.ts
@@ -3,6 +3,7 @@ import { PackageRules } from '..';
 let rules: PackageRules[] = [
   {
     package: '@ember-data/store',
+    semverRange: '<=4.11.0',
     addonModules: {
       '-private.js': {
         dependsOnModules: ['@ember-data/record-data/-private'],
--- a/packages/compat/src/compat-adapters/ember-data.ts
+++ b/packages/compat/src/compat-adapters/ember-data.ts
@@ -3,6 +3,7 @@ import { join } from 'path';
 import { Memoize } from 'typescript-memoize';
 import { Node } from 'broccoli-node-api';
 import { sync as resolveSync } from 'resolve';
+import semver from 'semver';
 
 export class EmberDataBase extends V1Addon {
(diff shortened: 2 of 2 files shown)


ef4:
Thanks. Does this means we should be disabling the compat adapters too? There are six ember-data-related ones:

https://github.com/embroider-build/embroider/blob/01575a4d3a86b8b329c75024aa77b2539a6dad5d/packages/compat/src/compat-adapters/ember-data.ts
https://github.com/embroider-build/embroider/tree/main/packages/compat/src/compat-adapters/%40ember-data

If so, here's an example of how a compat adapter can turn off based on the addon version:

https://github.com/embroider-build/embroider/blob/01575a4d3a86b8b329c75024aa77b2539a6dad5d/packages/compat/src/compat-adapters/ember-inflector.ts#L34-L36

mkszepp:
@ef4 yes we can disable all adapters, added this lines in PR

runspired:
fwiw probably this is true 4.8+. I think a few of the packages maybe weren't fully converted for 4.8 I'd have to go refresh my memory but it's at least true that far far less was needed for that release. Considering it was LTS we may want to also account for it.

PR #1373, opened 2023-03-07
app tree resolving

ef4:
This supersedes #1357 now that we've landed several prerequisites and can do it properly in one shot.

First step here is to implement app-tree resolving module-resolver.


ef4:
Latest note-to-self here: the latest commit streamlined how entry points get built but that is going to cause several tests to need to be updated, where the tests were making assumptions about the entrypoint formatting.

PR #1375, opened 2023-03-15
reduce coupling between stage1 and stage2 in preparation for moving t…

ef4:
…heir responsibilities into separate processes


ef4:
Superseded by #1435.

Issue #1382, opened 2023-03-28
`{{ensure-safe-component}}` not working with template tags


simonihmig:
We have an interoperability issue here:
* the `resolver-transform` checks for the existence `ensure-safe-component` helper by a simple [string comparison](https://github.com/embroider-build/embroider/blob/936fd6373339a6848260f6c0a52f426127023299/packages/compat/src/resolver-transform.ts#L233), allowing for the helper to be used _only_ when having exactly that dasherized name
* in a template tag environment, we need to import the helper instead of referring to it by that global string reference. But we cannot import it or alias it using that name, as dashes are not valid identifiers in JS scope

> Note: in my use case there is currently no way to _not_ use `ensure-safe-component`, as I need currying and the curried component is not resolvable using a string. And we don't have a variant of the component helper that does only the currying, without the string lookup (which would then be safe under Embroider)...

Reproduction in this PR: https://github.com/CrowdStrike/ember-headless-form/pull/84/files#diff-e61a0d763a79e50ad7c8eb97240e695edaacc3dde08a2f7669013a45a544f709R527. 
Works fine in classic, but Embroider fails with [`Unsafe dynamic component: cannot statically analyse this expression`](https://github.com/CrowdStrike/ember-headless-form/actions/runs/4542183105/jobs/8005314215#step:5:537) as it does not recognise the camelcased reference.

NullVoxPopuli:
Why use `this`?

I mean, it def should work, but if you need to be unblocked for now, i wouldn't say you need it. Since Field component is just an assignment 

https://github.com/CrowdStrike/ember-headless-form/pull/84/files#diff-e61a0d763a79e50ad7c8eb97240e695edaacc3dde08a2f7669013a45a544f709R195

You could?:
```
        (hash
          Field=(component
            FieldComponent
            data=this.etc
```
Though, maybe i'm missing something?

simonihmig:
> Why use this?

This was needed to fix the linting issue you also saw in your PR, due to the yielded component having a generic type (the `DATA` of the field needs to be the same as that of the parent form). 

But `this` is not even the problem here. Even without it we had that, as you [observed yourself](https://github.com/CrowdStrike/ember-headless-form/pull/90#issuecomment-1478444989). So Embroider will only consider `(component "some-string")` safe, but `(component FieldComponent)` and `(component this.FieldComponent)` are equally considered unsafe, AFAIK.

NullVoxPopuli:
[...]

PR #1383, opened 2023-03-28
Rename strict flag to strictMode

NullVoxPopuli:
Turns out we've had a discrepancy between strict and strictMode :upside_down_face: 

 - RFC update (and context / background): https://github.com/emberjs/rfcs/pull/920

There is another option to rename all the `strictMode` things to `strict`, but I think this would be more disruptive.

### DIFF
--- a/packages/util/tests/integration/helpers/ensure-safe-component-test.js
+++ b/packages/util/tests/integration/helpers/ensure-safe-component-test.js
@@ -188,7 +188,7 @@ module('Integration | Helper | ensure-safe-component', function (hooks) {
                 <Thing />
              {{/let}}`,
             {
-              strict: true,
+              strictMode: true,
               scope: () => ({
                 ensureSafeComponent,
               }),
--- a/packages/util/tests/integration/helpers/ensure-safe-component-test.js
+++ b/packages/util/tests/integration/helpers/ensure-safe-component-test.js
@@ -10,6 +10,7 @@ import { ensureSafeComponent } from '@embroider/util';
 import * as SomeComponentModule from 'dummy/components/some-component';
 import ColocatedExample from 'dummy/components/colocated-example';
 import { setOwner } from '@ember/application';
+import { precompileTemplate } from '@ember/template-compilation';
 
 const SomeComponent = SomeComponentModule.default;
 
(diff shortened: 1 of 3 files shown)


packages/util/tests/integration/helpers/ensure-safe-component-test.js
```diff
@@ -175,4 +176,34 @@ module('Integration | Helper | ensure-safe-component', function (hooks) {
     assert.strictEqual(this.element.textContent.trim(), 'second:goodbye');
     assert.deepEqual(log, ['target instantiated', 'new target instantiated']);
   });
+
+  module('with strict: true', function () {
+    test('reference to class property-as-component is allowed', async function (assert) {
```
NullVoxPopuli, line 181: passes on `release`
Fails on ember-source@4.4 and @3.28

NullVoxPopuli: now that I'm using the correct flag, `strictMode`, we get:
4.4 + 3.28 (still):
```
Error: Attempted to load a helper, but there wasn't a helper manager associated with the definition. The definition was: ensureSafeComponent
```                
release:
```
Unsafe dynamic component: cannot statically analyze this expression in tests/integration/helpers/ensure-safe-component-
```

ef4:
This seems like potentially a good bug fix but the way you're describing and testing it here is deeply misleading.

There is **never, ever a reason to use ensure-safe-component in strict mode**.

The existing [leaves strict mode templates alone](https://github.com/embroider-build/embroider/blob/e7c17891eef96ab35f75e7323ea39dc69a36a8b6/tests/scenarios/compat-resolver-test.ts#L1175-L1195) test is all you need. If it turns out it's really supposed to use `strictMode` instead of `strict`, then that test can be updated to be spelled that way. But other tests beyond that are just testing the same thing in a weird and misleading way.

NullVoxPopuli:
> If it turns out it's really supposed to use strictMode instead of strict,

yeah, I've discovered that the RFC was the only one using `strict` -- I have an update to that here: https://github.com/emberjs/rfcs/pull/920

tests/scenarios/compat-resolver-test.ts
```diff
@@ -525,7 +525,7 @@ Scenarios.fromProject(() => new Project())
           }
         );
 
-        expectTranspiled('templates/application.hbs').equalsCode(`
```
NullVoxPopuli, line 528: oof, lint configs that don't trim trailing invisible characters make for messy PRs :( 
Gonna try to PR a fix for this across the board

NullVoxPopuli: oh! apparently this isn't a lint thing, and it's an intentional choice of prettier to not trim trailing spaces within template s trings. :thinking: 
Guess I'll tweak my editor, and re-commit

NullVoxPopuli: no! it's the editorconfig that does this behavior! so we need an editorconfig lint :thinking:

simonihmig:
For the record: I just synced with Preston, as I also came to the conclusion when debugging https://github.com/embroider-build/embroider/issues/1382 that the reason for this was this mix up of the name of the strict mode flag. In the addon's dist output I had `strictMode`, and as you explained the component helper should not even try to resolve its argument and complain about dynamic usage. But it did. And by patching `resolver-transform.js` and renaming `strict` to `strictMode` as in this PR the error was gone, and Embroider was correctly identifying the compiled template tag coming from the addon as a strict mode template, bypassing all those resolver checks!

tests/scenarios/compat-resolver-test.ts
ef4: Whitespace inside strings is semantically important. If your editor is trimming it, it's wrong.

NullVoxPopuli: editorconfig is trimming it, so its wrong :shrug: 

why have editorconfig at all, I guess?

NullVoxPopuli: These changes are no longer a part of this PR, btw

PR #1386, opened 2023-03-29
Fix `Maximum call stack size exceeded` error

simonihmig:
Today I was running into the same issue that @void-mAlex has reported in yesterday's office hours, and as we discussed it removing the `exports` from the rewritten package.json indeed fixes the problem. At least for me, maybe @void-mAlex you can double check that on your side too?

### DIFF
--- a/packages/compat/src/v1-addon.ts
+++ b/packages/compat/src/v1-addon.ts
@@ -279,6 +279,7 @@ export default class V1Addon {
     // is just misleading to stage3 packagers that might look (rollup does).
     delete pkg.main;
     delete pkg.module;
+    delete pkg.exports;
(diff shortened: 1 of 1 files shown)


void-mAlex:
it was also desirable to give the better error in case the loop comes up for any other scenario, will try both this fix and the change for the error as soon as I get some time

ef4:
After our discussions yesterday, I'm pretty convinced this is correct. It's consistent with the way we already rewrite v1 addons.

> it was also desirable to give the better error in case the loop comes up for any other scenario,

Agreed, stopping the cycle is also good.

ef4 (approved):

PR #1384, opened 2023-03-28
Include prettier and .editorconfig settings in C.I.

NullVoxPopuli:
This'll help reduce the diff between developers.
Many editors pick up on .editorconfig settings, but not all do -- this leads to discrepancies during PR submission and creates extra noise.

Now, when local and CI run `yarn lint`, we get:
```
❯ yarn lint
yarn run v1.22.19
$ concurrently 'yarn:lint:*'
$ editorconfig-checker
$ eslint . --cache
$ cd packages/router && yarn lint:types
$ prettier . -c
$ tsc
[prettier] Checking formatting...
[editorconfig] yarn run lint:editorconfig exited with code 0
[eslint] yarn run lint:eslint exited with code 0
[router-types] yarn run lint:router-types exited with code 0
[prettier] All matched files use Prettier code style!
[prettier] yarn run lint:prettier exited with code 0
Done in 2.78s.
```

PR #1388, opened 2023-03-29
Enable prettier in ci

NullVoxPopuli:
Extracted from: https://github.com/embroider-build/embroider/pull/1384
since the editorconfig tooling may be suspicious (or misconfigured)


package.json
```diff
@@ -30,7 +30,11 @@
     "clean": "git clean -x -f",
     "compile": "tsc",
     "lint": "concurrently 'yarn:lint:*'",
+    "lint:fix": "concurrently 'npm:lint:*:fix' --names 'fix:'",
```
void-mAlex, line 33: any reason to have these called automatically by the lint command? it doesn't seem to exclude :fix as the default ember bp does

.prettierignore
```diff
@@ -0,0 +1,60 @@
...
+/test-packages/*/dist/
+/test-packages/*/tmp/
+
+# dependencies
+bower_components/
+node_modules/
+
+# misc
+coverage/
+!.*
```
void-mAlex, line 51: I think this is in the default bp but does it actually make sense? isn't this just re-adding all files?

void-mAlex (commented):
what's the motivation to change quotes in template/html files?
as far as I can tell default in ember is double quotes in templates single for js files

.prettierignore
NullVoxPopuli: it should only be re-adding `.`-prefixed files, like `.eslintrc.js`

NullVoxPopuli:
> what's the motivation to change quotes in template/html files?
as far as I can tell default in ember is double quotes in templates single for js files

I didn't change the prettier config -- this PR only hooks up prettier to C.I. and package.json#scripts. -- if there is a better config to use, lemme know.

packages/util/tests/dummy/app/templates/application.hbs
```diff
@@ -1,3 +1,3 @@
-<h2 id="title">Welcome to Ember</h2>
```
mansona, line 1: I feel like this shouldn't be changed 🤔 because it's not the same as a default output

mansona (changes requested):
we discussed this in the meeting

packages/util/tests/dummy/app/templates/application.hbs
NullVoxPopuli: resolved!

NullVoxPopuli: resolved for real now

mansona (approved):
Looks good to me 👍

void-mAlex (approved):

PR #1389, opened 2023-03-30
bump unstable versions by at least a patch

mansona:
I was having some issues where pnpm was picking up some wrong versions during an ember-try run and I was pulling my hair out 🙃 

I noticed that `@embroider/compat` version `2.1.1` was being used instead of the version `2.1.1-unstable.00ec2e7` that was specified in the package.json. No matter what I did I couldn't get pnpm to pick up correct version. Then I noticed that we're using carat dependencies in the package json on ember-try ever since https://github.com/embroider-build/embroider/pull/1328 and that seems to have caused my issue. 

Apparently the version `^2.1.1-unstable.00ec2e7` is satisfied by `2.1.1` 😭 

Screenshot 2023-03-30 at 17 09 43

This PR will make sure that all new packages that are deployed with the unstable packages will be at least one step higher and won't match any more

### DIFF
--- a/test-packages/unstable-release/version.js
+++ b/test-packages/unstable-release/version.js
@@ -33,7 +33,12 @@ const NEW_VERSIONS = {};
 
 async function setVersion(sha, filePath) {
   let json = await fse.readJSON(filePath);
-  json.version = `${json.version}-unstable.${sha}`;
+
+  // we need to at the very least bump the patch version of the unstable packages so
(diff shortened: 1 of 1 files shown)


void-mAlex (approved):

NullVoxPopuli (approved):
Discussed on the call today, and there is more to do (esp around test-setup capabilities), but those things can be separate PRs! :tada: 

nice work!!

PR #1390, opened 2023-03-31
Change namespace for cancel-in-progress

backspace:
This decouples `publish-unstable.yml` `cancel-in-progress` from that in `ci.yml`.

### DIFF
--- a/.github/workflows/publish-unstable.yml
+++ b/.github/workflows/publish-unstable.yml
@@ -9,7 +9,7 @@ on:
       - main
 
 concurrency:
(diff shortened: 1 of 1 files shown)


NullVoxPopuli (approved):
thanks for this!! 
accidental oversight! 
(though I would have expected merge commits to keep the concurrency different from PRs :see_no_evil: )

backspace:
ya I think `concurrency` is an all-repository namespace, I guess the event type would have to be part of the name for that to work 😯

PR #1393, opened 2023-04-05
continue deploying unstable packages even with an error

mansona:
One of the previous runs of NPM Publish for the unstable packages actually errored: https://github.com/embroider-build/embroider/actions/runs/4610016561/jobs/8147947998

In this case, the logical thing to do would be to rerun the process, but that wouldn't work if the release finished somewhere in the middle of the process because it would fail to deploy the first package again (because that version already exists).

This PR changes the semantics of the deploy step for unstable and it no longer stops on the first error it encounters. This means that we will be able to re-run the process without any issue if we have strange API problems again 👍

### DIFF
--- a/test-packages/unstable-release/.eslintrc.cjs
+++ b/test-packages/unstable-release/.eslintrc.cjs
@@ -12,7 +12,8 @@ module.exports = {
   },
   extends: ['eslint:recommended', 'plugin:prettier/recommended'],
   env: {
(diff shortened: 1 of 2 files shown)


NullVoxPopuli (approved):
lgtm. Only thing I'd change is use `console.error`, but when running in C.I. anyway, 🤷

PR #1394, opened 2023-04-10
Add @ember/owner to emberVirtualPackages

ef4:
For compatibility, we need to know the full set of virtual package names used by Ember. Historically we got that from the ember-rfc176-data package, but that is no longer updated for new APIs because the modules-api-polyfill that it was created to power is no longer needed.

Without this, v2 addons under embroider will fail to resolve these package names.


NullVoxPopuli:
Looks like this needs rebase (confirmed [here](https://github.com/embroider-build/embroider/pull/1396))

Issue #1399, opened 2023-04-19
Performance issue in macro-config's babelPluginConfig


raycohen:
I've got a monorepo using a yarn workspace with about 100 addons, on a classic (non-embroider) build. Booting takes a long time and I'm investigating performance improvements. I came across almost 3 seconds spent inside `babelPluginConfig`. About 1.5s is spent in crypto/hash#update, and then readFileSync and statSync (via findUp) are around 400ms each.

From what I can tell, it is repeatedly finding and reading the same `yarn.lock` file, and creating a hash with the same yarn.lock file contents and project dependency list.

I'm wondering if we can avoid some of this repetition. If so I can contribute a PR. It seems like at least the findUp result per `appRoot` value could be reused. I don't think the contents of `yarn.lock` or `gatherAddonCacheKey(appOrAddonInstance.project)` are changing during the time I'm looking at but it seems like this code is intended to catch changes to those?

https://github.com/embroider-build/embroider/blob/a6cd4aac9b7497991b7b1acc18d0e78dccb8cd78/packages/macros/src/macros-config.ts#L346-L370

![Screenshot 2023-04-19 at 1 54 01 PM](https://user-images.githubusercontent.com/20404/233159602-e53e4dd4-7939-4fa6-bd95-963d2468c947.png)

NullVoxPopuli:
I think your assumption is correct -- yeah, if you have time for a PR, that'd be super appreciated <3

PR #1400, opened 2023-04-19
memoize the generation of MacrosConfig cache keys per (project, appRoot)

raycohen:
Ref #1399 

This is the most aggressive memoization I can think of, computing the cacheKey only once per (project, appRoot) "tuple".


NullVoxPopuli:
What does performance look like in your app with this change?

SergeAstapov:
just a small note, seems related to #1102, not sure if optimisation made there was lost (we don't see that issue, although we may not be using the latest versions yet)

raycohen:
@SergeAstapov Thanks for the call out. That memoization is still in place from what I can tell.

@NullVoxPopuli Updated description with a speedscope image - this dropped time spent here to about 200ms (down from ~3s)

PR #1403, opened 2023-04-24
Remove deprecated dependency @types/source-map

francois2metz:
@types/source-map is deprecated https://www.npmjs.com/package/@types/source-map. So I removed it.


NullVoxPopuli:
All green -- thank you!!

PR #1409, opened 2023-05-02
Use a custom changelog line generator function to avoid adding meaningless changelog entries 

NullVoxPopuli:
In the Release Preview PR: https://github.com/embroider-build/embroider/pull/1408

We have a bunch of commits that are duplicated: 
image

This is because `changesets` is using the commit that the `changeset` file was added, rather than the commit of the actual change.

There are two problems:
 - repeated "dependency updated" entries
 - every change has the same commit.

This PR addresses both of those things.

Preview: 
image

This uses a tweaked version of the changelog-git plugin: https://github.com/changesets/changesets/blob/main/packages/changelog-git/src/index.ts

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,20 +59,21 @@
     "**/graceful-fs": "^4.0.0"
   },
   "devDependencies": {
+    "@changesets/changelog-git": "^0.1.14",
     "@changesets/changelog-github": "^0.4.7",
     "@changesets/cli": "^2.25.0",
     "@types/jest": "^29.2.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -59,20 +59,19 @@
     "**/graceful-fs": "^4.0.0"
   },
   "devDependencies": {
-    "@changesets/changelog-github": "^0.4.7",
```
NullVoxPopuli, line 62: unused

PR #1413, opened 2023-05-08
Fix @embroider/addon-dev keepAssets resolve priority

NullVoxPopuli:
Big project:
```
❯ tokei -t=TypeScript,JSON,JavaScript,SVG,Handlebars,CSS
===============================================================================
 Language            Files        Lines         Code     Comments       Blanks
===============================================================================
 CSS                    78         6857         5400          299         1158
 Handlebars             85         3702         3465           82          155
 JavaScript             10          553          427           47           79
 JSON                    2          456          456            0            0
 SVG                   164         1717         1717            0            0
 TypeScript            195        13396         9657         1476         2263
===============================================================================
 Total                 534        26681        21122         1904         3655
===============================================================================
```
With this in the rollup config:
```js
addon.keepAssets(['**/styles/*.css']), // resolves to only 5 files in the project
```

Before:
```
 → dist...
created dist in 28.7s
```

After:
```
 → dist...
created dist in 7.1s
```

### DIFF
--- a/packages/addon-dev/src/rollup-keep-assets.ts
+++ b/packages/addon-dev/src/rollup-keep-assets.ts
@@ -1,7 +1,7 @@
 import walkSync from 'walk-sync';
 import type { Plugin } from 'rollup';
 import { readFileSync } from 'fs';
-import { join } from 'path';
+import { join, dirname, relative } from 'path';
 import minimatch from 'minimatch';
 
 export default function keepAssets({
@@ -11,6 +11,8 @@ export default function keepAssets({
   from: string;
   include: string[];
 }): Plugin {
+  let sourceDir = join(process.cwd(), from);
+
   return {
     name: 'copy-assets',
 
@@ -23,6 +25,16 @@ export default function keepAssets({
       //
       // Without doing so can multiply the build time by 3x.
       let isMatch = include.some((pattern) => minimatch(source, pattern));
+
+      // It's possible that `source` is a relative import instead of absolute,
+      // so we need to generate that absolute path (or at least relative to the importer)
+      // and then see if that matches our globs
+      if (importer && !isMatch) {
+        let absolute = join(dirname(importer), source);
+        let rootRelative = relative(sourceDir, absolute);
+
+        isMatch = include.some((pattern) => minimatch(rootRelative, pattern));
+      }
 
       // Let some other plugin handle this source
       if (!isMatch) return null;
--- a/packages/addon-dev/src/rollup-keep-assets.ts
+++ b/packages/addon-dev/src/rollup-keep-assets.ts
@@ -17,16 +17,25 @@ export default function keepAssets({
     // imports of assets should be left alone in the source code. This can cover
     // the case of .css as defined in the embroider v2 addon spec.
     async resolveId(source, importer, options) {
+      // We need to check if the `source` (what the importer is trying to import)
+      // matches our patterns, because we don't want to hit the file system if
+      // the pattern doesn't match.
+      //
+      // Without doing so can multiply the build time by 3x.
+      let isMatch = include.some((pattern) => minimatch(source, pattern));
+
+      // Let some other plugin handle this source
+      if (!isMatch) return null;
+
       const resolution = await this.resolve(source, importer, {
         skipSelf: true,
         ...options,
       });
-      if (
-        resolution &&
-        include.some((pattern) => minimatch(resolution.id, pattern))
-      ) {
+
+      if (resolution) {
         return { id: source, external: true };
       }
+
       return resolution;
     },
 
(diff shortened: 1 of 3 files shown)


ef4:
This changes the behavior though. The patterns used to apply to your files, now they apply to your import statements, and those are definitely not the same thing, especially since people can add a lot of other config and plugins that adjust the behavior of how import statements map to files. 

The existing implementation was intended to not cost any extra resolutions because when we resolve something that turns out not to match our pattern, we still return it anyway. That resolution was going to happen with or without our plugin. It shouldn't add extra filesystem traversal. If your measurements show that it does, it would be good to figure out exactly why that is, because the same behavior is going to be very relevant to a rollup and vite version of the newer embroider module-resolver.

packages/addon-dev/src/rollup-keep-assets.ts
```diff
@@ -11,22 +11,43 @@ export default function keepAssets({
...
+      // the pattern doesn't match.
+      //
+      // Without doing so can multiply the build time by 3x.
+      let isMatch = include.some((pattern) => minimatch(source, pattern));
+
+      // It's possible that `source` is a relative import instead of absolute,
+      // so we need to generate that absolute path (or at least relative to the importer)
+      // and then see if that matches our globs
+      if (importer && !isMatch) {
+        let absolute = join(dirname(importer), source);
```
ef4, line 33: This stuff is definitely not OK. It has a bunch of broken edge cases. And if you fix all the edge cases, what you're going to end up with is `this.resolve()` anyway, the thing you're trying not to do.

NullVoxPopuli:
> This changes the behavior though.
>  especially since people can add a lot of other config and plugins that adjust the behavior of how import statements map to files.

Got an example? There weren't any tests around the usage of `keepAssets`, and we/I need to add them, but right now, since I'm not that familiar with `keepAssets` (or its intent), it feels like it's all _undefined_ behavior. 🙈 

> That resolution was going to happen with or without our plugin. 

_I'm_ not sure this is true.
the `resolveId` hook in this particular v2 addon is called 63332 times -- with a _lot_ of duplicates.
Maybe a different approach needs to happen, but that doesn't change the misalignment with what _should_/_shouldn't_ vs what _is_ happening.

There are more clues though!
Calling in to keepAssets' resolveId hook we have these `source`s:
```
src/components/my-component.css
src/components/my-component.css
src/components/my-component.css
src/components/my-component.css
src/components/my-component.css
src/components/my-component.css.gjs
src/components/my-component.css.gjs
src/components/my-component.css.gts
src/components/my-component.css.gts
src/components/my-component.css
src/components/my-component.css
src/components/my-component.css.gjs
src/components/my-component.css.gjs
src/components/my-component.css.gts
src/components/my-component.css.gts
src/components/my-component.css
src/components/my-component.css
src/components/my-component.css.gjs
src/components/my-component.css.gjs
src/components/my-component.css.gts
src/components/my-component.css.gts
src/components/my-component.css
src/components/my-component.css
src/components/my-component.css
src/components/my-component.css
src/components/my-component.css.gjs
src/components/my-component.css.gjs
src/components/my-component.css.gts
src/components/my-component.css.gts
src/components/my-component.css
src/components/my-component.css.gjs
src/components/my-component.css.gjs
src/components/my-component.css.gts
src/components/my-component.css.gts
src/components/my-component.css
src/components/my-component.css
```

about every component appears to produce similar output.

So this could be the result of this code here: https://github.com/NullVoxPopuli/rollup-plugin-glimmer-template-tag/blob/main/packages/rollup-plugin-glimmer-template-tag/src/rollup-plugin.js#L38

Which... from a standpoint of _not_ calling `resolve` too much -- doesn't seem ideal --- but you're saying its inevitable?
What's the path forward?

ef4:
>  it feels like it's all undefined behavior

You're trying to do an optimization here. That means you need to make the new code do the same thing as the old code. There's nothing undefined about it. Or instead call it `major` instead of `patch` and explain how you changed the behavior and why. 

> Got an example?

The user adds rollup-plugin-node-resolve and customizes `exports` in package.json, and starts writing imports in their code that get mapped to actual files through that `exports`. 

> So this could be the result of this code here:

Yeah, I think it's likely that rollup-plugin-glimmer-template-tag is the real source of your performance trouble. This pattern:

```js
let result = await this.resolve(source + ext, importer, {
  ...options,
  skipSelf: true,
});

if (result?.external) {
  return;
}
```

Seem very bad for performance, because `return` is going to fall through to the next plugin. But `this.resolve()` *already ran the next plugins*, and then you threw away their results. And you're doing this inside a loop, so you're probably running the whole stack of plugins many times per request.

NullVoxPopuli:
Thanks for explaining, I'll do more investigating 🎉

NullVoxPopuli:
I understand why the `this.resolve(source + ext` was done -- rollup doesn't resolve extentionless imports (except for js) -- mjs, cjs, etc all need explicit extensions.

PR #1414, opened 2023-05-09
Adjust macros and util types

ef4:
- we can export the macros types directly from index.ts because it only exists for types anyway
- ensure-safe-component doesn't come from macros, it belongs in its own registry in embroider/util

I'm only exporting the macros registry and not the individual macros because they don't actually work as importable strict mode helper anyway. That will need some additional implementation. Probably we would make the javascript macros work in that case anyway, eliminating the need for the distinction.


ef4:
@mansona volunteered to update the README's here to explain the type registries.

mansona (approved):
Looks good to me 🎉

PR #1416, opened 2023-05-10
Upgrade changeset-recover to support label-based changelog management

NullVoxPopuli:
Changes to the workflow
 - re-runs the changeset generation on label change
 - only allow one run at a time, cancelling old or pending runs 
 
 Changes to `changeset-recover`
 - supports setting `patch` / `minor` / `major` via PR labels (like release-it)
 - better changed-package detection

This should allow the entire release process to be managed in the GitHub UI without needing to have git credentials or anything (ie: phone-based maintenance)

### DIFF
--- a/.github/workflows/prepare-changelog.yml
+++ b/.github/workflows/prepare-changelog.yml
@@ -12,7 +12,7 @@ on:
       - main
 
 concurrency:
-  group: ci-${{ github.head_ref || github.ref }}
+  group: prepare-changelog-${{ github.head_ref || github.ref }}
   cancel-in-progress: true
 
 env:
--- a/.github/workflows/prepare-changelog.yml
+++ b/.github/workflows/prepare-changelog.yml
@@ -1,8 +1,20 @@
 name: 'Prepare Changelog'
 on:
+  # Use labels on PRs to update the changelog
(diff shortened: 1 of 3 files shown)


.github/workflows/prepare-changelog.yml
```diff
@@ -1,8 +1,20 @@
...
+  # - major
+  pull_request:
+    types:
+      - labeled
   push:
     branches:
       - main
+
+concurrency:
+  group: ci-${{ github.head_ref || github.ref }}
```
ef4, line 15: I don't think you meant this to share the same concurrency group with the test suite?

NullVoxPopuli: correct, thanks for catching that -- updated

.github/workflows/prepare-changelog.yml
```diff
@@ -1,8 +1,20 @@
 name: 'Prepare Changelog'
 on:
+  # Use labels on PRs to update the changelog
+  # - patch
+  # - minor
+  # - major
+  pull_request:
+    types:
+      - labeled
```
gilest, line 9: Could we run only when these specific labels are added?

PR #1429, opened 2023-05-16
Ember 5 compat

ef4:
Updates for ember 5 compat.

This includes unpinning ember-data in our test suite, because we need a newer one for ember 5 compat.


ef4:
The `release-engines-with-fastboot` scenario is failing due to https://github.com/emberjs/ember.js/pull/20461

PR #1430, opened 2023-05-16
Remove changeset, as @ef4 has some custom release tooling coming

NullVoxPopuli:
- Deletes the changeset folder
- Removes the changeset dependencies
- Removes changeset-recover github workflow
- Removes release job from ci.yml


.github/workflows/prepare-changelog.yml
```diff
@@ -1,48 +0,0 @@
-name: 'Prepare Changelog'
```
NullVoxPopuli, line 1: A new version of this will be created once the new tooling is merged

void-mAlex (approved):

PR #1436, opened 2023-05-18
prevent double ^ when using embroider test-setup

mansona:
After the last release `@embroider/test-setup` was outputting a double `^` for the embroider versions like this: 

```
"@embroider/core": "^^3.0.0",
"@embroider/webpack": "^^3.0.0",
"@embroider/compat": "^^3.0.0"
```

I don't know if any other packagers are more lenient about this but pnpm is very unhappy, so we should remove the extra carat 👍


NullVoxPopuli (approved):

void-mAlex (approved):

backspace:
Thanks for fixing this, I was puzzled seeing `ERR_PNPM_SPEC_NOT_SUPPORTED_BY_ANY_RESOLVER` in CI because of the `^^`.

PR #1439, opened 2023-05-19
Experiment: See if we can figure out a way to be kinder to the enviro…

NullVoxPopuli:
…nment

Very wip no looking (nothing even works atm, weird dep issues that seem unrelated) - I don't know if this is a good idea yet.
Need to run C.I. without meaningful changes at least once as a baseline though.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "jest": "^29.2.1",
     "patch-package": "^6.5.1",
     "prettier": "^2.3.1",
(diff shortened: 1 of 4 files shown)


NullVoxPopuli:
i dunno when it happened, but `tsc` from the root in this repo is pretty fast now-a-days, and the repo isn't big enough where adding turbo would provide any speed gains.

PR #1440, opened 2023-05-19
release preview workflow

NullVoxPopuli:
wip

The goal is to, on every update to `main`, update or create a PR that previews what the release would look like if we were to release at the moment.

In the future, we could add more automation where, if that PR is merged, a release occurs.

### DIFF
--- a/.github/workflows/release-plan-preview.yml
+++ b/.github/workflows/release-plan-preview.yml
@@ -1,0 +1,50 @@
+name: Release Plan Review
+on:
+  push:
+    branches:
+      - main
+
+env:
+  VOLTA_FEATURE_PNPM: 1
+  branch: "gh-workflow/release-preview"
+  title: "Preview Release"
+
+jobs:
+  update_unstable_branch:
+    name: Update Unstable Branch
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/release-plan-preview.yml
```diff
@@ -0,0 +1,50 @@
...
+        run: |
+          set +e
+          git branch -D ${{ env.branch }}
+          set -e
+          git switch -c ${{ env.branch }}
+
+      - id: explanation
+        run: |
+          echo "text=$(pnpm embroider-release explain-plan)" >> $GITHUB_OUTPUT
+      - run: pnpm embroider-release gather-changes
```
ef4, line 33: I don't think you need the `gather-changes` step. `prepare` will do the gather automatically if you aren't using `--stdin` to pass them in.

ef4:
Thanks, this is good. It would be nice if the output of `explain-plan` could go into the body of the pull request.

.github/workflows/release-plan-preview.yml
NullVoxPopuli: thanks!!

ef4 (approved):

PR #1443, opened 2023-05-22
linkto routable engine path in host app

void-mAlex:
this pr adds a failing test case for routable lazy engines

### DIFF
--- a/tests/fixtures/engines-host-app/app/templates/application.hbs
+++ b/tests/fixtures/engines-host-app/app/templates/application.hbs
@@ -1,6 +1,7 @@
 <nav>
   <LinkTo @route="use-eager-engine">Eager</LinkTo>
   <LinkTo @route="use-lazy-engine">Lazy</LinkTo>
+  <LinkTo @route="use-lazy-engine.inner-engine">Inner</LinkTo>
   <LinkTo @route="style-check">Style Check</LinkTo>
 </nav>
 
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -53,7 +53,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup
         with:
-          use_lockfile: false
+          use_lockfile: true
```
void-mAlex, line 56: this is temporary to show CI would be green if not for the meddling ts getting in the way

mansona: What do you mean by temporary? is it something that we need to remove in a few weeks when something upstream is fixed? do we want to merge it with this and create an issue to remove it? I assume we'll discuss more at the meeting 😂

void-mAlex: meaning without this any PR will be very bloody right now
turned this on as a means to validate that the pr itself with the changes passes, without interference from floating packages (@babel/types) that have nothing to do with this bugfix or change

mansona (approved):
I'm happy with what I see in this PR 👍 the only thing I'm not sure about is the `use_lockfile: true` change. I understand the reasoning for the change and I'm happy to merge it, I just want us to be clear on a longer-term strategy as a team i.e. next steps

void-mAlex:
> I'm happy with what I see in this PR 👍 the only thing I'm not sure about is the `use_lockfile: true` change. I understand the reasoning for the change and I'm happy to merge it, I just want us to be clear on a longer-term strategy as a team i.e. next steps

not intending to have that merged at all, but can't check CI for anything else right now as I don't know how to fix TS without that

PR #1449, opened 2023-05-25
[bugfix] Set output.experimentalMinChunkSize to 0, to counter a change in rollup@3.22.0

ijlee2:
### A few independent incidences

After updating `rollup` to `3.22.0`, the package [`embroider-css-modules`](https://github.com/ijlee2/embroider-css-modules/tree/0.1.6/packages/embroider-css-modules) worked, in the sense that [the consuming app](https://github.com/ijlee2/embroider-css-modules/tree/0.1.6/docs/embroider-css-modules) (a test app) continued to run and all tests for the test app continued to pass. I noticed, however, that the `lint:types` script of the test app failed because `embroider-css-modules` didn't create `dist/index.d.ts` in the same way as before.

<details>

<summary>Error messages</summary>

```sh
app/components/widgets/widget-4/memo/actions.gts:3:28 - error TS2306: File '/packages/embroider-css-modules/dist/index.d.ts' is not a module.

3 import { localClass } from 'embroider-css-modules';
                             ~~~~~~~~~~~~~~~~~~~~~~~

app/components/ui/form/field.gts:4:28 - error TS2306: File '/packages/embroider-css-modules/dist/index.d.ts' is not a module.

4 import { localClass } from 'embroider-css-modules';
```

</details>

<details>

<summary>What <code>dist</code> looked like before updating <code>rollup</code></summary>

</details>

<details>

<summary>What <code>dist</code> looked like after updating <code>rollup</code></summary>

</details>

After updating `rollup` to `3.22.0` in another package (in another repo), I noticed that a different file called `dist/template-registry.d.ts` had been created differently. The test app for that package also failed `lint:types` as a result.

<details>

<summary>What <code>dist/template-registry.d.ts</code> looked like before updating <code>rollup</code></summary>

```ts
import UiSpinnerComponent from "./components/ui/spinner.js";
interface UiSpinnerRegistry {
    'Ui::Spinner': typeof UiSpinnerComponent;
}
export { UiSpinnerRegistry as default };
```

</details>

<details>

<summary>What <code>dist/template-registry.d.ts</code> looked like after updating <code>rollup</code></summary>

```ts
export {};
```

</details>

Finally, on Discord, [`@jeffsawatzky` described encountering a similar issue](https://discord.com/channels/480462759797063690/491905849405472769/1111019357888319641).

> And it [Glint] is _mostly_ working, except that the registry that I define in `template-registry.ts` don't get saved in `template-registry.d.ts` after build. Instead that file is blank, and the code I put in `template-registry.ts` ends up in `index.d.ts`. I've looked at other examples, but I haven't been able to figure this out. 

## Possible root cause

I reported the issue to the maintainers of `rollup` in https://github.com/rollup/rollup/issues/5008.

@lukastaegert noticed, the problem seemed to be limited to `*.d.ts` files, e.g. `dist/index.d.ts`, `dist/template-registry.d.ts`. If so, `rollup-plugin-ts` (which we use to support TypeScript) or one of its dependencies, might have made an assumption about chunks that is not valid as of `rollup@3.22.0`.

## Proposed solution

Lukas (who maintains `rollup`) and I both think, a good option is for `@embroider/addon-dev` to explicitly set `output.experimentalMinChunkSize` to `0`.

> Maybe explicitly setting `output.experimentalMinChunkSize:0` in the shared config is not a bad idea for now. The reasoning for setting it to `1` was that there should not be adverse effects to users, but it should avoid some empty chunks. It would still be helpful if someone with more knowledge about the workings of `rollup-plugin-dts` could look into this and determine if the empty files are a bug in the plugin or in Rollup. In the worst case, we could change the default back to `0` until the next major version.

This way, an addon maintainer doesn't need to set the value themselves in `rollup.config.js`:

```js
/* rollup.config.{js,mjs} */

import typescript from 'rollup-plugin-ts';

const addon = new Addon({
  srcDir: 'src',
  destDir: 'dist',
});

export default {
  output: {
    ...addon.output(),
    experimentalMinChunkSize: 0,  // <-- Current workaround for addon developers
  },

  plugins: [ ... ],
};
```

At the same time, the maintainers of `rollup` could continue to iterate on the default value of `experimentalMinChunkSize` for `v3` and `v4`.

## Related links

- https://github.com/ijlee2/embroider-css-modules/pull/42#discussion_r1202018979
- https://rollupjs.org/configuration-options/#output-experimentalminchunksize
- https://github.com/rollup/rollup/issues/5008

### DIFF
--- a/packages/addon-dev/package.json
+++ b/packages/addon-dev/package.json
@@ -42,7 +42,7 @@
     "@types/fs-extra": "^9.0.12",
     "@types/minimatch": "^3.0.4",
     "@types/yargs": "^17.0.3",
-    "rollup": "^2.58.0",
+    "rollup": "^3.23.0",
     "tmp": "^0.1.0",
     "typescript": "^4.9.0"
   },
--- a/packages/addon-dev/src/rollup.ts
+++ b/packages/addon-dev/src/rollup.ts
@@ -75,6 +75,7 @@ export class Addon {
     return {
       dir: this.#destDir,
       entryFileNames: '[name]',
+      experimentalMinChunkSize: 0,
       format: 'es',
       hoistTransitiveImports: false,
       sourcemap: true,
--- a/packages/addon-dev/src/rollup.ts
+++ b/packages/addon-dev/src/rollup.ts
@@ -74,8 +74,8 @@ export class Addon {
   output() {
     return {
       dir: this.#destDir,
-      format: 'es',
       entryFileNames: '[name]',
+      format: 'es',
       hoistTransitiveImports: false,
       sourcemap: true,
     };
--- a/packages/router/package.json
+++ b/packages/router/package.json
@@ -53,9 +53,9 @@
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-prettier": "^4.0.0",
     "prettier": "^2.5.1",
-    "rollup": "^2.67.0",
+    "rollup": "^3.23.0",
     "rollup-plugin-copy": "^3.4.0",
-    "rollup-plugin-ts": "~3.0.2",
+    "rollup-plugin-ts": "^3.2.0",
     "typescript": "^4.9.0"
   },
   "peerDependencies": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -69,7 +69,7 @@ importers:
         version: 3.0.4
       rollup-plugin-copy-assets:
         specifier: ^2.0.3
-        version: 2.0.3(rollup@2.69.1)
+        version: 2.0.3(rollup@3.23.0)
       rollup-plugin-delete:
         specifier: ^2.0.0
         version: 2.0.0
@@ -96,8 +96,8 @@ importers:
         specifier: ^17.0.3
         version: 17.0.3
       rollup:
-        specifier: ^2.58.0
-        version: 2.69.1
+        specifier: ^3.23.0
+        version: 3.23.0
       tmp:
         specifier: ^0.1.0
         version: 0.1.0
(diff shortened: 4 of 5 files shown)


packages/addon-dev/src/rollup.ts
```diff
@@ -74,8 +74,9 @@ export class Addon {
   output() {
     return {
       dir: this.#destDir,
-      format: 'es',
       entryFileNames: '[name]',
+      experimentalMinChunkSize: 0,
+      format: 'es',
```
ijlee2, line 79: I alphabetized the keys. When there is a list of configuration keys, people tend to add a new one to the end of the list. I find unsorted lists not easy to understand and maintain.

NullVoxPopuli (approved):
Very nice write up!

NullVoxPopuli:
For testing, can you bump the rollup version in tests/scenarios to 3.22+?

ijlee2:
> For testing, can you bump the rollup version in tests/scenarios to 3.22+?

Yep, I can try.

packages/addon-dev/package.json
```diff
@@ -42,7 +42,7 @@
     "@types/fs-extra": "^9.0.12",
     "@types/minimatch": "^3.0.4",
     "@types/yargs": "^17.0.3",
-    "rollup": "^2.58.0",
+    "rollup": "^3.23.0",
```
ijlee2, line 45: See https://github.com/rollup/rollup/releases/tag/v3.0.0 for breaking changes that might affect `@embroider`.

ijlee2: @NullVoxPopuli Can you run CI for me?

ijlee2:
Yesterday, 2 out of 200 checks (both on windows env, if I recall) failed due to a test timeout. I rebased the branch to rerun the CI and 1 check failed because of a timeout.

Can you manually retry the failing check?

NullVoxPopuli:
Thank you!

PR #1454, opened 2023-05-30
Refactor scenario helpers to a shared location

NullVoxPopuli:
this happened mostly because v2 addon watch infra is large, and we don't want its responsibilities mingling with the other helpers.

Aside from changing a relative path reference to having an extra `..`, there are no code changes.

This should help out @void-mAlex with some addon-dev testing / watching tests


tests/scenarios/helpers.ts
```diff
@@ -1,66 +1 @@
...
-  const paths = globby.sync('**', { cwd: root, dot: true });
-  const fixtureStructure: any = {};
-
-  paths.forEach(path => {
-    set(fixtureStructure, path.split('/'), readFileSync(join(root, path), 'utf8'));
-  });
-
-  return fixtureStructure;
-}
+export * from './helpers/index';
```
NullVoxPopuli, line 1: at some point in node's history, implicit index is no longer available to use.

This file now emulates that behavior (and mostly to avoid having to change all the scenario tests to use `./helpers/index`)

mansona: I don't think this is a node thing in this context 😂 I'm pretty sure this whole `import *` and `export *` is a typescript specific thing that is pretty annoying if you ask me since it messes with true ESM semantics

PR #1459, opened 2023-06-02
Create embroider-architecture-overview.md

achambers:
[RENDERED](https://github.com/embroider-build/embroider/blob/2d87499c8a194b0f1b462c897225eaeae58d0eff/docs/embroider-architecture-overview.md)

I want to pull together a document that describes the embroider architecture to allow people to build a mental model of where it is and where it is going. I've had to do a lot of digging and debugging through code to try and piece together a mental model of things and I'd love it if there was a resource that could fast forward that understanding for others.

This document is very WIP and doesn't claim to be complete _or_ accurate. But it's the result of a good chunk of time from myself and @22a building up our mental models via code spelunking, debugging, pairing and conversations.

I'm keen to get this document out in the public so that we can run it past others that know more than we do and can iterate on it in the hope of having something that is consumable and accessible to others that would like to contribute to the embroider ecosystem but feel like they don't have a good grasp on how it all works.

All suggestions, edits, advice and corrections welcome.

### DIFF
--- a/docs/embroider-architecture-overview.md
+++ b/docs/embroider-architecture-overview.md
@@ -1,0 +1,402 @@
+# Embroider Architecture Overview (extremely WIP)
+
+
+## Summary
+What is Embroider? Why do we care about Embroider? What is Embroider trying to fix? How does Embroider work?
+
+This is an attempt at distilling the research @22a and myself have been doing the past while into something that's consumable and answers these questions. Or, at the very least, documents what we think we understand so we can iterate on it later.
+
+## Let's start at the goal
+Before we can talk about what Embroider is and how it works, it’s useful to talk about our end goal and why that is not currently possible with Ember apps and the build tools, specifically EmberCLI, we use today.
+
+In modern times, a javascript application is made up of a number of javascript files that reference each other via imports. Each of these files is an ECMAScript Module (ESM for short) and they are referenced using the ESM `import` specifier. For example:
+
+```js
+// src/full-name.js
+
+export default function fullName(first, last) {
+  return `${first} ${last}`;
+}
+```
+
+```js
+// src/main.js
+
+import fullName from ‘./full-name’;
+
+fullName(‘Bob’, ‘Marley”);
+```
+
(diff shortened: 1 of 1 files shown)


docs/embroider-architecture-overview.md
```diff
@@ -0,0 +1,402 @@
...
+
+```mermaid
+flowchart TD
+    A[Dependency] --> B{Is Ember addon?}
+    B -->|Yes| C{Is V2 addon?}
+    C -->|Yes| D{Has deps\nthat need\nrewriting?}
+    D -->|No| F[Symlink]
+    D -->|Yes| E[Rewrite] 
+    C -->|No| E[Rewrite]
+    B ---->|No| F[Symlink]
```
ijlee2, line 140: By reordering lines 134-140, we can avoid overlapping edges, which may confuse the reader.

If we pretend that symlink is good and rewrite is bad, I think we could add the ✅ and ❌ icons to hint which path the reader will want to take in the flowchart.

| Before | After |
|:--:|:--:|
|  |  |

```suggestion
flowchart TD
    A[Dependency] --> B{Is Ember addon?}
    B -->|Yes| C{Is V2 addon?}
    C -->|No| E["❌ Rewrite"]
    C -->|Yes| D{Has dependency\nthat needs rewrite?}
    D -->|Yes| E["❌ Rewrite"]
    D -->|No| F["✅ Symlink"]
    B -->|No| F["✅ Symlink"]
```

achambers: Great suggestion. Thanks @ijlee2 .

docs/embroider-architecture-overview.md
```diff
@@ -0,0 +1,402 @@
...
+  </body>
+</html>
+```
+
+Unlike the "perfect world" example above, this `index.html` points to `.js` and `.css` files that don't actually exist. This is because the are pointing to files in a location that will be created as a part of the EmberCLI build logic that runs when we run the `ember build` command. Let's look a bit deeper to see what else we can find.
+
+## The Addons
+**TODO**
+
+## Embroider Build Stages
```
villander, line 114: @achambers could you please add the stage with `ember-engines`?

villander (commented):
Thanks for writing that up. That's gonna make embroider way clear to our community. The only thing I'd love to see here is how embroider fits with **engines**.

PR #1463, opened 2023-06-06
fix resolution of files with .hbs extensions

ef4:
As discussed with @mansona, we have a bug here and need to make the app JS searching do a search for .js, .hbs, and .hbs.js resolvable extensions.

### DIFF
--- a/packages/core/src/module-resolver.ts
+++ b/packages/core/src/module-resolver.ts
@@ -306,7 +306,7 @@ export class Resolver {
 
     let pkg = this.owningPackage(match.filename);
     if (pkg) {
-      let rel = withoutJSExt(explicitRelative(pkg.root, match.filename));
+      let rel = explicitRelative(pkg.root, match.filename);
       let entry = this.mergeMap.get(pkg.root)?.get(rel);
       if (entry?.type === 'both') {
         return request.alias(entry[section].localPath).rehome(resolve(entry[section].packageRoot, 'package.json'));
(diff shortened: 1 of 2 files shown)


packages/core/src/module-resolver.ts
```diff
@@ -504,7 +504,7 @@ export class Resolver {
                 `addon ${addon.name} declares app-js in its package.json with the illegal name "${inAddonName}". It must start with "./" to make it clear that it's relative to the addon`
               );
             }
-            inEngineName = withoutJSExt(inEngineName);
+            inEngineName = inEngineName;
```
void-mAlex, line 507: this looks fishy;
what does this do?
should this not just be removed instead?

packages/core/src/module-resolver.ts
```diff
@@ -306,7 +306,7 @@ export class Resolver {
 
     let pkg = this.owningPackage(match.filename);
     if (pkg) {
-      let rel = withoutJSExt(explicitRelative(pkg.root, match.filename));
+      let rel = explicitRelative(pkg.root, match.filename);
```
ef4, line 309: To drop `withoutJSExt` here you must replace it with the search over all resolvable extensions. That's what's causing the failure in the `engine-relative resolving > file exists in both app-js and fastboot-js` test.

I would suggest taking your implementation in `searchAppTree` and making it a utility method that you can use here too.

mansona: Nice one, I completely missed that there was something else checking the mergemap 👍

PR #1469, opened 2023-06-13
update deprecated (and removed) blacklist config in test app

mansona:
It turns out that ember-cli@5.0 has merged https://github.com/ember-cli/ember-cli/pull/10175 which removes the blacklist key for addons and replaces it with `exclude`. 

I have a feeling we might need to use the right one based on a version but I'm trying this PR just updating it for now 👍


simonihmig (approved):
Nice catch! Given this is green, this seems to fix [release-stage-1-problematic-addon-zoo](https://github.com/embroider-build/embroider/actions/runs/5252300573/jobs/9488238002#step:4:38) failures on `master` and other feature branches!

PR #1470, opened 2023-06-13
Prepping release

ef4:


NullVoxPopuli (approved):

.release-plan.json
```diff
@@ -1,16 +1,7 @@
...
-      "impact": "patch",
-      "oldVersion": "3.1.0",
-      "newVersion": "3.1.1",
-      "constraints": [
-        {
-          "impact": "patch",
-          "reason": "Appears in changelog section :bug: Bug Fix"
-        }
-      ],
-      "pkgJSONPath": "/Users/edward/hacking/embroider/packages/addon-dev/package.json"
```
NullVoxPopuli, line 13: why include the `pkgJSONPath` here? (since it's machine specific?)

ef4: It's actually needed, but probably that means we have a bug and right now you need to run `prepare` and `publish` on the same machine.

PR #1471, opened 2023-06-15
use @swc/register instead of ts-node for tests

mansona:
when you're working on the tests and moving things around you sometimes have cases where ts-node is being super-strict and failing for unused variables (WHY!??!?)

This is a somewhat draft PR to open a discussion of using SWC instead of ts-node for tests, and we like it we can go ahead and merge it 🎉

### DIFF
--- a/tests/scenarios/package.json
+++ b/tests/scenarios/package.json
@@ -24,9 +24,9 @@
     "ts-node": "^10.9.1"
   },
   "scripts": {
-    "test": "qunit --require ts-node/register *-test.ts",
-    "test:list": "scenario-tester list --require ts-node/register --files=*-test.ts",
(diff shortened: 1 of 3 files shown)


ef4:
This seems fine to me as long as we've double-checked that a type error in the test files will fail our preflight CI job.

(I'm pretty sure it does, just pointing out that's the one potential difference.)

tests/scenarios/package.json
```diff
@@ -46,6 +46,7 @@
     "@embroider/router": "workspace:*",
     "@embroider/util": "workspace:*",
     "@rollup/plugin-babel": "^5.3.1",
+    "@swc/register": "^0.1.10",
```
lifeart, line 49: should  `"ts-node": "^10.9.1"` to be removed from deps?

PR #1474, opened 2023-06-20
cache bust index html on arbitrary dom changes

void-mAlex:
what this pr introduces is a cache key in the form of a render on first cache of the appInfo
subsequent runs render the  html file before it would be processed and altered by webpack to contain all chunks this providing a stable key to check against
we also have a final check in place once webpack has built to not write a file if the final content is unchanged which is done as a file system check

@ef4 this seems to be caused by the way we check if appInfo is the same and if we need to set up a new webpack instance 
I did not find a better way to query at start of the build or at another point during the process

### DIFF
--- a/packages/core/src/html-entrypoint.ts
+++ b/packages/core/src/html-entrypoint.ts
@@ -152,6 +152,11 @@ export class HTMLEntrypoint {
     }
     return this.dom.serialize();
   }
+
+  simpleRender(): string {
+    //used to cache bust during webpack build
+    return this.dom.serialize();
+  }
 }
 
 export interface BundleSummary {
--- a/packages/webpack/src/ember-webpack.ts
+++ b/packages/webpack/src/ember-webpack.ts
@@ -54,6 +54,7 @@ interface AppInfo {
   publicAssetURL: string;
   resolverConfig: ResolverOptions;
   packageName: string;
+  renderedEntrypointsDom: string[];
 }
 
 // AppInfos are equal if they result in the same webpack config.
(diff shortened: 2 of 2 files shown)


ef4:
But *do* we need a new webpack config? Just changing content in the HTML should not require that.

We already compare the output of HTMLEntrypoint.render from the previous run [here](https://github.com/embroider-build/embroider/blob/9b83c1139f53055c57dc51c7321c19fe8bf31974/packages/webpack/src/ember-webpack.ts#LL440C49-L440C49). This is supposed to accomplish the same thing that you're doing with simpleRender.

I suspect that your fix works by invalidating more than we really need to invalidate and that there's still something more targeted that is going wrong.

void-mAlex:
> But _do_ we need a new webpack config? Just changing content in the HTML should not require that.

short answer is no, I've also tested that it's fixed if I replace the entrypoints array on lastAppInfo and let it run like that it works but that seems riskier as it feels like missing out on side effects that happens in configureWebpack to entrypoints.modules

> 
> We already compare the output of HTMLEntrypoint.render from the previous run [here](https://github.com/embroider-build/embroider/blob/9b83c1139f53055c57dc51c7321c19fe8bf31974/packages/webpack/src/ember-webpack.ts#LL440C49-L440C49). This is supposed to accomplish the same thing that you're doing with simpleRender.

the content that is seen at that point is from the first cached lastAppInfo.entrypoints.dom which is what this tries to fix as that will never go into the if you pointed out due to correctly being the same at that point but not the one that want meaning the updated dom from newly examined appInfo.entrypoints.dom

> 
> I suspect that your fix works by invalidating more than we really need to invalidate and that there's still something more targeted that is going wrong.
I thought so as much as well but can't see anything more targeted than this that also caters to not change the semantics of code around it

void-mAlex:
@ef4 
specifically https://github.com/embroider-build/embroider/blob/9b83c1139f53055c57dc51c7321c19fe8bf31974/packages/webpack/src/ember-webpack.ts#L299
appInfo.entrypoints[0].dom is the correct updated html but because the equalAppInfo doesn't take that into account so it considers the lastAppInfo valid and this.lastAppInfo.entrypoints[0].dom does not have any of the updated html

ef4:
Closing in favor of https://github.com/embroider-build/embroider/pull/1597

PR #1478, opened 2023-06-27
Don't over-serialize options into `resolver.json`

### LINKED ISSUE #1441: Rebuilds fail when app is using WebpackBundleAnalyzer & probably others
## Reproduction
```sh
$ npx ember-cli new test-app
$ cd test-app/
$ yarn add --dev @embroider/core @embroider/compat @embroider/webpack webpack webpack-bundle-analyzer
```

edit `ember-cli-build.js` to look like:

```js
'use strict';

const { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer');

  const app = new EmberApp(defaults, {
    // Add options here
  });

    packagerOptions: {
      webpackConfig: {
        plugins: [new BundleAnalyzerPlugin()],
      },
    },
  });
};
```

run `ember serve`, then edit a file to cause a rebuild.

## Result

The rebuild fails with this error:

```
ERROR Summary:

  - broccoliBuilderErrorStack: TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Compiler'
    --- property 'root' closes the circle
    at JSON.stringify (<anonymous>)
    at stringify (/private/tmp/test-app/node_modules/fs-extra/node_modules/jsonfile/utils.js:3:20)
    at outputJsonSync (/private/tmp/test-app/node_modules/fs-extra/lib/json/output-json-sync.js:7:15)
    at AppBuilder.addResolverConfig (/private/tmp/test-app/node_modules/@embroider/core/src/app.js:966:35)
    at AppBuilder.build (/private/tmp/test-app/node_modules/@embroider/core/src/app.js:887:10)
    at async WaitForTrees.buildHook (/private/tmp/test-app/node_modules/@embroider/core/src/build-stage.js:33:13)
  - code: [undefined]
  - codeFrame: Converting circular structure to JSON
    --> starting at object with constructor 'Compiler'
    --- property 'root' closes the circle
  - errorMessage: Converting circular structure to JSON
    --> starting at object with constructor 'Compiler'
    --- property 'root' closes the circle
        at WaitForTrees (@embroider/compat/app)
-~- created here: -~-
    at new Plugin (/private/tmp/test-app/node_modules/@embroider/core/node_modules/broccoli-plugin/dist/index.js:47:36)
    at new WaitForTrees (/private/tmp/test-app/node_modules/@embroider/core/src/wait-for-trees.js:30:9)
    at CompatApp.get tree (/private/tmp/test-app/node_modules/@embroider/core/src/build-stage.js:25:16)
    at CompatApp.<anonymous> (/private/tmp/test-app/node_modules/typescript-memoize/dist/memoize-decorator.js:132:52)
    at new PackagerRunner (/private/tmp/test-app/node_modules/@embroider/core/src/to-broccoli-plugin.js:11:26)
    at Object.defaultPipeline (/private/tmp/test-app/node_modules/@embroider/compat/src/default-pipeline.js:50:12)
    at module.exports (/private/tmp/test-app/ember-cli-build.js:12:39)
    at Builder.readBuildFile (/private/tmp/test-app/node_modules/ember-cli/lib/models/builder.js:48:20)
    at async Builder.setupBroccoliBuilder (/private/tmp/test-app/node_modules/ember-cli/lib/models/builder.js:68:17)
    at async Builder.ensureBroccoliBuilder (/private/tmp/test-app/node_modules/ember-cli/lib/models/builder.js:56:7)
```

## Analysis
When the `webpack-bundle-analyzer` plugin's `apply()` is called, it [caches a copy of the compiler object](https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/37b406d65b3fe8219ce00afc641084b676e0a44b/src/BundleAnalyzerPlugin.js#L36) on itself. That object contains a reference back to the plugin, creating a circular reference. When embroider tries to [write out the resolver config](https://github.com/embroider-build/embroider/blob/8f4cd857fa1bc397ad4509c96e025fcfadde42dc/packages/core/src/app.ts#L998) on a rebuild, it fails because the config options now contain a circular reference and are not JSON stringifiable.

In general, I think it looks like the webpack config is not guaranteed to be JSON stringifiable, so trying to stringify the whole configuration isn't a safe thing to do. Perhaps `CompatAppAdapter.resolverConfig()` should not be returning its entire `CompatResolverOptions`, but instead filtering them down to only the fields needed to implement the [module resolver options](https://github.com/embroider-build/embroider/blob/8f4cd857fa1bc397ad4509c96e025fcfadde42dc/packages/core/src/module-resolver.ts#L49-L64)? Or `addResolverConfig()` should prune down the object it receives to just the set of fields that match the interface?

chancancode:
`this.options` can contain arbitrary unserializable data (such as Broccoli trees, Webpack plugins, etc).

Fixes #1441

Optionally, we can change the type of the `AppBuilder`:

```patch
diff --git a/packages/core/src/app.ts b/packages/core/src/app.ts
index 5ab0844a..b46bcf0f 100644
--- a/packages/core/src/app.ts
+++ b/packages/core/src/app.ts
@@ -223,14 +223,14 @@ class ConcatenatedAsset {

 type InternalAsset = OnDiskAsset | InMemoryAsset | BuiltEmberAsset | ConcatenatedAsset;

-export class AppBuilder<TreeNames> {
+export class AppBuilder<TreeNames, SpecificResolverConfig extends ResolverConfig = ResolverConfig> {
   // for each relativePath, an Asset we have already emitted
   private assets: Map<string, InternalAsset> = new Map();

   constructor(
     private root: string,
     private app: Package,
-    private adapter: AppAdapter<TreeNames>,
+    private adapter: AppAdapter<TreeNames, SpecificResolverConfig>,
     private options: Required<Options>,
     private macrosConfig: MacrosConfig
   ) {
@@ -994,7 +994,7 @@ export class AppBuilder<TreeNames> {
     );
   }

-  private addResolverConfig(config: ResolverConfig) {
+  private addResolverConfig(config: SpecificResolverConfig) {
     outputJSONSync(join(this.root, '.embroider', 'resolver.json'), config);
   }

```

Didn't commit it because I'm not sure if it's worth propagating the generic, and it wouldn't have detected this issue anyway, but it does document the intent of the `resolver.json` more correctly.

Happy to add a test if I can get some pointers.

### DIFF
--- a/packages/compat/src/compat-app.ts
+++ b/packages/compat/src/compat-app.ts
@@ -354,6 +354,13 @@ class CompatAppAdapter implements AppAdapter<TreeNames, CompatResolverOptions> {
       activeAddons[addon.name] = addon.root;
     }
 
+    let options: CompatResolverOptions['options'] = {
+      staticHelpers: this.options.staticHelpers,
+      staticModifiers: this.options.staticModifiers,
+      staticComponents: this.options.staticComponents,
+      allowUnsafeDynamicComponents: this.options.allowUnsafeDynamicComponents,
+    };
+
     let config: CompatResolverOptions = {
       // this part is the base ModuleResolverOptions as required by @embroider/core
       activeAddons,
@@ -379,8 +386,8 @@ class CompatAppAdapter implements AppAdapter<TreeNames, CompatResolverOptions> {
       // global template resolving
       modulePrefix: this.modulePrefix(),
       podModulePrefix: this.podModulePrefix(),
-      options: this.options,
       activePackageRules: this.activeRules(),
+      options,
     };
 
     return config;


ef4 (approved):
This seems good. I know we already had code like this somewhere else (to reduce the options down to the serializable parts we actually care about), but perhaps it got dropped in a recent refactor.

chancancode:
@ef4 do you have pointers testing wise or are we good as is?

ef4:
Picked this up in https://github.com/embroider-build/embroider/pull/1481 where I resolved new merge conflicts.

(This PR was not set to allow edits by maintainers, otherwise I would have done it here.)

PR #1435, opened 2023-05-18
Working to eliminate node_modules rewriting

ef4:
This is picking up the work explored in #1374. I'm going to probably land those same code changes, but in a different order so we can more easily keep the tests green through the refactor.

My plan here:

1. Extract reusable API for explicitly asking for rewritten packages, as in https://github.com/embroider-build/embroider/pull/1374/commits/aac1c754f85cfe63f6b4b6240bb647a3b4f6347a.
2. Use that API explicitly everywhere that wants to see rewritten packages.
3. This eliminates the need for MovedPackageCache and lets all package caches get unified as in https://github.com/embroider-build/embroider/pull/1374/commits/d7c17fee90a5c53704cd77b085b2c2fe2bb698ae.
4. Then it should be easier to land the simplified stage1 as in https://github.com/embroider-build/embroider/pull/1374/commits/cc3185e778d2651f183a4db2a163618058bf7521

PR #1488, opened 2023-06-29
fix this.import from node_modules in v1 addons

mansona:
As part of #1435 we have moved all vendor code from addons to `./node_modules/.embroider/rewritten-packages/@embroider/synthesized-vendor/`, which works when you call `this.import()` from an addon's `included()` hook.

Perhaps as a copy & paste error we were ignoring the case where `this.import()` was called on a file in `node_modules`, this PR will fix that.

Note: I have committed and pushed a failing test that I will fix in a subsequent commit so that we can see that the issue exists on main and is fixed by the commit 👍

### DIFF
--- a/packages/compat/src/compat-addons.ts
+++ b/packages/compat/src/compat-addons.ts
@@ -49,10 +49,7 @@ export default class CompatAddons implements Stage {
     if (!this.treeSync) {
       this.treeSync = new TreeSync(
         addons,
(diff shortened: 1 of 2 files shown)


void-mAlex (approved):

ef4:
Merged, but a thought for next time: it would be good to consolidate scenarios better. Not every test needs a whole app around it and the test suite overhead is considerable. 

We could make it clearer that there's a standard scenario for testing app features and a standard scenario for testing addon features and a test like this could be one new test in the addon scenario.

PR #1489, opened 2023-06-29
prepare to release patch of embroider/compat

mansona:


NullVoxPopuli (approved):

.release-plan.json
```diff
@@ -10,103 +10,31 @@
       "oldVersion": "3.0.0"
```
NullVoxPopuli, line 10: I'm confused by the diff in this file -- were its contents stale?

NullVoxPopuli: if the contents are stale, should we clear the config for each package after its release?

ef4 (approved):

mansona:
fyi I'm going to merge in spite of the flaky tests. Github seems to be stuck with a queued runner 😫

.release-plan.json
void-mAlex: would also work better when publishing bugfixes in lower versions

PR #1490, opened 2023-06-29
preparing patch 3.1.1

ef4:


.release-plan.json
```diff
@@ -11,15 +11,15 @@
...
+      "oldVersion": "3.1.1",
+      "newVersion": "3.1.2",
       "constraints": [
         {
           "impact": "patch",
           "reason": "Appears in changelog section :bug: Bug Fix"
         }
       ],
-      "pkgJSONPath": "/Users/mansona/git/opensource/ember/embroider/packages/compat/package.json"
+      "pkgJSONPath": "/Users/edward/hacking/embroider/packages/compat/package.json"
```
ef4, line 22: I'm also reminded that we have an open issue here where these paths are absolute, so you must run prepare and publish on the same machine.

PR #1494, opened 2023-06-29
Make proxied rewritten-package-cache methods clearer

ef4:
These worked like this but it's unnecessarily confusing to debug and read.


NullVoxPopuli:
like, some stack trace didn't include getter names, but does include function names? 

this would be surprising to me, because in codemod-land, they are the same:
image
I suppose I maybe haven't looked too deeply into actual error-time differences here

PR #1495, opened 2023-06-29
Eagerly virtualize emberVirtualPackages

ef4:
... and expand the list of virtual pacakges to includes some that are private-but-used-API.

Before this, we only provided these as fallbacks when normal resolving failed, and this mostly worked. But that was when we were also rewriting the entire node_modules graph on disk. Now that we don't do that, there are many more opportunities to accidentally resolve copies of packages that aren't really what you expected.

### DIFF
--- a/packages/shared-internals/src/ember-standard-modules.ts
+++ b/packages/shared-internals/src/ember-standard-modules.ts
@@ -37,6 +37,11 @@ emberVirtualPackages.add('@embroider/macros');
 // the modules-api-polyfill. Newer APIs need to be added here.
 emberVirtualPackages.add('@ember/owner');
 
+// these are not public API but they're included in ember-source, so for
+// correctness we still want to understand that they come from there.
+emberVirtualPackages.add('@glimmer/validator');
+emberVirtualPackages.add('@glimmer/manager');
+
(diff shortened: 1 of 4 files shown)


packages/shared-internals/src/ember-standard-modules.ts
```diff
@@ -37,6 +37,11 @@ emberVirtualPackages.add('@embroider/macros');
 // the modules-api-polyfill. Newer APIs need to be added here.
 emberVirtualPackages.add('@ember/owner');
 
+// these are not public API but they're included in ember-source, so for
+// correctness we still want to understand that they come from there.
+emberVirtualPackages.add('@glimmer/validator');
```
NullVoxPopuli, line 42: should all the glimmer packages from ember-source be listed here?
- `@glimmer/compiler`
- `@glimmer/syntax`
- etc?

NullVoxPopuli (approved):

packages/shared-internals/src/ember-standard-modules.ts
ef4: It would be good to be complete. Also it would be good to make these lists ember-version-dependent.

PR #1499, opened 2023-06-30
add staticAddonServices:false default option to control if addon serv…

### LINKED ISSUE #1498: addon owned service with staticAddonTrees breaks DI
Uses ember-scroll-modifiers
this relies on a service 
https://github.com/elwayman02/ember-scroll-modifiers/blob/master/addon/services/observer-manager.js

which is NOT re-exported but only used in the modifier inside the addon itself

https://github.com/elwayman02/ember-scroll-modifiers/blob/master/addon/modifiers/did-intersect.js#L14

reproduction can be found here 

* `git clone https://github.com/void-mAlex/full_static_flags_reproduction.git`
* `cd full_static_flags_reproduction/static_embroider_v3`
* `pnpm install`
* `pnpm ember serve`
* Visit your app at [http://localhost:4200](http://localhost:4200).

will see a js exception in the console and _brokewn_ text on screen

to 'fix'
in `ember-cli-build.js` comment out 
`staticAddonTrees` and `staticAddonTestSupportTrees` and restart ember server and it will show it works

core of the issue seems to be the mechanism by which DI system works

scroll modifiers exports a `did-intersect` modifier which when used in the host app gets an owner of the host app
when service injection looks up the requested service from inside the modifier it checks only on the owner where
due to staticAddonTrees we no longer register services from addons (re-exported or otherwise)

short of major enhancement of the DI system I am proposing we allow services to still be registered with staticAddonTrees enabled

any objections @ef4 ?
looking to expand on https://github.com/void-mAlex/embroider/commit/ac1d96eb52bfc892ba661249f72d919fba018d30

void-mAlex:
…ices should always be forced into the build

fixes #1498

### DIFF
--- a/packages/compat/src/options.ts
+++ b/packages/compat/src/options.ts
@@ -43,6 +43,18 @@ export default interface Options extends CoreOptions {
   // apply.
   staticAddonTestSupportTrees?: boolean;
 
+  // Controls whether your addon's "addonServices" trees should be resolved
+  // statically at build time.
+  //
+  //   false (the default): implies maximum backward compatibility at the cost
+  //   of bigger builds. All services files will be forced into your Ember
+  //   app, which is the legacy behavior.
+  //
+  //   true: produces smaller builds. Only files that are explicitly referenced
+  //   will end up in your app.
+  //
+  staticAddonServices?: boolean;
+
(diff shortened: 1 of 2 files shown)


void-mAlex:
closing in favor of #1543

PR #1500, opened 2023-06-30
Make the package-json path repo-relative

NullVoxPopuli:
Unblocks: https://github.com/embroider-build/embroider/pull/1492

Today, in the release-plan.json, the pkgJSONPath is the absolute path to each package.json _on the machine that ran the command last_.

To allow for anyone to generate and interpret a release-plan, we need to move to repo-relative paths.

All the tools that I saw that read the pkgJSONPath will respect the current working directory -- but that would mean that the `embroider-release` tool can only be ran from the repo root -- so whenever the relative path is read, I converted it to an absolute path.

To test:
- merge a pr, but don't release yet
- run `pnpm i`
- run `pnpm compile --watch` (or without watch)
- run `export GITHUB_AUTH=xyz` (I used a classic token with repo permissions that I've since deleted)
- run `pnpm exec embroider-release explain-plan` (to see what to expect in the .release-plan.json)
- run `pnpm exec embroider-release prepare` (to generate / update the .release-plan.json)

Diff should look something like this:
image

### DIFF
--- a/test-packages/release/src/interdep.ts
+++ b/test-packages/release/src/interdep.ts
@@ -1,5 +1,5 @@
 import glob from 'globby';
-import { resolve } from 'path';
+import { resolve, relative } from 'path';
 import { readFileSync, readJSONSync } from 'fs-extra';
 import yaml from 'js-yaml';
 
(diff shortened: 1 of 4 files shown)


mansona (approved):
I ran the steps and it worked 🤷 I feel like this needs to be tested on main 😂 LGTM

PR #1510, opened 2023-07-02
Replace rollup-plugin-ts with @rollup/plugin-typescript

ef4:
rollup-plugin-ts tries to do too much and has issues. @rollup/plugin-typescript is in the official rollup org and supports a mode that only emits declarations while leaving the build to babel.

### DIFF
--- a/packages/router/rollup.config.mjs
+++ b/packages/router/rollup.config.mjs
@@ -1,6 +1,6 @@
-import typescript from 'rollup-plugin-ts';
-import copy from 'rollup-plugin-copy';
+import typescript from '@rollup/plugin-typescript';
 import { Addon } from '@embroider/addon-dev/rollup';
+import { babel } from '@rollup/plugin-babel';
 
 const addon = new Addon({
   srcDir: 'src',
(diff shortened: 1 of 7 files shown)


ef4:
Also, I want to upgrade to typescript 5 and rollup-plugin-ts is blocking us from doing that.

NullVoxPopuli:
I'll make the corresponding changes on the v2 addon blueprint :tada:

packages/router/rollup.config.mjs
```diff
@@ -17,18 +17,18 @@ export default {
     // addon. Anything not listed here may get optimized away.
     addon.publicEntrypoints(['index.js']),
 
-    // Follow the V2 Addon rules about dependencies. Your code can import from
+    // Follow the V2 Addon rules about depecdndencies. Your code can import from
```
NullVoxPopuli, line 20:
```suggestion
    // Follow the V2 Addon rules about dependencies. Your code can import from
```

NullVoxPopuli (approved):

ef4:
Yeah I know that the case on this PR so far -- our router package -- is a simple case because it doesn't have any components. 

I agree that stand-alone glint for the declaration files is probably necessary.

Issue #1487, opened 2023-06-29
`externals: ["@glimmer/validator"]` does not work in apps


chancancode:
To reproduce:

1. `ember new --embroider zomg`
2. In `package.json`, add `"ember-addon": { "externals": ["@glimmer/validator"] }`
  Note: per spec, the key should be "ember" #1226 
3. In `app.js` or anywhere else, add an import for `@glimmer/validator`

Expected result: `@glimmer/validator` should only be included in once `vendor.js`, not in `chunks.*.js`

Actual result: `@glimmer/validator` is included in `vendor.js`, and then a second time in one of the chunks.

This is inefficient, but in the specific case of `@glimmer/validator` it actually causes an error when multiple copies are included. It seems like addons are [able to use the same configuration successfully](https://github.com/ember-migration-utils/tracked-built-ins-compat/pull/3) to avoid the problem, and per spec the `externals` key is one of the few configurations that are supposed to work in both apps and addons.

Using `require` instead works as a workaround.

ef4:
Another workaround is to tell webpack directly to treat it as commonjs (which means webpack itself will do the `require`):

```js
// ember-cli-build.js
compatBuild(app, Webpack, {
  packagerOptions: {
    webpackConfig: {
      externals: {
        '@glimmer/validator': 'commonjs @glimmer/validator',
      },
     },
   },
}
```

You're correct that the spec says this option is available to apps, however it's only available to apps that also declare `ember-addon.version === 2`, meaning they're already in v2 format. Embroider produces this format internally for apps, but it's not the format you author and even if you tried to opt-in to being v2 that is going to require a lot of other strictness that an app is not likely to be able to easily meet just to solve this bug.

We just shipped embroider 3.1 and have been noticing that this class of problem is likely to be more common in that release. The reason is that we no longer do a complete rewrite of node_modules, which leaves us more vulnerable to accidental resolutions now. Up until now, the ember-provided virtual packages have always just been a fallback when normal resolving fails, but I think we're forced to aggressively externalize all of them.

`@glimmer/validator` may not be a super common one, but `rsvp` definitely is and essentially every ember app has an accidentally-resolvable copy of `rsvp` in node_modules, since it's used by some of the broccoli tooling.

[...]

PR #1513, opened 2023-07-03
Add `@glimmer/reference` as a virtual package

chancancode:
Another fix to #1487

The fix in #1495 turned out to be incomplete and didn't solve my actual problem (which I reduced away in my report). The actual import in the app was for `@glimmer/reference`, which internally imports from `@glimmer/validator`. I think because in `@glimmer/reference`'s `package.json` it has its own dep on `@glimmer/validator` it escapes this rule and acquired the actual copy in `node_modules`, causing the duplication.

It seems like this would also be a problem if you, say, imported from `@glimmer/runtime` (which imports from `@glimmer/reference` and, in turns, `@glimmer/validator`)? I don't personally have that particular use case but I could see that happening. Do we just wait and see or is there something more targeted hard-coding that we should do specifically for `@glimmer/validator`?

### DIFF
--- a/packages/shared-internals/src/ember-standard-modules.ts
+++ b/packages/shared-internals/src/ember-standard-modules.ts
@@ -40,6 +40,7 @@ emberVirtualPackages.add('@ember/owner');
 // these are not public API but they're included in ember-source, so for
 // correctness we still want to understand that they come from there.
 emberVirtualPackages.add('@glimmer/validator');
+emberVirtualPackages.add('@glimmer/reference');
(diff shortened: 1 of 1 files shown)


chancancode:
Failure seems unrelated

ef4:
The more complete thing to do is to go through ember-source and capture the complete list of "packages" that it defines. Possibly per ember-source version. That would be the more reliable thing for us to use here.

For now this addition is good though. Thanks.

PR #1516, opened 2023-07-06
Move @embroider/macros from emberVirtualPackages to emberVirtualPeerDeps

ef4:
I noticed that `@embroider/macros/runtime` was getting force-externalized in an app. Which would make sense if it were really an emberVirtualPackage, but it's a real package so it should be listed in emberVirtualPeerDeps instead.

(It's true that most of what it exposes is "virtual" in the sense of having babel implementation, but the package does provide some runtime implementation too and that should be resolved to the real package)


ef4:
(This only started to matter due to the recent https://github.com/embroider-build/embroider/pull/1495)

void-mAlex (approved):

PR #1518, opened 2023-07-06
add a basic implementation of the gjs rollup plugin

mansona:
This is just testing out an end-to-end implementation of the swc-based content-tag preprocessor

There are a number of things that need to be done before we can merge this: 

- [x] merge the content-tag PRs
- [x] decide exactly where the content-tag npm package will be deployed to (i.e. what name and potentially what scope)
- [x] figure out why babel isn't happy with the output of content-tag 🤔 
- [x] decide if this is an independent rollup plugin or if it's part of the hbs one
- [x] https://github.com/embroider-build/content-tag/issues/1
- [x] https://github.com/embroider-build/content-tag/issues/3

If anyone can think of anything else let me know ;)

### DIFF
--- a/packages/addon-dev/src/rollup-gjs-plugin.ts
+++ b/packages/addon-dev/src/rollup-gjs-plugin.ts
@@ -1,0 +1,71 @@
+import { createFilter } from '@rollup/pluginutils';
+import type { Plugin, PluginContext, ResolvedId } from 'rollup';
+import { readFileSync } from 'fs';
+import { Preprocessor } from 'content-tag';
+
+const PLUGIN_NAME = 'rollup-gjs-plugin';
+
+const processor = new Preprocessor();
+// import { parse as pathParse } from 'path';
+
+export default function rollupGjsPlugin(): Plugin {
+  return {
+    name: PLUGIN_NAME,
+    async resolveId(source: string, importer: string | undefined, options) {
+      let resolution = await this.resolve(source, importer, {
+        skipSelf: true,
+        ...options,
+      });
+
+      if (resolution) {
+        return maybeRewriteGJS(resolution);
+      }
+    },
+
+    load(id: string) {
+      const meta = getMeta(this, id);
+      if (!meta) {
+        return;
+      }
+
+      this.addWatchFile(meta.originalId);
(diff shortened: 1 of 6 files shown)


packages/addon-dev/src/rollup-gjs-plugin.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+const gjsFilter = createFilter('**/*.gjs');
+
+function maybeRewriteGJS(resolution: ResolvedId) {
+  if (!gjsFilter(resolution.id)) {
+    return null;
+  }
+
+  // This creates an `*.gjs.js` that we will populate in `load()` hook.
```
ef4, line 61: This comment appears to be lying, yes? It doesn't turn foo.gjs into foo.gjs.js, it turns it into foo.js. 

I think the code is correct and it's just the comment that is misleading. We don't need to name the files foo.gjs.js. That was more important for foo.hbs.js because that needs to play nice beside foo.js.

packages/addon-dev/src/rollup-gjs-plugin.ts
```diff
@@ -0,0 +1,71 @@
...
+function getMeta(context: PluginContext, id: string): Meta | null {
+  const meta = context.getModuleInfo(id)?.meta?.[PLUGIN_NAME];
+  if (meta) {
+    return meta as Meta;
+  } else {
+    return null;
+  }
+}
+
+const gjsFilter = createFilter('**/*.gjs');
```
ef4, line 54: I think if you match gts here too (and rename gts to ts the same way you rename gjs to js) it will all just compose correctly for the typescript case too. content-tag already supports both automatically.

ef4 (approved):

packages/addon-dev/src/rollup-gjs-plugin.ts
mansona: so I figured it would be best to transform .gjs to .js files and .gts to .ts files so I expanded the filter and added a bit of code to rename the id correctly

I wanted to have a test but we actually don't have the TS setup correctly on the addon and rollup is complaining 🙈

PR #1521, opened 2023-07-08
ember-source as modules

ef4:
This restores a feature we did previously. At the time we backed it out because too many addons were relying on the `Ember` global in vendor.js. But after Ember 4 there is no `Ember` global, so it seems likely to work this time.

All our supported ember versions ship loose modules in their published package. The modules are normally not visible to the build though, because the ember-source addon manually packages them and stuffs them into vendor.js. This PR adds a compat adapter that stops that process.

The benefit of this is that each of the ember-source provided modules will only get pulled into the build if you actually use it.

One other problem we saw last time was that the deps that ship inside ember-source: 

```
.
├── @glimmer
│   ├── destroyable.js
│   ├── encoder.js
│   ├── env.js
│   ├── global-context.js
│   ├── low-level.js
│   ├── manager.js
│   ├── node.js
│   ├── opcode-compiler.js
│   ├── owner.js
│   ├── program.js
│   ├── reference.js
│   ├── runtime.js
│   ├── util.js
│   ├── validator.js
│   ├── vm.js
│   └── wire-format.js
├── @simple-dom
│   └── document.js
├── backburner.js
├── dag-map.js
├── route-recognizer.js
├── router_js.js
└── rsvp.js
```

started taking precedence globally over any copies of those packages that were consumed normally by other packages. This PR addresses that by making our `renamedModules` and `renamedPackages` compatibility features less aggressive -- they will now only take effect if the consuming package doesn't have an explicit dependency. That is, if a package tries to import from `rsvp`, if it doesn't have a dependency on `rsvp` we will redirect that to something like `ember-source/rsvp.js`. But if they do have an explicit dependency on rsvp we will leave it alone.

### DIFF
--- a/packages/compat/src/compat-adapters/ember-source.ts
+++ b/packages/compat/src/compat-adapters/ember-source.ts
@@ -2,12 +2,45 @@ import V1Addon from '../v1-addon';
 import buildFunnel from 'broccoli-funnel';
 import mergeTrees from 'broccoli-merge-trees';
 import AddToTree from '../add-to-tree';
-import { outputFileSync, unlinkSync } from 'fs-extra';
-import { join } from 'path';
+import { outputFileSync, readdirSync, unlinkSync } from 'fs-extra';
+import { join, resolve } from 'path';
+import { Memoize } from 'typescript-memoize';
 
 export default class extends V1Addon {
   get v2Tree() {
     return mergeTrees([super.v2Tree, buildFunnel(this.rootTree, { include: ['dist/ember-template-compiler.js'] })]);
+  }
+
+  // ember-source inlines a whole bunch of dependencies into itself
+  @Memoize()
+  private get includedDependencies() {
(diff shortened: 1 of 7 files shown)


ef4:
Ugh, why oh why does ember-source have router.js as both a hard-coded inlined dependency *and* a real NPM dependency?

ef4:
The final failures here are because ember-cli-fastboot uses `require` from within vendor.js. This is a thing we're going to have to work to deprecate across the ecosystem. I know other addons do it too, and that was one of the key findings in https://github.com/emberjs/ember.js/pull/20349, which is the upstream equivalent of the compat adapter in this PR.

We can probably still land this, but as a new option.

ef4:
This is landed as a new option `staticEmberSource`. Like all our options it defaults to the more-compatible, less-optimizable setting (`false`). It is included in the `embroiderOptimized` default test scenario.

PR #1524, opened 2023-07-11
Fix the `EMBROIDER_CONCAT_STATS` build error

### LINKED ISSUE #1523: `EMBROIDER_CONCAT_STATS` breaks the build
I'm following the [analyze.md guide ](https://github.com/embroider-build/embroider/blob/main/docs/analyzing.md#include-legacy-addons-in-analyzation) and when I try the `EMBROIDER_CONCAT_STATS=true npm run build` command the build fails with the following error:

`(0 , filesize_1.filesize) is not a function`

Looking at the stack trace and related code, it points to this line: https://github.com/embroider-build/embroider/blob/be02b4a40a7d240a07b2edfc613bb00c2c9e189c/packages/core/src/measure-concat.ts#L22

Or the compiled version:
```
            .map(([name, bytes]) => `  ${name}: ${(0, filesize_1.filesize)(bytes, {})}`)
```

It seems that v5 of the `filesize` package [only has a default export](https://github.com/avoidwork/filesize.js/blob/7b15687c0af9e716ab141e826fab16934f9b2c9e/lib/filesize.cjs.js#L137) but since it's imported as a [named import in the code](https://github.com/embroider-build/embroider/blob/be02b4a40a7d240a07b2edfc613bb00c2c9e189c/packages/core/src/measure-concat.ts#L4) it fails.

This is the commit that introduced that change: https://github.com/embroider-build/embroider/commit/1d350d93677218d4ff295bca147b2f9350557146#diff-cfbd9511262fbeca1aa6e2e563f54f6852e46b3ba0db0d1381d693d11fdca2a8L4

The latest version of filesize does support the named import, so I think updating to that should fix this problem. Manually updating the code in node_modules so it uses the default export fixed the issue locally. Forcing it to the latest version with overrides also worked.

I can create a PR but guidance would be appreciated if tests are needed.

Windvis:
Checking out the first commit should trigger the build error, the version bump fixes it.

Closes #1523

### DIFF
--- a/tests/scenarios/compat-app-embroider-concat-stats-test.ts
+++ b/tests/scenarios/compat-app-embroider-concat-stats-test.ts
@@ -1,0 +1,23 @@
+import { throwOnWarnings } from '@embroider/core';
+import { PreparedApp } from 'scenario-tester';
+import { appScenarios } from './scenarios';
+import QUnit from 'qunit';
+const { module: Qmodule, test } = QUnit;
+
+appScenarios
+  .map('compat-app-embroider-concat-stats', () => {})
+  .forEachScenario(function (scenario) {
+    Qmodule(scenario.name, function (hooks) {
+      throwOnWarnings(hooks);
+
+      let app: PreparedApp;
+
+      test('building with EMBROIDER_CONCAT_STATS works', async function (assert) {
+        app = await scenario.prepare();
+        let result = await app.execute('ember build', {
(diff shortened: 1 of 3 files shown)


tests/scenarios/compat-app-embroider-concat-stats-test.ts
Windvis, line 1: I have no idea what I'm doing here. I based this on the compat-app-dot-import.ts file and stripped everything that didn't seem needed. Let me know how I can improve this 😄.

ef4:
Thanks, this looks good.

PR #1528, opened 2023-07-12
Restore older node support

ef4:
`@embroider/shared-internals` is used by ember-auto-import, and ember-auto-import intentionally has a very wide support policy that includes node 12.

### DIFF
--- a/tests/scenarios/shared-internals-test.ts
+++ b/tests/scenarios/shared-internals-test.ts
@@ -46,7 +46,7 @@ Scenarios.fromProject(() => new Project('shared-internals-tests'))
           throw new Error('unable to locate our node version');
         }
         let nodebin = tryit.output.trim();
-        let result = await app.execute(`${nodebin} ./node_modules/.bin/qunit ./test.js`);
+        let result = await app.execute(`${nodebin} ./node_modules/qunit/bin/qunit.js ./test.js`);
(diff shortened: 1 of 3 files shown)


NullVoxPopuli (approved):

tests/scenarios/shared-internals-test.ts
```diff
@@ -0,0 +1,53 @@
...
+        let app = await scenario.prepare();
+
+        // if we just try to invoke node directly in a child process, our own
+        // volta settings dominate over the test app's
+        let tryit = await app.execute('volta which node');
+        if (tryit.exitCode !== 0) {
+          throw new Error('unable to locate our node version');
+        }
+        let nodebin = tryit.output.trim();
+        let result = await app.execute(`${nodebin} ./node_modules/.bin/qunit ./test.js`);
```
void-mAlex, line 49: consider my mind blown

void-mAlex (approved):

tests/scenarios/shared-internals-test.ts
ef4: Volta is pretty good at keeping you inside its little fence. Need to be tricky if you actually want to get out like this.

ef4: Maybe there's actually real Volta public API for doing this in a way that doesn't need tricks...

PR #1534, opened 2023-07-14
Fixes case when podModulePrefix is set to `my-app/routes`

### LINKED ISSUE #1533: Route splitting on pods found under `./app/routes/`
Repro repo: https://github.com/evoactivity/embroider-route-split-pods-routes-dir

When pods are found in the `./app/routes/` directory by setting `podModulePrefix` to `my-app/routes` _and_ they are route split the routes do not load with no runtime or build errors.

evoactivity:
Resolves #1533

I copied the pods scenarios test so this case is explicity tested against. Not sure if that is required.

### DIFF
--- a/packages/core/src/app-files.ts
+++ b/packages/core/src/app-files.ts
@@ -108,8 +108,8 @@ export class AppFiles {
       }
 
       if (
-        this.handleClassicRouteFile(relativePath) ||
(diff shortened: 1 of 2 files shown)


evoactivity:
I didn't realise the test name needed to be hyphenated. Once CI finishes I'll adjust to test.

ef4:
Thanks, nice work.

PR #1550, opened 2023-07-20
Initial test of vite integration

ef4:
This is not an automated test yet, just an app that you can run and the beginning of vite integrations to make it run.


packages/vite/index.d.ts
```diff
@@ -0,0 +1,5 @@
+export * from './src/resolver.js';
```
NullVoxPopuli, line 1: Was this supposed to be comitted? 🙃

ef4: Yes. This is working around the fact that our packages all compile to commonjs and the vite config is a real node ES module. Both the implementation and the types go through these top-level reexports, which happens to make the cjs interop work out ok.

ef4: Someone please do the big yak shave to get all our packages dual compiling as CJS and ESM so I don't have to. 😅

PR #1555, opened 2023-07-21
Fix release script

mansona:
I've had a chance to play with #1492 and I saw a few places where it could be improved: 

- the things that are checking github and npm should not be noisy if there is an **expected** 404 error
- --dryRun should not stop after the first package 🙈 
- we need a way to pass --otp=something to the publish stage

This PR fixes all of those 🎉


NullVoxPopuli (approved):
Beautiful! Thanks for following up!

PR #1556, opened 2023-07-21
support packages that use modules

void-mAlex:
.mjs files can be imported directly by an app
we should allow babel to handle any features the library is using to match the targets set in the app

babel is more than capable of transpiling .mjs files as it's an unambiguous extension that could use features the user is not targeting like private fields or the next hottest ecma script proposal.

this change optionally matches those files as well for processing by babel when stage3 asks about it.

### DIFF
--- a/packages/shared-internals/src/babel-filter.ts
+++ b/packages/shared-internals/src/babel-filter.ts
@@ -26,9 +26,9 @@ export default function babelFilter(skipBabel: { package: string; semverRange?: 
 }
 
 function babelCanHandle(filename: string) {
-  // we can handle .js and .ts files with babel. If typescript is enabled, .ts
-  // files become resolvable and stage3 will be asking us if they should get
+  // we can handle .mjs, .js and .ts files with babel. If typescript is enabled,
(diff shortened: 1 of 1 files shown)


ef4:
I think this a fine incremental improvement and we can merge it.

But in the medium term, I expect this to be wrong because depending on the specifics of your build system, [babel may see all kinds of different extensions](https://github.com/embroider-build/embroider/blob/b8ac17b67836d5274412e63bcdeb23caa9bb9151/tests/vite-app/vite.config.mts#L23). But also, this code only exists to power the skipBabel option, which is only needed when users don't control their own babel config, which I expect to stop being a thing.

void-mAlex:
it's fine, more than happy to make this change in my own webpack/babel config one in able to but right now can't due to this being hardcoded
if my understanding of what your saying is correct I would be able to handle this config once it's a stand-alone file

this came up for me while trying to get embroider working for an app with a requiement for ie 11 as a target and not having any success with passing webpack config for targets : ['web','es5']
this change worked to handle a module only package that is a dependency of the app
would love to get this as an incremental improvement

Issue #1514, opened 2023-07-04
Attempted to resolve `did-insert`, which was expected to be a modifier, but nothing was found


simonihmig:
Seeing [this error](https://github.com/simonihmig/ember-stargate/actions/runs/5457107753/jobs/9930794915#step:5:97) pop up (in this [PR](https://github.com/simonihmig/ember-stargate/pull/655), second [70c78db](https://github.com/simonihmig/ember-stargate/pull/655/commits/70c78db19c8cb9c006d2a9b68b0f3102acb65b51) commit). 

I am using `@ember/render-modifiers` (which I probably should get rid off, but they should work nevertheless), but they don't seem to be included into the build. This does get "fixed" by [downgrading to Embroider 2](https://github.com/simonihmig/ember-stargate/pull/655/commits/d4c72a74ffcf674a158dab5a93dc09b12199eb79)!

mansona:
I just confirmed this issue and I have a fix for it (but it is not a good permanent solution) 

Essentially this is the same case that I brought up in the embroider meeting yesterday, the components that are being re-exported via the app exports are depending on something that is in the **dependencies** of the addon. This doesn't work and we agreed that this should be considered a bug from the embroider perspective

The "fix" that I was talking about was to move `@ember/render-modifiers` to be a peer-dependency of the addon and install it directly in the app, but that isn't something that you probably want to ship and we should just fix it upstream

PR #1560, opened 2023-07-25
Fix rewritten package cache encapsulation

ef4:
Fixes https://github.com/embroider-build/embroider/issues/1514

A v2 addon with a v1 addon dependency was not resolving its dependencies correctly, in two places.
 - in virtual-content.ts where we emit implicit-modules
 - in compat-app-builder.ts when doing partitionEngines

The fix to both is adjusting the meaning of RewrittenPackageCache so that it always provides packages that see moved dependencies, even when a given package *itself* has not moved (because it's a v2 addon).

Along with that I took the opportunity to make `packageCache.original()` more like `maybeMoved()` in that returns its argument if no change was needed, which simplified things for the majority of callers.

### DIFF
--- a/packages/babel-loader-8/package.json
+++ b/packages/babel-loader-8/package.json
@@ -12,6 +12,9 @@
     "@babel/core": "^7.14.5",
     "babel-loader": "8"
   },
+  "devDependencies": {
+    "@embroider/core": "workspace:^"
+  },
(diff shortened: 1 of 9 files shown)


NullVoxPopuli (approved):

mansona (approved):
🎉

PR #1563, opened 2023-07-28
Add semverRange <=4.11.0 for ember-data debug

### LINKED ISSUE #1506: Embroider optimized fails with Ember Data 4.12.x
In an addon with ember-data installed the [embroider optimized scenario](https://github.com/ilios/common/actions/runs/5426907988/jobs/9869666714) fails with:

```
Uncaught Error: Could not find module `@ember-data/store` imported from `(require)
```

Steps to reproduce

1. `$ ember addon emberFiveDataFourEmbroider`
2. `$ ember install ember-data@4.12.3`
3. `$ ember try:one embroider-optimized`

Just in case I tried it after the release of `v3.1.4` and it still fails. It passes with ember-data v5.x (or least doesn't fail in this obvious way).

mkszepp:
fix #1506
There is missing the `semverRange` rule for ember-data/debug in (https://github.com/embroider-build/embroider/pull/1435 - commit https://github.com/embroider-build/embroider/commit/0382837a2ccf74ad6e5f885bc399b9b31bdafd52)

### DIFF
--- a/packages/compat/src/compat-adapters/@ember-data/debug.ts
+++ b/packages/compat/src/compat-adapters/@ember-data/debug.ts
@@ -1,5 +1,6 @@
 import { AddonMeta } from '@embroider/core';
 import V1Addon from '../../v1-addon';
+import semver from 'semver';
 
 export default class EmberDataDebug extends V1Addon {
   get packageMeta(): Partial<AddonMeta> {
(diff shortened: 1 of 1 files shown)


ef4:
Thanks.

PR #1587, opened 2023-09-01
Add a v2 addon "nested layout" component test

Windvis:
Discord discussion: https://discord.com/channels/480462759797063690/1147119814444318863

When trying out Embroider in one of our apps I noticed an error related to ember-velcro. The issue seems related to the "nested layout" that is sometimes used. It fails on Embroider v3 under the strict scenario.


Windvis:
It seems v1 addons can also have this issue: https://discord.com/channels/480462759797063690/483601670685720591/1149081856088756235

Which points to this component: https://github.com/josemarluedke/frontile/blob/main/packages/forms/addon/components/form-field/index.ts

Windvis:
Superseded by #1596

PR #1588, opened 2023-09-02
Update macros README with v2 addon instructions

tcjr:
This clarifies where to put build-time options for v2 addons without a root `index.js` file.

### DIFF
--- a/packages/macros/README.md
+++ b/packages/macros/README.md
@@ -57,6 +57,34 @@ module.exports = {
 };
 ```
 
+## Setting Configuration: from a v2 Ember Addon
+
+1. Add `@embroider/macros` as `dependency`.
+2. In `addon-main.js`, do:
+
+```js
+'use strict';
+
+const { addonV1Shim } = require('@embroider/addon-shim');
+
+module.exports = {
+  ...addonV1Shim(__dirname),
+  options: {
+    '@embroider/macros': {
+      setOwnConfig: {
+        // your config goes here
+      },
+      setConfig: {
+        'some-dependency': {
+          // config for some-dependency
+        },
+      },
+    },
+  },
+};
+
+```
+
 ## The Macros
 
 ### macroCondition


ef4:
This not correct. If a V2 addon does this, it's not going to work under non-broccoli build systems, where the classical ember-cli addon main will never even be evaluated.

The V2 addon RFC [describes new build hooks](https://github.com/emberjs/rfcs/blob/master/text/0507-embroider-v2-package-format.md#build-hooks) for this purpose, though so far we have held off on implementing them.

What is your specific use case?

tcjr:
Ah, thanks @ef4.

> What is your specific use case?

I am trying to update the [ember-tippy](https://github.com/nag5000/ember-tippy) addon to the V2 format.  I managed to get everything working, except the build-time configuration setting (in this case, `shouldIncludeTippyCoreCss`) was not setup correctly.  The call to `getOwnConfig()` imported from `@ember/macros` returned undefined, so it was clear to me I was missing the options. 

In the porting guide, [Part 4, Step 13](https://github.com/embroider-build/embroider/blob/main/docs/porting-addons-to-v2.md#part-4-convert-addon-to-v2) it says:
> Create a new addon/addon-main.js file (this replaces addon/index.js)

But this addon has a build-time config option, so I consulted this macros doc where it says 
> in `index.js`, do this:

and shows where to put the options.

Since addon-main.js replaced index.js, I tried putting the options in that file (merging them with the rest) and it worked! The call to `getOwnConfig()` now returned the options as expected.  Is there a correct place to put build-time config options?

I don't really need this addon to be v2, I was just trying to get some experience by updating an addon I had some familiarity with and ran into this.

Would you recommend holding off updating addons with config options to v2 for now?

ef4:
> Would you recommend holding off updating addons with config options to v2 for now?

No, I think this addon is probably fine to update. 

The source of the difficulty here is that an option like `shouldIncludeTippyCoreCss` is only needed because the import of the tippy core CSS is in the wrong place: if the app is allowed to decide whether or not to include Tippy Core CSS, then the app should do the import of it.

tcjr:
> then the app should do the import of it

Ok, I'll submit a PR to remove the config option and require the consuming app to import the base css. It looks like this is [the approach ember-pikaday uses](https://github.com/adopted-ember-addons/ember-pikaday#styles) to avoid this issue as well. Thank you.

PR #1590, opened 2023-09-05
Move away from jest to vitest

NullVoxPopuli:
Motivation:
- get the tests out of "the build"
- simpler typescript setup
- no need to compile before testing
- "faster"
- explicit imports
  - however, this revealed we have test stuff showing up in published JS :( --- so that needs untangling

Blockers: 
- https://github.com/embroider-build/embroider/pull/1591 (thanks, @mansona !!)
- resolving module-import goofiness (test stuff being in src, etc)


NullVoxPopuli:
too many conflicts

PR #1591, opened 2023-09-05
Lay groundwork for esm/cjs multi build from TS

mansona:
This is following a guide in the TS docs to setup "project references" https://www.typescriptlang.org/docs/handbook/project-references.html and allows us to have local builds for each of the packages but global config.

This is the same setup that the typescript monorepo is using so I'm pretty sure it's ok 😂 

This PR gives us something that @ef4 has been wanting for a while now: when developing your build artifacts don't litter your src dirs any more 🎉 but the main idea behind this is that we should be able to add a second build for MJS output to each of the packages without too much difficulty once this is merged 👍


.github/workflows/ci.yml
```diff
@@ -31,8 +31,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup
-      - name: lint
-        run: pnpm lint
+      # - name: lint
```
NullVoxPopuli, line 34: why skip lint?
only looks like prettier issues: https://github.com/embroider-build/embroider/actions/runs/6090510549/job/16525470139#step:4:36

mansona: because lint is hanging forever at the moment 😞 I haven't invested the time into fixing it yet, but It could be because of this change https://github.com/embroider-build/embroider/pull/1591/files#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5L26-L28 which was necessary

NullVoxPopuli: Probs gotta update some ignore files

PR #1593, opened 2023-09-06
[draft] Add strict mode to app scenarios

lolmaus:


ef4:
Lots of tests have specific opinions about these settings, I don't think we can force them across all tests.

PR #1594, opened 2023-09-06
remove volta from CI

mansona:
This PR attempts to just use the standard `actions/setup-node` infra to setup node because the volta infra keeps crashing on us


.github/actions/setup/action.yml
```diff
@@ -8,19 +8,14 @@ inputs:
...
-    - name: Get pnpm cache directory path
-      id: pnpm-cache-dir-path
-      run: echo "STORE_PATH=$(pnpm store path)" >> $GITHUB_OUTPUT
-      shell: bash
-    - uses: actions/cache@v3
-      id: pnpm-cache
+    - uses: pnpm/action-setup@v2
+      name: Install pnpm
+      with:
+        version: 7
```
void-mAlex, line 14: don't we want pnpm version 8?

mansona: lol yes, but also I just noticed that our volta was pinning us to before the `resolution-mode setting is changed to highest` shenanigans so we'll see if anything breaks 🙃

NullVoxPopuli (approved):

.npmrc
```diff
@@ -1 +1,2 @@
-auto-install-peers=false
\ No newline at end of file
+auto-install-peers=false
```
NullVoxPopuli, line 1: we probably also want: https://github.com/ember-cli/ember-cli/blob/master/blueprints/addon/files/.npmrc#L8

can be separate, ofc

void-mAlex:
seems like the one test that is failing is related directly to volta shenanigans ... we may need to see how that translates in a world without volta
https://github.com/void-mAlex/embroider/blob/main/tests/scenarios/shared-internals-test.ts#L44

tests/scenarios/shared-internals-test.ts
```diff
@@ -39,14 +39,7 @@ Scenarios.fromProject(() => new Project('shared-internals-tests'))
...
 
-        // if we just try to invoke node directly in a child process, our own
-        // volta settings dominate over the test app's
-        let tryit = await app.execute('volta which node');
-        if (tryit.exitCode !== 0) {
-          throw new Error('unable to locate our node version');
-        }
-        let nodebin = tryit.output.trim();
-        let result = await app.execute(`${nodebin} ./node_modules/qunit/bin/qunit.js ./test.js`);
+        let result = await app.execute(`node ./node_modules/qunit/bin/qunit.js ./test.js`);
```
void-mAlex, line 42: the above test is names testing on node 12 ... from qunit

mansona: I added testing this with node 12 as a separate job, I was going to remove it from the main matrix but I thought it wouldn't help 🤷

ef4:
Volta maintainers discussing a fix: https://github.com/volta-cli/volta/issues/1523#issuecomment-1732187410

mansona:
I'm submitting this PR for consideration again. I have implemented the pnpm node version management for the tests that need it but I have not implemented it across the board.

My reasoning for this is that we specifically have tests that use `--no-lockfile` because we want to track when our CI is failing because of changes in our dependencies. With volta and pnpm's `use-node-version` implementation you have to pin an exact version of node which does not seem in the spirit of `--no-lockfile`. This PR also identified the problem that I describe in #1649 because it also didn't pin an exact version of pnpm

I will add this PR to the meeting agenda for discussion 👍

NullVoxPopuli (approved):

ef4 (approved):

PR #1596, opened 2023-09-10
Resolve "nested layout" components

Windvis:
This adds a new (failing) "nested-layout-components" scenario and the fix for it in a second commit.

Supersedes #1587 since it's not only a v2 addon issue.

### DIFF
--- a/packages/core/src/module-resolver.ts
+++ b/packages/core/src/module-resolver.ts
@@ -565,6 +565,7 @@ export class Resolver {
 
   private *componentJSCandidates(inPackageName: string) {
     yield { prefix: '/components/', suffix: '' };
+    yield { prefix: '/components/', suffix: '/index' };
     yield { prefix: '/components/', suffix: '/component' };
 
     let pods = this.podPrefix(inPackageName);
@@ -954,7 +955,7 @@ export class Resolver {
         throw new Error(
           `A module tried to resolve "${request.specifier}" and didn't find it (${label}).
 
- - Maybe a dependency declaration is missing? 
+ - Maybe a dependency declaration is missing?
  - Remember that v1 addons can only import non-Ember-addon NPM dependencies if they include ember-auto-import in their dependencies.
  - If this dependency is available in the AMD loader (because someone manually called "define()" for it), you can configure a shim like:
 
--- a/tests/scenarios/nested-layout-components-test.ts
+++ b/tests/scenarios/nested-layout-components-test.ts
@@ -1,0 +1,116 @@
+import { appScenarios, baseAddon, baseV2Addon } from './scenarios';
+import type { PreparedApp } from 'scenario-tester';
+import QUnit from 'qunit';
+import merge from 'lodash/merge';
+
+const { module: Qmodule, test } = QUnit;
+
+appScenarios
+  .map('nested-layout-components', project => {
+    const v1Addon = baseAddon();
+    v1Addon.pkg.name = 'v1-addon';
+    merge(v1Addon.files, {
+      app: {
+        components: {
+          'v1-nested-layout-component': {
+            'index.js': 'export { default } from "v1-addon/components/v1-nested-layout-component";',
+          },
+        },
+      },
+      addon: {
+        components: {
+          'v1-nested-layout-component': {
+            'index.hbs': '<div>nested layout components in v1 addons work</div>',
+          },
+        },
+      },
+    });
+    project.addDevDependency(v1Addon);
+
+    const v2Addon = baseV2Addon();
+    v2Addon.pkg.name = 'v2-addon';
+    (v2Addon.pkg as any)['ember-addon']['app-js']['./components/v2-nested-layout-component/index.js'] =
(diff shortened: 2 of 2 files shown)


tests/scenarios/nested-layout-components-test.ts
```diff
@@ -0,0 +1,112 @@
...
+  })
+  .forEachScenario(scenario => {
+    Qmodule(scenario.name, function (hooks) {
+      let app: PreparedApp;
+      hooks.before(async () => {
+        app = await scenario.prepare();
+      });
+
+      test(`pnpm test`, async function (assert) {
+        let result = await app.execute('EMBROIDER_TEST_SETUP_OPTIONS="optimized" pnpm test');
```
Windvis, line 108: Not sure if there are better ways to enable the "optimized" preset for a specific scenario, but this seemed to work.

packages/core/src/module-resolver.ts
```diff
@@ -565,6 +565,7 @@ export class Resolver {
 
   private *componentJSCandidates(inPackageName: string) {
     yield { prefix: '/components/', suffix: '' };
+    yield { prefix: '/components/', suffix: '/index' };
```
Windvis, line 568: When I added a similar line to the `componentTemplateCandidates` generator method I ran into an issue with the v2 component. Something about not being allowed to call `setComponentTemplate` twice. My guess is that the v2 component is resolved twice if that scenario, since the extensions logic checks for both .js and .hbs files.

I haven't dug deeper to find the origins of the current setup. Only adding it here made the tests pass and I noticed `yield { prefix: '/components/', suffix: '' };` wasn't added to `componentTemplateCandidates` either, which might be because of similar reasons?

tests/scenarios/nested-layout-components-test.ts
Windvis: Nevermind, that doesn't work on Windows, but I found other examples in different scenarios.

tests/scenarios/nested-layout-components-test.ts
Windvis, line 1: This adds a lot of extra CI jobs. Should I try to fit it in another scenario?

ef4: Yes please. If this fix is entirely in `module-resolver.ts` it can get tested in `core-resolve-test.ts`.

ef4:
Thanks for working on this. 

I was guessing the fix for this would actually be at a slightly different spot. But I haven't dug in myself yet, so your solution may also be right.

Is the problematic case the one where `componentJSCandidates()` is called by `tryReverseComponent()` or is it the one where it's called by `resolveComponent()`? I think in `resolveComponent()` this fix would be superfluous because there the output is sent to `nodeResolve`, and that would necessarily handle `index.js` automatically.

If the problematic case happens only inside `tryReverseComponent` then the fix may be better inside that method.

I was also suspecting the problem could be in `getEntryFromMergeMap()`, because there we match over multiple candidate extensions but potentially should also include the `index` variant of each extension as well. If this is an issue for addon-provided components but not app-provided components, it's the mergeMap I'd expect to be the problem.

Windvis:
> Is the problematic case the one where componentJSCandidates() is called by tryReverseComponent() or is it the one where it's called by resolveComponent()? I think in resolveComponent() this fix would be superfluous because there the output is sent to nodeResolve, and that would necessarily handle index.js automatically.

As far as I can see, it's the `resolveComponent` code path where it goes wrong. `nodeResolve` does get called, but it tries to look up the app reexports of those addon components and they don't seem to be there when I check the folder locally.

For example, in the tests I added here it looks for `app-template/components/v1-nested-layout-component` in a tmp folder, so it should in theory look for the `/index.*` version as well, but that folder isn't there. It would then go to the `fallbackResolve` code path where the `/index` suffixed one does get found.

> I was also suspecting the problem could be in getEntryFromMergeMap(), because there we match over multiple candidate extensions but potentially should also include the index variant of each extension as well. If this is an issue for addon-provided components but not app-provided components, it's the mergeMap I'd expect to be the problem.

Yea, I originally planned to do it like that, but then I saw that there was also special handling for pod layouts in the `componentJSCandidates` method and I figured this was more in line with that.

In any case, it's probably going to take me a while to grasp the logic we're dealing with here, so no hard feelings if you would rather look into it yourself to find the real underlying issue :smile:.

lolmaus:
@Windvis If you happen to need a strict mode test (as you mentioned in your [previous PR](https://github.com/embroider-build/embroider/pull/1587)), you can use [this draft](https://github.com/embroider-build/embroider/pull/1593/files?w=1) as a pattern, except enable the strict mode for an individual test case.

ef4:
I don't fully understand why this would be the correct place to fix this bug. `/index` is already implicitly allowed by `nodeResolve`. Maybe the bug here is really in the app-js matching?

We can use the test from here to debug it.

Windvis:
Closing since I won't have the time to look into this again and it wasn't the correct fix anyways.

Issue #1602, opened 2023-09-14
[Addon V2] Dependencies conflict between the addon and the test-app with the same package version for both (ember-source)


Shishouille:
Using `pnpm`, when installing `ember-source: ~5.2.0` in both addons and test-app, I got this error when launching tests:

```
test-app:test: Stack Trace and Error Report: /var/folders/jg/tyc48ncs3c99g10_vpqcckyr0000gn/T/error.dump.463e47596f7c3b8f29591ef187f10b92.log
test-app:test: Some V1 ember addons are resolving as incorrect peer dependencies. This makes it impossible for us to safely convert them to v2 format.
test-app:test: See https://github.com/embroider-build/embroider/blob/main/docs/peer-dependency-resolution-issues.md for an explanation of the problem and suggestions for fixing it.
test-app:test:
test-app:test: test-app@0.0.0 -> ember-formidable@0.0.0-beta.13 -> @ember/render-modifiers@2.1.0
test-app:test:     sees peerDep ember-source@5.2.0
test-app:test:       at /Users/home/DEV/BEVOLTA/ember-formidable/node_modules/.pnpm/ember-source@5.2.0_@babel+core@7.22.10_@glimmer+component@1.1.2_@glint+template@1.0.2/node_modules/ember-source
test-app:test:     but ember-formidable@0.0.0-beta.13 is using ember-source@5.2.0
test-app:test:       at /Users/home/DEV/BEVOLTA/ember-formidable/node_modules/.pnpm/ember-source@5.2.0_@babel+core@7.22.10_@glimmer+component@1.1.2_@glint+template@1.0.2_webpack@5.88.2/node_modules/ember-source
test-app:test:
test-app:test: test-app@0.0.0 -> ember-formidable@0.0.0-beta.13 -> @ember/render-modifiers@2.1.0
test-app:test:     sees peerDep ember-source@5.2.0
test-app:test:       at /Users/home/DEV/BEVOLTA/ember-formidable/node_modules/.pnpm/ember-source@5.2.0_@babel+core@7.22.10_@glimmer+component@1.1.2_@glint+template@1.0.2/node_modules/ember-source
test-app:test:     but test-app@0.0.0 is using ember-source@5.2.0
[...]

PR #1603, opened 2023-09-15
Add some action items to the peer errors when node_modules is messed up

NullVoxPopuli:
Resolves: https://github.com/embroider-build/embroider/issues/1602

### DIFF
--- a/packages/compat/src/standalone-addon-build.ts
+++ b/packages/compat/src/standalone-addon-build.ts
@@ -29,6 +29,11 @@ export function convertLegacyAddons(compatApp: CompatApp) {
         `Some V1 ember addons are resolving as incorrect peer dependencies. This makes it impossible for us to safely convert them to v2 format.
 See https://github.com/embroider-build/embroider/blob/main/docs/peer-dependency-resolution-issues.md for an explanation of the problem and suggestions for fixing it.
 
+Some potential ways to address the following issues:
+  - ensure your package-manager version is up to date
+  - delete all node_modules directories and re-install your dependencies
+  - deduplicate dependencies (e.g.: pnpm dedupe)
+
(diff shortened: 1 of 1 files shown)


packages/compat/src/standalone-addon-build.ts
```diff
@@ -29,6 +29,11 @@ export function convertLegacyAddons(compatApp: CompatApp) {
         `Some V1 ember addons are resolving as incorrect peer dependencies. This makes it impossible for us to safely convert them to v2 format.
 See https://github.com/embroider-build/embroider/blob/main/docs/peer-dependency-resolution-issues.md for an explanation of the problem and suggestions for fixing it.
 
+tl;dr: action items from the above documentation
+  - ensure your package-manager version is up to date
+  - delete all node_modules directories and re-install your dependencies
```
void-mAlex, line 34: would in most cases also need to remove your lock file

as a whole I don't feel confident this is the correct advice to give out as it lacks context of how a particular package manager behaves and which one of the many problems you could be encountering this may or may not fix;

doing a reroll of your dependencies could actually be harmful depending on your declared ranges in in package json and send people chasing down different non-errors

tl;dr: it's complicated and doing these steps without understanding why, likely not helpful

NullVoxPopuli: > would in most cases also need to remove your lock file

real projects can't seriously consider this option. It's perfectly fine in OSS tho

> [remove lock file] as a whole I don't feel confident this is the correct advice to give out

I agree :) 

> tl;dr: it's complicated and doing these steps without understanding why, likely not helpful

if the lockfile removal was included, sure, but the 3 steps listed "should be safe"

void-mAlex (approved):

PR #588, opened 2020-10-25
Passthrough HTML attributes

bobisjan:
- closes #456 
- [ ] add tests?

PR #1605, opened 2023-09-19
Ensure custom HTML attributes are passed-through

### LINKED ISSUE #456: Should Embroider passthrough script attributes?
Hi,

should Embroider passthrough the HTML attributes like `defer` on scripts from `app/index.html` into `dist/index.html` for `ember build --environment=production`?

```html
<!-- app/index.html -->
<script src="{{rootURL}}assets/vendor.js" defer></script>
<script src="{{rootURL}}assets/embroider-playground.js" defer></script>

<!-- dist/index.html -->
<script src="/assets/vendor.js"></script>
<script src="/assets/chunk.c1fdfc6b793b9d023c96.js"></script>
<script src="/assets/chunk.4aecffffbd029f26d4de.js"></script>
```

Thanks ☺️

---

Just for curiosity why the `assets/embroider-playground.js` is split into two chunks by default?

chancancode:
This ensures user-defined attributes on `<script>` and `<link>` tags in the compat `index.html` are propagated to the final HTML file. Replaces #588

Fixes #456

### DIFF
--- a/packages/compat/src/compat-app-builder.ts
+++ b/packages/compat/src/compat-app-builder.ts
@@ -223,16 +223,15 @@ export class CompatAppBuilder {
         rootURL: this.rootURL(),
         prepare: (dom: JSDOM) => {
           let scripts = [...dom.window.document.querySelectorAll('script')];
-          let styles = [...dom.window.document.querySelectorAll('link[rel="stylesheet"]')] as HTMLLinkElement[];
-
+          let styles = [...dom.window.document.querySelectorAll('link[rel*="stylesheet"]')] as HTMLLinkElement[];
           return {
-            javascript: definitelyReplace(dom, this.compatApp.findAppScript(scripts, entrypoint)),
-            styles: definitelyReplace(dom, this.compatApp.findAppStyles(styles, entrypoint)),
-            implicitScripts: definitelyReplace(dom, this.compatApp.findVendorScript(scripts, entrypoint)),
-            implicitStyles: definitelyReplace(dom, this.compatApp.findVendorStyles(styles, entrypoint)),
-            testJavascript: maybeReplace(dom, this.compatApp.findTestScript(scripts)),
-            implicitTestScripts: maybeReplace(dom, this.compatApp.findTestSupportScript(scripts)),
-            implicitTestStyles: maybeReplace(dom, this.compatApp.findTestSupportStyles(styles)),
+            javascript: this.compatApp.findAppScript(scripts, entrypoint),
+            styles: this.compatApp.findAppStyles(styles, entrypoint),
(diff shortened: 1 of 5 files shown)


packages/compat/src/compat-app-builder.ts
```diff
@@ -1365,18 +1364,6 @@ export class CompatAppBuilder {
...
-    return definitelyReplace(dom, element);
-  }
-}
-
-function definitelyReplace(dom: JSDOM, element: Element): Node {
-  let placeholder = dom.window.document.createTextNode('');
-  element.replaceWith(placeholder);
-  return placeholder;
-}
-
```
chancancode, line 1379: As far as I can tell these are somewhat duplicated by the same thing in core (previously known as `NodeRange`).

packages/core/src/ember-html.ts
```diff
@@ -25,57 +26,84 @@ export interface EmberHTML {
   implicitTestStyles?: Node;
 }
 
-class NodeRange {
```
chancancode, line 28: I think there are multiple bugs in the original implementation. For example, we never updated `this.end` when we insert something, so the clear code probably does not work as intended.

packages/core/src/html-entrypoint.ts
```diff
@@ -66,7 +66,7 @@ export class HTMLEntrypoint {
   }
 
   private handledStyles() {
-    let styleTags = [...this.dom.window.document.querySelectorAll('link[rel="stylesheet"]')] as HTMLLinkElement[];
+    let styleTags = [...this.dom.window.document.querySelectorAll('link[rel*="stylesheet"]')] as HTMLLinkElement[];
```
chancancode, line 69: This fixes a bug where previously `rel="prefetch stylesheet"` (or similar) in the compat HTML did not get picked up.

packages/core/src/html-placeholder.ts
```diff
@@ -55,32 +117,31 @@ export default class Placeholder {
...
-        // we always convert modules to scripts
-        continue;
-      }
-      // all other attributes are copied forward unchanged
-      newTag.setAttribute(name, value);
-    }
-    newTag.src = src;
+    let newTag = makeTag(this.end.ownerDocument, { from: this.target, attributes: { src } });
+    normalizeScriptTag(newTag);
+
```
chancancode, line 122: It seems like we do need both the changes in `ember-html.ts` as well as in this file. (@ef4 was pondering if we needed both in the other PR). I tried removing these changes, but the new tests I added started failing on the `<link>` tags, even though the `<script>` tags were fine. I did not investigate further.

chancancode:
What failing check ?_____?

PR #1608, opened 2023-09-21
cache realpathSync access in package-cache.get

raycohen:
With a large, interconnected app (~40 engines, ~100 addons), this gets hit a _lot_ with the same relatively small set of roots. Caching it saves as much as 6 minutes for us.

### DIFF
--- a/packages/shared-internals/src/package-cache.ts
+++ b/packages/shared-internals/src/package-cache.ts
@@ -4,6 +4,19 @@ import { getOrCreate } from './get-or-create';
 import resolvePackagePath from 'resolve-package-path';
 import { dirname, sep } from 'path';
 
+const realpathSyncCache = new Map<string, string>();
+
+function getCachedRealpath(path: string): string {
+  let cached = realpathSyncCache.get(path);
+  if (cached) {
+    return cached;
+  }
+
+  let root = realpathSync(path);
+  realpathSyncCache.set(path, root);
+  return root;
+}
+
 export default class PackageCache {
   constructor(public appRoot: string) {}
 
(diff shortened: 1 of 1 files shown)


ef4:
Thanks. 

This is probably good and we can land it. But I suspect that this spot being so hot may be the symptom of a different performance problem, and fixing that would yield an even bigger win.

I've had reports that suggest that the PackageCache is getting cache misses when we wouldn't expect it to.  If you'd like to check for that in your app, see how many times `Package` is constructed for the same `root`. We'd expect a small number (maybe two per OS-level process). The constructor is [here](https://github.com/embroider-build/embroider/blob/17cf74c90573a22648807786d20360a5715ae038/packages/shared-internals/src/package.ts#L13).

raycohen:
That `Package` constructor is getting hit 4201 times for us, but I don't see any cases where there are duplicates.

ef4:
I'm not quite sure what the CI failures are here, a quick attempt didn't reproduce them locally.

raycohen:
To provide more color on this particular slowness, when we hit these `realpathSync` calls we're inside the recursive `findActiveAddons` logic inside `partitionEngines`:
Screenshot 2023-09-22 at 1 26 17 PM

Is it possible that it is safe to cache some of this work? We have lots of engines that have lots of addon dependencies in common, so I think effectively there's a lot of redundant work happening.

ef4:
Can you please test https://github.com/embroider-build/embroider/pull/1609 in your app as well to see what the build performance impact is?

I am hoping that PR makes a significant improvement, and also it *might* mean we don't need the new cache that's in this PR.

raycohen:
@ef4 #1609 seems to make a massive improvement to how much time our build spends in webpack, but I'm not sure if it cuts the number of hits I'm seeing to realpath, it still looks to be in the millions.

PR #1609, opened 2023-09-24
Fix WrappedPackage caching for app

ef4:
The app's WrappedPackage was not getting cached correctly, causing it to repeat a lot of duplicate work.


mansona (approved):
Looks great 🎉

Issue #1610, opened 2023-09-25
[perf] Avoid fastboot regex cost in app not using fastboot?


raycohen:
When profiling our build (about 8 minutes), one of the things that jumped out at me immediately is a 28 seconds spent on a regex that appears to be fastboot specific. Our app does not use fastboot, and `get activeFastboot` is returning false. Can we avoid paying this cost?

<img width="849" alt="Screenshot 2023-09-25 at 9 58 19 AM" src="https://github.com/embroider-build/embroider/assets/20404/58840a9a-052e-4edc-aa07-a7ee42e10244">

The cost is close to entirely downstream from `beforeResolve > handleFastbootSwitch > decodeFastbootSwitch` with negligible time from `virtualLoader`.

SergeAstapov:
@raycohen would you like to create a file under /doc with steps you've used as other folks in community may find it helpful to analyze their builds?

raycohen:
For me its `node --cpu-prof  node_modules/.bin/ember serve` which creates a bunch of `.cpuprofile` files. I then import the main `.cpuprofile` file into  https://www.speedscope.app/

ef4:
Probably the easiest solution here is to change `decodeFastbootSwitch` to check `filename.includes('embroider_fastboot_switch')` before bothering with the regex exec. I would guess the string `includes` is cheap enough to be negligible. 

If that helps, we could do the same for `decodeVirtualPairComponent` and `decodeImplicitModules` too.

raycohen:
Thanks, I'll give it a try!

raycohen:
Yeah, that had a big impact. It took the time for that RegExp down to 3 seconds, and reduced the overall build time similarly.

```
function decodeFastbootSwitch(filename) {
  // I added these 3 lines:
  if (!filename.includes('embroider_fastboot_switch')) {
    return;
  }
```

raycohen:
Oops, that isn't the same regexp. In fact, the fastboot switch regexp no longer shows at all in my speedscope view. Looks like that completely eliminated the time here for me!

<img width="822" alt="Screenshot 2023-09-25 at 11 24 51 AM" src="https://github.com/embroider-build/embroider/assets/20404/ee133c49-b4fa-4bad-a090-694d0bb2fe18">

ef4:
Yup, that would be the one in `decodeImplicitModules`, which is one of the two others I mentioned above. We can improve that one the same way.

mansona:
[...]

PR #1611, opened 2023-09-25
performance: avoid paying decodeFastbootSwitch regex cost unless needed

raycohen:
Ref https://github.com/embroider-build/embroider/issues/1610

Before this change, my large app was spending 20+ seconds on the regex:

Screenshot 2023-09-25 at 9 58 19 AM

After this change, it spends no time here (my app does not use fastboot, so the condition is never met). I think this condition will be correct for users that do use fastboot.

### DIFF
--- a/packages/core/src/virtual-content.ts
+++ b/packages/core/src/virtual-content.ts
@@ -150,6 +150,10 @@ export function fastbootSwitch(specifier: string, fromFile: string, names: Set<s
 }
 
 export function decodeFastbootSwitch(filename: string) {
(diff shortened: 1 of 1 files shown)


ef4 (commented):
Thanks, this looks good. Can you do the same fix for `decodeVirtualPairComponent` and `decodeImplicitModules`? 

I don't want us to forget those ones as well.

raycohen:
Sure, I can add those to this PR

void-mAlex (approved):

PR #1612, opened 2023-09-26
performance: cache existsSync results in PackageCache.ownerOfFile

raycohen:
Similar to https://github.com/embroider-build/embroider/pull/1608

A lot of time is being spent in `PackageCache#ownerOfFile` and a good chunk of that is from the calls to `existsSync`:

Before:
Screenshot 2023-09-25 at 11 17 06 PM

By caching, the amount of time in `existsSync` drops considerably in my large app, From ~12 seconds to under 500ms. The total time spent in `ownerOfFile` drops less, from 20 to 14 seconds.

After:
Screenshot 2023-09-25 at 11 17 18 PM

Even with the caching, this path is still so hot that there's a good deal of time in `ownerOfFile` and the new `getCachedExistsSync` methods that isn't accounted for by `existsSync`. There's probably more that can be done here.

### DIFF
--- a/packages/shared-internals/src/package-cache.ts
+++ b/packages/shared-internals/src/package-cache.ts
@@ -17,6 +17,21 @@ function getCachedRealpath(path: string): string {
   return root;
 }
 
+const existsCache = new Map<string, boolean>();
+
+function getCachedExists(path: string): boolean {
+  if (existsCache.has(path)) {
+    const cachedExists = existsCache.get(path);
+    if (cachedExists !== undefined) {
+      return cachedExists;
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


raycohen:
Looks like CI is having issues downloading volta

NullVoxPopuli:
fix for that is here: https://github.com/embroider-build/embroider/pull/1594

ef4 (approved):
Thanks, this looks good to me.

A nice-to-have in this area for someone to followup would be to refactor the existsCache and the rootCache into one thing, because I think an improved design would do both jobs. But that's not a blocker here.

PR #1613, opened 2023-09-26
More v2 addon FAQs

simonihmig:
* integrates the content from https://github.com/embroider-build/addon-blueprint/blob/main/AUTHORING_LIBRARIES.md
* adds other FAQs (based on my EmberFest talk)
* adds a TOC and some better grouping

### DIFF
--- a/docs/v2-faq.md
+++ b/docs/v2-faq.md
@@ -6,6 +6,71 @@ The push-based `/styles` folder that v1 addons used to have is not available for
 
 Given that your addon's code is only pulled into the app when you import it, your CSS will also only be used when the module importing that is used itself. Therefore, we would recommend to import only the CSS that you need (say for a specific component) in the place where you need it. A common pattern is to colocate the CSS used for a component next to it, and import it from the component's JavaScript module. In case of a template-only component, you can create a JavaScript module for it that exports [templateOnly()](https://api.emberjs.com/ember/5.2/functions/@ember%2Fcomponent%2Ftemplate-only/templateOnly) and import the CSS from there or convert to a `<template>` tag component.
 
+## How can I ship other static assets with my addon?
+
+Similarily to the way we can ship CSS, we can do the same for other static assets like images by importing them. 
+Let's say an addon wants to ship an SVG file and refer to it in a component:
+
+```js
+import logo from '../assets/logo.svg';
+
+<template>
+  <img src={{logo}} alt='ACME' />
+</template>
+```
+
+Contrary to the CSS example, this is now not a side-effect only import anymore, but we actually get a value back as the default export of that imported asset: its public URL in the final build output. That's why we can pass this as the value of the `src` attribute in the example above.
+
+A few caveats though! 
+
+First, the ability to import static assets other than CSS is not enabled by default it Ember apps yet. So any Ember app that follows this pattern itself, or consumes a v2 addon doing this, will need to get this set up correctly first. To do so, the user would have to add a module rule to their webpack config, that configures [assets modules](https://webpack.js.org/guides/asset-modules/) for the given file extension(s):
+
+```js
+module: {
+  rules: [
+    {
+      test: /\.(svg)$/i,
+      type: 'asset/resource', // could also be just 'asset', if you want to have small assets be inlined
+    },
+  ],
+}
+```
+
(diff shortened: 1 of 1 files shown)


NullVoxPopuli (approved):

docs/v2-faq.md
```diff
@@ -29,10 +188,11 @@ For example, when your addon exposes a few components at the root level of `src/
 addon.publicEntrypoints('index.js', 'components/*.js'),
 ```
 
-Additionally, there is a feature supported in node.js and modern bundlers to define an `exports` key in your `package.json` with a mapping of export paths to the actual files on disk, that lets you further tweak or constrain your public exports. This is explained in more detail [here](https://nodejs.org/api/packages.html#exports).
+Additionally, there is a feature supported in node.js and modern bundlers to define an `exports` key in your `package.json` with a mapping of export paths to the actual files on disk, that lets you further tweak or constrain your public exports. This is explained in more detail here:
+- https://dev.to/binjospookie/exports-in-package-json-1fl2
```
void-mAlex, line 192: I don't feel right linking to `random` dev.to blog posts that can change on a whim their content
with the other two at least they are an org that gives us some assurance of the content being stable

docs/v2-faq.md
```diff
@@ -1,22 +1,181 @@
...
+### Why do v2 addons need a build step?
+
+The [v2 addon blueprint](https://github.com/embroider-build/addon-blueprint) uses [Rollup](https://rollupjs.org/) for assembling and transpiling your code in to a native npm package that can be imported from anywhere.
+
+While having a build step is not strictly required for v2 addons, there are a few reasons we would want it in general nevertheless:
+- use of pre-shipped JS features (decorators, and other in-progress ECMA proposals)
+- use of TypeScript
+- co-located components are not modules-by-default
+
+The important thing to remember here though is that this build step is very different from the build integration that v1 addons allow! While v1 addons integrate with the app's build and as such all add a bit of overhead to it, the build step of v2 addons happens ahead of time, before they are published to npm, so at the time they are consumed by the app, they are fully static and do _not_ integrate with the app's build.
```
void-mAlex, line 157: the wording about not integrating with the app build at the end is a bit confusing to me. I know what you mean and I don't immediately have a different way of phrasing it, but still wanted to raise this

docs/v2-faq.md
```diff
@@ -1,22 +1,181 @@
...
+
+### How can I integrate with the app's build?
+
+If you really need to add behaviour to the app's build that your addon needs to rely on, the way to go is to provide a plugin to the bundler used by the app (through `ember-auto-import` in a classic build or Embroider), which for now will most likely be [Webpack](https://webpack.js.org/).
```
void-mAlex, line 161: I think the third paragraph is better suited to replace the first one as we want to get across what people should do if they really need to do it.

docs/v2-faq.md
```diff
@@ -29,10 +188,11 @@ For example, when your addon exposes a few components at the root level of `src/
...
+- https://nodejs.org/api/packages.html#package-entry-points
+- https://webpack.js.org/guides/package-exports/
 
-## How can I write code depending on the context of the app or its dependencies?
-
-v2 addons are static packages, that do not integrate with the app's build, thus cannot know anything about the app, its context or its dependencies ahead of time.
-
-For the rare cases where you really need to know something about the app to be able to do the right thing, there is an escape hatch in the form of the `@embroider/macros` package, which is a set of macros that are able to transform your code at build-time of the app. Please refer to its [documentation](../packages/macros/README.md).
+When using `package.json#exports` make sure that:
+- the `addon.publicEntrypoints(...)` plugin in `rollup.config.mjs` includes _at least_ whatever is defined in `package.json#exports`
```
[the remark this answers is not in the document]
NullVoxPopuli: once type=module is supported in embroider, we can have a lint cover the enforcement for us

docs/v2-faq.md
simonihmig: I rephrased this a bit, hope makes better sense now.

docs/v2-faq.md
simonihmig: Removed this link. This was from the original [AUTHORING_LIBRARIES.md](https://github.com/embroider-build/addon-blueprint/blob/main/AUTHORING_LIBRARIES.md) file. However, I am not sure I agree with your reasoning, but I think the other links are sufficient and more authoritative...

docs/v2-faq.md
simonihmig: Note sure, I think the first paragraph is important to point out that the solution is to integrate with webpack (or whatever other bundlers we will have). If you don't mind, I would merge this as-is. Or you make a suggestion how the whole item should be phrased, that makes more sense...

PR #1614, opened 2023-09-27
esbuild resolver for Vite dependency optimization

lolmaus:
Taking over @ef4's branch `esbuild-resolver`. Rebased it against `main`.


lolmaus:
@mansona and me have found the reason why the build of `ember-vite-app` [build](https://github.com/mansona/ember-vite-app/pull/1) was failing after using @ef4's `EsbuildResolver`.

Instead of having `EsbuildResolver` resolve external dependencies, we have have it mark as `external: true` and do nothing, which lets Vite take over and do the right thing via Embroider.

Notes:
1. This may not be the most performance-efficient thing to do, but it seems to be *at least correct*, so we should consider merging this and iterate/improve on the implementation.
2. There may be more dead code in @ef4's [commit](https://github.com/embroider-build/embroider/pull/1614/commits/745e51a0bf8c19db86dfca3193a7ed2ae1aae122) after we added the early return.

NullVoxPopuli (approved):

NullVoxPopuli (approved):
I would feel better about this if we have some tests that used rollup build, but I understand there is a bit of a catch-22 here, and progress does need to be made.

PR #1607, opened 2023-09-20
Add v2 addon FAQs

simonihmig:
Not necessarily complete, just a first starting point...

PR #1617, opened 2023-10-01
Docs: refine `v2-faq.md` and fix a few typos

MrChocolatine:
As I was reading the changes in #1607 and #1613 (thanks @simonihmig), I took the opportunity to refine a few things on this document.

Initially I spotted a few typos, then decided to improve the overall readability of the document.

### DIFF
--- a/docs/v2-faq.md
+++ b/docs/v2-faq.md
@@ -79,9 +79,9 @@ that code, but to use a dynamic `import()`.
 Unlike v1 addons, v2 addons can dynamically import not only external packages but also their own
 code, like simple modules or even components. You need to make sure though, that these modules are
 not eagerly loaded elsewhere through static imports. This especially applies to "app re-exports",
-which basically make the consuming app (statically) import your addon code on your behalf. So when
+which basically makes the consuming app (statically) import your addon code on your behalf. So when
 using the common rollup setup with `addon.appReexports()`, make sure the components you want to be
-able to load lazilly are not covered by the glob patterns supplied to that rollup plugin!
+able to load lazily are not covered by the glob patterns supplied to that rollup plugin!
 
 
 ### How can I write code depending on the context of the app or its dependencies?
@@ -114,7 +114,7 @@ and import the CSS from there or convert to a `<template>` tag component.
 
 ### How can I ship other static assets with my addon?
 
-Similarily to the way we can ship CSS, we can do the same for other static assets like images by
+Similarly to the way we can ship CSS, we can do the same for other static assets like images by
 importing them. Let's say an addon wants to ship an SVG file and refer to it in a component:
 
 ```js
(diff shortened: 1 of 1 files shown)


simonihmig (commented):
Thanks for working on this!

I just wonder about the manual line breaks added here. Is that something we really still have to do nowadays? Like editors can wrap lines. And if you have explicit line breaks, editing the text can become very tedious, right? 

We have prettier explicitly disabled for .md files in this repo. So I think either we set up prettier to do this stuff for us, or we don't care!? Thoughts?

NullVoxPopuli (approved):

NullVoxPopuli:
> And if you have explicit line breaks, editing the text can become very tedious, right?
> We have prettier explicitly disabled for .md files in this repo. So I think either we set up prettier to do this stuff for us, or we don't care!? Thoughts?

ye, if we merge this I'd like to have this style of formatting be automatic.
it _is_ tedious to do manually.

NullVoxPopuli (changes requested):
can prettier do this for us?

docs/v2-faq.md
```diff
@@ -20,23 +20,34 @@
...
+  it is recommended in general to use relative imports only.
+* including the file extension in the import path so that rollup knows which set of plugins to run
+  over the file.
 
 ```ts
 import { Something } from './path/to/file.ts';
 ```
 
-If you've done ESM in node, this should feel familiar, and we can be be consistent with JS imports as well:
+If you've done ESM in node, this should feel familiar, and we can be be consistent with JS imports
```
lolmaus, line 49: Let's fix `be be`.

lolmaus (changes requested):
Me, @mansona and @void-mAlex agree that line breaks add little to no value and cause problems:

* bloated diffs (both for current and upcoming PRs),
* difficulty reading the source for dyslexic people,
* difficulty editing and reformatting.

We believe a paragraph should be a paragraph. You can tweak your IDE for a line wrap (for a specific extension) and preferred line width.

MrChocolatine:
> line breaks add little to no value and cause problems:
> - bloated diffs (both for current and upcoming PRs),
> - difficulty reading the source for dyslexic people,
> - difficulty editing and reformatting.

It actually facilitates reading.

MrChocolatine:
- https://insidegovuk.blog.gov.uk/2014/08/04/sentence-length-why-25-words-is-our-limit/
- https://www.gov.uk/guidance/content-design/writing-for-gov-uk#how-people-read

mansona:
Hey @MrChocolatine 👋 I just want to explain the comment that @lolmaus made a little bit more. We are currently in a meeting for the Embroider Triage and Andrey is "driving" so he expressed an opinion that I had with his comment but it is not exactly clear the reasons for that opinion so I would like to explain a bit better.

While you are correct that sentence length is important for facilitating reading, there is a difference between the display of line length and **forcing** line length by manually (or automatically) putting newlines in the source material. I'm speaking as someone with dyslexia that this sort of forced line break will make it significantly harder for me to read the Markdown source, so much so that I would be forced to read only a rendered output and not be able to easily read diffs in things like github.

I personally appreciate the changes that you were making in this PR so I would like to try to get it merged if you wouldn't mind re-opening it and removing the linebreak change?

Issue #1619, opened 2023-10-02
Caching issues (leading to Webpack crashing or component templates not being picked up)


chancancode:
### Setup

1. `ember new --embroider --yarn zomg` (yarn optional)
2. `JOBS=0 ember s` (`JOBS=0` optional, but may get better stack traces/easier to step through the code without the workers)

### Scenario 1

1. `touch app/components/foo.hbs`
2. observe it rebuilds successfully
3. observe that `.../rewritten-app/components/foo.hbs` exists (as expected)
4. observe that `.../rewritten-app/components/foo.js` exists (in addition to `foo.hbs`)
5. observe that `.../rewritten-app/assets/zomg.js` has an entry for `components/foo.js`
6. `rm app/components/foo.hbs`
7. observe that the build crashes
8. observe that `.../rewritten-app/components/foo.hbs` does not exists (as expected)
9. observe that `.../rewritten-app/components/foo.js` exists (unexpected)
10. observe that `.../rewritten-app/assets/zomg.js` has an entry for `components/foo.js` (unexpected)

The problem here is that the code the reflects the FS changes (deletion, in this case) into `rewritten-app`, in response to a sole `.hbs` file being deleted, should have also deleted the "virtual" js file (which it added in the first place), but did not.

While it is hard to "see" (not sure if there is a way to get the babel output written to disk), but because of the content-based caching (see the scenario below), the babel-transformed output of the leftover `foo.js` has a reference to `foo.hbs`, leading to the crash.

### Scenario 2

(This scenario is to illustrate an independent issue to the previous one, so make sure you git reset the changes, restart the server, etc, to being from a fresh state)

1. create `app/components/foo.js` with the following content:
    ```js
    import Component from "@ember/component";
    export default Component.extend({ tagName: "span" });
    ```
2. observe it rebuilds successfully
3. observe that the expected files and entries are present in `rewritten-app`
4. observe that if you invoke the component (from `application.hbs`) it renders an empty span in the browser
5. create `app/components/foo.hbs` with any non-empty content (e.g. `hello world!`)
6. observe it rebuilds successfully
7. observe that the expected files and entries are present in `rewritten-app`
8. observe that if you invoke the component it renders an empty span without the template's content (unexpected)
9. edit the component js file to have `tagName: "p"`
10. observe it rebuilds successfully
11. observe that if you invoke the component it does render an p tag with the template's content (as expected)

[...]

PR #1620, opened 2023-10-03
Fix synthesized template-only components crash

chancancode:
This fixes ~~the first scenario detailed in~~ #1619. See the individual commit messages for details.

### DIFF
--- a/packages/compat/src/synthesize-template-only-components.ts
+++ b/packages/compat/src/synthesize-template-only-components.ts
@@ -1,16 +1,27 @@
 import Plugin from 'broccoli-plugin';
 import type { Node } from 'broccoli-node-api';
 import { join, basename } from 'path';
-import walkSync from 'walk-sync';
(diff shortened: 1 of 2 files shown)


chancancode:
Not sure if we currently have infrastructure for testing rebuilds?

chancancode:
This only fixed half of the issues (which is fine, this can still be reviewed/merged/released as-is). I had incorrectly assumed that the second problem in #1619, while incorrect, does not result in crashes. I was wrong and I have added a third scenario (which may even be the dominate kind of crashes) in the issue.

chancancode:
That would be great!

mansona:
@chancancode we finally ironed out the issues with https://github.com/embroider-build/embroider/pull/1624 (I'm done with windows 😭) so you can create the watch mode tests now 👍 

Let me know if you have any questions about the setup 👍

chancancode:
Oops I accidentally pushed everything to #1632 so I’ll just close this

PR #1622, opened 2023-10-04
use realpath of engine's route when building resolver.json

mansona:
This solves a problem that we face when we try to run embroider in an app where `node_modules` has been symlinked. The issue arises because all webpack requests normalise the path of a package using fs.realpath so when we're trying to find the [owningEngine](https://github.com/embroider-build/embroider/blob/main/packages/core/src/module-resolver.ts#L707C7-L707C7) of a package we will always get the wrong answer because package passed into that request will have the real path on disk (after removing symbolic link information)

I have tested this locally and it also seems to unblock https://github.com/ember-cli/ember-cli/pull/10370 because the ember-cli CI also does some shenanigans with linking node_modules


mansona:
I've tested this against https://github.com/mansona/ember-inline-svg-bug-recreation and I've verified that it does indeed fix the issue that we're facing

PR #1623, opened 2023-10-04
Implement the optimizeDeps() helper

### LINKED ISSUE #1621: maybe have optimizeDeps come from embroider i.e. don't add @embroider/macros but we add it in our function

lolmaus:
...that can be used in `vite.config.js`.

It would prevent users from accidentally overriding the important keys (for now it's just `exclude`).

Comes with the very basic Jest setup, copied from a sibling folder.

Fixes #1621.

### DIFF
--- a/packages/vite/index.d.ts
+++ b/packages/vite/index.d.ts
@@ -3,3 +3,4 @@ export * from './src/hbs.js';
 export * from './src/scripts.js';
 export * from './src/template-tag.js';
 export * from './src/addons.js';
+export * from './src/optimize-deps.js';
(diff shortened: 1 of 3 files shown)


mansona (changes requested):
I think you need to add a re-export for this so that it can be consumed from a test app like this: https://github.com/embroider-build/embroider/blob/main/packages/vite/index.mjs

On that note I think you should use this function in the actual test app and make sure that it works as expected https://github.com/embroider-build/embroider/blob/main/tests/vite-app/vite.config.mts#L26-L28

lolmaus:
@mansona Done.

mansona (approved):
I've manually tested the vite app and I'm happy that it's working 👍

PR #1624, opened 2023-10-04
Add app-level watch-mode tests

NullVoxPopuli:
Adds our first app-level watch-mode tests so we can start testing against some more involved scenarios that @chancancode has been fixing :tada:

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1566,7 +1566,7 @@ importers:
         version: 2.19.2
       ember-auto-import:
         specifier: ^2.6.3
-        version: 2.6.3(webpack@5.88.2)
+        version: 2.6.3
       fastboot:
         specifier: ^4.1.1
         version: 4.1.1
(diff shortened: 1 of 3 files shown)


mansona (approved):
Awesome work 🎉

pnpm-lock.yaml
NullVoxPopuli, line 1: oofta. I cannot win today

tests/scenarios/watch-mode-test.ts
```diff
@@ -0,0 +1,76 @@
...
+      });
+    });
+
+    test(`pnpm ember test`, async function (assert) {
+      await waitFor('Serving on');
+      const content = 'TWO IS A GREAT NUMBER< I LKE IT A LOT< IT IS THE POWER OF ALL  OF ELECTRONICS, MATH, ETC';
+
+      assert.false(await checkScripts(/js$/, content), 'file has not been created yet');
+
+      fs.writeFile(path.join(app.dir, 'app/simple-file.js'), `export const two = "${content}";`);
```
lifeart, line 70: should we here `await` for file creation?

mansona: I guess yes technically 😂 but we have a thing just after that that is explicitly looking for the "app rebuild" notification which only happens after the file is written so it's safe to not await here 👍

PR #1627, opened 2023-10-05
Resolver refactor

mansona:
This PR started as an effort myself and @ef4 were pairing on to see if we can figure out how to teach vite and esbuild about our dependencies a bit better

It ended up being a nice refactor of the resolver so that we could remove some codepaths and simplify what the resolver was doing. It will also (hopefully) help esbuild understand what is going on a bit better, but I have split this PR so that it only contains the resolver refactor and the Vite/esbuild stuff will follow on in a new PR 👍

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,8 +21,7 @@
   },
   "jest": {
     "projects": [
-      "<rootDir>/packages/*",
-      "<rootDir>/test-packages/"
+      "<rootDir>/packages/*"
     ],
     "testEnvironment": "node",
     "testMatch": [
--- a/packages/compat/src/standalone-addon-build.ts
+++ b/packages/compat/src/standalone-addon-build.ts
@@ -44,6 +44,10 @@ ${summarizePeerDepViolations(violations)}`
     return new Funnel(interior, { destDir: index.packages[pkg.root] });
   });
 
+  let fakeTargets = Object.values(index.packages).map(dir => {
+    return writeFile(join(dir, '..', 'moved-package-target.js'), '');
+  });
+
   return broccoliMergeTrees([
     ...exteriorTrees,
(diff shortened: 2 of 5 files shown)


ef4:
In https://github.com/embroider-build/embroider/pull/1627/commits/a31e2d6b7aff59a79828f933f848aa5099701efa we realized that the changes this PR makes to `resolveWithinPackage` are good when resolving between packages, but unhelpful when we're just adjusting self-imports inside a package. Those are not places where we want to make our imports look like inter-package dependencies to Vite, and by moving the requests out of their owning package we just complicate the rest of the module-resolver checks.

So in that commit we changed it so that self-imports continue to use relative paths as they do on main, and we only use the new superFakeTarget.js for inter-package imports.

With this change, it's possible that we no longer need this PR to move the app from `node_modules/.embroider/rewritten-app` to `node_modules/.embroider/rewritten-app/node_modules/my-app-name`, because I think it's really only the self-import case that applies to the app. We should investigate whether we can roll that part of the PR back and still see things working.

packages/compat/src/compat-addons.ts
```diff
@@ -35,7 +35,12 @@ export default class CompatAddons implements Stage {
 
   async ready(): Promise<{ outputPath: string }> {
     return {
-      outputPath: resolve(locateEmbroiderWorkingDir(this.compatApp.root), 'rewritten-app'),
+      outputPath: resolve(
+        locateEmbroiderWorkingDir(this.compatApp.root),
+        'rewritten-app',
+        'node_modules',
```
ef4, line 41: This should get backed out of the PR. Since you didn't need the fake target for the app, we also don't need to put the app inside this structure.

This will mean a bunch of test changes in the PR also won't be needed.

ef4 (approved):

PR #1629, opened 2023-10-07
Randomize port and improve watch mode test

chancancode:


tests/scenarios/watch-mode-test.ts
```diff
@@ -15,20 +15,235 @@ let app = appScenarios.map('watch-mode', () => {
...
+  private matchLine(line: string): boolean {
+    if (typeof this.output === 'string') {
+      return this.output === line;
+    } else {
+      return this.output.test(line);
+    }
+  }
+}
+
+type Status = { type: 'starting' } | { type: 'ready' } | { type: 'errored'; error: unknown } | { type: 'completed' };
```
NullVoxPopuli, line 107: It kinda feels like this whole util belongs in ember-cli and should have its own tests.
But that would mean ember-cli would need to formally support a js api, which could probably be implemented without stream peaking
:upside_down_face: idk.

Looks solid tho!

NullVoxPopuli (approved):

tests/scenarios/watch-mode-test.ts
chancancode: I called it EmberCLI but it really is just a stateful child process thing with facilities for testing outputs, which I am sure could be useful for testing the vite integrations and what not as well

mansona (approved):
This looks great 🎉 I think the helper classes have gotten a bit complicated now that we could benefit from them being moved into a separate file so that the test are easier to follow, but that's not a blocker for merging this 👍

PR #1632, opened 2023-10-08
Fix co-located components regressions (#1619)

chancancode:
~~These tests are currently written to exhibit the buggy behavior. When I fix the issue I'll adjust the tests to the correct/expected behavior. These tests should pass right now (since the expectations in the tests are for the buggy behavior), so this can be merged even before the fix is ready.~~ see the individual commits and #1619 for details

### DIFF
--- a/tests/scenarios/watch-mode-test.ts
+++ b/tests/scenarios/watch-mode-test.ts
@@ -311,6 +311,45 @@ class AssertFile {
   }
 }
 
+function d(strings: TemplateStringsArray, ...values: unknown[]): string {
+  let buf = '';
+  for (let string of strings) {
+    if (values.length) {
+      buf += string + values.shift();
+    } else {
+      buf += string;
+    }
+  }
+  return deindent(buf);
+}
+
+function deindent(s: string): string {
+  if (s.startsWith('\n')) {
+    s = s.slice(1);
+  }
+
+  let indentSize = s.search(/\S/);
+
+  if (indentSize > 0) {
+    let indent = s.slice(0, indentSize);
+
+    s = s
+      .split('\n')
+      .map(line => {
+        if (line.startsWith(indent)) {
+          return line.slice(indentSize);
+        } else {
+          return line;
+        }
+      })
+      .join('\n');
+  }
+
+  s = s.trimEnd();
+
+  return s;
+}
+
 app.forEachScenario(scenario => {
   Qmodule(scenario.name, function (hooks) {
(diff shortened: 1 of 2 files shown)


tests/scenarios/watch-mode-test.ts
```diff
@@ -239,13 +226,115 @@ class EmberCLI {
   }
 }
 
+class File {
+  constructor(readonly label: string, readonly fullPath: string) {}
+
+  async exists(): Promise<boolean> {
```
NullVoxPopuli, line 232: I think we have fs-extra available, idk if that'd help clean this up with pathExists?

chancancode: A tiny bit, but probably not enough to watch CI run again 🤷🏼‍♂️ if it fails for other reasons I'll refactor it, otherwise I'll do it when I fix the actual issue

NullVoxPopuli (approved):
thanks for adding so many tests!

tests/scenarios/watch-mode-test.ts
```diff
@@ -263,37 +391,224 @@ app.forEachScenario(scenario => {
...
+      await appFile('app/simple-file.js').write(`export const two = "${updatedContent}";`);
+      await changed('simple-file.js');
       await waitFor(/Build successful/);
 
       // TODO: find a better way to test this; this seems to linger around
       // assert.false(await checkScripts(/js$/, originalContent), 'the original file does not exists');
       assert.true(await checkScripts(/js$/, updatedContent), 'the updated file now exists');
     });
+
+    Qmodule('[GH#1619] co-located components regressions', function (hooks) {
```
NullVoxPopuli, line 427: I really like this style of test naming - it gives future viewers additional context as to why a test was added

tests/scenarios/watch-mode-test.ts
```diff
@@ -239,13 +227,153 @@ class EmberCLI {
...
+
+    if (actual === null) {
+      this.assert.ok(false, `${this.file.label} does not exists`);
+    } else {
+      this.assert.false(actual.includes(expected), `content of ${this.file.label}`);
+    }
+  }
+}
+
+function d(strings: TemplateStringsArray, ...values: unknown[]): string {
```
NullVoxPopuli, line 314: common-tags does all this, yeah?
https://www.npmjs.com/package/common-tags
`stripIndent`?

NullVoxPopuli (approved):

PR #1633, opened 2023-10-10
Perf: cache macros lockfile lookup and read operations

raycohen:
This shaves about 2 seconds off of the total time spent in `babelPluginConfig` in my large app. 

Follow-up ideas:
 - There's also about 400ms from the use of `resolve` in `babelPluginConfig` which seems like it could be done once
 - It looks safe to cache the result of `hash.digest` which is currently taking 4s in my build

Before:

Screenshot 2023-10-09 at 8 51 38 PM

After:

Screenshot 2023-10-09 at 9 06 09 PM

### DIFF
--- a/packages/macros/src/macros-config.ts
+++ b/packages/macros/src/macros-config.ts
@@ -303,6 +303,40 @@ export default class MacrosConfig {
     };
   }
 
+  private static lockFilePathForAppRoot: Map<string, string | undefined> = new Map();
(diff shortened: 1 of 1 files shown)


raycohen:
Closing as caching the read operation is potentially unsafe - the lockfile can change between rebuilds and currently this will result in a new `cacheKey` being used. I will open a new PR that caches the lockfile and package.json paths

PR #1636, opened 2023-10-10
Increase CI timeout

chancancode:
The windows watch mode tests takes just about 20 minutes and often get killed when they are just about to finish


mansona:
I'm not exactly sure we're seeing a real case of this taking > 10 mins 🤔 I suspect the thing you're going to see is that it's going to `get killed when [its] just about to finish` but at 20 mins. I don't know why we would be having this issue only with release but I suspect that something isn't cleaning up a stream and that's keeping the subprocess open

mansona (approved):
This change is obviously needed for our CI but I'm very uncomfortable merging it 🙈

PR #1641, opened 2023-10-13
Allow to pass additional config to addonV1Shim

krasnoukhov:
We're in the process of converting our internal addon to v2 and one thing that we're missing is a way to specify additional options for the ember-cli addon. This PR introduces a way to forward any custom stuff to the ember-cli addon config. In our case this would look like this:

```js
const bodyParser = require('body-parser');
const path = require('path');
const { addonV1Shim } = require('@embroider/addon-shim');

module.exports = addonV1Shim(__dirname, {
  additionalConfig: {
    options: {
      svgJar: {
        sourceDirs: [path.join(__dirname, 'src/assets/icons')],
      },
    },

    testemMiddleware(app) {
      app.post('/write-junit', bodyParser.text({ limit: '100mb' }), function (req, res) {
        // ...implementation omitted...
        res.end();
      });
    },
  },
});
```

One thing is `options` for the addons we're depend on - in this case it's telling `svgJar` to look for icons inside `src` so we can do both things: use icons in the addon and package them up to be used in the apps directly.

Another thing is `testemMiddleware` - there does not seem to be another way to have this middleware, because configuring it via `testem.js` in the apps does not work at all (ember-cli overrides middleware with delegation to the addons).

Not sure if there needs to be a test for this here, any guidance is appreciated. Cheers

### DIFF
--- a/packages/addon-shim/src/index.ts
+++ b/packages/addon-shim/src/index.ts
@@ -12,6 +12,7 @@ import { satisfies } from 'semver';
 
 export interface ShimOptions {
   disabled?: (options: any) => boolean;
+  additionalConfig?: object;
 }
 
 function addonMeta(pkgJSON: PackageInfo): AddonMeta {
@@ -167,6 +168,7 @@ export function addonV1Shim(directory: string, options: ShimOptions = {}) {
         (this.parent as EAI2Instance).registerV2Addon(name, root);
       }
     },
+    ...(options.additionalConfig || {}),
   };
 }
 


krasnoukhov:
Ok thanks for the explanation @mansona. Would you mind looking into the `testemMiddleware` case? I don't think it's possible to have that configured in the app unfortunately. I guess for us it would mean we'd create another v1 addon just for this stuff.

krasnoukhov:
@mansona Good idea, but I tried this already, it seems like ember-cli overwrites `middleware` with its own thing [likely here](https://github.com/ember-cli/ember-cli/blob/e57b6bd7959f45be0b0a2a0ad55e60f018a1f48b/lib/tasks/test.js#L59), which makes it impossible currently to use from within the app...

mansona:
This is something that we should fix, you need to have some way to register these things 🤔 for now having a v1 addon that does this should be fine. You can always make it an in-repo addon since that is a way to have something in the app interact with the addon apis 👍 

I'll add this PR to the agenda for next week's core meeting to discuss potential solutions

krasnoukhov:
@mansona Sounds good, thanks. I think maybe the right thing would be to make a patch to ember-cli to respect original app's `options.middleware`? Do you think there's going to be any problem importing addon code into testem.js to have addon provide a middleware function?

mansona:
so that's exactly what I wanted to discuss at the core meeting, it's not just an Embroider meeting any more it covers all of the Ember Tooling so chatting about fixing ember-cli is within our remit 👍 

Personally I think it's "on brand" with what we're trying to do, getting people to import things and apply them to config files themselves. As for `any problem importing addon code into testem.js` it's not really going to be a problem because we're truly just talking about standard Javascript here 🤷 It doesn't need to be an "addon" per say, it could just be a package. But if you didn't want a new package you can still have it importable from an addon 👍

PR #1642, opened 2023-10-13
Allow for more flexible addon-dev appReexports

krasnoukhov:
While `appReexports` is pretty great at what it's doing it's not very flexible when you don't just `export { default }`. For example, if you have an initializer in v2 addon and you want to make it available to app without any additional config. But initializers require `initialize` function to be exported too. In our case some modules need to have `export *` as well.

This change creates a way to have support for such scenarios:
- `addon.appReexports` can be called multiple times in the rollup config (we reset `app-js` initially when addon instance is created so there's no leftover stuff in there)
- Optionally, for a specific glob, `defaultExport` can be also passed in which can be set to `*` or `{ default, initialize }`

Tests were adjusted to account for all this. Let me know how this looks, any guidance will be appreciated. Cheers

### DIFF
--- a/packages/addon-dev/src/rollup-app-reexports.ts
+++ b/packages/addon-dev/src/rollup-app-reexports.ts
@@ -8,6 +8,7 @@ export default function appReexports(opts: {
   to: string;
   include: string[];
   mapFilename?: (filename: string) => string;
+  defaultExport?: string;
 }): Plugin {
   return {
     name: 'app-reexports',
@@ -25,21 +26,21 @@ export default function appReexports(opts: {
           this.emitFile({
             type: 'asset',
             fileName: `_app_/${appFilename}`,
-            source: `export { default } from "${pkg.name}/${addonFilename.slice(
-              0,
-              -extname(addonFilename).length
-            )}";\n`,
+            source: `export ${opts.defaultExport} from "${
+              pkg.name
+            }/${addonFilename.slice(0, -extname(addonFilename).length)}";\n`,
           });
         }
       }
       let originalAppJS = pkg['ember-addon']?.['app-js'];
-
-      let hasChanges = JSON.stringify(originalAppJS) !== JSON.stringify(appJS);
+      let newAppJs = Object.assign({}, originalAppJS || {}, appJS);
+      let hasChanges =
+        JSON.stringify(originalAppJS) !== JSON.stringify(newAppJs);
 
       // Don't cause a file i/o event unless something actually changed
       if (hasChanges) {
         pkg['ember-addon'] = Object.assign({}, pkg['ember-addon'], {
-          'app-js': appJS,
+          'app-js': newAppJs,
         });
         writeJsonSync('package.json', pkg, { spaces: 2 });
       }
--- a/packages/addon-dev/src/rollup-app-reexports.ts
+++ b/packages/addon-dev/src/rollup-app-reexports.ts
@@ -8,7 +8,7 @@ export default function appReexports(opts: {
   to: string;
   include: string[];
   mapFilename?: (filename: string) => string;
-  defaultExport?: string;
+  exports?: (filename: string) => string[] | string | undefined;
 }): Plugin {
   return {
     name: 'app-reexports',
@@ -17,6 +17,12 @@ export default function appReexports(opts: {
       let appJS: Record<string, string> = {};
       for (let addonFilename of Object.keys(bundle)) {
         let appFilename = opts.mapFilename?.(addonFilename) ?? addonFilename;
+        let appExports = opts.exports?.(addonFilename) || ['default'];
+
+        let computedExports =
+          typeof appExports === 'string'
+            ? appExports
+            : `{ ${appExports.join(', ')} }`;
 
         if (
           opts.include.some((glob) => minimatch(addonFilename, glob)) &&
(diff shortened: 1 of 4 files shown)


tests/scenarios/v2-addon-dev-typescript-test.ts
```diff
@@ -100,6 +104,9 @@ appScenarios
           ], {
             mapFilename: (name) => reexportMappings[name] || name,
           }),
+          addon.appReexports(['initializers/**/*.js'], {
```
NullVoxPopuli, line 107: ya know, I never thought of just using `appReexports` twice. 🤔

is there a risk that the other copy of this plugin would overwrite / erase the work of the other plugin?

tests/scenarios/v2-addon-dev-typescript-test.ts
```diff
@@ -167,6 +174,18 @@ appScenarios
             `,
           },
         },
+        initializers: {
+          'demo.js': `
+            export function initialize() {
```
NullVoxPopuli, line 179: it's been a while since I've used initializers, but why is this export needed? (since it's included in the default export below)

tests/scenarios/v2-addon-dev-typescript-test.ts
krasnoukhov: I don't think so, given that we merge `app-js` with original. It also does `emitFile`, which is fine to call many times. In case you have duplicate files processing with different `appReexports` calls the rollup will give you a warning

tests/scenarios/v2-addon-dev-typescript-test.ts
krasnoukhov: I'm not sure where exactly ember does it but I'm sure it does. If I don't specify this I'm getting runtime errors like this one:

image

Do you want me to dig for it to see where/why/how it's done?

krasnoukhov: Maybe it's related to a fact that we're still on ember 3.x for some of the apps...

packages/addon-dev/src/rollup-app-reexports.ts
```diff
@@ -8,6 +8,7 @@ export default function appReexports(opts: {
   to: string;
   include: string[];
   mapFilename?: (filename: string) => string;
+  defaultExport?: string;
```
ef4, line 11: I think we need a better name for this. It's not only about default exports, it's about exports in general.

packages/addon-dev/src/rollup-app-reexports.ts
```diff
@@ -25,21 +26,21 @@ export default function appReexports(opts: {
           this.emitFile({
             type: 'asset',
             fileName: `_app_/${appFilename}`,
-            source: `export { default } from "${pkg.name}/${addonFilename.slice(
-              0,
-              -extname(addonFilename).length
-            )}";\n`,
+            source: `export ${opts.defaultExport} from "${
```
ef4, line 29: This is going to lead to very confusing failures when somebody puts the wrong thing in the opts.

I don't think we should ask the user to supply a fully-valid snippet of javascript that happens to work in this context, I think they should give us a `string[]` of exports and we should emit the correct javascript.

packages/addon-dev/src/rollup.ts
```diff
@@ -16,20 +17,28 @@ export class Addon {
...
+      mapFilename?: (fileName: string) => string;
+      defaultExport?: string;
+    } = {}
   ): Plugin {
     return appReexports({
       from: this.#srcDir,
       to: this.#destDir,
       include: patterns,
       mapFilename: opts.mapFilename,
+      defaultExport: opts.defaultExport || '{ default }',
```
ef4, line 41: The default value of this option should be in rollup-app-reexports.ts itself, not this callsite.

packages/addon-dev/src/rollup.ts
```diff
@@ -16,20 +17,28 @@ export class Addon {
   constructor(params: { srcDir?: string; destDir?: string } = {}) {
     this.#srcDir = params.srcDir ?? 'src';
     this.#destDir = params.destDir ?? 'dist';
+
+    let pkg = readJsonSync('package.json');
+    delete pkg['ember-addon']['app-js'];
```
ef4, line 22: This scheme of only clearing `app-js` at startup and then forever merging it forward is not going to work correctly.

It means that deleted files will continue to exists in `app-js`, for example, which is a hard build error.

packages/addon-dev/src/rollup-app-reexports.ts
```diff
@@ -25,21 +26,21 @@ export default function appReexports(opts: {
...
+              pkg.name
+            }/${addonFilename.slice(0, -extname(addonFilename).length)}";\n`,
           });
         }
       }
       let originalAppJS = pkg['ember-addon']?.['app-js'];
-
-      let hasChanges = JSON.stringify(originalAppJS) !== JSON.stringify(appJS);
+      let newAppJs = Object.assign({}, originalAppJS || {}, appJS);
```
ef4, line 36: This never removes deleted things from app-js.

Instead of trying to make it possible to layer multiple copies of this plugin together, I think it would be better to offer an API so that it can generate a correct app-js in one pass.

ef4 (changes requested):
Thanks, I hadn't noticed that initializers need this, but I see you are correct.

(It's arguably a mistake that initializers ever worked this way, given how everything else in the traditional system works, but that ship has sailed and this is a feature for achieving compatibility, so I agree we should support it.)

My suggestion for how to address the concerns I've pointed out here is an API like:

```diff
function appReexports(opts: {
  from: string;
  to: string;
  include: string[];
  mapFilename?: (filename: string) => string;
+ exports?: (filename: string) => string[] | undefined;
}) {}
```

Returning `undefined` from `exports` would be treated the same as returning `['default']`.

A usage for initializers would look like:

```ts
addon.appReexports(['**/*.js'], {
  exports(filename) {
    if (/\/initializers\//.test(filename)) {
       return ['default', 'initialize'];
    }
  }
})
```

krasnoukhov:
Thanks @ef4, good feedback, everything is addressed. One note: I changed function signature to `string[] | string | undefined` in order to also allow `export *` syntax. This is also covered by the test. I hope this is not a problem, but let me know. Cheers

krasnoukhov:
@ef4 Any other feedback here?

NullVoxPopuli (approved):

ef4 (approved):

PR #1643, opened 2023-10-15
cache macros lockfile and package.json lookup

raycohen:
The `findUp.sync` operation gets called repeatedly during the ember-cli addon initialization phase. Each time it receives the same `appRoot` to search in. For projects with no lockfile we fall back to using package.json.

This becomes expensive in apps with substantial numbers of addons. Caching it avoids paying the cost repeatedly.

### DIFF
--- a/packages/macros/src/macros-config.ts
+++ b/packages/macros/src/macros-config.ts
@@ -303,6 +303,26 @@ export default class MacrosConfig {
     };
   }
 
+  private static lockFilePathForAppRoot: Map<string, string | undefined> = new Map();
+  private static getLockFilePath(appRoot: string): string | undefined {
+    if (this.lockFilePathForAppRoot.has(appRoot)) {
+      return this.lockFilePathForAppRoot.get(appRoot);
+    }
+    let path = findUp.sync(['yarn.lock', 'package-lock.json', 'pnpm-lock.yaml'], { cwd: appRoot });
+    this.lockFilePathForAppRoot.set(appRoot, path);
+    return path;
+  }
+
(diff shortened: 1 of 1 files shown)


ef4:
Two things:

1. I think we can cache at the level of the whole `babelPluginConfig()` instead. Weakmapped off the `appOrAddonInstance`. 

2. All of this work only does anything classic builds. In embroider, it literally doesn't matter, because embroider natively supports `@embroider/macros` and actively detects and removes any custom per-package installations of this plugin. Unfortunately given the timing in ember-cli, there's no obvious way to know whether an EmberApp instance will be used by embroider or not at the time when `included` is running. Open to suggestions. One idea would be to make the return value from babelPluginConfig even lazier so it does no work until it's actually consumed.

raycohen:
One thing I noticed while debugging was if you change the contents of your lockfile while the server is running (say by running yarn install), the next rebuild's computed `cacheKey` would be different. Am I understanding (1) correctly that that doesn't matter?

Another thing I saw was that one of the places that calls `babelPluginConfig` doesn't pass a param. 

https://github.com/embroider-build/embroider/blob/65d906734b3c70cf4aee8ad71bdd82b83f93cf04/packages/macros/src/ember-addon-main.ts#L80

vs 

https://github.com/embroider-build/embroider/blob/65d906734b3c70cf4aee8ad71bdd82b83f93cf04/packages/compat/src/compat-app-builder.ts#L440

Is that intentional? It seemed odd that we get to
https://github.com/embroider-build/embroider/blob/65d906734b3c70cf4aee8ad71bdd82b83f93cf04/packages/macros/src/macros-config.ts#L377
and have `@embroider/macros cache buster: null` in that case.

PR #1645, opened 2023-10-15
AntiPerf: Memoize cache key generation per (project,lockFile) tuple

raycohen:
Trades hash generation work for toString + map lookup work. This is actually worse performance. What can we do instead?

Note that gatherAddonCacheKey is already only [performed once](https://github.com/embroider-build/embroider/blob/dfe4bc37a0947f217c06ab2257b6a6b93449ce86/packages/macros/src/macros-config.ts#L54-L58) per `project`.

This fails to improve the speed because the cost of using the lockfile's contents as a lookup key in a map is too high.

Before: About 700ms of `hash.update`
Screenshot 2023-10-14 at 10 49 16 PM

After: About 600ms of map lookup and 400ms of `buffer.toString`
Screenshot 2023-10-14 at 10 45 07 PM

### DIFF
--- a/packages/macros/src/macros-config.ts
+++ b/packages/macros/src/macros-config.ts
@@ -350,23 +350,29 @@ export default class MacrosConfig {
 
     let lockFileBuffer = lockFilePath ? fs.readFileSync(lockFilePath) : 'no-cache-key';
 
-    // @embroider/macros provides a macro called dependencySatisfies which checks if a given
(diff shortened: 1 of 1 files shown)


raycohen:
see https://github.com/embroider-build/embroider/pull/1643#issuecomment-1765093914 for discussion of how we can cache more here

PR #1646, opened 2023-10-16
docs: fix typo of publicAssetURL

camerondubas:
Just noticed this example config didn't match the actual property name

### DIFF
--- a/packages/webpack/src/options.ts
+++ b/packages/webpack/src/options.ts
@@ -41,8 +41,8 @@ export interface Options {
   //     filename: `mychunk.[chunkhash].js`,
   //     chunkFilename: `mychunk.[chunkhash].js`,
   //   },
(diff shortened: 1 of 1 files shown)


lifeart (approved):

ef4:
I think this is still incorrect because it's showing `publicAssetURL` inside `webpackConfig`. `publicAssetURL` is an option to `@embroider/webpack` not to `webpack`.

camerondubas:
> I think this is still incorrect because it's showing `publicAssetURL` inside `webpackConfig`. `publicAssetURL` is an option to `@embroider/webpack` not to `webpack`.

ah, you're right! I updated the commit

void-mAlex:
I wonder if this was not refering to https://webpack.js.org/configuration/output/#outputpublicpath instead which is very similar in function

PR #1648, opened 2023-10-19
use package paths instead of relative paths for app tree resolving

mansona:
This is blocked on https://github.com/embroider-build/embroider/issues/1651 because we can't really release code that depends on a shakey implementation of reverse-exports. We considered feature-flagging the change but it felt like a lot of work with very little gain 🤔 

~Edit: I made this PR draft since it's blocked on needing `reverse-exports` 👍~

Edit: reverse-exports was added in https://github.com/embroider-build/embroider/pull/1652 so this PR is good to go now

### DIFF
--- a/packages/compat/src/compat-app-builder.ts
+++ b/packages/compat/src/compat-app-builder.ts
@@ -289,9 +289,10 @@ export class CompatAppBuilder {
         root: realpathSync(index === 0 ? this.root : appFiles.engine.package.root),
         fastbootFiles: appFiles.fastbootFiles,
         activeAddons: [...appFiles.engine.addons]
-          .map(a => ({
-            name: a.name,
-            root: a.root,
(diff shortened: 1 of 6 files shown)


packages/reverse-exports/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "reverse-exports",
+  "version": "0.0.0",
+  "description": "",
+  "main": "src/index.js",
+  "scripts": {
+    "test": "jest"
```
void-mAlex, line 7: by what magic does this (`jest`) not need to be in dev deps?

mansona: it's in the monorepo deps, it's the same way that the other packages that use jest do things

NullVoxPopuli (approved):

PR #1650, opened 2023-10-26
[WIP] Esbuild resolver

mansona:
This PR is covering all of the experimentation currently going on to get Vite working.

This depends on https://github.com/embroider-build/embroider/pull/1648


packages/vite/src/esbuild-resolver.ts
```diff
@@ -0,0 +1,132 @@
...
+        let src = virtualContent(path, resolverLoader.resolver);
+        if (!macrosConfig) {
+          macrosConfig = readJSONSync(
+            resolve(locateEmbroiderWorkingDir(root), 'rewritten-app', 'macros-config.json')
+          ) as PluginItem;
+        }
+        return { contents: runMacros(src, path, macrosConfig) };
+      });
+
+      build.onLoad({ filter: /\.gjs$/ }, async ({ path: filename }) => {
```
NullVoxPopuli, line 51: should this also handle gts?

or is that automatic somehow?

mansona: For now I would like to not add typescript to this setup because we might have other considerations that we would need to cover. I would prefer if we added specific support for it in its own PR if that's ok 👍

NullVoxPopuli (approved):

void-mAlex (approved):

PR #1652, opened 2023-10-30
create new @embroider/reverse-exports package

lolmaus:
Some references:
* https://nodejs.org/api/packages.html#package-entry-points
* https://webpack.js.org/guides/package-exports/

### DIFF
--- a/packages/reverse-exports/package.json
+++ b/packages/reverse-exports/package.json
@@ -10,5 +10,10 @@
   "author": "",
   "license": "ISC",
   "devDependencies": {
+    "@types/minimatch": "^3.0.4"
+  },
+  "dependencies": {
+    "minimatch": "^3.0.4",
+    "resolve.exports": "^2.0.2"
   }
 }
--- a/packages/reverse-exports/package.json
+++ b/packages/reverse-exports/package.json
@@ -1,0 +1,14 @@
+{
+  "name": "@embroider/reverse-exports",
+  "version": "0.0.0",
+  "description": "",
+  "main": "src/index.js",
+  "scripts": {
+    "test": "jest"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "devDependencies": {
+  }
+}
--- a/packages/reverse-exports/src/index.ts
+++ b/packages/reverse-exports/src/index.ts
@@ -1,0 +1,14 @@
+import { posix } from 'path';
+
+export default function reversePackageExports(
+  packageJSON: { exports?: any; name: string },
+  relativePath: string
+): string {
+  // TODO add an actual matching system and don't just look for the default
+  if (packageJSON.exports?.['./*'] === './dist/*.js') {
+    return posix.join(packageJSON.name, relativePath.replace(/^.\/dist\//, `./`).replace(/\.js$/, ''));
+  }
+
+  // TODO figure out what the result should be if it doesn't match anything in exports
+  return posix.join(packageJSON.name, relativePath);
+}
--- a/packages/reverse-exports/src/index.ts
+++ b/packages/reverse-exports/src/index.ts
@@ -1,14 +1,116 @@
 import { posix } from 'path';
+import minimatch from 'minimatch';
+import { exports as resolveExports } from 'resolve.exports';
 
-export default function reversePackageExports(
-  packageJSON: { exports?: any; name: string },
-  relativePath: string
-): string {
-  // TODO add an actual matching system and don't just look for the default
-  if (packageJSON.exports?.['./*'] === './dist/*.js') {
-    return posix.join(packageJSON.name, relativePath.replace(/^.\/dist\//, `./`).replace(/\.js$/, ''));
+type Exports = string | string[] | { [key: string]: Exports };
+
+/**
+ * An util to find a string value in a nested JSON-like structure.
+ *
+ * Receives an object (a netsted JSON-like structure) and a matcher callback
+ * that is tested against each string value.
+ *
+ * When a value is found, returns an object containing a `value` and a `key`.
+ * The key is one of the parent keys of the found value — the one that starts
+ * with `.`.
+ *
+ * When a value is not found, returns `undefined`.
+ */
+export function _findPathRecursively(
+  exportsObj: Exports,
+  matcher: (path: string) => boolean,
+  key = '.'
+): { key: string; value: Exports } | undefined {
+  if (typeof exportsObj === 'string') {
+    return matcher(exportsObj) ? { key, value: exportsObj } : undefined;
   }
 
-  // TODO figure out what the result should be if it doesn't match anything in exports
-  return posix.join(packageJSON.name, relativePath);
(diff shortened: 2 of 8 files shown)


packages/reverse-exports/tests/reverse-exports.test.ts
```diff
@@ -0,0 +1,87 @@
...
+        './glob/*': './grod/**/*.js',
+      },
+    };
+
+    expect(reversePackageExports(packageJson, './main.js')).toBe('my-addon');
+    expect(reversePackageExports(packageJson, './secondary.js')).toBe('my-addon/sub/path');
+    expect(reversePackageExports(packageJson, './directory/some/file.js')).toBe('my-addon/prefix/some/file.js');
+    expect(reversePackageExports(packageJson, './other-directory/file.js')).toBe('addon/prefix/deep/file.js');
+
+    expect(reversePackageExports(packageJson, './yet-another/deep/file/deep/file.js')).toBe(
```
mansona, line 67:
```suggestion
    expect(reversePackageExports(packageJson, './yet-another/deep/file.js')).toBe(
```

packages/reverse-exports/tests/reverse-exports.test.ts
```diff
@@ -0,0 +1,87 @@
...
+      'addon/other-prefix/deep/file.js'
```
mansona, line 68:
```suggestion
      'addon/other-prefix/deep/file'
```

packages/reverse-exports/tests/reverse-exports.test.ts
```diff
@@ -0,0 +1,87 @@
...
+    );
+
+    expect(reversePackageExports(packageJson, './grod/very/deep/file/deep/file.js')).toBe(
```
mansona, line 71:
```suggestion
    expect(reversePackageExports(packageJson, './grod/very/deep/file.js')).toBe(
```

packages/reverse-exports/tests/reverse-exports.test.ts
```diff
@@ -0,0 +1,87 @@
...
+      'addon/glob/very/deep/file.js'
```
mansona, line 72:
```suggestion
      'addon/glob/very/deep/file'
```

packages/reverse-exports/tests/reverse-exports.test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  it('alternative exports', function () {
+    const packageJson = {
+      name: 'my-addon',
+      exports: {
+        './things/': ['./good-things/', './bad-things/'],
+      },
+    };
+
+    expect(reversePackageExports(packageJson, './good-things/apple.js')).toBe('my-addon/things/apple');
```
mansona, line 84:
```suggestion
    expect(reversePackageExports(packageJson, './good-things/apple.js')).toBe('my-addon/things/apple.js');
```

packages/reverse-exports/tests/reverse-exports.test.ts
```diff
@@ -0,0 +1,87 @@
...
+    expect(reversePackageExports(packageJson, './bad-things/apple.js')).toBe('my-addon/things/apple');
```
mansona, line 85:
```suggestion
    expect(reversePackageExports(packageJson, './bad-things/apple.js')).toBe('my-addon/things/apple.js');
```

mansona (commented):
Looking good so far 🎉 we need to add the conditional exports stuff (at least for import, export, and default) and then "just" add the implementation 😂

packages/reverse-exports/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "reverse-exports",
```
mansona, line 2: @ef4 how do you feel about this not being `@embroider/reverse-exports`? I figured it would be nice to just have it as `reverse-exports` in case anyone else needed it or in case we wanted to extract the package from the monorepo at any point

packages/reverse-exports/src/index.ts
```diff
@@ -0,0 +1,118 @@
...
+  const { key, value } = maybeKeyValuePair;
+
+  if (typeof value !== 'string') {
+    throw new Error('Expected value to be a string');
+  }
+
+  const maybeResolvedPaths = resolveExports({ name, exports: { [value]: key } }, relativePath);
+
+  if (!maybeResolvedPaths) {
+    throw new Error('Expected path to be found at this point');
```
mansona, line 112:
```suggestion
    throw new Error('Bug Discovered! `_findPathRecursively()` must always return a string value but instead it found a 'typeof value'. Please report this as an issue to <link to new issue on embroider>');
```

packages/reverse-exports/src/index.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  throw new Error(`Unexpected type of obj: ${typeof exportsObj}`);
+}
+
+export default function reversePackageExports(
+  { exports: exportsObj, name }: { exports?: Exports; name: string },
+  relativePath: string
+): string {
+  // // TODO add an actual matching system and don't just look for the default
```
mansona, line 80: delete me!

packages/reverse-exports/tests/reverse-exports.test.ts
```diff
@@ -0,0 +1,251 @@
+import reversePackageExports, { _findPathRecursively } from '../src';
+
+describe('reverse exports', function () {
+  //   it('correctly reversed exports', function () {
+  //     // TODO figure out what the result should be if it doesn't match anything in exports
+  //     expect(reversePackageExports({ name: 'best-addon' }, './dist/_app_/components/face.js')).toBe(
```
mansona, line 6: we need to keep this test i.e. a package with no exports

mansona (changes requested):

mansona:
as for the case where we have an exports but don't match it correctly, I think we should throw with an error like: 

`You tried to reverse exports for the file <filename> in package <package name> but it does not match any of the packages exports rules. This means it should not be possible to access directly.`

packages/reverse-exports/src/index.ts
lolmaus: Done.

packages/reverse-exports/src/index.ts
lolmaus: Done.

packages/reverse-exports/tests/reverse-exports.test.ts
lolmaus: Done.

lolmaus:
> as for the case where we have an exports but don't match it correctly, I think we should throw with an error like:

Done.

packages/reverse-exports/package.json
mansona: @lolmaus let's reset this back to `@embroider/reverse-exports` and updated where it needs to be 👍

lolmaus: @mansona Done.

PR #1653, opened 2023-11-07
Refactor embroider-implicit-modules export pojo

mansona:
This PR is part of the work to a) remove the need for rewritten_app and b) eventually support vite

PR #1654, opened 2023-11-08
don't run CI for all branches starting with v

mansona:
We noticed that CI was originally running twice for https://github.com/embroider-build/embroider/pull/1653 because the branch name was `virtual-implicit-modules` 🤦 

This PR makes sure that branch pushes only run if it starts with a `v` that is immediately followed by at least one number, it's not a perfect solution but it's a lot better than what we had 😂

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ on:
   push:
     branches:
       - main
-      - 'v*' # prior release branches (e.g. `v0.30.x` branch)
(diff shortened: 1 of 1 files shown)


void-mAlex (approved):

void-mAlex:
should we do the same for tags?

mansona:
> should we do the same for tags?

No i think we're fine, most of the time we're not going to "accidentally" create a tag that starts with v but isn't in the format of vX.Y.Z 👍

PR #1655, opened 2023-11-09
fix node version for @embroider/util

mansona:
I noticed while updating another ember app that `@embroider/util` accidentally dropped support for Node 12 in a minor version so this just fixes that 👍

### DIFF
--- a/packages/util/ember-cli-build.js
+++ b/packages/util/ember-cli-build.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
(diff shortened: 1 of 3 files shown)


SergeAstapov (approved):

void-mAlex (approved):

void-mAlex:
looks like 'genuine' windows test failures

mansona:
Yea it's not just failing in windows I'm afraid 😞 the issue with this package is that it's not actually possible (I think) to test it in isolation from an ember app 🤔 and we might have to so a bit of work to get tests passing for an Ember app in Node 12 😫 I'll take another look today 👍

PR #1663, opened 2023-11-14
adjust moved-package-target for namespaced packages

ef4:
I don't think this was actually breaking anything but it produced more confusing error messages when a module is missing.


packages/compat/src/standalone-addon-build.ts
```diff
@@ -52,7 +52,12 @@ ${summarizePeerDepViolations(violations)}
   });
 
   let fakeTargets = Object.values(index.packages).map(dir => {
-    return writeFile(join(dir, '..', 'moved-package-target.js'), '');
+    let segments = dir.split('/');
```
void-mAlex, line 55: @ef4 is it ok for this to not split on `sep` ?

ef4: I actually tried that first and it breaks, because the index of moved packages here is always assembled in unix format.

PR #1664, opened 2023-11-14
reorder extension precedence when searching app-js and fastboot-js trees

ef4:
As far as I can tell, this should always have been .js before .hbs and we just missed it when adding this.

### DIFF
--- a/packages/core/src/module-resolver.ts
+++ b/packages/core/src/module-resolver.ts
@@ -1130,7 +1130,7 @@ export class Resolver {
     if (filename.match(/\.(hbs|js|hbs\.js)$/)) {
       yield filename;
     } else {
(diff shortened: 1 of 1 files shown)


chancancode (approved):
Worth noting that this was previously said to be deliberate:

https://github.com/embroider-build/embroider/commit/da744d9ea68da0dfa4b21ed29e3e2d1e4bbee984#diff-61967680c203e14cb9a3b2c252caca7d6b0ee9b285991dba6b728f6c54112a12L960-R1032

This order seems wrong, and tests pass, so 👍🏼 but if anyone somehow gets bitten by this, it may be worth tracing it deeper into why the original code had it done this way

ef4:
The origin was https://github.com/embroider-build/embroider/pull/1463. It was definitely not intentional.

PR #1665, opened 2023-11-15
Needs love: Update porting-addons-to-v2.md to mention "app-js"

Alonski:
Where is this "app-js" explained?
Should this be explained in this porting guide?

### DIFF
--- a/docs/porting-addons-to-v2.md
+++ b/docs/porting-addons-to-v2.md
@@ -255,7 +255,10 @@ module.exports = addonV1Shim(__dirname);
     "ember-addon": {
       "main": "addon-main.js",
       "type": "addon",
-      "version": 2
+      "version": 2,
+      "app-js": {
+        "./components/path/component-name.js
+      }
     }
     ```
 16. In the `addon` directory, run `yarn start` to start building the addon.


mansona:
So i've you're using the v2 addon blueprint those entries are added automatically by rollup. I'm not sure we should be implying that people should be updating them manually 🤔

NullVoxPopuli:
people still need to know about it for debugging purposes.

NullVoxPopuli:
maybe, I think we could cross-link to the spec tho. like, "for more information on {this thing}, see {this specific location in the RFC/spec}"

mansona:
yes but the line right above this says "add this to your package.json" so it really can't go here. These are not intended to be added manually

Alonski:
Hmm but this is a porting guide not a ran v2 blueprint guide ☺

I went over this to see if a v2 addon I had locally was actually a v2 addon. So I went through this guide and the only thing that was different was what I added here.

I'm not saying my change was good, I'm trying to surface that right now it could be confusing not to include any info on this in this guide.

Thanks friends ♥

Alonski:
@mansona what did you mean by this should not be added manually?

Is the app-js property supposed to be generated?

ef4:
It's public API. The v2 addon blueprint has a utility for automatically updating them, but that is optional.

The addon porting guide was created before we had a blueprint. Somebody should redo it in light of the blueprint existing.

The best existing reference for `app-js` is https://github.com/emberjs/rfcs/blob/master/text/0507-embroider-v2-package-format.md.

PR #1676, opened 2023-11-20
fix single asterisk replacement in reverse-exports

mansona:
I took the reverse-exports package out for its first real-world spin today by combining https://github.com/embroider-build/embroider/pull/1648 https://github.com/embroider-build/embroider/pull/1650 and https://github.com/embroider-build/embroider/pull/1653 into one giant mega branch 💪 ... but it didn't work 😂 

I first looked at the error and thought "on no we need to update the blueprint because the exports are wrong" but thinking about it a bit better it looks like our reverse-exports could do with a little bit of love. This is totally expected in the dev process of the package since real-world examples will always beat lab-built ones 👍 

@lolmaus if you could take a look at this and get the test passing that would be super 🎉

### DIFF
--- a/packages/reverse-exports/tests/reverse-exports.test.ts
+++ b/packages/reverse-exports/tests/reverse-exports.test.ts
@@ -140,6 +140,24 @@ describe('reverse exports', function () {
       'You tried to reverse exports for the file `./foo.bar` in package `my-addon` but it does not match any of the exports rules defined in package.json. This means it should not be possible to access directly.'
     );
   });
+
+  it('breaks TODO rename this test to something better', function () {
+    const packageJson = {
+      name: 'my-v2-addon',
+      exports: {
+        '.': './dist/index.js',
+        './*': {
+          types: './dist/*.d.ts',
+          default: './dist/*.js',
+        },
+        './addon-main.js': './addon-main.js',
+      },
+    };
+
+    expect(reversePackageExports(packageJson, './dist/_app_/components/welcome-page.js')).toBe(
+      'my-v2-addon/_app_/components/welcome-page'
+    );
+  });
 });
 
 describe('_findKeyRecursively', function () {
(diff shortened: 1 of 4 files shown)


lolmaus:
I have found this in [the documentation](https://nodejs.org/api/packages.html#subpath-patterns):

> `*` maps expose nested subpaths as it is a string replacement syntax only.

Also, the mention of `**` the double asterisk glob in the documentation does not actually allow using it in keys or values!

> The property of exports being statically enumerable is maintained with exports patterns since the individual exports for a package can be determined by treating the right hand side target pattern as a `**` glob against the list of files within the package.

Thank you for the heads up, I know exactly how to fix it.

lolmaus:
Fixed now!

packages/reverse-exports/tests/reverse-exports.test.ts
```diff
@@ -260,3 +278,18 @@ describe('_findKeyRecursively', function () {
     expect(_findPathRecursively(exports, str => str === './missing-path.js')).toBe(undefined);
   });
 });
+
+///
```
mansona, line 282: delete this empty comment

packages/reverse-exports/tests/reverse-exports.test.ts
```diff
@@ -140,6 +140,24 @@ describe('reverse exports', function () {
       'You tried to reverse exports for the file `./foo.bar` in package `my-addon` but it does not match any of the exports rules defined in package.json. This means it should not be possible to access directly.'
     );
   });
+
+  it('breaks TODO rename this test to something better', function () {
```
mansona, line 144: I know this is my line but I hoped you would replace the todo with a better name once you figured out what the issue was 😂 maybe something like "works with single * replacement" ? 🤔

mansona (commented):
I can't "request changes" since it's my PR 🙃 one minor thing and one that I would like to fix before we merge 👍

packages/reverse-exports/tests/reverse-exports.test.ts
lolmaus: Fixed.

packages/reverse-exports/tests/reverse-exports.test.ts
lolmaus: Deleted.

mansona (commented):
Looks good to me 👍

lolmaus (approved):
Great job, @lolmaus! 😂

PR #1679, opened 2023-11-27
[vite entry point] add gjs + gts extension for resolvable extensions

patricklx:
add gjs+gts as resolveable extensions for vite entry point asset file

### DIFF
--- a/packages/compat/src/compat-app-builder.ts
+++ b/packages/compat/src/compat-app-builder.ts
@@ -199,7 +199,7 @@ export class CompatAppBuilder {
     // For TS, we defer to ember-cli-babel, and the setting for
     // "enableTypescriptTransform" can be set with and without
     // ember-cli-typescript
-    return ['.wasm', '.mjs', '.js', '.json', '.ts', '.hbs', '.hbs.js'];
+    return ['.wasm', '.mjs', '.js', '.json', '.ts', '.hbs', '.hbs.js', '.gts', '.gjs'];
   }
 
   private *emberEntrypoints(htmlTreePath: string): IterableIterator<Asset> {


ef4:
Hi, I need more detail to understand what you're trying to do here.

`resolvableExtensions` is trying to match existing ember-cli behaviors and those don't actually resolve gjs/gts. I suspect there might be better place to put this fix to achieve what you're trying to achieve.

patricklx:
It's possible to have gts/gjs files for templates with ember-route-templates. 
It's possible to transform them before starting vite, but  i would like vite to do the transform instead.
Also because I'm working to make it work to load everything from app dir instead of rewritten-app.
Which is mostly working now, except for this.

ef4:
Sure but the [gjs plugin you've also been working on](https://github.com/embroider-build/embroider/pull/1680/files) already resolve gjs/gts. I still don't understand why it would be needed here in compat-app-builder too.

patricklx:
i think this also makes sense without having vite load everything from root.
it can load gts/gjs files directly and we can also then remove `enableTypeScriptTransform` in ember-cli-build as vite can do it. making hot reload even faster

ef4:
It's not OK to sometimes load from root and sometimes from the rewritten app, and so long as the files existing in the rewritten app they are resolvable. 

If you have both, vite will duplicate them.

We are quite close to eliminating rewritten app entirely.

patricklx:
Right, what i meant in my last statement is that embroider could just copy the gts/gjs files to the rewritten-app and have vite resolve them. 
But the issue would remain that the route.gts would be missing from the entry point defines.
It would then lead to issues in the app, because ember would use the default router instead.

It's good to hear that rewritten-app is soon gone.

patricklx:
Wip here: https://github.com/embroider-build/embroider/pull/1695

patricklx:
reopen to separte out of big vite  pr

patricklx:
moved to #1718

PR #1680, opened 2023-11-27
[vite] use transform instead of load for gjs & hbs files

patricklx:
use transform function, this simplifies loading of those files without the need to handle paths

### DIFF
--- a/packages/vite/src/template-tag.ts
+++ b/packages/vite/src/template-tag.ts
@@ -2,13 +2,13 @@ import { createFilter } from '@rollup/pluginutils';
 import type { Plugin } from 'vite';
 import { Preprocessor } from 'content-tag';
 
-const gjsFilter = createFilter('**/*.{gjs,gts}?(\\?)*');
+const gjsFilter = createFilter('**/*.gjs?(\\?)*');
 
 export function templateTag(): Plugin {
   let preprocessor = new Preprocessor();
 
   function candidates(id: string) {
-    return [id + '.gjs', id + '.gts'];
+    return [id + '.gjs'];
   }
 
   return {
--- a/packages/vite/src/template-tag.ts
+++ b/packages/vite/src/template-tag.ts
@@ -1,15 +1,14 @@
 import { createFilter } from '@rollup/pluginutils';
 import type { Plugin } from 'vite';
-import { readFileSync } from 'fs';
 import { Preprocessor } from 'content-tag';
 
-const gjsFilter = createFilter('**/*.gjs?(\\?)*');
+const gjsFilter = createFilter('**/*.{gjs,gts}?(\\?)*');
 
 export function templateTag(): Plugin {
   let preprocessor = new Preprocessor();
 
   function candidates(id: string) {
-    return [id + '.gjs'];
+    return [id + '.gjs', id + '.gts'];
   }
 
   return {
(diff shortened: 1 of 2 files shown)


patricklx:
@mansona @ef4

packages/vite/src/template-tag.ts
```diff
@@ -1,15 +1,14 @@
 import { createFilter } from '@rollup/pluginutils';
 import type { Plugin } from 'vite';
-import { readFileSync } from 'fs';
 import { Preprocessor } from 'content-tag';
 
-const gjsFilter = createFilter('**/*.gjs?(\\?)*');
+const gjsFilter = createFilter('**/*.{gjs,gts}?(\\?)*');
```
mansona, line 5: I'm not comfortable adding gts to this for now. I know we should deal with the Typescript situation but I feel like that should be a wholistic PR that explicitly adds support rather than changing this one place

mansona (changes requested):
I think overall this is a good PR, it also potentially allows us to tap into some caching of modules because of the caching behaviour of rollup

> If the Rollup cache is used (e.g. in watch mode or explicitly via the JavaScript API), Rollup will skip the [transform](https://rollupjs.org/plugin-development/#transform) hook of a module if after the [load](https://rollupjs.org/plugin-development/#transform) hook, the loaded code is identical to the code of the cached copy.

However I'm not too comfortable expanding the gjs filter to gts because I can imagine it would have a whole list of unintended consequences down the line 🙈 I'm sure it's probably fine but I would prefer that to be in a different PR if that's ok?

packages/vite/src/template-tag.ts
NullVoxPopuli: What is 'the typescript situation'?

For running ts code, babel (or whatever) removes the type annotations and that's it

mansona: yes but my understanding is that at this level we should always only be seeing gjs files no? and by "situation" I mean fully end-to-end tests on it with typechecking, that should be a different PR in my opinion

patricklx: Right, as of now it would not make any difference. Since currently it's expected that the previous stage compiles the gts files.
It would work together with #1679 . Which would work for vite and then i could make it work in the tests in #1718

NullVoxPopuli: Why wouldn't we see the input file ext, if gjs is here already?

If it were compiled, we'd just have js, yeah?

patricklx: Potentially could also work with webpack if content-tag is added as a loader. Or is there already one?
Why would it work with gjs but not gts?

patricklx: Also, .ts is already part of the entry points and vite can load it directly. E.g the vite-app is not setup to transform ts files for the previous stage.
But sure, i can move this gts part elsewhere.
Shall i add it to #1679 ?

patricklx: > Why wouldn't we see the input file ext, if gjs is here already?
> 
> If it were compiled, we'd just have js, yeah?

True, somehow i was thinking gts transforms to gjs first... But it would be js directly, right?

NullVoxPopuli: Right, there is no gts to gjs conversion.
I think it goes: gts->ts->js

For content-tag -> babel/whatever

patricklx:
@mansona i reverted the gts part

mansona (approved):
great work 🎉

PR #1684, opened 2023-11-28
Workaround typescript regression in router package

ef4:
We're hitting https://github.com/microsoft/TypeScript/issues/56571 such that the router package is emitting a d.ts containing a `<reference types` that refers to internal private paths in ember-source that are not always valid.

Pinning the typescript version in that package.


mansona (approved):
Nice one 🎉

PR #1468, opened 2023-06-08
macroCondition: do branch elimination if no runtime impl. is involved

simonihmig:
This will make `macroCondition()` do full branch elimination even in development, as long as no macro is involved that has a runtime implementation (`get*Config()`, `isTesting` have).

Closes https://github.com/embroider-build/ember-auto-import/issues/582

PR #1688, opened 2023-11-29
Fix branch elimination for `macroDependencySatisfies`

mike-engel:
This is a follow up to #1468, and should fix https://github.com/ember-animation/ember-animated/pull/647.

The problem is that #1468 only fixed branch elimination for non-runtime macros in JS/TS code via babel, but didn't fix it for templates via glimmer. This fixes `macroDependencySatisfies` by transforming it into a `macroCondition` in the first pass, which will eliminate the branch in the second pass when it's used as a sub-expression.

Writing code like `macroCondition (macroDependencySatisfies 'qunit' '^2.8.0')` will work, but is not ergonomic and is unfriendly.

I don't think this needs to be done for `macroGetConfig` or `macroGetOwnConfig`.

### DIFF
--- a/packages/macros/src/glimmer/ast-transform.ts
+++ b/packages/macros/src/glimmer/ast-transform.ts
@@ -77,7 +77,7 @@ export function makeFirstTransform(opts: FirstTransformParams) {
             }
           },
         },
-        SubExpression(node: any) {
+        SubExpression(node: any, walker: { parent: { node: any } }) {
           if (node.path.type !== 'PathExpression') {
             return;
           }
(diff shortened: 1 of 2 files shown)


NullVoxPopuli:
thanks for submitting this! looks like you have some test failures there!

mike-engel:
Whoops, I thought they were all passing before I pushed.

@NullVoxPopuli I added an extra case to handle macro compositions, but it's starting to feel a little hacky. An alternative to this PR would be to expand the docs to call out explicitly that `macroDependencySatisfies` requires a parent `macroCondition` in templates to properly prune branches. That also feels a little hacky, but it's potentially less brittle.

mansona:
Hey tests were failing so I rebased this 👍 I also re-targeted it to stable so we can get the bugfix out the door

mansona (approved):
Looks great 🎉

mike-engel:
Wow, that was simple! Thanks @mansona

PR #1692, opened 2023-11-30
force importSync to always be eager

mansona:


ef4:
I ultimately had to drop all of ember-data < 5 from the test coverage, so as it stands that is part of the breaking change. We can separately assess whether / how we can restore some of that support with new compat adapters.

PR #1695, opened 2023-12-01
vite dev/test/build from app

patricklx:
this can now already load most from root folder ./app and tests folder instead of rewritten_app folder.  a few hacks where required.

the only files read from rewritten_app are :
* config/environment.js
* assets/test-support.js
* assets/*.css
* index.html

most problematic where asset loading.
vite does not allow loading static files from outside of root dir, so setting root to /app will not work.
using the CWD as root, will not load the index.html file.

this is where the custom middleware comes in to handle this parts.
* load index.html foim app dir
* load assets from public dir
* load assets from node-module package
* load from rewritten-packages

missing parts to allow editing live reload:
- [x] environment.js -> i think we should just call the `vite-app/config/environment.js` again at runtime (browser)  and merge it again , like this, developers can edit the env and have it load by vite.
- [x] assets/vite-app.js 
- [x] transform index.html in vite --> need replacements for `content-for` in a separate file so we can do it our selfs and expose api to build index.html
- [x] working tests
- [x] working vite build 
- [x] rollup plugin for assets
- [x] auto rebuild when lock file changed or no . embroider dir


patricklx:
some tests are failing for unknown reason.
when I run test with `pnpm test "--" "--filter" "release-compat-namespaced-app"` it fails
but when I run it with `node \\wsl$\Ubuntu\home\patrick\projects\embroider\node_modules\.pnpm\qunit@2.20.0\node_modules\qunit\bin\qunit.js --require ts-node/register *-test.ts "--filter" "release-compat-namespaced-app"` it works...

patricklx:
Should i add a Test for vite app?
Can just be `vite build `

patricklx:
@ef4 this is ready

packages/vite/src/build.ts
```diff
@@ -0,0 +1,96 @@
...
+    return hashSum.digest('hex');
+  }
+  return null;
+}
+
+export function emberBuild(): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const child = fork('./node_modules/ember-cli/bin/ember', ['build']);
+    child.on('exit', code => (code === 0 ? resolve() : reject()));
+  });
```
ef4, line 47: I don't think this is the best place to manage rebuilds of the addons. This kind of rebuild is extremely expensive compared with `ember s`.

We could still do the persistent caching you have here in order to avoid even a first-time build of the addons if none have changed since your last build. But that logic would go some place more like https://github.com/embroider-build/embroider/blob/3480aa65fa33b937748c9f94d8b334a1d52d8610/packages/compat/src/standalone-addon-build.ts#L13

packages/vite/src/build.ts
```diff
@@ -0,0 +1,96 @@
...
+      console.log(path + ' change requires rebuild, rebuilding...');
+      await emberBuild();
+      updateCacheKey(path, key);
+    }
+  }
+}
+
+export function build(): Plugin {
+  let resolverLoader = new ResolverLoader(process.cwd());
+  const engine = resolverLoader.resolver.options.engines[0];
```
ef4, line 63: engines other than the first need this feature too.

packages/vite/src/build.ts
```diff
@@ -0,0 +1,96 @@
...
+  return {
+    name: 'embroider-builder',
+    enforce: 'pre',
+    writeBundle(options) {
+      engine.activeAddons.forEach(addon => {
+        const pkg = resolverLoader.resolver.packageCache.ownerOfFile(addon.root);
+        if (!pkg) return;
```
ef4, line 70: If you do `!pkg?.isV2Addon()` you won't need to cast `pkg.meta` below.

ef4 (changes requested):
Thanks for giving this a go. Your spike showing this working is a helpful guide for several things, especially the middleware setup.

But this is not mergeable as is. A lot of stuff here is hard-coded in ways that might work in your app but definitely break in others.  And `@embroider/vite` should only need to do vite-specific things. 

Can we find time for you to pair with either me or @mansona? Because your help on this would be great, but it's so far not aligned with what we're trying to do. The effort that went into this PR could go directly into fixing these things in `@embroider/core` instead of hacking around them.

packages/util/package.json
```diff
@@ -52,6 +52,7 @@
     "@ember/test-helpers": "^2.9.1",
     "@embroider/compat": "workspace:*",
     "@embroider/core": "workspace:*",
+    "@embroider/shared-internals": "workspace:*",
```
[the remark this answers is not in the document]
patricklx: Yes, not sure why tests where suddenly failing. It was failing we this missing dependency...

packages/vite/src/build.ts
patricklx: Yes,  well it would also be quite rare to do build. Only on package change and index.html change

packages/vite/src/build.ts
patricklx: okay

packages/vite/src/build.ts
patricklx: Will try it

packages/vite/src/build.ts
```diff
@@ -0,0 +1,96 @@
...
+        const assets = (pkg.meta as AddonMeta)['public-assets'] || {};
+        Object.entries(assets).forEach(([path, dest]) => {
```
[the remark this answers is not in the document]
patricklx: Middleware works only in dev mode and will not be setup in build mode

packages/vite/src/hbs.ts
```diff
@@ -16,11 +16,16 @@ export function hbs(): Plugin {
     name: 'rollup-hbs-plugin',
     enforce: 'pre',
     async resolveId(source: string, importer: string | undefined) {
+      // prevent resolve loop during vite build
+      if (source.match(/\.(gjs|gts)$/)) {
```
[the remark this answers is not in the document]
patricklx: Somehow when building, skip self does not work reliable.
Then template-tag.js and hbs.js are invoking each other

packages/vite/src/hbs.ts
```diff
@@ -16,11 +16,16 @@ export function hbs(): Plugin {
...
+        return null;
+      }
       let resolution = await this.resolve(source, importer, {
         skipSelf: true,
       });
 
       if (!resolution) {
+        if (source.endsWith('.hbs')) return null;
```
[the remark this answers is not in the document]
patricklx: It's because in build mode, it was invoking itself again

packages/vite/src/resolver.ts
```diff
@@ -1,17 +1,125 @@
...
     name: 'embroider-resolver',
     enforce: 'pre',
+    configureServer(server) {
+      const files = readdirSync('.');
+      files.forEach(f => {
+        if (lockFiles.includes(f)) {
+          server.watcher.add('./' + f);
+        }
+      });
+      server.watcher.add('./app/index.html');
```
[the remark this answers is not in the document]
patricklx: moving to root will not work, because the ember build will not work

packages/vite/src/build.ts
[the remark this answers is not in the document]
patricklx: this is only for building so it ends up in the dist folder.
I only found another implementation in embroider/webpack...

packages/vite/src/virtual-files.ts
```diff
@@ -0,0 +1,125 @@
+import { relative } from 'path';
```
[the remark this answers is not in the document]
patricklx: replacing this with compat app builder

packages/compat/src/compat-app-builder.ts
```diff
@@ -57,6 +57,7 @@ import type { TransformOptions } from '@babel/core';
 import { MacrosConfig } from '@embroider/macros/src/node';
 import SourceMapConcat from 'fast-sourcemap-concat';
 import escapeRegExp from 'escape-string-regexp';
+import { configReplacePatterns } from 'ember-cli/lib/utilities/ember-app-utils';
```
patricklx, line 60: Im not sure if there is another place where embroider does the replacements

patricklx:
looks like the pr #1686 broke this... 
Cannot find components/example. Any idea why?

ef4:
Perhaps it was relying on the resolvableExtensions in the node resolver, which is not there anymore after #1686. 

For normal module imports that would not be an issue, since we can just configure vite to resolve .ts. But it is potentially an issue for discovering components in non-strict templates, since we're using nodeResolve there. 

That gives me an idea: we could probably use `defaultResolve` instead of `nodeResolve` for component discovery. That would mean that our component search would understand `.ts` automatically when the prevailing packager does.

patricklx:
Also, I needed to add back the addons.ts for the depOptimizer as otherwise i was getting duplicates for glimmer-vm which does allow it.

patricklx:
@mansona maybe you can already have a look at this now.
I tried to implement the required changes without affecting normal builds

packages/compat/src/compat-app.ts
```diff
@@ -200,7 +209,7 @@ export default class CompatApp {
 
   private get indexTree() {
     let indexFilePath = this.legacyEmberAppInstance.options.outputPaths.app.html;
-    let index = buildFunnel(this.legacyEmberAppInstance.trees.app, {
+    let index = buildFunnel(this.legacyEmberAppInstance.trees.indexHtml || this.legacyEmberAppInstance.trees.app, {
```
patricklx, line 212: required to have index.html at root for vite

packages/vite/src/addons.ts
```diff
@@ -0,0 +1,23 @@
+import { ResolverLoader, packageName } from '@embroider/core';
```
patricklx, line 1: added this back because i was getting duplicate packages, like glimmer-vm which does not like it

tests/vite-app/tests/setup-harness.js
```diff
@@ -0,0 +1,59 @@
+/* eslint-disable no-console */
```
patricklx, line 1: this is to support ci tests with vite

packages/vite/src/resolver.ts
```diff
@@ -43,22 +72,44 @@ function defaultResolve(context: PluginContext): ResolverFunction<RollupModuleRe
...
+          },
         },
-      },
-    });
+      });
+    }
+    if (!result) {
+      result = await context.resolve(
+        request.specifier,
+        request.fromFile.replace('/package.json', '/app/package.json'),
```
patricklx, line 101: need to look into /app folder

patricklx (commented):
@ef4 my current wip to virtualize generated assets. Am I om the right track?
https://github.com/embroider-build/embroider/pull/1695/commits/149d5b76737703a2bf79e8c71de22a3fe36d0e36
Now i only need to make some files paths absolute instead of relative in the assets.
When this works i extract this into a separate pr.

PR #1696, opened 2023-12-04
Use rollup's `addWatchFile` API to mark dependencies

chancancode:
This solves a bunch of watch-mode (not) rebuilding issues; as a general rule everytime we probe or read from the file system we should at least consider whether such an annotation is needed.

Unfortunately this package doesn't appear to be tested currently, but I have applied the same patch locally and it appears to be working fine.

If someone has the bandwidth to set up new tests for this package (similar to https://github.com/embroider-build/embroider/blob/main/tests/scenarios/watch-mode-test.ts), I am happy to add the regression tests, but otherwise it probably shouldn't block landing this, I don't think it's making things worse.


chancancode:
One example of the issue this fixes is that, currently, in a v2 addon, if you add a new file in the `src` folder after the initial build, it won't be picked up by rollup at all until you restart the build.

ef4:
Based on your comments it look like this probably bumps the minimum supported rollup version? That would be OK but we should document it and consider whether this should be a server major of `@embroider/addon-dev`.

SergeAstapov:
if this end up requiring major, may be bundled with #1697

ef4:
Possibly place to add a test if it's not hard: https://github.com/embroider-build/embroider/blob/main/tests/scenarios/v2-addon-dev-watch-test.ts

chancancode:
I couldn't get the test to work on CI so I split it out into #1711

PR #1702, opened 2023-12-05
reduce duplication in resolver config

ef4:
And add top-level app as a valid "activeAddon" fallback.


ef4:
This landed in #1686.

PR #1706, opened 2023-12-06
Support engine's lazyLoading boolean config

fengb:
Engine's lazyLoading previously used boolean configs `lazyLoading: true`. This re-enables the backwards compatibility inside embroider.

Source: https://github.com/ember-engines/ember-engines/blob/master/packages/ember-engines/lib/utils/ensure-lazy-loading-hash.js


SergeAstapov:
@fengb seems like CI is not happy with this change

raycohen:
seems like the CI issue is not just impacting this PR

void-mAlex (approved):

PR #1714, opened 2023-12-07
update scenario-tester

mansona:
~I'm not exactly sure why typescript can't find the types from the new version of scenario-tester 🤔 anyone have any clues?~

Thanks to @SergeAstapov we found the issue with types. And thanks to @ef4 we fixed the upstream issue and got scenario-tester@3.0.1 released 🎉

### DIFF
--- a/packages/compat/package.json
+++ b/packages/compat/package.json
@@ -87,7 +87,7 @@
     "broccoli-node-api": "^1.7.0",
     "code-equality-assertions": "^0.9.0",
     "ember-engines": "^0.8.19",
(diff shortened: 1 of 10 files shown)


SergeAstapov:
@mansona maybe you need to add `typesVersions` to package.json with smth like

```
"typesVersions": {
    "*": {
      "*": [
        "dist/*"
      ]
    }
  },
```

mansona:
Yep that did it 🎉 thanks @SergeAstapov

PR #1717, opened 2023-12-12
Always load ember-testing package eagerly

ef4:
In ember-source 5.6 (now in beta), we refactored the way Ember and its test-only pieces find each other to no longer be dependent on the AMD loader.

But this mean it's no longer sufficient to `define()` *and not evaluate* the `ember-testing` package in order to install the test support. It must actually get evaluted first.

Note that this bug does not manifest on embroider `main`, because there implicit-test-modules are eagerly evaluated. This is only a problem on stable, when using `staticEmberSource: true`.

### DIFF
--- a/packages/compat/src/compat-adapters/ember-source.ts
+++ b/packages/compat/src/compat-adapters/ember-source.ts
@@ -124,11 +124,16 @@ export default class extends V1Addon {
         meta['implicit-modules'] = [];
       }
       meta['implicit-modules'].push('./ember/index.js');
-
-      if (!meta['implicit-test-modules']) {
-        meta['implicit-test-modules'] = [];
+      // before 5.6, Ember uses the AMD loader to decide if it's test-only parts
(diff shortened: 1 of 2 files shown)


ef4:
I manually tested this against the beta. We don't currently have the ember beta in CI, and should also address that.

NullVoxPopuli (approved):

patricklx:
i tried to fix it for ember < 5.6 like this now:
```
let amdModules: { runtime: string; buildtime: string }[] = [
      {
        runtime: 'ember-testing',
        buildtime: 'ember-testing'
      }
    ];
```
but thats not working since the imports are already loaded before ember-testing is set with define

mansona:
So there is a bit of a similarity between the remaining failures on this PR and my pojo PR https://github.com/embroider-build/embroider/pull/1653 🤔 I don't know if that gives us a clue to what is going on here

PR #1718, opened 2023-12-12
vite test ci

patricklx:
adds ci tests for vite dev
* testing: opening /tests/ and checking the tests output 
* fixes optimizer issue

### DIFF
--- a/packages/core/src/module-resolver.ts
+++ b/packages/core/src/module-resolver.ts
@@ -453,18 +453,18 @@ export class Resolver {
   private resolveComponent<R extends ModuleRequest>(path: string, inEngine: EngineConfig, request: R): R {
     let target = this.parseGlobalPath(path, inEngine);
 
-    let hbsModule: string | null = null;
-    let jsModule: string | null = null;
+    let hbsModule: { requested: string; found: string } | null = null;
+    let jsModule: { requested: string; found: string } | null = null;
 
     // first, the various places our template might be.
     for (let candidate of this.componentTemplateCandidates(target.packageName)) {
-      let candidateSpecifier = `${target.packageName}${candidate.prefix}${target.memberName}${candidate.suffix}`;
+      let candidateSpecifier = `${target.packageName}${candidate.prefix}${target.memberName}${candidate.suffix}.hbs`;
       let resolution = this.nodeResolve(
         `${target.packageName}${candidate.prefix}${target.memberName}${candidate.suffix}`,
         target.from
       );
       if (resolution.type === 'real') {
-        hbsModule = candidateSpecifier;
+        hbsModule = { requested: candidateSpecifier, found: resolution.filename };
         break;
       }
     }
@@ -478,7 +478,7 @@ export class Resolver {
       // It matches as a priority lower than .js, so finding an .hbs means
       // there's definitely not a .js.
       if (resolution.type === 'real' && !resolution.filename.endsWith('.hbs')) {
-        jsModule = candidateSpecifier;
+        jsModule = { requested: candidateSpecifier, found: resolution.filename };
         break;
       }
     }
@@ -487,10 +487,14 @@ export class Resolver {
       return logTransition(
         `resolveComponent found legacy HBS`,
         request,
-        request.virtualize(virtualPairComponent(hbsModule, jsModule))
+        request.virtualize(virtualPairComponent(hbsModule.found, jsModule?.found))
       );
     } else if (jsModule) {
-      return logTransition(`resolveComponent found only JS`, request, request.alias(jsModule).rehome(target.from));
+      return logTransition(
+        `resolveComponent found only JS`,
+        request,
+        request.alias(jsModule.requested).rehome(target.from)
+      );
     } else {
       return logTransition(`resolveComponent failed`, request);
     }
--- a/packages/core/src/module-resolver.ts
+++ b/packages/core/src/module-resolver.ts
@@ -458,27 +458,27 @@ export class Resolver {
 
     // first, the various places our template might be.
     for (let candidate of this.componentTemplateCandidates(target.packageName)) {
+      let candidateSpecifier = `${target.packageName}${candidate.prefix}${target.memberName}${candidate.suffix}`;
       let resolution = this.nodeResolve(
         `${target.packageName}${candidate.prefix}${target.memberName}${candidate.suffix}`,
         target.from
       );
       if (resolution.type === 'real') {
-        hbsModule = resolution.filename;
+        hbsModule = candidateSpecifier;
         break;
       }
     }
 
     // then the various places our javascript might be.
     for (let candidate of this.componentJSCandidates(target.packageName)) {
-      let resolution = this.nodeResolve(
-        `${target.packageName}${candidate.prefix}${target.memberName}${candidate.suffix}`,
-        target.from
-      );
+      let candidateSpecifier = `${target.packageName}${candidate.prefix}${target.memberName}${candidate.suffix}`;
+
+      let resolution = this.nodeResolve(candidateSpecifier, target.from);
       // .hbs is a resolvable extension for us, so we need to exclude it here.
       // It matches as a priority lower than .js, so finding an .hbs means
       // there's definitely not a .js.
       if (resolution.type === 'real' && !resolution.filename.endsWith('.hbs')) {
-        jsModule = resolution.filename;
+        jsModule = candidateSpecifier;
         break;
       }
     }
--- a/packages/vite/src/esbuild-resolver.ts
+++ b/packages/vite/src/esbuild-resolver.ts
@@ -7,10 +7,10 @@ import {
   virtualContent,
   locateEmbroiderWorkingDir,
 } from '@embroider/core';
-import { readFileSync, readJSONSync } from 'fs-extra';
+import { existsSync, readFileSync, readJSONSync } from 'fs-extra';
 import { EsBuildModuleRequest } from './esbuild-request';
 import assertNever from 'assert-never';
-import { dirname, resolve, join } from 'path';
+import { dirname, resolve, join, isAbsolute } from 'path';
 import { hbsToJS } from '@embroider/core';
 import { Preprocessor } from 'content-tag';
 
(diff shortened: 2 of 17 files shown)


packages/compat/src/compat-app-builder.ts
```diff
@@ -366,7 +366,9 @@ export class CompatAppBuilder {
         kind: 'in-memory',
         relativePath: '_testing_suffix_.js',
         source: `
-        var runningTests=true;
+        import * as EmberTesting from 'ember-testing';
+        define('ember-testing', () => EmberTesting);
+        runningTests=true;
```
patricklx, line 371: i'm not sure if there is a better place

patricklx: i reverted this to get the ci tests passing...

mansona:
@patricklx I don't see this running in CI 🤔 are you just working on this to be able to run tests locally?

patricklx:
the target is to be able to run it in ci. this is just the setup for it.
No idea where to setup the ci test here in embroider repo...

patricklx:
@mansona also tests will not work without
```
import * as EmberTesting from 'ember-testing';
define('ember-testing', () => EmberTesting);
```
unless we use ember-source 5.6 beta

mansona:
> No idea where to setup the ci test here in embroider repo...

So everything happens in scenario-tester which you can see in `tests/scenarios/*-test.ts`. for now I think we could start up a new scenario that would run this

patricklx:
i might need to extract more from my other PR to make this work

patricklx:
If you could setup the scenario, then i can fix the rest :)

packages/core/src/module-resolver.ts
```diff
@@ -542,12 +542,22 @@ export class Resolver {
   }
 
   private *componentJSCandidates(inPackageName: string) {
+    const extensions = ['.ts', '.gjs', '.gts'];
```
patricklx, line 545: vite can handle this files. webpack can also already handle ts files, for gjs/gts a loader could be added

packages/vite/src/template-tag.ts
```diff
@@ -2,39 +2,53 @@ import { createFilter } from '@rollup/pluginutils';
...
 
   return {
     name: 'embroider-template-tag',
     enforce: 'pre',
 
     async resolveId(id: string, importer: string | undefined) {
-      let resolution = await this.resolve(id, importer, {
-        skipSelf: true,
-      });
+      // prevent resolve loop during vite build
```
patricklx, line 21: this parts are only needed during vite build. otherwise it will hang

tests/scenarios/vite-app-test.ts
```diff
@@ -3,8 +3,10 @@ import type { PreparedApp } from 'scenario-tester';
 import QUnit from 'qunit';
 
 const { module: Qmodule, test } = QUnit;
+const isWin = process.platform === 'win32';
```
mansona, line 6: so it's a nice solution but I'm not 100% sure it will work 🤔 it might, but it might also not be a supported scenario-tester technique

We probably want to filter over here https://github.com/embroider-build/embroider/blob/main/test-packages/support/suite-setup-util.ts#L31 because that's where we filter things for windows already.

It does beg the question: are we not intending to support Vite on windows? is there something specific that is causing a problem right now?

packages/vite/src/resolver.ts
```diff
@@ -50,15 +50,18 @@ function defaultResolve(context: PluginContext): ResolverFunction<RollupModuleRe
...
     }
-    let result = await context.resolve(request.specifier, request.fromFile, {
-      skipSelf: true,
-      custom: {
-        embroider: {
-          enableCustomResolver: false,
-          meta: request.meta,
+    let result;
+    try {
+      result = await context.resolve(request.specifier, request.fromFile, {
```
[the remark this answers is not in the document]
mansona: that's probably indicating that there is a real problem and that these requests might be causing some issues in prebuild too 🤔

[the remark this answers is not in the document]
patricklx: No, it's a difference between esbuild and vite. Esbuild reenters the same resolveId if request changes. Vite doesn't.
I saw an issue about this on the repo

tests/scenarios/vite-app-test.ts
patricklx: I want to also support windows, but currently it fails.
I do not want to make this pr bigger than it is.
Let's try to get this to work good on unix. Afterwards i can try to fix it on windows. I do mostly work on windows anyway:)

patricklx: I just realised I didn't try to fix thr path from process.cwd(). Will give it another try

patricklx:
the tests pass locally when i run them directly in the repo. But through the scenario setup it fails. Anyway dep optimizer is not working correctly.
Disabled windows for now. Will look into it later again.
@mansona

patricklx:
@mansona found a fix for windows.

patricklx:
@mansona this is ready and also shows that its working while also building cached deps

packages/vite/src/resolver.ts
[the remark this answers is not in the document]
patricklx: https://github.com/vitejs/vite/issues/13852

ef4 (changes requested):
Adding the CI test for vite is great. But that can be one PR by itself. The tests in `vite-app` **are already passing on main**, so no other changes should be needed to enable the tests in CI.

To make these other changes to core features like PackageCache or module-resolver we need tests showing why they're needed, and to ensure we don't break those cases in the future.

packages/shared-internals/src/package-cache.ts
```diff
@@ -67,7 +67,7 @@ export default class PackageCache {
   }
 
   ownerOfFile(filename: string): Package | undefined {
-    let segments = filename.split(sep);
+    let segments = filename.replace(/\\/g, '/').split(posix.sep);
```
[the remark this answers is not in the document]
patricklx: This is required for windows tests to pass.
The vite requests are always with '/', also on windows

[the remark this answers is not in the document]
patricklx: It's to fix the windows tests. The requests in vite are always with `/`.
So we actually don't need to replace the `\` slash here

packages/vite/src/template-tag.ts
```diff
@@ -2,39 +2,53 @@ import { createFilter } from '@rollup/pluginutils';
...
 
-const gjsFilter = createFilter('**/*.gjs?(\\?)*');
+const gjsFilter = createFilter('**/*.{gjs,gts}?(\\?)*');
 
 export function templateTag(): Plugin {
   let preprocessor = new Preprocessor();
 
   function candidates(id: string) {
-    return [id + '.gjs'];
+    if (id.endsWith('.gjs')) return id;
```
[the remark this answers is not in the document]
patricklx: It's the same issue during build. It renters here with an additional `.gts` or `.gjs` every time

patricklx:
> Adding the CI test for vite is great. But that can be one PR by itself. The tests in `vite-app` **are already passing on main**, so no other changes should be needed to enable the tests in CI.
> 
> To make these other changes to core features like PackageCache or module-resolver we need tests showing why they're needed, and to ensure we don't break those cases in the future.

The tests where not really testing much. Only that it loads... There where issues with decorators and vite build was also not working. 
But yes, i can minimise this to only do vite test first on unix only, and extract the rest to other prs like this:
* test windows 
* add gjs/gts entr for tests
* build (if i cannot find a simple solution)

packages/vite/src/esbuild-resolver.ts
```diff
@@ -30,6 +30,9 @@ export function esBuildResolver(root = process.cwd()): EsBuildPlugin {
         let resolution = await resolverLoader.resolver.resolve(request, defaultResolve(build, kind));
         switch (resolution.type) {
           case 'found':
+            if (resolution.result?.path?.includes('rewritten-app')) {
```
patricklx, line 33: @mansona I i think i found the real reason we have the optimizer issue.
We call context.resolve in the default resolve here https://github.com/embroider-build/embroider/pull/1718/commits/0aafe02b99bb2f5a54621c42d6722d789f46871a#diff-a6c8071691a61acc68be864b34d3f338fd92e9bf5217c2d12403ac6029bca885R130 
with the full absolute path.
which then goes to the vite dep:scan plugin which marks it as external here: https://github.com/vitejs/vite/blob/main/packages/vite/src/node/optimizer/scan.ts#L681
so either we do not call that and use e.g resolve from npm or unmark them as external here

patricklx: i also created a PR on vite as i think its a vite bug: https://github.com/vitejs/vite/pull/15402
lets see

patricklx: Btw, this was working in windows without issues... Will verify later again why this is. I suspect the generated path has somr difference in slashes

packages/core/src/module-resolver.ts
```diff
@@ -458,27 +458,27 @@ export class Resolver {
...
     // first, the various places our template might be.
     for (let candidate of this.componentTemplateCandidates(target.packageName)) {
+      let candidateSpecifier = `${target.packageName}${candidate.prefix}${target.memberName}${candidate.suffix}`;
       let resolution = this.nodeResolve(
         `${target.packageName}${candidate.prefix}${target.memberName}${candidate.suffix}`,
         target.from
       );
       if (resolution.type === 'real') {
-        hbsModule = resolution.filename;
+        hbsModule = candidateSpecifier;
```
ef4, line 467: This is similar to what we did in https://github.com/embroider-build/embroider/pull/1648. Both are cases where we were rewriting requests to absolute paths, which ends up looking to vite like the user actually typed an absolute path in their import, which gets different behavior than a normal relative- or package-name import.

In this case when we successfully resolve a specifier to a component file to check that it exists, rather than rewriting the request to point directly at that file we rewrite the request to use the specifier that we just successfully resolved.

(This duplicate resolving is only needed when supporting the ambiguity of loose-mode templates. As people migrate to template tag they stop needing it.)

ef4 (approved):

PR #1719, opened 2023-12-12
Only apply ember composable helpers compat adapter to before v5

NullVoxPopuli:
VERY WIP: I don't know what the solution is yet -- I also haven't been able to reproduce the problem I'm having in a big app. 

<details><summary>old info</summary>

Local testing shows that with `ember-composable-helpers`
- `>= 2.3.0`:
  - before this change, infinite rebuild loop occurs due to compat-adapter writing to a watched directory:
    https://github.com/embroider-build/embroider/blob/efd91174e7c27cf5365d54567b3b998f65a4b9d4/packages/compat/src/compat-adapters/ember-composable-helpers.ts#L36
  - after this change, build is stable
- `<= 2.2.0`:
  - same as >= 2.3. I wonder if an embroider change _enabled_ dependency watching.
  
  
</details>


NullVoxPopuli:
Gonna close this for now, as I don't have any solid leads, and it seems to only be affecting an internal set of projects.

PR #1720, opened 2023-12-13
add test scenarios for beta

mansona:
In https://github.com/embroider-build/embroider/pull/1717 we mention that we should have some scenarios for beta, this PR is the most naive way of adding those 😂 

If we're happy with the time hit on CI then we should consider just merging this 👍


mansona:
As it turns out we can't merge this because we've actually hit the limit of tests on github 🙃 

> Strategy expansion exceeded 256 results for job 'test'

ef4:
There's a lot of stuff that could get pruned back. Many of the tests don't really need to run in every supported ember version. It would be enough to pick a few of the end-to-end tests. Same goes for duplicating tests per OS.

PR #1724, opened 2023-12-13
fix publish for stable branch

mansona:
we [needed to update relase-plan](https://github.com/embroider-build/release-plan/pull/38) to add an option `--publish-branch` because by default pnpm only allows you to run `pnpm publish` on [main or master](https://pnpm.io/cli/publish#--publish-branch-branch). We are running it on `stable` and it was hanging in CI because it was prompting to know if you wanted to continue.

This PR prevents that from happening again 👍

### DIFF
--- a/.github/workflows/publish-stable.yml
+++ b/.github/workflows/publish-stable.yml
@@ -49,7 +49,7 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
 
       - name: npm publish
-        run: pnpm release-plan publish
+        run: pnpm release-plan publish --publish-branch=stable
 
         env:
           GITHUB_AUTH: ${{ secrets.GITHUB_TOKEN }}
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "eslint-plugin-prettier": "^4.2.1",
     "jest": "^29.2.1",
     "prettier": "^2.3.1",
-    "release-plan": "^0.5.1",
+    "release-plan": "^0.6.0",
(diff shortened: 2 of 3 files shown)


.github/workflows/publish-stable.yml
```diff
@@ -49,7 +49,7 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
 
       - name: npm publish
-        run: pnpm release-plan publish
+        run: pnpm release-plan publish --publish-branch=stable
```
NullVoxPopuli, line 52: don' we want `--publish-branch=stable|main`, since we merge `stable` into `main` frequently?

mansona: once we sync this it won't be a problem any more. For example https://github.com/embroider-build/embroider/pull/1721 didn't even try to update the yaml branches

also it's better to be explicit with this. 

and also we didn't test if that syntax works (if the pnpm docs says it does that's fine but I'm extra cautious around releasing things). We know what this PR has does work because I've done it manually twice.

So in summary, no i don't think we should change this 😂

NullVoxPopuli: ah ok, I see [what you mean](https://github.com/embroider-build/embroider/pull/1724/files/fbffd7a761762f6da8683dfb235e2acac457d7ee#diff-ee4f3ba7c21547934a6f659fce2ab1c2a42251b904c6844fad629c9f28a991aeR11)

Legit

NullVoxPopuli (approved):

PR #1727, opened 2023-12-14
use meta tag to set global EmberENV

patricklx:
have a single source for ember env.
same as https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/app-config-from-meta.js

it could also help with rebuilds, as an edit to the config only needs to rebuild the html file

### DIFF
--- a/packages/compat/src/compat-app-builder.ts
+++ b/packages/compat/src/compat-app-builder.ts
@@ -359,7 +359,7 @@ export class CompatAppBuilder {
               return e
           }
           var metaName = '${appName}' + '/config/environment';
-          var metaTag = typeof document !== 'undefined && document.querySelector('meta[name="' + metaName + '"]');
+          var metaTag = typeof document !== 'undefined' && document.querySelector('meta[name="' + metaName + '"]');
           if (metaTag) {
             try {
               var rawConfig = document.querySelector('meta[name="' + metaName + '"]').getAttribute('content');
--- a/packages/compat/src/compat-app-builder.ts
+++ b/packages/compat/src/compat-app-builder.ts
@@ -322,7 +322,11 @@ export class CompatAppBuilder {
     return extensionsPattern(this.resolvableExtensions());
   }
 
-  private impliedAssets(type: keyof ImplicitAssetPaths, engine: AppFiles): (OnDiskAsset | InMemoryAsset)[] {
+  private impliedAssets(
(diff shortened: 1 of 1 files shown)


packages/compat/src/compat-app-builder.ts
```diff
@@ -350,7 +351,24 @@ export class CompatAppBuilder {
...
-        source: `window.EmberENV={ ...(window.EmberENV || {}), ...${JSON.stringify(emberENV, null, 2)} };`,
+        source: `
+        (function() {
+          function merge(e, t) {
+              for (var r in t)
+                  e[r] = t[r]
+              return e
+          }
+          try {
+            var metaName = '${appName}' + '/config/environment';
```
void-mAlex, line 362: doesn't this include module prefix as well if it's declared?

packages/compat/src/compat-app-builder.ts
```diff
@@ -350,7 +351,24 @@ export class CompatAppBuilder {
...
+            var rawConfig = document.querySelector('meta[name="' + metaName + '"]').getAttribute('content');
+            var config = JSON.parse(decodeURIComponent(rawConfig));
+            window.EmberENV = merge(window.EmberENV || {}, config.EmberENV);
```
void-mAlex, line 365: need to make sure that this propagates variables already existing on EmberENV especially in the catch clause for ember Inspector

we most likely want to leave EmberENV intact instead of overriding it in case it errors

patricklx:
i wasn't aware that there is an option to disable the meta generation. in that case this does not make sense

void-mAlex:
I would say parts of this pr would still be very much desired, like the little merge function;

you can still be trolled by addons that using ember-get-config - can't remember exactly the case but I know that I had to stick a meta tag into the index html even though I set the config to not ever try to use it
that said it was on a much older version of things so it may have been fixed

void-mAlex:
stumbled upon the place that has the problematic config code requiring a valid meta tag and it's in engines
https://github.com/ember-engines/ember-engines/blob/master/packages/ember-engines/lib/engine-config-from-meta.js#L9
the fact that it throws if it's not there

patricklx:
if i build with `ember build` and open url `/testing` things will not work because the config uses the generated config in `node_modules/.embroider/rewritten-app/config/environment.js`

But the meta tag actually has the correct config, which would allow one to switch between app and testing and having correct environment

PR #1728, opened 2023-12-14
[vite] fix for duplicate modules with dep optimizer

patricklx:
Because we do imports from virtual modules, some imports gets rewritten to absolute paths. Those will then be ignored by depa optimizer, since the absolute path contains node_modules
https://github.com/vitejs/vite/issues/13538

https://github.com/vitejs/vite/blob/711dd807610b39538e9955970145d52e4ca1d8c0/packages/vite/src/node/plugins/resolve.ts#LL832C5-L832C44

### DIFF
--- a/packages/vite/src/esbuild-resolver.ts
+++ b/packages/vite/src/esbuild-resolver.ts
@@ -18,6 +18,8 @@ export function esBuildResolver(root = process.cwd()): EsBuildPlugin {
   let resolverLoader = new ResolverLoader(process.cwd());
   let macrosConfig: PluginItem | undefined;
   let preprocessor = new Preprocessor();
+  let workingDir = locateEmbroiderWorkingDir(root);
+  let rewrittenApp = join(workingDir, 'rewritten-app');
 
   return {
     name: 'embroider-esbuild-resolver',
@@ -28,6 +30,7 @@ export function esBuildResolver(root = process.cwd()): EsBuildPlugin {
           return null;
         }
         let resolution = await resolverLoader.resolver.resolve(request, defaultResolve(build, kind));
+
         switch (resolution.type) {
           case 'found':
             return resolution.result;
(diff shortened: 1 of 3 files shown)


patricklx:
From referenced issue:

> virtual module cannot use bare imports (e.g. import "dep1/init1.js") because transitive dependencies may not be reachable from the virtual modules "location" (nested node_modules etc. should be taken into account). This is why the plugin resolves the module ids passing some importer from where the import is possible. This results in an absolute path like /@fs/.../dep1/init1.js?v=d7007810, which will then be ignored by the depsOptimizer.

> The problem here arises when the unoptimized modules mentioned above use node modules that are also used by the application directly (and are thus optimized in node_modules/.vite/...). There will be two copies of any modules used in this way: raw files in node_modules and a copy bundled by esbuild in .vite, with different parts of the application observing one or the other. In the reproduction repository, the module shared/inner is duplicated.

mansona:
so that issue that you're quoting does explain the issue quite well but our architecture is (currently) taking a slightly different approach in trying to deal with this. We are trying to build with vite and issolate all places that drop down to `@fs` imports and fix them, essentially teaching esbuild how to load them properly.

Does that make sense?

patricklx:
Probably, depends how to fix them.
(What i did before only broke the dep optimizer, which is why it then loaded...)
What I'm trying now is to resolve everything in rewritten-app to a relative path.
I had some success, but it was failing on ember-data. ember-data has some circular dependencies that it solves with inline require calls. But when optimized its all eager imports at the top...

patricklx:
found a fix:
https://esbuild.github.io/plugins/#on-load-results:~:text=built%2Din%20loaders.-,resolveDir,This%20is%20the%20file%20system%20directory%20to%20use%20when%20resolving,-an%20import%20path

packages/shared-internals/src/package-cache.ts
```diff
@@ -67,7 +67,7 @@ export default class PackageCache {
   }
 
   ownerOfFile(filename: string): Package | undefined {
-    let segments = filename.split(sep);
+    let segments = filename.replace(/\\/g, '/').split(posix.sep);
```
patricklx, line 70: required fix for windows system

patricklx:
fixed in vite 5

PR #1739, opened 2023-12-21
support ts, gts and gjs in vite

patricklx:
this adds support for loading ts & gts.

it also adds the gts/gjs extensions to compat-app-builder so that they can be used in tests


packages/core/src/module-resolver.ts
```diff
@@ -542,12 +542,22 @@ export class Resolver {
   }
 
   private *componentJSCandidates(inPackageName: string) {
+    const extensions = ['.ts', '.gjs', '.gts'];
     yield { prefix: '/components/', suffix: '' };
     yield { prefix: '/components/', suffix: '/component' };
 
+    for (const ext of extensions) {
+      yield { prefix: '/components/', suffix: ext };
+      yield { prefix: '/components/', suffix: `index${ext}` };
```
NullVoxPopuli, line 551:
```suggestion
      yield { prefix: '/components/', suffix: `/index${ext}` };
```
if component has a leading slash, should index as well?

tests/vite-app/tests/integration/fany2-test.js
```diff
@@ -0,0 +1,19 @@
...
+
+module('Integration | Component | Fany2', (hooks) => {
+  setupRenderingTest(hooks);
+
+  test('should have Yay for gjs!', async function (assert) {
+    await render(hbs`
+    <Fancy @type="primary2"></Fancy>
+    <Fancy2 @type="primary2"></Fancy2>
+    `);
+    await rerender();
```
NullVoxPopuli, line 14: Non blocking, but rerender isn't needed here

NullVoxPopuli (approved):
Looks good to me!

PR #1743, opened 2023-12-22
Don't rely on htmlbars internals

ef4:
I'm working on a cleanup in ember-cli-htmlbars that is blocked on embroider touching some internals.


packages/compat/src/prepare-htmlbars-ast-plugins.ts
```diff
@@ -1,32 +1,7 @@
...
-
-      // if the cacheKey is set on the wrapper, but not on the plugin, add it
-      if (cacheKey && !plugin.cacheKey) {
-        plugin.cacheKey = cacheKey;
-      }
-    }
-  }
+export default function loadAstPlugins(registry: any): Transform[] {
+  let plugins = registry.load('htmlbars-ast-plugin').map((wrapper: any) => wrapper.plugin);
+  plugins.reverse();
```
NullVoxPopuli, line 5: why's the reverse needed?

ef4: The plugins in the registry have historically always run in the backwards order for dumb reasons. *sigh*.

I should make sure the comment doesn't get dropped by this PR though, if you look through the diff you can see the spot where it's getting removed.

PR #1744, opened 2023-12-22
control ember build through vite

patricklx:
and also force optimizer rebundle when lock file changed.
this also solves an issue where the watcher did not detect changes if embroider build --watch was called after starting vite

### DIFF
--- a/packages/vite/src/build.ts
+++ b/packages/vite/src/build.ts
@@ -1,0 +1,40 @@
+import { fork } from 'child_process';
+import type { Plugin } from 'vite';
+
+export function emberBuild(mode: string): Promise<void> {
+  if (mode === 'build') {
+    return new Promise((resolve, reject) => {
+      const child = fork('./node_modules/ember-cli/bin/ember', ['build', '--production'], { silent: true });
+      child.on('exit', code => (code === 0 ? resolve() : reject()));
+    });
+  }
+  return new Promise((resolve, reject) => {
+    const child = fork('./node_modules/ember-cli/bin/ember', ['build', '--watch'], { silent: true });
+    child.on('exit', code => (code === 0 ? resolve() : reject(new Error('ember build --watch failed'))));
+    child.on('spawn', () => {
+      child.stderr?.on('data', data => {
+        console.error(data.toString());
+      });
+      child.stdout!.on('data', data => {
+        console.log(data.toString());
+        if (data.toString().includes('Build successful')) {
+          resolve();
+        }
+      });
+    });
+  });
+}
+
+export function compatPrebuild(): Plugin {
+  let mode = 'build';
+  return {
+    name: 'embroider-builder',
+    enforce: 'pre',
+    configureServer() {
+      mode = 'development';
+    },
+    async buildStart() {
+      await emberBuild(mode);
+    },
+  };
+}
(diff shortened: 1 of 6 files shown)


tests/vite-app/vite.config.mjs
```diff
@@ -15,11 +16,13 @@ export default defineConfig({
   root,
   // esbuild in vite does not support decorators
   esbuild: false,
+  cacheDir: resolve("node_modules", ".vite"),
```
mansona, line 19: do we actually need this change? or is this just a "cosmetic" thing?

packages/vite/src/build.ts
```diff
@@ -0,0 +1,99 @@
...
+        console.log(data.toString());
+        if (data.toString().includes('Build successful')) {
+          resolve();
+        }
+      });
+    });
+  });
+}
+
+export async function needForcedRebundle(path: string | null | undefined): Promise<boolean> {
```
mansona, line 64: can we not add this? is this working around a bug (possibly something not being watched correctly in broccoli?) or is this actually needed for a specific reason?

tests/vite-app/vite.config.mjs
```diff
@@ -5,6 +5,7 @@ import {
   scripts,
   templateTag,
   optimizeDeps,
+  build,
```
mansona, line 8: let's call this `emberAddonPrebuild` or `compatPrebuild`. using `build` here is too generic

mansona (changes requested):
We took a look at this during the Embroider meeting and there are a few things that we would like to see changed before this is mergable, but overall we're super excited about not needing two independent scripts running 🎉 

Let me know if my comments don't make any sense, and we can open a thread in Discord to discuss 👍

tests/vite-app/vite.config.mjs
patricklx: Yes, it's required, since ut defaults to . embroider directory, which does not exist initially

packages/vite/src/build.ts
patricklx: Its required if deps changed. Otherwise you will get issues with outated deps, embroider will import addons directly instead of rewritten packages

patricklx: The outdated dep is common. And some who have experience with vite already now that they need to start vite with --force. But there are akso other issues with the embroider rewritten packages if they are outdated

mansona: > Its required if deps changed

But ember-cli has never supported changing deps during a running server 🤔 this seems like a lot of complexity to add for a new functionality like this

tests/vite-app/vite.config.mjs
mansona: Yea that's a good point, I hadn't considered that 🙃 we can leave this one in then 👍

packages/vite/src/build.ts
patricklx: well, it can be done in another pr if the need arises. removed

NullVoxPopuli: Counterpoint, everyone i've talked to outside of (and new to) the ember community expects this behavior

mansona (approved):
I tested this locally and it works a treat! We even had the opportunity to have the Ember prebuild fail, and it handled that perfectly too 🎉

PR #1751, opened 2024-01-10
Cache `resolveComponent`-lookups

LevelbossMike:
To speed up `resolveComponent` we can cache the lookups instead using node resolution every time we use a component.

As discussed with @mansona , this speeds up builds significantly. I'm not 💯 if the cache-key is the best but it works in our fine in our rather biggish application.


packages/core/src/module-resolver.ts
```diff
@@ -485,31 +489,42 @@ export class Resolver {
...
+        // .hbs is a resolvable extension for us, so we need to exclude it here.
+        // It matches as a priority lower than .js, so finding an .hbs means
+        // there's definitely not a .js.
+        if (resolution.type === 'real' && !resolution.filename.endsWith('.hbs')) {
+          jsModule = resolution.filename;
+          break;
+        }
       }
+
+      this.componentCache.set(`${target.from}:${target.packageName}:${target.memberName}`, { hbsModule, jsModule });
```
NullVoxPopuli, line 527:
```suggestion
      this.componentCache.set(cacheKey, { hbsModule, jsModule });
```

to help prevent typos

NullVoxPopuli (dismissed):

packages/core/src/module-resolver.ts
```diff
@@ -159,7 +159,11 @@ export type SyncResolverFunction<R extends ModuleRequest = ModuleRequest, Res ex
 ) => Res;
 
 export class Resolver {
-  constructor(readonly options: Options) {}
+  componentCache: Map<String, any>;
```
simonihmig, line 162: Why `any`? 

Wouldn't that work? 👇

```suggestion
  componentCache: Map<String, { hbsModule?: string, jsModule?: string }>;
```

simonihmig (commented):
Those numbers look amazing! 😍
Thank you!

LevelbossMike:
@simonihmig done

ef4 (changes requested):
It's good that this identifies such an expensive hot path. But we can't land this without some cache invalidation strategy. Otherwise people will get permanently broken builds when they delete a file that is in the cache.

PR #1756, opened 2024-01-16
Allow route templates to be used in v2 addons

balinterdi:
When compiling `hbs` files, the `rollup-hbs-plugin` in `@embroider/addon-dev` creates a template-only module definition (it “synthesises a module”) when a JS module is attempted to be resolved and no module exists yet:

https://github.com/embroider-build/embroider/blob/289f83d80d133da3de2b342e63f937986248c07d/packages/addon-dev/src/rollup-hbs-plugin.ts#L79-L88

This makes a lot of sense if the only kind of `hbs` file we can encounter is a component template.

Unfortunately, this rules out using route templates which are also `hbs` files: they don't have a corresponding `.js` file and so will be considered a template-only component making them unusable.

I believe that on the long term, the current route templates where the corresponding controller is considered the backing JS class and the route provides the `@model` property will give way to something else (something like [ember-route-template](https://github.com/discourse/ember-route-template)) but that's still some way off and it'd be nice to have an interim solution.

----

I'm aware the PR as it is currently is not mergeable: the assumption that if the import path (the source) has `templates` in the name is too rigid. The PR aims to serve more as a conversation starter as we need this feature in a project and I'd like to explore the correct way to do this.

And this is just one piece of the puzzle: it'd need a "rollup-ember-route-template" Rollup plugin which looks like this:

```js
import { readFileSync } from 'fs';
import { hbsToJS } from '@embroider/core';

function getMeta(context, id) {
  const meta =
    context.getModuleInfo(id)?.meta?.['rollup-ember-route-templates-plugin'];
  if (meta) {
    return meta;
  } else {
    return null;
  }
}

export default function routeTemplates() {
  return {
    name: 'rollup-ember-route-templates-plugin',
    load(id) {
      if (getMeta(this, id)) {
        const hbsFile = id.replace(/\.js$/, '.hbs');
        let input = readFileSync(hbsFile, 'utf8');
        let code = hbsToJS(input);
        return {
          code,
        };
      }
    },
  };
}
```

With this change, Embroider emits a file like this for a route template:

```js
import { precompileTemplate } from '@ember/template-compilation';

var TEMPLATE = precompileTemplate("The Caro-Kann Opening\n\n<div class=\"image-container\">\n  <img src=\"/grand-prix/assets/images/caro-kann.webp\" alt=\"Caro Kann Opening\" />\n</div>");

var caroKann = setComponentTemplate(TEMPLATE, precompileTemplate("The Caro-Kann Opening\n\n<div class=\"image-container\">\n  <img src=\"/grand-prix/assets/images/caro-kann.webp\" alt=\"Caro Kann Opening\" />\n</div>"));

export { caroKann as default };
//# sourceMappingURL=caro-kann.js.map
```

You'll notice that a call a`setComponentTemplate` call is inserted by [the Babel template-colocation-plugin](https://github.com/embroider-build/embroider/blob/main/packages/shared-internals/src/template-colocation-plugin.ts) which is incorrect. I'm not exactly sure what should be exported for route templates (maybe just the result of `precompileTemplate`?) but for some reason, the above, incorrect build output works fine 🤷🏼‍♂️ so finding out how to make it right is a next, bonus step :)

### DIFF
--- a/packages/addon-dev/src/rollup-hbs-plugin.ts
+++ b/packages/addon-dev/src/rollup-hbs-plugin.ts
@@ -76,6 +76,18 @@ async function maybeSynthesizeComponentJS(
   if (!templateResolution) {
     return null;
   }
+  // A rather primivite approach that will probably need to be refined.
+  // If an `hbs` file has `templates` in its path, it's considered a route template.
+  if (source.includes('templates')) {
+    return {
+      id: templateResolution.id.replace(/\.hbs$/, '.js'),
+      meta: {
+        'rollup-ember-route-templates-plugin': {
+          type: 'route-template',
+        },
+      },
+    };
+  }
   // we're trying to resolve a JS module but only the corresponding HBS
   // file exists. Synthesize the template-only component JS.
   return {


void-mAlex:
hello @balinterdi 
I can confirm that something like ember-route-template works today as an alternative to the changes you wrote out here
you would need to add `'routes/**/*.js'` and `'templates/**/*.js'` to your v2 addons rollup config and use RouteTemplate in a `src/routes/my-route.gjs` file as described in their readme
that way you can use route templates from your v2 addon

balinterdi:
@void-mAlex Hey Alex,

> I can confirm that something like ember-route-template works today as an alternative to the changes you wrote out here
you would need to add 'routes/**/*.js' and 'templates/**/*.js' to your v2 addons rollup config and use RouteTemplate in a src/routes/my-route.gjs file as described in their readme
that way you can use route templates from your v2 addon

Thank you.

Indeed, and you also have to add `addon.gjs()` to your rollup config to make that work. I have a working example [here](https://github.com/balinterdi/app-files-from-v2-addons/tree/use-ember-route-template-to-make-route-template-available) for people interested.

In my opinion, this is a very clean and nice solution. However, it implies a lot of work as you have to convert the route template to a format expected by `ember-route-template`, make it a gjs, and add all the explicit imports that strict mode templates require (which has still further ramifications that I can elaborate on, if needed, mostly stemming from the fact that v2 addons can't import from the host app).

When migrating >100 route templates (as in the project I'm working on) this is just too much work up front. That's why I think a temporary, uglier, interim solution might be needed to help folks migrate before the clean & nice solution is established.

@NullVoxPopuli I'm not very intimate with the Embroider roadmap and decisions but to me, using `ember-route-template` seems a good choice as the long-term solution for this problem (see above for more details)

void-mAlex:
> 
> Indeed, and you also have to add `addon.gjs()` to your rollup config to make that work. I have a working example [here](https://github.com/balinterdi/app-files-from-v2-addons/tree/use-ember-route-template-to-make-route-template-available) for people interested.
> 

I would recommend updating to the latest blueprint as that should be there already for you see
https://github.com/embroider-build/addon-blueprint/blob/main/files/__addonLocation__/rollup.config.mjs#L55
which you can do using ember-cli-update

closing this pr as we have a workable solution while we work on further improvements

thank you for raising this

ef4:
I do think it's reasonable to add optional arguments [here](https://github.com/embroider-build/embroider/blob/289f83d80d133da3de2b342e63f937986248c07d/packages/addon-dev/src/rollup.ts#L50) that would let people provide globs for including or excluding files from being treated as template-only components.

balinterdi:
@ef4 Can I take this to mean that you'd accept PRs that would implement that including/excluding glob param? ;)

PR #1763, opened 2024-01-18
Add new watches API

ef4:
~This is just a toy example to motivate a watch api for virtual content.~ Edit: this has been trimmed down to only include the API we want

Along the way we discovered that vite incorectly implements rollups `addWatchFile` API because (1) it's supposed to accept directories and doesn't and (2) it's supposed to resolve relative paths relative to the current working directory, not the current request.

This example also needs https://github.com/embroider-build/embroider/pull/1762

### DIFF
--- a/packages/core/src/virtual-content.ts
+++ b/packages/core/src/virtual-content.ts
@@ -10,29 +10,29 @@ const externalCJSPrefix = '/@embroider/ext-cjs/';
 // this produces the corresponding contents. It's a static, stateless function
 // because we recognize that that process that did resolution might not be the
 // same one that loads the content.
-export function virtualContent(filename: string, resolver: Resolver): string {
+export function virtualContent(filename: string, resolver: Resolver): { src: string; watches: string[] } {
   let cjsExtern = decodeVirtualExternalCJSModule(filename);
   if (cjsExtern) {
-    return renderCJSExternalShim(cjsExtern);
+    return { src: renderCJSExternalShim(cjsExtern), watches: [] };
   }
 
   let extern = decodeVirtualExternalESModule(filename);
   if (extern) {
-    return renderESExternalShim(extern);
+    return { src: renderESExternalShim(extern), watches: [] };
   }
   let match = decodeVirtualPairComponent(filename);
   if (match) {
-    return pairedComponentShim(match);
+    return { src: pairedComponentShim(match), watches: [] };
   }
 
   let fb = decodeFastbootSwitch(filename);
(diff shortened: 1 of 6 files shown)


packages/core/src/virtual-content.ts
```diff
@@ -10,29 +11,40 @@ const externalCJSPrefix = '/@embroider/ext-cjs/';
...
 // because we recognize that that process that did resolution might not be the
 // same one that loads the content.
-export function virtualContent(filename: string, resolver: Resolver): string {
+export function virtualContent(filename: string, resolver: Resolver): { src: string; watches: string[] } {
+  if (filename === '/the-file-list') {
+    let dir = join(resolver.options.appRoot, 'app');
+    let files = readdirSync(dir);
+    return {
+      src: `export default [
+      ${files.map(f => `"${f}"`).join(',')}
```
patricklx, line 20: this specific case could probably also be solved by using vite `import.meta.glob('./dir/*')` . would it sill not do auto reload in that case?
other cases could solve it by creating imports to files, vite also supports `?raw` which would allow to include any file that needs to be in a virtual file.

packages/vite/src/resolver.ts
```diff
@@ -29,7 +50,10 @@ export function resolver(): Plugin {
     },
     load(id) {
       if (id.startsWith(virtualPrefix)) {
-        return virtualContent(id.slice(virtualPrefix.length), resolverLoader.resolver);
+        let { src, watches } = virtualContent(id.slice(virtualPrefix.length), resolverLoader.resolver);
+        virtualDeps.set(id, watches);
```
patricklx, line 54: prefix here the src with `import.meta.glob(...watches)`. and filter watches for the ones containing globs. 
files should just work with addWatchFile.
like that you do not need the `configureServer` code

ef4: Our actual use cases need more sophisticated filtering than we can do with import.meta.glob. For example, we need a way to include "node_modules/some-addon/_app_/components/foo.js" but only if "./components/foo.js" doesn't exist. If we do that with globs you will always evaluate the addon copy, even when it should not evaluate.

patricklx: mport.meta.glob is not eager, but can be configured to be so with params, so it shouldn't evaluate. but not sure if thats enough for the use case you are looking for

ef4: Right, in the case of entrypoints we need the `eager: true` variant.

PR #1766, opened 2024-01-23
Provide Glint utility plugin for `@embroider/addon-dev`

NullVoxPopuli:
Provides "a" solution to:
- https://github.com/embroider-build/addon-blueprint/issues/248
- https://github.com/embroider-build/addon-blueprint/issues/215

### DIFF
--- a/packages/addon-dev/src/rollup-glint-plugin.ts
+++ b/packages/addon-dev/src/rollup-glint-plugin.ts
@@ -1,0 +1,25 @@
+import type { Plugin } from 'rollup';
+import { execaCommand } from 'execa';
+import { fixBadDeclarationOutput } from 'fix-bad-declaration-output';
+
+export default function rollupGlintPlugin(pattern: string): Plugin {
+  return {
+    name: 'rollup-glint-plugin',
+
+    async closeBundle() {
+      /**
+       * Generate the types (these include /// <reference types="ember-source/types"
+       * but our consumers may not be using those, or have a new enough ember-source that provides them.
+       */
+      await execaCommand(`pnpm glint --declaration`, { stdio: 'inherit' });
+      /**
+       * https://github.com/microsoft/TypeScript/issues/56571#
+       * README: https://github.com/NullVoxPopuli/fix-bad-declaration-output
+       */
+      await fixBadDeclarationOutput(pattern || 'declarations/**/*.d.ts', [
+        'TypeScript#56571',
+        'Glint#628',
+      ]);
+    },
+  };
+}
--- a/packages/addon-dev/package.json
+++ b/packages/addon-dev/package.json
@@ -29,6 +29,8 @@
     "@embroider/core": "workspace:^",
     "@rollup/pluginutils": "^4.1.1",
     "content-tag": "^1.1.2",
+    "execa": "^8.0.1",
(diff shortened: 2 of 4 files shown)


packages/addon-dev/src/rollup-glint-plugin.ts
```diff
@@ -0,0 +1,25 @@
...
+export default function rollupGlintPlugin(pattern: string): Plugin {
+  return {
+    name: 'rollup-glint-plugin',
+
+    async closeBundle() {
+      /**
+       * Generate the types (these include /// <reference types="ember-source/types"
+       * but our consumers may not be using those, or have a new enough ember-source that provides them.
+       */
+      await execaCommand(`pnpm glint --declaration`, { stdio: 'inherit' });
```
SergeAstapov, line 14: `glint` does not seem to be listed in `dependencies`?

NullVoxPopuli: no need, this command would run from the cwd, so at best we could declare an optional peer dependency on `@glint/core`

simonihmig, line 14: Why `pnpm`? execa should be able to find the `glint` executable using `preferLocal: true`

simonihmig (commented):
> Looks like this is blocked on @embroider/addon-dev needing to be ESM.

I think you could use dynamic import inside `closeBundle` to not be blocked on esm conversion:
```
const { execaCommand } = await import('execa');
```

NullVoxPopuli:
> I think you could use dynamic import inside closeBundle to not be blocked on esm conversion:

oh! that's a good idea! I always forget that's a thing. been so long since I've actually had to worry about cjs :sweat_smile:

NullVoxPopuli:
Nevermind -- our build setup is removing all imports...........
image

packages/addon-dev/src/rollup-glint-plugin.ts
NullVoxPopuli: thanks!

NullVoxPopuli:
Been trying a few different compilation techniques, and it seems that there is a lot of:
```
[!] SyntaxError: Named export 'readJsonSync' not found. The requested module 'fs-extra' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'fs-extra';
const { readJsonSync, writeJsonSync } = pkg;
```

style of errors, including from `@embroider/core`

packages/addon-dev/src/rollup-glint-plugin.ts
NullVoxPopuli, line 1: At this point, this PR is converting to proper type=module and this one file _happens to be here_.

Should probably remove it and introduce it later.

Though, since it's only ~ 28 lines that'd be removed from this PR, if tests go green, I think folks could start reviewing.

packages/addon-dev/sample-babel.config.js
NullVoxPopuli, line 1: js means "ESM" now.

Do we need this file?

NullVoxPopuli:
woah, it's green

NullVoxPopuli:
addon-dev is still not ESM, but I don't have any motivation to continue this work -- for my newer libraries, I'm also not using addon-dev anymore

PR #1768, opened 2024-01-24
docs(porting addons to v2): change the recommended package manager to pnpm

BlueCutOfficial:
This PR is a proposal to change the recommended package manager to pnpm in the [`Porting Addons to V2`](https://github.com/embroider-build/embroider/blob/main/docs/porting-addons-to-v2.md) guide.

My understanding is that pnpm is better than yarn to work with Embroider in general, so it may be preferable to recommend pnpm in this document rather than making it more generic about the package manager part.

### DIFF
--- a/docs/porting-addons-to-v2.md
+++ b/docs/porting-addons-to-v2.md
@@ -16,7 +16,7 @@ If your addon is a mix of both build-time and run-time features, consider replac
 
 Traditionally, an Ember addon is a single NPM package that combines both the actual addon code _and_ a "dummy" app for hosting tests and docs. This was [problematic for several reasons](https://github.com/ember-cli/rfcs/issues/119). V2 addons instead require clean separation between addon and app, so you're going to be working with more than one distinct NPM package: one for the addon, one for the test-app, and optionally one for the documentation site.
 
-Our recommended way to manage these multiple packages is using a monorepo, via Yarn, NPM, or PNPM Workspaces. The example in this guide assumes a Yarn Workspaces monorepo.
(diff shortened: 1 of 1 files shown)


NullVoxPopuli (approved):
Thanks! Looks great!

sandstrom:
I'm a bit late to the party here, but I would recommend this approach:

> I think Ember should pick the package manager that most people are familiar with as its default. **Currently, that is NPM.** Yes, there are better alternatives (which one is better depend on the project/team, for some the "best manager" is Yarn, for others it is pnpm).
> 
> But NPM, with all its flaws, is still the de-facto standard NPM package manager. To keep friction down, I think that should be the Embroider default.
> 
> This doesn't preclude us from making it easy to use other package managers, or even offer it as an command line option `--package-manager=pnpm`.
> 
> *Also, for everyone discussing this. Keep in mind that this choice shouldn't be optimized on what you (all tech savvy people with tons of JS/front-end experience) knows is "the best choice". Instead, it should be optimized on what makes Ember easy to get started with for the average developer. For all of you, it'll be a breeze to pick another package manager.*
> 
> -- https://github.com/emberjs/rfcs/issues/820#issuecomment-1190657424

mansona:
thanks @sandstrom for the link. That concern isn't entirely relevant for this PR because this is just for the internal documentation that is being used to guide people through the process of upgrading to a v2 addon. When we open this process to a RFC (which will happen once we are ready to make it the default) then we can start sweating the details of our choices 👍

PR #1771, opened 2024-01-25
Simplify vite-app template

ef4:
Move ember-data into the specific test scenario thats wants it, so that the base vite-app template does't need to have ember-data.


mansona (approved):
Looks good 🎉

PR #1780, opened 2024-02-01
vite: fix initializers not running

patricklx:
loadInitializers was called before initalizers where registered via define.
So initializers where not able to run


patricklx:
fixed by https://github.com/embroider-build/embroider/pull/2006

PR #1786, opened 2024-02-01
Bump content-tag and add inline_source_map option for rollup and vite

vstefanovic97:
Backport https://github.com/embroider-build/embroider/pull/1784 to `stable`

PR #1787, opened 2024-02-02
Fix source map option not being passed to plugin

vstefanovic97:
Sorry about that, checked and now it works 100%
Missed it in https://github.com/embroider-build/embroider/pull/1786 thought plugin was just getting re-exported, my mistake

### DIFF
--- a/packages/addon-dev/src/rollup.ts
+++ b/packages/addon-dev/src/rollup.ts
@@ -57,8 +57,8 @@ export class Addon {
     return hbs();
   }
 
-  gjs() {
-    return gjs();
+  gjs(options?: { inline_source_map: boolean }) {
+    return gjs(options);
   }
 
   // By default rollup does not clear the output directory between builds. This
--- a/packages/addon-dev/src/rollup-gjs-plugin.ts
+++ b/packages/addon-dev/src/rollup-gjs-plugin.ts
@@ -9,7 +9,7 @@ const processor = new Preprocessor();
 // import { parse as pathParse } from 'path';
 
 export default function rollupGjsPlugin(
-  { inline_source_map } = { inline_source_map: false }
+  { inline_source_map } = { inline_source_map: true }
 ): Plugin {
   return {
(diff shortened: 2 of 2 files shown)


packages/addon-dev/src/rollup.ts
```diff
@@ -57,8 +57,8 @@ export class Addon {
     return hbs();
   }
 
-  gjs() {
-    return gjs();
+  gjs(options: { inline_source_map: true }) {
```
SergeAstapov, line 60: @vstefanovic97 shouldn’t we mark this as optional?
```suggestion
  gjs(options?: { inline_source_map: true }) {
```

vstefanovic97: 🤦‍♂️ , I'll fix it right away

packages/addon-dev/src/rollup.ts
```diff
@@ -57,8 +57,8 @@ export class Addon {
     return hbs();
   }
 
-  gjs() {
-    return gjs();
+  gjs(options?: { inline_source_map: true }) {
+    return gjs(options);
```
NullVoxPopuli, line 61: Is there any downside to defaulting this to true?

We already have rollup's sourcemaps enabled

vstefanovic97: hmm well probably not, I think I can default it to true in this case if you want @NullVoxPopuli

vstefanovic97: I changed the default to true @NullVoxPopuli

NullVoxPopuli (approved):

packages/addon-dev/src/rollup.ts
NullVoxPopuli: I mean, in your defense, we don't have sufficient testing if for addon-dev in this repo

vstefanovic97:
@NullVoxPopuli can we merge and release patch ?

PR #1788, opened 2024-02-02
Fix source map option not being passed to plugin

vstefanovic97:
Sorry about that, checked and now it works 100%


mansona:
I'm going to close this PR, the pattern that we have been following is to make sure that any stable changes first go into stable (i.e. we merge https://github.com/embroider-build/embroider/pull/1787) and then we merge stable into main 👍

PR #1790, opened 2024-02-03
Update @embroider/macros README.md real-world examples

machty:
Remove/replace real-world examples of `@embroider/macros` that no longer use `@embroider/macros`.


NullVoxPopuli (approved):
Thank you!

PR #1793, opened 2024-02-08
refactor: change isProduction check strategy

colenso:
In this PR I aim to essentially change the way Embroider checks whether we're building for `production`. In an `EmberApp` there's a `isProduction` which basically does a `environment === 'production'`. This shouldn't affect anything really.
It's just that we aim to override this `isProduction` in our app because we have a different notion of `isProduction`. We have environment === 'staging', environment === 'native-app' for which we still want Embroider to run it's production optimization logic.

### DIFF
--- a/packages/compat/src/compat-app.ts
+++ b/packages/compat/src/compat-app.ts
@@ -65,6 +65,10 @@ export default class CompatApp {
     return this.legacyEmberAppInstance.env;
   }
 
+  get isProduction(): boolean {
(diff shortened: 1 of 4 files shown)


colenso:
@mansona Can you please have a look at these changes? Maybe I shouldn't make this PR against `main`?

void-mAlex:
@colenso are you not able to use https://github.com/embroider-build/embroider/tree/main/packages/macros#istesting-isdevelopingapp
macros?

PR #1794, opened 2024-02-08
Refactor the resolve function to be the only public api to module-resolver

mansona:


void-mAlex:
any chance this refactor could land on stable first? I'd like to have one api to call from the code mod and not support maybe Promises maybe sync style nodeResolve

mansona:
> any chance this refactor could land on stable first? 

To be honest, that's quite unlikely 😞 This is in the class of things where it could potentially break things but has very little gain to people who are not using Vite, and is the main reason why we forked stable from main. 

> I'd like to have one api to call from the code mod

I mean... I don't see any reason why you need to use stable for the codemod 🤔 why not use unstable?

void-mAlex:
> I mean... I don't see any reason why you need to use stable for the codemod 🤔 why not use unstable?

because I want to use the codemod and I use stable :) 

also it's more reasonable to give it to people who can't get to main and this could be the push/unblocker to get them there

ef4:
This is getting close. The remaining failure is in the vite-specific test, because as of this change we use the real esbuild and vite resolvers instead of the node resolver when locating components, and that is causing some trouble.

 - vite has a plugin during depscan that turns "not found" into "external" so it will be silently ignored by esbuild. But this is bad for us because when we're checking "does this file exist?" we want honest answers, rather than always getting "yes (but external)". It's probably OK to always treat external as not_found, at least during depscan.
 - we're also running into the dreaded https://github.com/vitejs/vite/issues/9731 because now component resolution uses the real vite resolver and the vite resolver is wrong for self-references in v2 addons (v1 addons get adjusted automatically by embroider to handle self-references, v2 addons are able to use the full package.json `exports` API and are supposed to be able to rely on it)

NullVoxPopuli:
> v2 addons are able to use the full package.json exports API and are supposed to be able to rely on it)

in my testing, this has only been true when `type: module`
demo here: 
https://stackblitz.com/edit/stackblitz-starters-l1ujez?description=Starter%20project%20for%20Node.js,%20a%20JavaScript%20runtime%20built%20on%20Chrome%27s%20V8%20JavaScript%20engine&file=type-module-pkg%2Fpackage.json,pnpm-workspace.yaml,type-default-pkg%2Fpackage.json,type-default-pkg%2Ffoo.js,type-default-pkg%2Fbar.js,type-module-pkg%2Ffoo.js,type-module-pkg%2Fbar.js,a-consumer%2Fpackage.json,a-consumer%2Findex.js,package.json&title=node.new%20Starter

ef4:
The remaining failure here seems to be because our esbuild external-to-not-found workaround is too broad. It's causing a dep like qunit to be reported as not_found, when I think vite is telling esbuild that's it's external during depscan.

patricklx:
@ef4 one possible solution is to make esbuild-resolver and resolver communicate in some way, since all requests also go trough  the resolver we can know from there if the request actually resolved to something or not:

```diff
diff --git a/packages/vite/src/esbuild-request.ts b/packages/vite/src/esbuild-request.ts
index 67f62518..e8a227c4 100644
--- a/packages/vite/src/esbuild-request.ts
+++ b/packages/vite/src/esbuild-request.ts
@@ -4,6 +4,8 @@ import { dirname } from 'path';
 
 import type { Resolution } from '@embroider/core';
 
+export const resolvedImports = {};
+
 export class EsBuildModuleRequest implements ModuleRequest {
   static from(
     context: PluginBuild,
@@ -157,7 +159,7 @@ export class EsBuildModuleRequest implements ModuleRequest {
     });
     if (result.errors.length > 0) {
       return { type: 'not_found', err: result };
-    } else if (result.external) {
+    } else if (result.external && !resolvedImports[request.specifier]) {
       // during depscanning, vite configures esbuild with a very loose plugin
       // that turns all not_found into external so they're ignored. But that
       // means when we're doing something like checking for the existence of a
diff --git a/packages/vite/src/resolver.ts b/packages/vite/src/resolver.ts
index 441ed567..439efd67 100644
--- a/packages/vite/src/resolver.ts
+++ b/packages/vite/src/resolver.ts
@@ -3,6 +3,7 @@ import { virtualContent, ResolverLoader } from '@embroider/core';
 import { RollupModuleRequest, virtualPrefix } from './request';
 import assertNever from 'assert-never';
 import makeDebug from 'debug';
+import { resolvedImports } from './esbuild-request';
 
 const debug = makeDebug('embroider:vite');
 
@@ -42,8 +43,10 @@ export function resolver(): Plugin {
       let resolution = await resolverLoader.resolver.resolve(request);
       switch (resolution.type) {
         case 'found':
+          resolvedImports[request.specifier] = true;
           return resolution.result;
         case 'not_found':
+          resolvedImports[request.specifier] = resolvedImports[request.specifier] || false;
           return null;
         default:
           throw assertNever(resolution);

```

PR #1795, opened 2024-02-09
fix vite build

patricklx:


packages/vite/src/hbs.ts
```diff
@@ -14,9 +14,17 @@ export function hbs(): Plugin {
...
     name: 'rollup-hbs-plugin',
     enforce: 'pre',
-    async resolveId(source: string, importer: string | undefined) {
+    async resolveId(source: string, importer: string | undefined, options) {
+      if (options.custom?.embroider?.isExtensionSearch) {
+        return null;
+      }
       let resolution = await this.resolve(source, importer, {
         skipSelf: true,
+        custom: {
```
mansona, line 23: I think in the discussion we agreed to only add this when you're actually doing an extension search, so in the resolve inside `maybeSynthesizeComponentJS` on line 87

packages/vite/src/template-tag.ts
```diff
@@ -15,9 +15,17 @@ export function templateTag({ inline_source_map } = { inline_source_map: false }
...
     enforce: 'pre',
 
-    async resolveId(id: string, importer: string | undefined) {
+    async resolveId(id: string, importer: string | undefined, options) {
+      if (options.custom?.embroider?.isExtensionSearch) {
+        return null;
+      }
       let resolution = await this.resolve(id, importer, {
         skipSelf: true,
+        custom: {
```
mansona, line 24: same with this one, you already have it added during the extension search so you just need to remove this one

tests/vite-app/vite.config.mjs
```diff
@@ -42,7 +43,13 @@ export default defineConfig({
     },
   },
   build: {
+    outDir: resolve(process.cwd(), "dist"),
     rollupOptions: {
+      plugins: [
+        rollupResolve({
+          extensions: [".gjs", ".js", ".hbs", ".ts", ".gts"],
+        }),
+      ],
```
mansona, line 52: I'm pretty sure that we can't add this plugin and expect things to work 🤔 even if we needed to we should only add this in a new PR. Can you remove it from this one?

mansona (changes requested):
Almost there 🎉 3 little changes and then we are good 👍

tests/vite-app/vite.config.mjs
patricklx: the build fails without this

patricklx: or it did, its working now without it...

mansona (approved):
LGTM 🎉 🚀

PR #1799, opened 2024-02-10
Define local types for @embroider/core, since we have no guarantees of the consuming test environment

NullVoxPopuli:
This will help with ESM conversion because when our packages type check, they don't (and can't) know if they're going to be used in a Jest or QUnit test environment (or maybe an environment without either of those global types).

The approach here is that we define only what we use for when we receive args, and globals are accessed on `globalThis`, which we can cast to _any_, because we can't know what's on there.

If we want to generate _all_ types for `@embroider/addon-dev` in #1766, we'll need to make sure no package that `@embroider/addon-dev` depends on (nor sub-packages) rely on globally provided types.

This has worked historically for us so far due to a global and single type-checking command with every set of types available at once (which is representative of how our packages can be used :sweat_smile: )

### DIFF
--- a/packages/core/src/messages.ts
+++ b/packages/core/src/messages.ts
@@ -35,6 +35,18 @@ export function warn(message: string, ...params: any[]) {
   }
 }
 
+/**
+ * This type is normally from QUnit as a global.
+ * But dependents on `@embroider/core` may not be testing with QUnit,
+ * so we can't rely on the global availability of the NestedHooks interface.
+ *
+ * Here, we define only what we use in a way that is compatible with QUnit's types.
+ */
+interface NestedHooks {
+  before: (callback: () => void | Promise<void>) => void;
+  after: (callback: () => void | Promise<void>) => void;
+}
+
 // for use in our test suites
 let hardFailMode = 0;
 export function throwOnWarnings(hooks?: NestedHooks) {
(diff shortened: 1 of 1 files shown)


packages/core/src/messages.ts
```diff
@@ -48,8 +60,14 @@ export function throwOnWarnings(hooks?: NestedHooks) {
     });
   } else {
     // Jest mode
-    beforeAll(() => hardFailMode++);
-    afterAll(() => hardFailMode--);
+    if ('beforeAll' in globalThis && 'afterAll' in globalThis) {
```
ef4: We don't want this guard, it should crash if we get here with `hook` undefined and no `globalThis.beforeAll`.

NullVoxPopuli:
this hasn't come up in a while. It might come up again if we ever convert anything to ESM, but I don't have any motivation to do so, so as a part of cleaning up my PRs so my [PR view](https://github.com/search?q=is%3Apr+author%3ANullVoxPopuli++is%3Aopen+-is%3Adraft++org%3Aember-learn++org%3Aembroider-build++org%3Aemberjs++org%3Aember-cli++org%3Aember-tooling+org%3Arelease-plan+org%3Amainmatter&type=pullrequests&s=updated&o=desc&p=10) is more focused on what I have energy for, I am not misled by things I'm not going to follow up with.

PR #1801, opened 2024-02-13
Module resolver: virtualize vendor.js

BlueCutOfficial:
Virtualization of the vendor endpoint. 

This PR replaces the script tag `src="assets/vendor.js"` with the virtual entry point `src="@embroider/core/vendor"` in the `index.html` of the rewritten app. The content that was previously written in `assets/vendor.js` is now generated by the `virtual-vendor`'s functions.

**How to test**
- On the main branch:
  - Start the vite-app
  - Visit http://127.0.0.1:4200/assets/vendor.js
- On the `virtual-vendor` branch:
  - Start the vite-app
  - Visit http://127.0.0.1:4200/@embroider/core/vendor.js
- 👀 both files should have the same content
- On the `virtual-vendor` branch, build the vite app
- 👀 There should be a file `dist/@embroider/core/vendor.js` with the same content

### DIFF
--- a/packages/core/src/module-resolver.ts
+++ b/packages/core/src/module-resolver.ts
@@ -948,12 +948,7 @@ export class Resolver {
 
   private handleVendor<R extends ModuleRequest>(request: R): R {
     //TODO move the extra forwardslash handling out into the vite plugin
-    const candidates = [
-      '#embroider/core/vendor',
-      '@embroider/core/vendor',
-      '/@embroider/core/vendor',
-      './@embroider/core/vendor',
-    ];
+    const candidates = ['@embroider/core/vendor.js', '/@embroider/core/vendor.js', './@embroider/core/vendor.js'];
 
     if (!candidates.includes(request.specifier)) {
       return request;
@@ -962,7 +957,7 @@ export class Resolver {
     let pkg = this.packageCache.ownerOfFile(request.fromFile);
     if (pkg?.root !== this.options.engines[0].root) {
       throw new Error(
-        `bug: found an import of ${request.specifier} in ${request.fromFile}, but this is not the top-level Ember app. The top-level Ember app is the only one that has support for @embroider/core/vendor. If you think something should be fixed in Embroider, please open an issue on https://github.com/embroider-build/embroider/issues.`
+        `bug: found an import of ${request.specifier} in ${request.fromFile}, but this is not the top-level Ember app. The top-level Ember app is the only one that has support for @embroider/core/vendor.js. If you think something should be fixed in Embroider, please open an issue on https://github.com/embroider-build/embroider/issues.`
       );
     }
 
--- a/packages/core/src/module-resolver.ts
+++ b/packages/core/src/module-resolver.ts
@@ -197,6 +197,7 @@ export class Resolver {
     request = this.handleVendorStyles(request);
     request = this.handleTestSupportStyles(request);
     request = this.handleRenaming(request);
+    request = this.handleVendor(request);
     // we expect the specifier to be app relative at this point - must be after handleRenaming
     request = this.generateFastbootSwitch(request);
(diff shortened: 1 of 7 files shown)


packages/core/src/module-resolver.ts
```diff
@@ -870,6 +871,31 @@ export class Resolver {
...
+      './@embroider/core/vendor',
+    ];
+
+    if (!candidates.includes(request.specifier)) {
+      return request;
+    }
+
+    let pkg = this.packageCache.ownerOfFile(request.fromFile);
+    if (pkg?.root !== this.options.engines[0].root) {
+      return request;
```
mansona, line 889: so the check here is correct but the thing that you're doing is not quite. Essentially, we only want to support somebody adding the `@embroider/core/vendor` to their app code, i.e. the package that owns the file that contains the reference to `@embroider/core/vendor` is their actual top-level app.

If we just return the request here you would get some strange error from vite or webpack saying `could not resolve @embroider/core/vendor` and it wouldn't communicate anything to the end user that something was wrong. We have an opportunity here to tell the end user **exactly** what is wrong so we should take it, and we should throw an error along the lines of `we found an import of @embroider/core/vendor in ${fromFile} but that is not in your top-level ember app. We only support blah blah blah. Open an issue etc.`

packages/core/src/virtual-vendor.ts
```diff
@@ -0,0 +1,88 @@
...
+        addon.canResolveFromFile,
+      ])
+    ),
+    isApp: true,
+    modulePrefix: resolver.options.modulePrefix,
+    appRelativePath: 'NOT_USED_DELETE_ME',
+  };
+
+  // TODO - From where do we get this dynamically? in compat-app-builder:
+  // let emberENV = this.configTree.readConfig().EmberENV;
```
mansona, line 38: note for when we're discussing this: all communication between stage 1 & 2 and stage 3 has to go through a file on disk in the `.embroider` folder. Right now we only have a `resolver.json` file but we might need to discuss having the config tree from addons be output on disk 👍

mansona (commented):
This is looking amazing! so close to the finishing line now 🎉

BlueCutOfficial:
The scenarios [lts_3_28-util](https://github.com/embroider-build/embroider/actions/runs/8723912589/job/23934579633?pr=1801) are failing, but for now, I am not sure how that relates to my changes. Locally, I have this scenario failing on both `main` and `virtualize-vendor` but the output is completely different from the CI.

BlueCutOfficial:
I put the branch back to the commit that was green ✅ and `lts_3_28-util` is now red ❌ So whatever the problem is, it's probably unrelated to my changes (unless my changes no longer work with the latest things we merged). Pushing back the 2 last commits as they are innocent 😇

tests/scenarios/compat-addon-classic-features-test.ts
```diff
@@ -131,3 +131,43 @@ appScenarios
...
+
+    Qmodule(`${scenario.name} - dev mode`, function (hooks) {
+      hooks.before(async () => {
+        app = await scenario.prepare();
+      });
+
+      test('vendor.js script is served', async function (assert) {
+        const server = CommandWatcher.launch('vite', ['--clearScreen', 'false'], { cwd: app.dir });
+        try {
+          const [, url] = await server.waitFor(/Local:\s+(https?:\/\/.*)\//g);
```
patricklx, line 160:
```suggestion
          const [, url] = await server.waitFor(/Local:\s+(https?:\/\/.*)\//g, 180 * 1000);
```

BlueCutOfficial: Thanks, I'll skip the test as discussed during the office hours, but that was probably the issue 👍

mansona (approved):
Nice work 🎉 LGTM 💪

PR #1802, opened 2024-02-13
fix vite auto reload on windows

patricklx:
ember build needs to be done before vite starts the server.
that is because vite already starts watching .embroider dir on server start. 

But its a problem for windows if the dir does not exist at that moment or if it gets deleted/replaced. Which is what an ember build does. Windows ends up not receiving updates for files in that directory...


patricklx:
after discussion decided to leave as is, since rewritten-app will be removed anyway

PR #1805, opened 2024-02-15
Virtual vendor styles entry point

BlueCutOfficial:
This draft is about the virtualization of `vendor.css`


mansona:
With a bit of experimentation on a vanilla Vite app @BlueCutOfficial discovered that we can probably get away with properly virtualising a raw css file making use of `?direct` when internally redirected. 

We've been struggling to make sure that the content-type is correct in an embroider test, there are a few things to try out but we could also try a "post middleware" to edit the content type of css responses. https://vitejs.dev/guide/api-plugin.html#configureserver

if adding a post middleware to edit the content type doesn't work then we could just implement a normal middleware to handle the virtual css response correctly. Then for build mode we would just use the emit file as we wanted to but couldn't because it doesn't work in dev mode.

packages/compat/src/compat-app-builder.ts
```diff
@@ -391,20 +391,9 @@ export class CompatAppBuilder {
...
       if (implicitScripts) {
-        let styles = [];
         let options = { basedir: addon.root };
         for (let mod of implicitScripts) {
-          if (type === 'implicit-styles') {
-            // exclude engines because they will handle their own css importation
-            if (!addon.isLazyEngine()) {
-              styles.push(resolve.sync(mod, options));
-            }
-          } else {
```
BlueCutOfficial, line 402: Note for myself: that may not be ready for clean-up after all, the function may be called with `implicit-styles` parameter somewhere else

tests/scenarios/compat-app-html-attributes-test.ts
```diff
@@ -59,9 +59,9 @@ appScenarios
         expectFile('./index.html').doesNotMatch('rel="stylesheet"', 'does not have rel=stylesheet');
         expectFile('./index.html').matches('<script defer', 'has script defer');
         expectFile('./index.html').doesNotMatch('<script src', 'does not have script src');
-        expectFile('./index.html').doesNotMatch(
+        expectFile('./index.html').matches(
           'data-original-filename="vendor.css">',
-          'does not have data-original-filename vendor.css'
+          'has data-original-filename vendor.css'
```
BlueCutOfficial, line 64: The compat-app-builder used to compute a boolean `implicitStyles` and add the stylesheet only if `true`. With this PR, we simplify this and add the stylesheet without condition.

PR #1806, opened 2024-02-17
resolver transform to emit imports for helper and modifiers that need…

void-mAlex:
… them in strict mode templates

### DIFF
--- a/packages/compat/src/resolver-transform.ts
+++ b/packages/compat/src/resolver-transform.ts
@@ -43,51 +43,73 @@ export interface Options {
   appRoot: string;
 }
 
-export const builtInKeywords = [
-  '-get-dynamic-var',
-  '-in-element',
-  '-with-dynamic-vars',
-  'action',
-  'array',
-  'component',
-  'concat',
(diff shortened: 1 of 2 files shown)


mansona (dismissed):
This looks good to me 👍 I think that this is in line with what we discussed in meetings and the fact that tests are passing gives me the confidence that this can be merged and released without much issue 👍

void-mAlex:
Thank you for this @ef4 ,
I do have to point out that in my work I've found that `#embroider_compat/components/power-select` resolves directly to a file in the re-written packages and not something from the app scope, so I'm not convinced about the addon provided components needing this treatment as I have tested this scenario  (though not very extensively) and it behaves differently - as far as I've seen I get straight to the path after the re-export but more testing is needed.

also `uniqId` is exported as `uniqueId` - I'll fix that up unless there is a specific reason the former one was used.

I'll work to fix up the tests as well

mansona:
FYI I just rebased this against stable 👍

ef4 (approved):

PR #1807, opened 2024-02-20
Module resolver: virtualize test-support.js

BlueCutOfficial:


packages/core/src/module-resolver.ts
```diff
@@ -421,6 +422,32 @@ export class Resolver {
...
+    ];
+
+    if (!candidates.includes(request.specifier)) {
+      return request;
+    }
+
+    let pkg = this.packageCache.ownerOfFile(request.fromFile);
+    if (pkg?.root !== this.options.engines[0].root) {
+      throw new Error(
+        `bug: found an import of ${request.specifier} in ${request.fromFile}, but this is not the top-level Ember app. The top-level Ember app is the only one that has support for @embroider/core/vendor. If you think something should be fixed in Embroider, please open an issue on https://github.com/embroider-build/embroider/issues.`
```
BlueCutOfficial, line 440:
```suggestion
        `bug: found an import of ${request.specifier} in ${request.fromFile}, but this is not the top-level Ember app. The top-level Ember app is the only one that has support for @embroider/core/test-support.js. If you think something should be fixed in Embroider, please open an issue on https://github.com/embroider-build/embroider/issues.`
```

packages/core/src/virtual-test-support.ts
```diff
@@ -0,0 +1,88 @@
...
+  if (!hasEmbroiderMacrosTestSupport) {
+    result.unshift(`${readFileSync(require.resolve('@embroider/macros/src/vendor/embroider-macros-test-support'))}`);
+  }
+
+  return result.join('') as string;
+}
+
+function impliedAddonTestSupport(engine: Engine): string[] {
+  let result: Array<string> = [];
+  for (let addon of sortBy(Array.from(engine.addons.keys()), pkg => {
```
ef4, line 69: I think we can delete the sortBy. It was only being used by `implicit-scripts` not `implicit-test-scripts`, and now that this function only handles `implicit-test-scripts` this part should never have an effect.

packages/core/src/module-resolver.ts
```diff
@@ -421,6 +422,32 @@ export class Resolver {
...
+      );
+    }
+
+    if (!pkg?.isV2Ember()) {
+      throw new Error(`bug: an import of ${request.specifier} in non-ember package at ${request.fromFile}`);
+    }
```
ef4, line 446: We can remove this check, the check up above for being the first engine is sufficient.

ef4:
Thanks, nice work!

PR #1808, opened 2024-02-20
fix vite ts

patricklx:
ts transform was actually not enabled for vite app


patricklx:
Will be solved when ts-app turns into a vite app

PR #1813, opened 2024-02-22
vite: fix hbs loading for virtual pair components

patricklx:
it happens when embroider generates a virtual pair component, the resolved id then looks like `/template.hbs/component.ts-embroider-pair-component`

### DIFF
--- a/packages/vite/src/hbs.ts
+++ b/packages/vite/src/hbs.ts
@@ -1,12 +1,13 @@
 // TODO: I copied this from @embroider/addon-dev, it needs to be its own package
 import type { PluginContext, ResolvedId } from 'rollup';
 import type { Plugin } from 'vite';
-import { hbsToJS } from '@embroider/core';
+import { hbsToJS, ResolverLoader } from '@embroider/core';
 import assertNever from 'assert-never';
 import { parse as pathParse } from 'path';
 import makeDebug from 'debug';
 
 const debug = makeDebug('embroider:hbs-plugin');
+const resolverLoader = new ResolverLoader(process.cwd());
 
 export function hbs(): Plugin {
   return {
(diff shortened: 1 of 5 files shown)


mansona:
This is likely going to be solved as part of https://github.com/embroider-build/embroider/pull/1794, we literally just hit a similar error

patricklx:
@mansona i think that will fix a specific issue. But not the filter. I think the filter is not good.
It would also match any  extension like `.hbs.md`

patricklx:
first error https://github.com/embroider-build/embroider/actions/runs/8252373750/job/22571786702?pr=1813

packages/vite/src/hbs.ts
```diff
@@ -99,6 +98,24 @@ async function maybeSynthesizeComponentJS(context: PluginContext, source: string
   if (!templateResolution) {
     return null;
   }
+
+  const resolvedId = templateResolution.id.split('?')[0];
+  templateResolution.id = resolvedId;
+
+  const pkg = resolverLoader.resolver.packageCache.ownerOfFile(resolvedId);
+  const isInComponents = pkg?.isV2App() && resolvedId.slice(pkg?.root.length).startsWith('/components');
```
patricklx, line 106: this will always be false for addons, which is good, because they should not do template colocation

packages/vite/src/hbs.ts
```diff
@@ -130,8 +132,10 @@ async function maybeSynthesizeComponentJS(context: PluginContext, source: string
   };
 }
 
+const hbsFilter = createFilter('**/*.hbs?(\\?)*');
+
 function maybeRewriteHBS(resolution: ResolvedId) {
-  if (!resolution.id.split('?')[0].endsWith('.hbs')) {
+  if (!hbsFilter(resolution.id)) {
```
patricklx, line 138: @mansona this will match  `/rewritten-app/components/old/template.hbs?v=45cfc9c0/..%2Fcomponent.js%3Fv%3D45cfc9c0-embroider-pair-component`

mansona (approved):
This looks good 👍 myself and @ef4 reviewed it this morning and we're happy to get it merged. I added a commit to reset the hbs extension handling to use createFilter again, but tests pass and I'm happy that it's doing the same thing

packages/vite/src/hbs.ts
mansona: Doesn't your test cover this issue? 🤔

patricklx: mm, its passing, ill have a look

PR #1815, opened 2024-02-23
Make macro configs static

simonihmig:


packages/macros/src/babel/evaluate-json.ts
```diff
@@ -388,13 +389,53 @@ export class Evaluator {
...
+      let maybeFastbootMemberExpression = path.parentPath;
+      if (
+        maybeFastbootMemberExpression.isMemberExpression() ||
+        maybeFastbootMemberExpression.isOptionalMemberExpression()
+      ) {
+        let maybeFastbootProperty = maybeFastbootMemberExpression.isMemberExpression()
+          ? maybeFastbootMemberExpression.get('property')
+          : maybeFastbootMemberExpression.isOptionalMemberExpression()
+          ? maybeFastbootMemberExpression.get('property')
+          : assertNever(maybeFastbootMemberExpression);
```
simonihmig, line 409: This looks weird, but I found no better way to make TS happy. I had it first as simple as this:

image

Those tertiary expressions narrow down the type before hitting `.get()` to make it return the correct type.


The type definition of `.get()` would not deal well with a union type (`NodePath<MemberExpression> | NodePath<OptionalMemeberExpression>`  it seems, falling back to the second loosely typed one:

image

simonihmig:
Failing tests are the same as those on master, fixed in https://github.com/embroider-build/embroider/pull/1817...

ef4 (approved):

PR #1816, opened 2024-02-26
Remove silent option from ember build in compatPrebuild

enspandi:
Vite does not complete otherwise and no ember build output is shown.

### DIFF
--- a/packages/vite/src/build.ts
+++ b/packages/vite/src/build.ts
@@ -4,7 +4,7 @@ import type { Plugin } from 'vite';
 export function emberBuild(mode: string): Promise<void> {
   if (mode === 'build') {
     return new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


patricklx:
found this on stackoverflow:

`
When you pass silent: true to child_process.fork(), pipes are used between the parent and child. In your parent script you are never reading from the child's streams (especially stdout), so at some point the internal buffer is filling up and the built-in backpressure mechanism causes the process to stop accepting output from the child. Which also means the exit signal.
`

so that explains why it could fail sometimes. I think this makes sense

mansona (approved):
Looks great 🎉

mansona:
Thanks @patricklx! I really had no idea why this could be causing any issues, so I'm glad you found a potential mechanism 🎉

PR #1817, opened 2024-02-26
pin ember-data to fix issue in CI

mansona:
It looks like ember-data@5.3.1 has a few issues that make it incompatible with embroider right now 😭 I'm pinning it for now to fix CI and I'll open a new PR right after this to unpin it and we can fix from there 👍

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -15,13 +15,13 @@ importers:
     devDependencies:
       '@types/jest':
         specifier: ^29.2.0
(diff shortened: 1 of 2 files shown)


simonihmig:
@mansona should we do the same for `stable`? Just picking up those errors in https://github.com/embroider-build/embroider/pull/1815...

mansona:
Yes I guess that's totally reasonable to do @simonihmig, although I would really like to get this fixed ASAP upstream 🙈

PR #1819, opened 2024-02-27
Add a new prebuild function with strict defaults

BlueCutOfficial:


packages/compat/src/default-pipeline.ts
```diff
@@ -51,6 +51,37 @@ export default function defaultPipeline<PackagerOptions>(
...
+  options: PipelineOptions<PackagerOptions> = {
+    staticAddonTrees: true,
+    staticAddonTestSupportTrees: true,
+    staticComponents: true,
+    staticHelpers: true,
+    staticModifiers: true,
+    staticEmberSource: true,
+    amdCompatibility: {
+      es: [],
+    },
```
mansona, line 65: How do we feel about passing this default object here? should this PR be going deeper and passing nothing here, but switching the defaults under the hood? That way we would have to set the default options for `defaultPipeline` to be in inverse of this.

PR #1823, opened 2024-02-27
Make macro configs static

simonihmig:
Same as #1815, this time targeting `main`

### DIFF
--- a/packages/macros/src/babel/evaluate-json.ts
+++ b/packages/macros/src/babel/evaluate-json.ts
@@ -409,25 +409,11 @@ export class Evaluator {
           : assertNever(maybeFastbootMemberExpression);
 
         if (maybeFastbootProperty.isIdentifier() && maybeFastbootProperty.node.name === 'fastboot') {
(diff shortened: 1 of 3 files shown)


mansona (commented):
Shouldn't we just merge stable into main after we release stable and then the changes will come along? 🤔

simonihmig:
yeah, I wasn't sure how we plan that!? I guess it depends how far the two branches diverged, and where the changes on stable happened? Changes in the macro package should be easy to merge, but not sure if we have others that cause crazy merge conflicts, or that we even don't want in `main`? More on you to decide I guess what makes more sense! :)

mansona:
essentially we agreed collectively that we want to do this stuff as "forward merges" from stable to main, I'll give it a go and see how far I get :)

mansona:
Didn't seem like too hard a merge 👍 https://github.com/embroider-build/embroider/pull/1825

simonihmig:
Even better, closing this!

PR #1827, opened 2024-02-29
fix extension resolving for esbuild

mansona:
We have implemented extension searching on the Vite side of a build, but in https://github.com/embroider-build/embroider/pull/1779 our templates are starting to get analysed in esbuild and need the same "extension candidate" implementation 👍

### DIFF
--- a/packages/vite/src/esbuild-resolver.ts
+++ b/packages/vite/src/esbuild-resolver.ts
@@ -8,6 +8,13 @@ import { resolve, join } from 'path';
 import { hbsToJS } from '@embroider/core';
 import { Preprocessor } from 'content-tag';
 
+function* candidates(path: string) {
+  yield path;
+  yield path + '.hbs';
+  yield path + '.gjs';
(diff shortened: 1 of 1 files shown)


patricklx:
i'm not sure what changed, but vite adds its own plugin to esbuild `vite:dep-scan`, which then calls all vite plugins. So it should go through `vite/src/hbs.ts` etc

patricklx:
this could also be fixed by setting the extensions resolving in vite config.

https://vitejs.dev/config/shared-options#resolve-extensions

which would also need to be done for rollup with an extra plugin...

ef4:
We discussed this and @patricklx's questions are good ones. We think vite does not add custom rollup plugins to the esbuild when doing dep pre-building. We can probably rely on resolve-extensions for gjs and gts, but the hbs rollup plugin has more complex behaviors. We can experiment with slightly simplifying our plugins in favor of resolve-extensions but it would not completely replace them.

PR #1828, opened 2024-02-29
add ts, js extension to dep optimizer

patricklx:


patricklx:
What this actually does is that vite checks those extensions and does a build with esbuild on those before scanning it for deps. That why this breaks

PR #1832, opened 2024-03-04
Update to `ember-cli-babel` v8

bertdeblock:


mansona (approved):
I guess this looks good 😂 tests are passing so LGTM 💪

mansona:
It turns out that this should have been a breaking change 😞 the only reason this was working for the most part was because the dependencies of ember-cli-babel weren't using any dependencies that would break tests. 

Recently babel-plugin-module-resolver bumped their glob major version (in a patch 😫) that brought in minimatch@8 which needs at least Node 14 (for private fields) https://github.com/tleunen/babel-plugin-module-resolver/pull/445

@embroider/util is being used by ember-auto-import tests that need to work in Node 12 so we should have made this a major 😞

PR #1836, opened 2024-03-05
Replace content-for using a Vite plugin

BlueCutOfficial:
This PR moves `{{content-for}}` from stage 2 to stage 3: 

We no longer replace the `{{content-for}}` snippets in the `index.html` during the build of the rewritten app. Instead, we leave the `{{content-for}}` in place and it's a new Vite plugin that performs the replacement. To do so, the Vite plugin relies on a new file `content-for.json` output during the pre-build. 

**Breaking change**

The new approach supposes that Embroider knows in advance what types of `{{content-for}}` are provided by the addons, because the Broccoli plugin that builds the dictionary needs to be agnostic from `rewritten-app/index.html` if we don't want to be tied with stage 2.

The new build option `availableContentForTypes` allows to extend the list of `{{content-for}}` types in case an addon provides something custom. For instance, let's suppose your app relies on a classic addon that provides `{{content-for  "custom"}}`, you should add the following build option so the content can be replaced:
```js
// ember-cli-build.js
const { prebuild } = require('@embroider/compat');

  const app = new EmberApp(defaults, {});
  return prebuild(app, {
    availableContentForTypes: ['custom'],
  });
};
```

### DIFF
--- a/packages/compat/src/compat-app.ts
+++ b/packages/compat/src/compat-app.ts
@@ -150,10 +150,13 @@ export default class CompatApp {
 
   @Memoize()
   private get contentFor(): ContentForConfig {
-    const configPath = join('environments', `${this.legacyEmberAppInstance.env}.json`);
+    const configPaths = [
+      { file: '/index.html', path: join('environments', `${this.legacyEmberAppInstance.env}.json`) },
+    ];
+    if (this.shouldBuildTests) configPaths.push({ file: '/tests/index.html', path: join('environments', `test.json`) });
     return new ContentForConfig(this.configTree, {
       availableContentForTypes: this.options.availableContentForTypes,
-      configPath,
+      configPaths,
       pattern: this.filteredPatternsByContentFor.contentFor,
     });
   }
--- a/packages/compat/src/compat-app.ts
+++ b/packages/compat/src/compat-app.ts
@@ -180,6 +180,14 @@ export default class CompatApp {
     });
   }
 
+  private get filteredPatternsByContentFor() {
+    const filter = '/{{content-for [\'"](.+?)["\']}}/g';
+    return {
+      contentFor: this.configReplacePatterns.find((pattern: any) => filter.includes(pattern.match.toString())),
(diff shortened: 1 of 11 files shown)


packages/compat/src/compat-app.ts
```diff
@@ -195,12 +196,29 @@ export default class CompatApp {
...
+
+    return mergeTrees([
+      new this.configReplace(new ContentForSnippets(index, {}), this.configTree, {
+        configPath: join('environments', `${this.legacyEmberAppInstance.env}.json`),
+        files: [
+          // TODO: get dynamically the name of the files output by ContentForSnippets
+          'embroider-content-for-body-footer.html',
+          'embroider-content-for-head-footer.html',
+          'embroider-content-for-body.html',
+          'embroider-content-for-head.html',
```
patricklx, line 208: there are also test-head etc in tests/index.html

and you will need to use a different `env` for `tests/index.html`.
so you will also need `content-for-head` and `content-for-test-head` etc with test env

patricklx: ah, missed that one

packages/compat/src/compat-app.ts
```diff
@@ -146,6 +147,16 @@ export default class CompatApp {
     }
   }
 
+  @Memoize()
+  private get contentFor(): ContentForConfig {
+    const configPath = join('environments', `${this.legacyEmberAppInstance.env}.json`);
```
BlueCutOfficial, line 152: It feels like the test failures are related to this line. The code comes from `htmlTree` getter, but `testIndexTree` getter has:
```
  configPath: join('environments', `test.json`)
```
might be related.

BlueCutOfficial:
The [`canary-static-app`](https://github.com/embroider-build/embroider/actions/runs/8556664525/job/23447115028?pr=1836) error sounds a bit like the tests not being present but I am not sure how it's related to the PR and how it happens because the command because `pnpm test: production` uses the option `FORCE_BUILD_TESTS: 'true'`.

mansona (approved):
Fantastic work 🎉

PR #1840, opened 2024-03-12
Use Vite for all tests

mansona:


ef4:
The merge from main above fixed part of the failures in canary-static-app because 

The above commit fixed the development mode canary-static-app. The next thing we debugged is the production mode canary-static-app. Before, that was doing `EMBER_ENV=production ember test`, which put ember into production mode while still including the tests in the build. Now the build happens separately from `ember test` so there's nothing telling the build to include tests, so they're not in dist.

Other issues identified while debugging (some of which only appear if you try to run one of the scenarios in `vite dev`, particularly with staticAddonsTrees===false):
 - when `@ember/test-waiters` gets upgraded to v2 we include types/index.js in the implicit modules. An empty JS file is produced for it, but the browser considers it an error to try to `import * from` an empty module. This is not a new bug, but webpack was sloppy enough to not care.
 - we saw some of the package rules in the static-app not applying and causing test failures in dev only. We suspect this is a query params matching issue that could be fixed by https://github.com/embroider-build/embroider/pull/1846

We may want to consider not supporting staticAddonsTrees===false at all in the new major. It often introduces bugs that are otherwise not present. It's trying to give behavior closer to classic, but it can hurt more than it helps.

We also started debugging the macro-tests failure, which is again a staticAddonsTrees===false issue where the implicit-modules appears to have a module cycle involving `@ember/modifier/index`.

ef4:
Leading edge of work:
 - debugging in canary-macro-tests, there's still the presumed module cycle involving `@ember/modifier/index`
 - it would be nice to debug that cycle in `vite dev` rather than `vite build`, but `vite dev` is failing for its own reasons that need to be debugged
 - the macro-tests app was historically a dumping ground for testing weird app behaviors, and it includes several custom outputPaths using app.import. Those appear not to be working at least in `vite dev`. Unclear whether we care, but it's something you'll see immediately when you try to debug it.

ef4:
I figured out the `@ember/test-waiters` issue. It's not a bug in embroider -- the addon actually includes a javascript module containing only `export {}`. Vite's dep optimizer [doesn't handle that correctly](https://github.com/vitejs/vite/issues/16315).

(Arguably the addon shouldn't include such a  module. It seems to exist as a workaround for other code doing a poor job of distinguishing type-only imports.)

ef4:
Down to only the watch mode tests.

I started reimplementing them to use a direct fetch for a given module, but that is going to be problematic. Vite controls rebuilds by propagating query params downward through the module graph. To get a realistic result we want to follow those same imports. This is a good fit for the audit assertions, but we'll need to extend those to be able to follow URLs and not just files.

PR #1844, opened 2024-03-19
Add @glimmer/runtime to ember-standard-modules

NullVoxPopuli:
#compatibility 'n'such


NullVoxPopuli:
because of embroider's implicit-modules, this is no longer needed

PR #1845, opened 2024-03-19
support import sync with dynamic expression

patricklx:


patricklx:
e.g. ember-svg-jar
image
compiles to
image

patricklx:
The screenshot shows keys with extension. That is fixed.
Open question: should we filter by files, so exclude folders? It could be possible that folders have a index.js file

patricklx:
@ef4 removed the binary expression support

packages/macros/src/babel/macros-babel-plugin.ts
```diff
@@ -123,7 +125,53 @@ export default function main(context: typeof Babel): unknown {
...
           let specifier = path.node.arguments[0];
           if (specifier?.type !== 'StringLiteral') {
-            throw new Error(`importSync eager mode doesn't implement non string literal arguments yet`);
+            let relativePath = '';
+            let property;
+            if (specifier.type === 'TemplateLiteral' && specifier.expressions[0].type === 'Identifier') {
+              relativePath = specifier.quasis[0].value.cooked!;
+              property = specifier.expressions[0];
+            }
+            // babel might transform template form `../my-path/${id}` to '../my-path/'.concat(id)
```
ef4, line 134: I didn't notice before that this stuff is distinct from the BinaryExpression support.  I don't think it's correct.

Consider:

```js
importSync(`one${two}three${four}`)
// becomes
importSync("one".concat(two, "three").concat(four));
```

(It would be even better to figure out how to run before preset-env, but I recognize that this is already in an `exit` hook because it's trying to interoperate with other plugins that adjust the import specifiers themselves.)

patricklx: Right, i only support `"one".concat`. this is done by ensuring that the object of the member expression is a string.
`one${two}three${four}` is also not supported

ef4 (approved):

PR #1846, opened 2024-03-20
Prevent query-params added by vite to be passed to core logic

BlueCutOfficial:
**Context**
PR #1805 highlighted how Vite sometimes adds internal query parameters to requests. For instance, if you include a CSS file _directly_ in `index.html` with a `link` tag pointing to a stylesheet, then Vite will internally add the query parameter `?direct` to the request so it knows it has to return the stylesheet as is without wrapping it in some hot-reload-related JS. This is just one example of the sort of things Vite can do for its own reason. 

The problem is that Embroider doesn't expect this kind of query params to exist, for instance when dealing with virtual files, and we can't let the core part of Embroider know about that: it is bundler-agnostic, it should behave the same way if you use Vite, or Webpack, or whatever.

**Problem to solve**
Everything that has to do with Vite query params should stay in `packages/vite/src` and from this point, we need to make sure that: 
- query-params added by Vite keep being propagated from request to request,
- but are always removed from the specifier when we call the `packages/core/src` functions that are bundler-agnostic.

**How to test**
- [x] No regression when testing the vite-app
- [x] If the changes are used in PR #1805, then we should be able to get rid of the last commit that handles `isDirect` in Vite resolver, and the `vendor.css` should still be generated correctly.

### DIFF
--- a/packages/vite/src/request.ts
+++ b/packages/vite/src/request.ts
@@ -1,5 +1,5 @@
 import type { ModuleRequest, Resolution } from '@embroider/core';
-import { cleanUrl } from '@embroider/core';
+import { cleanUrl, cleanUrlQueryParams, getUrlQueryParams } from '@embroider/core';
 import type { PluginContext, ResolveIdResult } from 'rollup';
 
 export const virtualPrefix = 'embroider_virtual:';
@@ -28,7 +28,20 @@ export class RollupModuleRequest implements ModuleRequest {
 
       // strip query params off the importer
       let fromFile = cleanUrl(nonVirtual);
-      return new RollupModuleRequest(context, source, fromFile, custom?.embroider?.meta, false, undefined);
+
+      // strip query params off the source but keep track of them
+      let cleanSource = cleanUrlQueryParams(source);
+      let queryParams = getUrlQueryParams(source);
+
+      return new RollupModuleRequest(
+        context,
+        cleanSource,
+        fromFile,
+        custom?.embroider?.meta,
+        false,
+        undefined,
+        queryParams
+      );
     }
   }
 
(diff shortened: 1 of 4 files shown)


packages/vite/src/request.ts
```diff
@@ -28,7 +28,20 @@ export class RollupModuleRequest implements ModuleRequest {
 
       // strip query params off the importer
       let fromFile = cleanUrl(nonVirtual);
-      return new RollupModuleRequest(context, source, fromFile, custom?.embroider?.meta, false, undefined);
+
+      // strip query params off the source but keep track of them
+      let cleanSource = cleanUrlQueryParams(source);
+      let queryParams = getUrlQueryParams(source);
```
BlueCutOfficial, line 34: @ef4 I tried to replace these regexp-based methods with a URL object, but it caused trouble because the `source` can be a relative path. Here are examples of specifiers I get with both methods:
```
URL   : /app, 
Regexp: ../app, 

URL   : /-embroider-implicit-modules.js, 
Regexp: ./-embroider-implicit-modules.js, 

URL   : /packages/macros/src/addon/es-compat2, 
Regexp: ../../../../../../packages/macros/src/addon/es-compat2, 
```

packages/vite/src/resolver.ts
```diff
@@ -52,7 +52,10 @@ export function resolver(): Plugin {
     },
     load(id) {
       if (id.startsWith(virtualPrefix)) {
-        let { src, watches } = virtualContent(id.slice(virtualPrefix.length), resolverLoader.resolver);
+        let { src, watches } = virtualContent(
+          cleanUrlQueryParams(id.slice(virtualPrefix.length)),
+          resolverLoader.resolver
+        );
```
BlueCutOfficial, line 58: The only parameter I have on `load` is the `id`, which corresponds - I guess - to a `result.id` as it contains the query params. I don't see any obvious way to avoid cleaning again the query params before passing to `virtualContent`.

PR #1847, opened 2024-03-21
Refactor watch mode tests

ef4:
@mansona and I are simplifying this because we need to add features in the vite-tests branch and it will be easier if we simplify first.


mansona (approved):
LGTM 🎉

PR #1850, opened 2024-03-21
vite: do not optimize dynamic addons 

patricklx:
with dynamic addons I mean addons which override the treeFor methods.
they can potentially depend on app tree and rebuild whenever an app file changes. e.g ember-svg-jar, ember-cli-addon-docs.

With dep optimization that will not be updated. they need to be excluded

### DIFF
--- a/packages/compat/src/v1-addon.ts
+++ b/packages/compat/src/v1-addon.ts
@@ -561,7 +561,7 @@ export default class V1Addon {
         version: 2,
         'auto-upgraded': true,
         type: 'addon',
-        'is-dynamic': this.hasCustomizedTree()
+        'is-dynamic': this.hasCustomizedTree(),
       },
       built.staticMeta,
       ...built.dynamicMeta.map(d => d())
--- a/packages/compat/src/v1-addon.ts
+++ b/packages/compat/src/v1-addon.ts
@@ -384,8 +384,12 @@ export default class V1Addon {
     );
   }
 
+  hasCustomizedTree(): boolean {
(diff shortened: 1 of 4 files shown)


mansona (changes requested):
So I am pretty sure that we don't want to go in this direction. A lot of thought has gone into the [V2 addon spec](https://github.com/embroider-build/embroider/blob/main/docs/spec.md) to define what should or shouldn't be possible. This sort of addition to the spec feels like it should be a new RFC 🙈 but at the very least it should be discussed at a Tooling team meeting before we move forward. 

I can see what you're trying to do here, you're trying to keep track of auto-upgraded addons that have customised their trees, but any feature that you add to the v2 addon spec could theoretically be consumed directly by a v2 addon and there is no way to prevent that and limit it to auto-upgraded v1 addons only. 

If you want to discuss this idea further I recommend adding an Agenda item to the Tooling meeting or bringing it up at office-hours 👍

patricklx:
I actually have to re verify this with #1876 first if this is still an issue.
Maybe i only had bad dep optimizer issues.

Can v2 addons also override trees? Then they also would need this.
The issue was currently with v1 addons that merge into themselves the app tree and do some app files processing. They create new files or recreate files based on app filed during dev. And if they are optimized, vite wouldn't re-optimize them. But that's also an issue without dynamic addons. Vite wouldn't optimize when a previous unused dep was imported  during development

NullVoxPopuli:
> Can v2 addons also override trees? 

they cannot :tada:

patricklx:
this behaviour  was never supported

PR #1853, opened 2024-03-23
add vite compat styles resolver

patricklx:
allows resolving v1 scss, less styles. They are currently copied to rewritten-app/assets.

### DIFF
--- a/packages/vite/src/compat-styles.ts
+++ b/packages/vite/src/compat-styles.ts
@@ -1,0 +1,45 @@
+import { ResolverLoader } from '@embroider/core';
+
+export const function scss() {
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
> allows resolving v1 styles

what do you mean? :sweat_smile:

patricklx:
> > allows resolving v1 styles
> 
> what do you mean? 😅

There are.scss or .less files in some addons, which currently can only be consumed because they are copied to rewritten-app/assets.

But i could now do something like this in vite:
```scss 
// my-app/app.scss
@use 'my-v1-addon'
```

PR #1854, opened 2024-03-24
add vite compat styles resolver

patricklx:
There are.scss or .less files in some v1 addons, which currently can only be consumed because they are copied to rewritten-app/assets.

But i could now do something like this in vite:
```scss 
// my-app/app.scss
@use 'my-v1-addon'
```
And have app.js import app.scss


ef4:
Discussed this together in office hours.

This is implementing a feature that might be useful for people who are ready to completely opt out of the classic style pipeline. But it doesn't mix well with that pipeline.

Today, sass from v1 addons does work in the classic `app/styles/app.scss` and gets handled entirely in the broccoli-based prebuild for compatibility. This is different from configuring vite to directly build sass, which does not have access to sass from v1 addons.

An `@embroider/sass` plugin that uses embroider's resolver to locate files in classic addons would be a potentially useful new addition, but would be mutually exclusive with the use of `@embroider/synthesized-vendor`.

PR #1855, opened 2024-03-26
addon-dev: incremental updates to output 

patricklx:
this now only updates changed files in dist. And deletes removed files in dist.
otherwise all files are deleted on each change, which makes HMR less useful

### DIFF
--- a/packages/addon-dev/src/rollup-gjs-plugin.ts
+++ b/packages/addon-dev/src/rollup-gjs-plugin.ts
@@ -1,6 +1,5 @@
 import { createFilter } from '@rollup/pluginutils';
 import type { Plugin } from 'rollup';
-import { readFileSync } from 'fs';
 import { Preprocessor } from 'content-tag';
 
 const PLUGIN_NAME = 'rollup-gjs-plugin';
@@ -14,11 +13,10 @@ export default function rollupGjsPlugin(
   return {
     name: PLUGIN_NAME,
 
-    load(id: string) {
+    transform(input: string, id: string) {
       if (!gjsFilter(id)) {
         return null;
       }
-      let input = readFileSync(id, 'utf8');
       let code = processor.process(input, {
         filename: id,
         inline_source_map,
--- a/packages/addon-dev/src/rollup-hbs-plugin.ts
+++ b/packages/addon-dev/src/rollup-hbs-plugin.ts
@@ -1,8 +1,15 @@
+import type { Plugin, PluginContext } from 'rollup';
 import { createFilter } from '@rollup/pluginutils';
-import type { Plugin, PluginContext, CustomPluginOptions } from 'rollup';
-import { readFileSync } from 'fs';
-import { correspondingTemplate, hbsToJS } from '@embroider/core';
 import minimatch from 'minimatch';
+import {
+  hbsToJS,
+  templateOnlyComponentSource,
+  needsSyntheticComponentJS,
+  syntheticJStoHBS,
+} from '@embroider/core';
+import { extname } from 'path';
+
+const hbsFilter = createFilter('**/*.hbs?([?]*)');
 
 export default function rollupHbsPlugin({
   excludeColocation,
@@ -12,48 +19,88 @@ export default function rollupHbsPlugin({
   return {
     name: 'rollup-hbs-plugin',
     async resolveId(source: string, importer: string | undefined, options) {
+      if (options.custom?.embroider?.isExtensionSearch) {
+        return null;
+      }
+
       let resolution = await this.resolve(source, importer, {
         skipSelf: true,
-        ...options,
       });
 
-      if (resolution) {
-        return resolution;
-      } else {
-        return maybeSynthesizeComponentJS(
-          this,
-          source,
-          importer,
-          options,
-          excludeColocation
+      if (!resolution && extname(source) === '') {
+        resolution = await this.resolve(source + '.hbs', importer, {
+          skipSelf: true,
+        });
+      }
+
+      if (!resolution) {
+        let hbsSource = syntheticJStoHBS(source);
+        if (hbsSource) {
+          resolution = await this.resolve(hbsSource, importer, {
+            skipSelf: true,
+            custom: {
+              embroider: {
+                isExtensionSearch: true,
+              },
+            },
+          });
+        }
+
+        if (!resolution) {
+          return null;
+        }
+      }
+
+      if (resolution && resolution.id.endsWith('.hbs')) {
+        let isExcluded = excludeColocation?.some((glob) =>
+          minimatch(resolution!.id, glob)
         );
+        if (isExcluded) {
+          return resolution;
+        }
+      }
+
+      let syntheticId = needsSyntheticComponentJS(source, resolution.id);
+      if (syntheticId) {
+        this.addWatchFile(source);
+        return {
+          id: syntheticId,
+          meta: {
+            'rollup-hbs-plugin': {
+              type: 'template-only-component-js',
+            },
+          },
+        };
       }
     },
 
     load(id: string) {
-      if (hbsFilter(id)) {
-        return getHbsToJSCode(id);
-      }
-      let meta = getMeta(this, id);
-      if (meta) {
-        if (meta?.type === 'template-js') {
-          const hbsFile = id.replace(/\.js$/, '.hbs');
-          return getHbsToJSCode(hbsFile);
-        }
+      if (getMeta(this, id)?.type === 'template-only-component-js') {
+        this.addWatchFile(id);
         return {
-          code: templateOnlyComponent,
+          code: templateOnlyComponentSource(),
         };
       }
+    },
+
+    transform(code: string, id: string) {
+      let hbsFilename = id.replace(/\.\w{1,3}$/, '') + '.hbs';
+      if (hbsFilename !== id) {
+        this.addWatchFile(hbsFilename);
+        if (getMeta(this, id)?.type === 'template-only-component-js') {
+          this.addWatchFile(id);
+        }
+      }
+      if (!hbsFilter(id)) {
+        return null;
+      }
+      return hbsToJS(code);
     },
   };
 }
 
-const templateOnlyComponent =
-  `import templateOnly from '@ember/component/template-only';\n` +
-  `export default templateOnly();\n`;
-
 type Meta = {
-  type: 'template-only-component-js' | 'template-js';
+  type: 'template-only-component-js';
 };
 
 function getMeta(context: PluginContext, id: string): Meta | null {
@@ -64,44 +111,3 @@ function getMeta(context: PluginContext, id: string): Meta | null {
     return null;
   }
 }
-
-function getHbsToJSCode(file: string): { code: string } {
-  let input = readFileSync(file, 'utf8');
-  let code = hbsToJS(input);
-  return {
-    code,
-  };
-}
-
-async function maybeSynthesizeComponentJS(
-  context: PluginContext,
-  source: string,
-  importer: string | undefined,
-  options: { custom?: CustomPluginOptions; isEntry: boolean },
-  excludeColocation: string[] | undefined
-) {
-  let hbsFilename = correspondingTemplate(source);
-  let templateResolution = await context.resolve(hbsFilename, importer, {
-    skipSelf: true,
-    ...options,
-  });
-  if (!templateResolution) {
-    return null;
-  }
-  let type = excludeColocation?.some((glob) => minimatch(hbsFilename, glob))
-    ? 'template-js'
-    : 'template-only-component-js';
-  // we're trying to resolve a JS module but only the corresponding HBS
-  // file exists. Synthesize the JS. The meta states if the hbs corresponds
-  // to a template-only component or a simple template like a route template.
-  return {
-    id: templateResolution.id.replace(/\.hbs$/, '.js'),
-    meta: {
-      'rollup-hbs-plugin': {
-        type,
-      },
-    },
-  };
-}
-
-const hbsFilter = createFilter('**/*.hbs');
--- a/packages/addon-dev/src/rollup-incremental-plugin.ts
+++ b/packages/addon-dev/src/rollup-incremental-plugin.ts
@@ -1,0 +1,79 @@
+import walkSync from 'walk-sync';
+import { rmSync } from 'fs';
+import { join } from 'path';
+import type { Plugin } from 'rollup';
+import { existsSync } from 'fs-extra';
+
+export default function incremental(): Plugin {
+  const generatedAssets = new Map();
+  const generatedFiles = new Set<string>();
+
+  function isEqual(v1: string | Uint8Array, v2: string | Uint8Array): boolean {
+    if (typeof v1 === 'string' && typeof v2 === 'string') {
+      return v1 === v2;
+    }
+    if (Buffer.isBuffer(v1) && Buffer.isBuffer(v2)) {
+      return v1.equals(v2);
+    }
+    return false;
+  }
+
+  let firstTime = true;
+
+  function initGeneratedFiles(outDir: string) {
+    if (existsSync(outDir)) {
+      const files = walkSync(outDir, {
+        globs: ['*/**'],
+        directories: false,
+      });
+      for (const file of files) {
+        generatedFiles.add(file);
+      }
+    }
+  }
+
+  function deleteRemovedFiles(bundle: Record<string, any>, outDir: string) {
+    for (const file of generatedFiles) {
+      if (!bundle[file]) {
+        generatedAssets.delete(file);
+        rmSync(join(outDir, file));
+      }
+    }
+    generatedFiles.clear();
+    for (const file of Object.keys(bundle)) {
+      generatedFiles.add(file);
+    }
+  }
+
+  function syncFiles(bundle: Record<string, any>) {
+    for (const checkKey of Object.keys(bundle)) {
+      if (bundle[checkKey]) {
+        let module = bundle[checkKey] as any;
+        let code = module.source || module.code;
+        if (
+          generatedAssets.has(checkKey) &&
+          isEqual(code, generatedAssets.get(checkKey))
+        ) {
+          delete bundle[checkKey];
+        } else {
+          generatedAssets.set(checkKey, code);
+        }
+      }
+    }
+  }
+
+  return {
+    name: 'incremental',
+    generateBundle(options, bundle) {
+      if (firstTime) {
+        firstTime = false;
+        initGeneratedFiles(options.dir!);
+      }
+      if (existsSync(options.dir!)) {
+        deleteRemovedFiles(bundle, options.dir!);
+      }
+
+      syncFiles(bundle);
+    },
+  };
+}
--- a/packages/addon-dev/src/rollup.ts
+++ b/packages/addon-dev/src/rollup.ts
@@ -2,14 +2,13 @@ import { default as hbs } from './rollup-hbs-plugin';
 import { default as gjs } from './rollup-gjs-plugin';
 import { default as publicEntrypoints } from './rollup-public-entrypoints';
 import { default as appReexports } from './rollup-app-reexports';
-import type { Options as DelOptions } from 'rollup-plugin-delete';
-import { default as clean } from 'rollup-plugin-delete';
 import { default as keepAssets } from './rollup-keep-assets';
 import { default as dependencies } from './rollup-addon-dependencies';
 import {
   default as publicAssets,
   type PublicAssetsOptions,
 } from './rollup-public-assets';
+import { default as clean } from './rollup-incremental-plugin';
 import type { Plugin } from 'rollup';
 
 export class Addon {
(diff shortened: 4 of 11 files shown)


NullVoxPopuli (dismissed):
Logically, it seems good.
Do we have any tests for this behavior? I don't remember: 😅

mansona (changes requested):
so one thing that I would like to do before we actually move forward on this is that we verify that this Plugin works correctly with the bug scenario that https://github.com/embroider-build/embroider/pull/1448 was originally trying to fix. Ideally we would cover this in the v2 addon dev watch tests

patricklx:
> so one thing that I would like to do before we actually move forward on this is that we verify that this Plugin works correctly with the bug scenario that #1448 was originally trying to fix. Ideally we would cover this in the v2 addon dev watch tests

I think what this actually does now is more like a `sync`. When a file gets deleted in src, the corresponding generated file in dist also gets deleted.
there is the option `runOnce`, which only deletes the files only once at start. Which is already better, but it leaves the deleted files during dev.

mansona:
> I think this also fixes the original issue

The problem I have with this PR is that I want more confidence that it fixes the original issue before we move forward with it. Myself and @NullVoxPopuli tried to recreate the issue on the Triage meeting and we couldn't get it to fail so we can't tell if this makes any difference 🙃 

I would really like a failing test that shows the problem in a repeatable way before we try to fix it by altering the underlying implementation. If this is not possible (because maybe it's an OS level race condition 🤷) then at the very least we need to coordinate with @simonihmig because he said he would be looking into the original issue over the next while: https://github.com/embroider-build/addon-blueprint/issues/32#issuecomment-2079582665

patricklx:
I'm not sure how to add a test for this

patricklx:
@mansona i improved this even more and added some tests

packages/addon-dev/src/rollup-incremental-plugin.ts
```diff
@@ -0,0 +1,70 @@
...
+          checkKey = key.replace('.js.map', '.js');
+          if (!bundle[checkKey]) {
+            delete bundle[key];
+            continue;
+          }
+        }
+        if (
+          bundle[checkKey]?.type === 'asset' &&
+          (bundle[checkKey] as OutputAsset).source ===
+            generatedAssets.get(checkKey)
```
simonihmig, line 48: Trying this locally, this if clause was not catching a lot of asset files as expected, causing them to be written to disk even when unchanged. The issue seems to be that depending on the file type, `.source` could point to a Buffer, in which case you would have to compare them via [.equals()](https://nodejs.org/api/buffer.html#bufequalsotherbuffer).

packages/addon-dev/src/rollup-incremental-plugin.ts
```diff
@@ -0,0 +1,70 @@
...
+        ) {
+          delete bundle[key];
+          continue;
+        } else {
+          generatedAssets.set(
```
simonihmig, line 53: This else block would also be executed for non-asset files, which is not expected, right?
I would think there should be an out-if checking of the current file is an asset, and another inner-if that checks for quality, so that this block is only reachable for assets that are not equal to the previous run.

simonihmig (changes requested):
I tested this locally with a real project and `yarn link`. I found some minor issues, see inline comments. But the good part is that this seems to also fix some of the issues discussed in https://github.com/embroider-build/addon-blueprint/issues/32#issuecomment-2079582665.

The main issue I was hoping to get fixed by a revised clean plugin was the fact that an app watching the dist folder of an addon would often pick up changes as they were written but before the addon's rebuild has finished. 
Leading to multiple app rebuilds, with the first ones often picking up interim build errors. Which would quickly go away, when the addon's rebuild has finished, but that's still annoying...

And testing this change here locally indeed seems to fix that issue! :tada:

packages/addon-dev/src/rollup.ts
```diff
@@ -64,10 +63,11 @@ export class Addon {
...
 
-  // By default rollup does not clear the output directory between builds. This
-  // does that.
-  clean(options: DelOptions) {
-    return clean({ targets: `${this.#destDir}/*`, ...options });
+  // this does incremental updates to the dist files and also deletes files that are not part of the generated bundle
+  // rollup already supports incremental transforms of files,
+  // this extends it to the dist files
+  clean() {
+    return clean();
```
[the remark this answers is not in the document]
simonihmig: If we consider this breaking, we could still release this on stable as a minor version, by exposing this as a separate plugin, and updating the blueprint to use the new one instead of `addon.clean()`! 🤔

tests/scenarios/v2-addon-dev-watch-test.ts
```diff
@@ -49,15 +50,15 @@ Scenarios.fromProject(() => baseV2Addon())
           output: addon.output(),
 
           plugins: [
-            addon.publicEntrypoints(['components/**/*.js']),
-            addon.appReexports(['components/**/*.js']),
+            addon.publicEntrypoints(['components/**/*.{gts,js}']),
+            addon.appReexports(['components/**/*.{gts,js}']),
```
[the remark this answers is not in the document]
patricklx: Ah, maybe it wasn't needed

packages/addon-dev/src/rollup-incremental-plugin.ts
```diff
@@ -0,0 +1,70 @@
...
+      // could also be done directly in the babel plugin
+      // by passing rollup context into it
+      let hbsFilename = id.replace(/\.\w{1,3}$/, '') + '.hbs';
+      if (hbsFilename !== id && existsSync(hbsFilename)) {
+        this.addWatchFile(hbsFilename);
+      }
+    },
+    generateBundle(options, bundle) {
+      if (existsSync(options.dir!)) {
+        const files = walkSync(options.dir!, {
```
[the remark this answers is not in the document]
patricklx: It's required to delete files that are not part of the bundle

packages/addon-dev/src/rollup-incremental-plugin.ts
```diff
@@ -0,0 +1,70 @@
...
+          globs: ['*/**'],
+          directories: false,
+        });
+        for (const file of files) {
+          if (!bundle[file]) {
+            generatedAssets.delete(file);
+            rmSync(join(options.dir!, file));
```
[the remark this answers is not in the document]
patricklx: generated assets do not pass through transform. So it's not detectable

packages/addon-dev/src/rollup-gjs-plugin.ts
```diff
@@ -14,11 +13,10 @@ export default function rollupGjsPlugin(
   return {
     name: PLUGIN_NAME,
 
-    load(id: string) {
+    transform(input: string, id: string) {
```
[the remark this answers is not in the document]
patricklx: It's required to have hbs pass through transform to have thrm incremental.

packages/addon-dev/src/rollup-incremental-plugin.ts
```diff
@@ -42,14 +53,22 @@ export default function incremental(): Plugin {
...
+              generatedAssets.get(checkKey)
+            )
+          ) {
             delete bundle[key];
             continue;
           } else {
             generatedAssets.set(
               checkKey,
-              (bundle[checkKey] as OutputAsset).source
+              (bundle[checkKey] as OutputAsset).source.toString()
```
simonihmig, line 71: Isn't that string casting breaking your `isEqual` implementation? Like when the plugin sees a jpeg asset, `(bundle[checkKey] as OutputAsset).source` will be a Buffer, but you put it into `generatedAssets` as a string, so `isEqual` will always return false, won't it?

patricklx: your right! thanks

packages/addon-dev/src/rollup-incremental-plugin.ts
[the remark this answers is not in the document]
patricklx: I improved this ti only walk once

tests/scenarios/scenarios.ts
```diff
@@ -1,5 +1,28 @@
-import { Scenarios, Project } from 'scenario-tester';
+import { Project, Scenario, Scenarios } from 'scenario-tester';
 import { dirname } from 'path';
+import { exec } from 'child_process';
+
+let prepare = Scenario.prototype.prepare;
```
patricklx, line 5: waiting on https://github.com/stefanpenner/node-fixturify-project/pull/98

packages/addon-dev/src/rollup-hbs-plugin.ts
```diff
@@ -30,30 +29,34 @@ export default function rollupHbsPlugin({
...
-  `export default templateOnly();\n`;
+function templateOnlyComponent(hbsCode: string) {
+  const code = hbsCode.replace(/`/g, '\\`').replace(/\$/g, '\\$');
+  return `
+  import templateOnly from '@ember/component/template-only';
+  import { precompileTemplate } from '@ember/template-compilation';
+  import { setComponentTemplate } from '@ember/component';
+  export default setComponentTemplate(precompileTemplate(\`${code}\`), templateOnly());
+  `;
+}
```
ef4, line 55: We were avoiding this kind of inlining on purpose. I would prefer to keep letting rollup do the inlining automatically.

packages/addon-dev/src/rollup-incremental-plugin.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  return {
+    name: 'clean',
+    transform(_code, id) {
+      changed.add(id);
+      // support colocation changes
+      // could also be done directly in the babel plugin
+      // by passing rollup context into it
+      let hbsFilename = id.replace(/\.\w{1,3}$/, '') + '.hbs';
+      if (hbsFilename !== id && existsSync(hbsFilename)) {
```
ef4, line 95: This feature doesn't belong here, because the dependency was created by the hbs plugin but is being patched in by the incremental plugin.

And the only reason you need this is the manual inlining that I mentioned above. It should not be needed at all.

packages/addon-dev/src/rollup-gjs-plugin.ts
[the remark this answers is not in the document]
ef4: I'm glad to use `transform` over `load` whenever possible. `transform` composes more easily with other plugins. 👍

ef4 (changes requested):

packages/addon-dev/src/rollup-hbs-plugin.ts
patricklx: I do not see how it can be done when using `transform` in the hbs plugin.

the hbs gets transformed by embroiders babel plugin to
e.g. if we get the file `demo.hbs`
we would have
```js
import default as TEMPLATE from './demo.hbs'
export default setComponentTemplate(precompileTemplate(TEMPLATE), templateOnly());
```

but ` './demo.hbs'` is already cached with the transform, so we would get the same thing back

patricklx: i fixed it

packages/addon-dev/src/rollup-incremental-plugin.ts
patricklx: i moved it to the hbs plugin. adding watch file is required when deleting or creating files that take part in the colocation

tests/scenarios/compat-stage2-test.ts
```diff
@@ -40,6 +40,9 @@ stage2Scenarios
         'check-resolution-target.js': 'export {}',
       },
     });
+
+    (depC.pkg['ember-addon'] as any).before = ['dep-a', 'dep-b', 'in-repo-d', 'in-repo-a'];
```
simonihmig, line 44: Why do we have this?

tests/scenarios/v2-addon-dev-test.ts
```diff
@@ -73,6 +73,8 @@ appScenarios
               exclude: ['**/-excluded/**/*'],
             }),
 
+            addon.clean(),
```
simonihmig, line 76: Is this change in order required? If so, then this is clearly a breaking change!

tests/scenarios/v2-addon-dev-watch-test.ts
```diff
@@ -134,6 +137,139 @@ Scenarios.fromProject(() => baseV2Addon())
...
+              // to guess for how long it'll take for the file system to update our file.
+              //
+              // the `stat` is measured in `ms`, so it's still pretty fast
+              await fs.writeFile(demoHbs, someContent + `\n\n`);
+              await aBit(10);
+              await watcher?.nextBuild();
+            },
+          });
+
+          // updating hbs content should update resulting app re-exported component
```
simonihmig, line 221: Is this a typo, or am I misunderstanding?

```suggestion
          // updating hbs content should not update resulting app re-exported component
```

tests/scenarios/v2-addon-dev-watch-test.ts
```diff
@@ -134,6 +137,139 @@ Scenarios.fromProject(() => baseV2Addon())
...
+          await fs.rm(demoJs);
+          await watcher?.nextBuild();
+          let distPathDemoCompContent = await fs.readFile(distPathDemoComp);
+          assert.ok(distPathDemoCompContent.includes('templateOnly'));
+
+          // creating demo.js should make demo a template colocated component
+          await fs.writeFile(demoJs, demoJsContent);
+          await watcher?.nextBuild();
+          distPathDemoCompContent = await fs.readFile(distPathDemoComp);
+          assert.ok(!distPathDemoCompContent.includes('templateOnly'));
```
simonihmig, line 270: nit pick:

```suggestion
          assert.false(distPathDemoCompContent.includes('templateOnly'));
```

(this was initially confusing to me, as it took me a while to see the `!` 😅 )

tests/scenarios/v2-addon-dev-watch-test.ts
```diff
@@ -134,6 +137,139 @@ Scenarios.fromProject(() => baseV2Addon())
...
+              await watcher?.nextBuild();
+            },
+          });
+
+          // deleting demo.js should make demo a template only component
+          const demoJsContent = await fs.readFile(demoJs);
+          await fs.rm(demoJs);
+          await watcher?.nextBuild();
+          let distPathDemoCompContent = await fs.readFile(distPathDemoComp);
+          assert.ok(distPathDemoCompContent.includes('templateOnly'));
```
simonihmig, line 264: nit pick:

```suggestion
          assert.true(distPathDemoCompContent.includes('templateOnly'));
```

tests/scenarios/v2-addon-dev-watch-test.ts
```diff
@@ -134,6 +137,139 @@ Scenarios.fromProject(() => baseV2Addon())
...
+              // the `stat` is measured in `ms`, so it's still pretty fast
+              await fs.writeFile(demoHbs, someContent + `\n`);
+              await aBit(10);
+              await watcher?.nextBuild();
+            },
+          });
+
+          // updating template only hbs should update the dist output
+          distPath = path.join(addon.dir, 'dist/components/button.js');
+          await isNotModified({
```
simonihmig, line 242: why is this _not_ getting modified?

tests/scenarios/compat-stage2-test.ts
patricklx: Using windows real path instead of short path alias caused this

tests/scenarios/v2-addon-dev-test.ts
patricklx: I think it is. I need to revalidate

tests/scenarios/v2-addon-dev-watch-test.ts
patricklx: that's actually correct. The rexport file did not change.
Wrong comment

patricklx: it should update---

packages/addon-dev/src/rollup.ts
```diff
@@ -2,14 +2,13 @@ import { default as hbs } from './rollup-hbs-plugin';
 import { default as gjs } from './rollup-gjs-plugin';
 import { default as publicEntrypoints } from './rollup-public-entrypoints';
 import { default as appReexports } from './rollup-app-reexports';
-import type { Options as DelOptions } from 'rollup-plugin-delete';
-import { default as clean } from 'rollup-plugin-delete';
```
ef4, line 6: Let's remove this from package.json. I think this was probably the only use.

packages/addon-dev/src/rollup-hbs-plugin.ts
```diff
@@ -30,9 +30,24 @@ export default function rollupHbsPlugin({
...
 
+    transform(code: string, id: string) {
+      let hbsFilename = id.replace(/\.\w{1,3}$/, '') + '.hbs';
+      if (hbsFilename !== id) {
+        this.addWatchFile(hbsFilename);
+      }
+      if (hbsFilter(id)) {
+        let basename = id.replace(/\.\w{1,3}$/, '');
+        this.addWatchFile(basename + '.ts');
+        this.addWatchFile(basename + '.js');
```
ef4, line 41: I think these belong beside the spot where we actually attempted to resolve the corresponding JS files in `maybeSynthesizeJS`. Having them together may address the caching behaviors you mentioned that caused you to have to move code between load and transform phases.

patricklx: I tried to move it to the resolve hook. But then the transform for js files does not get updated. It might need to be in the transform hook for the cache to be invalidated

tests/scenarios/v2-addon-dev-test.ts
simonihmig: Any news on this @patricklx? 

With ~250 addons in our monorepo (not all v2 though), I'd rather not have to write a codemod to change the order! 😏

patricklx: hi @simonihmig I reverted the change and the tests passed. (other unrelated failed)

patricklx: hold that, i forgot one, lets see

simonihmig: There was a timeout in one test job. I restarted and it is all green now! :tada: 

Thanks for getting back to this, I think this revert is helpful, as it shows we are not introducing breaking changes here!

simonihmig (approved):
LGTM!

I think we are close to getting this merged finally,. Thanks for working on this, Patrick, I am looking forward to get to use this!

@ef4 @mansona @NullVoxPopuli wanna give this a final review?

AskerFED:
image

I have faced the ENOENT error for the addon.clean() which triggers file deletion event while rollup watch but it is not build stopping issue just error thrown and started build once dist was replaced. After @NullVoxPopuli suggestion to try this addon-dev branch now ENOENT error resolved. But the below warning is shown but is harmless I guess...

image

packages/addon-dev/src/rollup-gjs-plugin.ts
patricklx, line 16: @ef4 I noticed that now the gjs plugin must come before the babel plugin in the rollup config. That would make this a breaking change

packages/addon-dev/src/rollup-hbs-plugin.ts
```diff
@@ -12,6 +12,10 @@ export default function rollupHbsPlugin({
   return {
     name: 'rollup-hbs-plugin',
     async resolveId(source: string, importer: string | undefined, options) {
+      let hbsFilename = source.replace(/\.\w{1,3}$/, '') + '.hbs';
+      if (hbsFilename !== source) {
+        this.addWatchFile(hbsFilename);
```
patricklx, line 17: @ef4 I had to add addWatchFile to all 3 methods to get this to work correctly on windows

ef4:
This is hard for me to follow. Can we take advantage of the cleanup that has happened in the vite hbs plugin to use the same patterns here?

patricklx:
> This is hard for me to follow. Can we take advantage of the cleanup that has happened in the vite hbs plugin to use the same patterns here?

I updated the hbs plugin to be similar to the one in vite hbs.
only a with a few changes. there is no generic extension search setup, it could be setup.
I added the extension search for .hbs extension.

also added the addWatchFile calls to get tests working

packages/addon-dev/src/rollup-hbs-plugin.ts
```diff
@@ -12,48 +19,88 @@ export default function rollupHbsPlugin({
...
+      if (resolution && resolution.id.endsWith('.hbs')) {
+        let isExcluded = excludeColocation?.some((glob) =>
+          minimatch(resolution!.id, glob)
         );
+        if (isExcluded) {
+          return resolution;
+        }
+      }
+
+      let syntheticId = needsSyntheticComponentJS(source, resolution.id);
```
patricklx, line 63: we do not have a packageCache here

packages/addon-dev/src/rollup-hbs-plugin.ts
```diff
@@ -63,6 +63,7 @@ export default function rollupHbsPlugin({
       let syntheticId = needsSyntheticComponentJS(source, resolution.id);
       if (syntheticId) {
         this.addWatchFile(source);
+        this.addWatchFile(syntheticId);
```
patricklx: this change did not work @ef4 .
I think transform does have its own specific cache.
https://rollupjs.org/plugin-development/#transform

packages/vite/src/hbs.ts
```diff
@@ -58,8 +59,8 @@ export function hbs(): Plugin {
         }
       }
 
-      let syntheticId = needsSyntheticComponentJS(source, resolution.id, resolverLoader.resolver.packageCache);
-      if (syntheticId) {
+      let syntheticId = needsSyntheticComponentJS(source, resolution.id);
+      if (syntheticId && isInComponents(resolution.id, resolverLoader.resolver.packageCache)) {
```
patricklx, line 63: @ef4 is this okay like this?

ef4 (approved):

Issue #1776, opened 2024-01-31
Add an option for hbs files to be considered route templates and compiled as such


balinterdi:
### Problem description

(I'm taking up the thread from #1756 as requested by @mansona during an Embroider Initiative call.)

Currently, all `hbs` files in v2 addons are considered component templates. If there's no co-located `.js` file, [they will be considered a template-only component](https://github.com/embroider-build/embroider/blob/289f83d80d133da3de2b342e63f937986248c07d/packages/addon-dev/src/rollup-hbs-plugin.ts#L79-L88), otherwise `setComponentTemplate` will be called with the template content and the component class.

This prevents using *route templates* (which are also `.hbs`) in v2 addons but, at least in the initial, migration period there exist apps which would need that feature (the need came from an actual client project).

On the longer term, using a gjs [`ember-route-template`](https://github.com/discourse/ember-route-template) elegantly side-steps that problem (I have a small, working example [here](https://github.com/balinterdi/app-files-from-v2-addons/tree/use-ember-route-template-to-make-route-template-available)), but since it requires the template to be in strict-mode, migrating existing route templates to ember-route-templates is quite an undertaking.

In #1756 I provided a working modification to Embroider and a simple Rollup plugin which allowed compiling route templates. However, it had a few drawbacks:

- It had a hard-wired path to a folder in which templates are considered route templates
- It produced a built output which was "wrong" in the sense that it contained a line that made no sense (see the `setComponentTemplate` call in the below code snippet), but it still worked!

```
import { precompileTemplate } from '@ember/template-compilation';
import { setComponentTemplate } from '@ember/component';

var TEMPLATE = precompileTemplate("The Caro-Kann Opening\n\n<div class=\"image-container\">\n  <img src=\"/grand-prix/assets/images/caro-kann.webp\" alt=\"Caro Kann Opening\" />\n</div>");

var caroKann = setComponentTemplate(TEMPLATE, precompileTemplate("The Caro-Kann Opening\n\n<div class=\"image-container\">\n  <img src=\"/grand-prix/assets/images/caro-kann.webp\" alt=\"Caro Kann Opening\" />\n</div>"));

export { caroKann as default };
```

### Possible solution

In the same PR, [@ef4 suggested](https://github.com/embroider-build/embroider/pull/1756#issuecomment-1895912346) to allow "people "o provide globs for including or excluding files from being treated as template-only components.", which I think is a great idea.

[...]

PR #1856, opened 2024-03-26
Compile Hbs route templates correctly

BlueCutOfficial:
Relates to #1776

**Problem**
In `main` branch, any hbs that is not colocated with its js is compiled as a template-only component. This behavior breaks the possibility for addons to provide hbs templates that are not components, like route templates.

**Solution**
This PR introduces the possibility of specifying in config the hbs files that should not be compiled as template-only components. 

Two different layers transform the standalone `.hbs` in the codebase into a template-only component `.js` in the addon build: the Rollup plugin `rollup-hbs-plugin` and the Babel plugin `template-colocation-plugin`. As Rollup config and Babel config are currently not shared, the glob patterns that prevent this compilation from happening will have to be specified for both tools.

In the `rollup.config.mjs` of the v2 addon:
```diff 
// Ensure that standalone .hbs files are properly integrated as Javascript.
- addon.hbs(),
+ addon.hbs({
+  excludeColocation: ['templates/**/*'] // all files in the templates folder are not colocated components
+ })
```

In the `babel.config.json` of the v2 addon:
```diff
"plugins": [
- "@embroider/addon-dev/template-colocation-plugin",
+ ["@embroider/addon-dev/template-colocation-plugin", {
+   exclude: ['templates/**/*'] 
+ }],
],
```

**Required refactorings**
Helpers that are not provided by the addon itself need to be explicitly imported.

For instance, let's assume your v1 addon used to have a dependency to [ember-root-url](https://github.com/ef4/ember-root-url) and the helper was called in a hbs route template:
```hbs
<img src={{root-url "images/hello.png"}} />
```
You'll be able to build the v2 addon, but when starting the app you'll run into `Attempted to resolve 'root-url', which was expected to be a component or helper, but nothing was found.`

To get rid of the error, one solution is to import the helper into the route controller so it's passed down to the template through the context:
```js
import Controller from '@ember/controller';
import rootUrl from 'ember-root-url/helpers/root-url';

export default class ApplicationController extends Controller {
  rootUrl = rootUrl;
}
```
```hbs
<img src={{this.root-url "images/hello.png"}} />
```
(This example doesn't apply to built-in helpers like the `{{component}}` helper, the hbs route template your v2 addon provides can use such a helper and it will continue to work with your classic Ember app.)

### DIFF
--- a/docs/v2-faq.md
+++ b/docs/v2-faq.md
@@ -17,6 +17,7 @@
   * [Why do v2 addons need a build step?](#why-do-v2-addons-need-a-build-step)
   * [How can I integrate with the app's build?](#how-can-i-integrate-with-the-apps-build)
   * [How can I define the public exports of my addon?](#how-can-i-define-the-public-exports-of-my-addon)
+  * [How can I provide route templates with my v2 addon?](#how-can-i-provide-route-templates-with-my-v2-addon)
 
 <!-- tocstop -->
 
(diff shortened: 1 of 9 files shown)


mansona (changes requested):
This looks great 🎉 it's ready to go on the tech side, we need at least some basic docs before we can merge though. It doesn't need to be too in-depth, even a placeholder will do for now.

mansona (approved):
Looks great 🎉

PR #1858, opened 2024-03-28
Add support for `{{unless}}` to the `macroCondition` macro

Windvis:
I noticed the `macroCondition` macro could only be used with the `{{if}}` helper / keyword. This change makes it possible to use `{{unless}}` as well.

(Not a very DRY change, but it seemed better to keep things simple and duplicate the if version 😄.)

### DIFF
--- a/packages/macros/src/glimmer/ast-transform.ts
+++ b/packages/macros/src/glimmer/ast-transform.ts
@@ -2,7 +2,14 @@ import literal from './literal';
 import getConfig from './get-config';
 import dependencySatisfies from './dependency-satisfies';
 import { maybeAttrs } from './macro-maybe-attrs';
-import { macroIfBlock, macroIfExpression, macroIfMustache } from './macro-condition';
+import {
+  macroIfBlock,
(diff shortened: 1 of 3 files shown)


ef4 (approved):
Thanks, this looks nice.

Windvis:
@ef4 is there anything that needs to happen before this can be merged or did this simply fall through the cracks 😄?

mansona:
@Windvis one thing that might be slowing this down is that you're targeting main with this change 🤔 are you intending to make use of this soon? `main` isn't going to be released any time soon and there are a lot of breaking changes, if you want to use this soon then I would target `stable`

mansona:
@Windvis once things are released in stable we forward-port them to main. It tends to be a lot easier to do it this way, also since we're keeping things up to date often enough 👍 

Essentially you would not need to do anything extra, it would be on us :)

ef4:
Let's try to retarget this to `stable`, @mansona has volunteered, it looks like it will apply cleanly.

PR #1859, opened 2024-03-29
Public assets handling/ clean up stage 2

BlueCutOfficial:
Relates to #1736 

Thanks to the new `assets` plugin, Vite can now serve and emit the public assets provided by the Ember app and addons without using the rewritten-app. It means that copying these assets in the rewritten-app has become an unnecessary step. 

This PR does the clean-up by removing the related code from the `compat-app-builder` along with the tests that use to assert the presence of the public assets in the rewritten-app.

### DIFF
--- a/packages/compat/src/compat-app-builder.ts
+++ b/packages/compat/src/compat-app-builder.ts
@@ -822,21 +822,6 @@ export class CompatAppBuilder {
   private gatherAssets(inputPaths: OutputPaths<TreeNames>): Asset[] {
     // first gather all the assets out of addons
     let assets: Asset[] = [];
-    for (let pkg of this.allActiveAddons) {
-      if (pkg.meta['public-assets']) {
-        for (let [filename, appRelativeURL] of Object.entries(pkg.meta['public-assets'] || {})) {
-          let sourcePath = resolvePath(pkg.root, filename);
-          let stats = statSync(sourcePath);
-          assets.push({
-            kind: 'on-disk',
-            sourcePath,
-            relativePath: appRelativeURL,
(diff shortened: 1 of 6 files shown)


BlueCutOfficial:
[canary-compat-dummy-app-tests](https://github.com/embroider-build/embroider/actions/runs/8566905008/job/23477683446?pr=1859) concerns a `robots.txt` file that is read directly in the rewritten-app. However the comment is not quite clear to me and I don't know if the test should be removed or adjusted to the new way of doing things.

packages/compat/src/compat-app-builder.ts
```diff
@@ -855,19 +837,19 @@ export class CompatAppBuilder {
...
+      for (let [filename, appRelativeURL] of Object.entries(pkg.meta['public-assets'] || {})) {
+        let sourcePath = resolvePath(pkg.root, filename);
+        let stats = statSync(sourcePath);
+        assets.push({
+          kind: 'on-disk',
+          sourcePath,
+          relativePath: appRelativeURL,
+          mtime: stats.mtimeMs,
+          size: stats.size,
+        });
```
BlueCutOfficial, line 852: I would have deleted this whole block so we no longer generate the public-assets coming from addons in the rewritten-app. 

However, it seems the CSS of the app itself (`dist/assets/my-app.css`) is generated here out of the `@embroider/synthesized-styles` package. So I left the handling we had before for this particular package.

mansona (approved):
This is great 🎉 I am feeling very suspicious of the synthesised-styles 🙈 but we should probably fix that in a new PR since it's going to be a bunch of work to unravel 👍

PR #1864, opened 2024-04-04
Make the addon-template use vite

BlueCutOfficial:
Apps used in test scenarios are built out of an app template. Recently, one of these templates, `app-template`, has been modified so the output apps are built with vite. But `addon-template` hasn't been modified the same way. With the work on virtualization, we start to see test cases where we require the `addon-template` to output a dummy app building with vite as well.

This PR aligns `addon-template` with `app-template` to use vite:
- aligns the dependencies
- aligns vite and babel configs (+ ignores prettier)
- change the test so there's no helper to be discovered at runtime (this no longer work because vite build use `staticHelpers`)


mansona (approved):
LGTM 💪

Issue #1862, opened 2024-04-02
Ability to control order of addon implicit styles?


raycohen:
In a classic build, my addons css was being concatenated into vendor.css in alphabetical order based on package name, which I think is a result of file system behavior.

With embroider, I'm getting a very different order that I don't fully understand. I tried using package.json "ember-addon"."before"/"after" config which didn't seem to have any impact.

I have a very large monorepo with ~100 addons, so migrating fully off of implicit styles would be a large effort. But with the changed css order we've found issues where same-specificity rules changed order and a different one is taking precedence with embroider than with classic.

Is there any way to control the order that I'm missing?

ef4:
vendor.css is intended as a backward compatibility feature so if it's not producing the same order that's a bug.

PR #1866, opened 2024-04-04
Sort addon css by package name

raycohen:
Fixes https://github.com/embroider-build/embroider/issues/1862

While broccoli-concat [doesn't guarantee order](https://github.com/broccolijs/broccoli-concat/blob/8c35038f012f57c791f7c9ad370a54b467e0e0cc/README.md#L41), the effective order of addon css in (most?) classic builds seems to be alphabetical by package name.

Also, the order of `app.import`'d styles was being reversed from what the classic build had, in addition to not coming before addon css.

Updates:
- Targets `stable`
- Added tests
- Simpler logic change, only impacts `implicit-styles`

### DIFF
--- a/packages/compat/src/compat-app-builder.ts
+++ b/packages/compat/src/compat-app-builder.ts
@@ -396,7 +396,15 @@ export class CompatAppBuilder {
 
   private impliedAddonAssets(type: keyof ImplicitAssetPaths, { engine }: AppFiles): string[] {
     let result: Array<string> = [];
-    for (let addon of sortBy(Array.from(engine.addons), this.scriptPriority.bind(this))) {
+    let addons: Array<AddonPackage> = sortBy(Array.from(engine.addons), this.scriptPriority.bind(this));
+    if (type === 'implicit-styles') {
+      const synthesizedVendor = addons.find(pkg => pkg.name === '@embroider/synthesized-vendor');
+      if (synthesizedVendor) {
+        addons = sortBy(addons, pkg => pkg.name).filter(pkg => pkg !== synthesizedVendor);
+        addons.unshift(synthesizedVendor);
+      }
+    }
+    for (let addon of addons) {
       let implicitScripts = addon.meta[type];
       if (implicitScripts) {
         let styles = [];
--- a/tests/scenarios/compat-addon-styles-test.ts
+++ b/tests/scenarios/compat-addon-styles-test.ts
@@ -1,6 +1,6 @@
 import type { ExpectFile } from '@embroider/test-support/file-assertions/qunit';
 import { expectFilesAt, expectRewrittenFilesAt } from '@embroider/test-support/file-assertions/qunit';
-import type { PreparedApp } from 'scenario-tester';
+import type { PreparedApp, Project } from 'scenario-tester';
 import { throwOnWarnings } from '@embroider/core';
 import { appScenarios, baseAddon } from './scenarios';
 import QUnit from 'qunit';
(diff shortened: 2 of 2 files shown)


ef4 (approved):
This looks good to me. 

We may want to retarget it against the `stable` branch so it's easier to release sooner. I think it will apply cleanly there.

packages/compat/src/compat-app-builder.ts
```diff
@@ -560,8 +581,7 @@ export class CompatAppBuilder {
     if (!asset) {
       let implicitStyles = this.impliedAssets('implicit-styles', application);
       if (implicitStyles.length > 0) {
-        // we reverse because we want the synthetic vendor style at the top
-        asset = new ConcatenatedAsset('assets/vendor.css', implicitStyles.reverse(), this.resolvableExtensionsPattern);
+        asset = new ConcatenatedAsset('assets/vendor.css', implicitStyles, this.resolvableExtensionsPattern);
```
raycohen, line 584: we need this reverse back, but I can impact just the imported stuff somewhere else

raycohen:
[stable](https://github.com/embroider-build/embroider/blob/stable/packages/compat/src/compat-app-builder.ts#L399) and [main](https://github.com/embroider-build/embroider/blob/main/packages/compat/src/compat-app-builder.ts#L369) differ slightly inside `impliedAddonAssets`

packages/compat/src/compat-app-builder.ts
```diff
@@ -365,25 +365,42 @@ export class CompatAppBuilder {
...
             if (!addon.isLazyEngine()) {
-              styles.push(resolve.sync(mod, options));
+              result.push(resolve.sync(mod, options));
             }
           } else {
             result.push(resolve.sync(mod, options));
           }
         }
-        if (styles.length) {
-          result = [...styles, ...result];
```
raycohen, line 385: this was unshifting the group of styles from each package to the overall list, which reverses the order and we no longer want. then in `implicitScriptsAsset` the entire list was getting (re-)reversed. We don't need that reverse either now.

BlueCutOfficial:
`impliedAddonAssets` code is reused to replace the assets generated by `compat-app-builder.ts` with virtual contents. If this PR is merged we probably want to double-check #1805 and #1811.

mansona:
Just a quick note on how to make sure this continues to work for the upcoming major release: @ef4 is correct that you should target stable with a change like this so that it could be released sooner than the next major, but we can't ensure that this is going to keep working unless we make sure to add a test that verifies the output is in the correct order. If this gets merged to stable and we then forward-port the test to the main branch we can ensure that the sort ordering is still correct even though the implementation is completely different 👍

tests/scenarios/compat-addon-styles-test.ts
```diff
@@ -132,3 +132,83 @@ appScenarios
...
+    const addonC = baseAddon();
+    setAddonNameAndCSS(addonC, 'c');
+
+    const addonD = baseAddon();
+    setAddonNameAndCSS(addonD, 'd');
+
+    project.addDependency(addonA);
+    project.addDependency(addonB);
+    addonA.addDependency(addonC);
+    project.addDependency(addonD);
```
raycohen, line 164: this test without the logic change gave c->a->b->d ordering

tests/scenarios/compat-addon-styles-test.ts
```diff
@@ -132,3 +132,83 @@ appScenarios
...
+    addon: {
+      styles: {
+        'addon.css': `.addon-${letter} {}`,
+      },
+    },
+  });
+}
+
+appScenarios
+  .map('compat-addon-styles', project => {
```
raycohen, line 148: seems to work like this but this name should probably be different from the above tests?

tests/scenarios/compat-addon-styles-test.ts
```diff
@@ -132,3 +132,83 @@ appScenarios
...
+    merge(project.files, {
+      'ember-cli-build.js': `
+        'use strict';
+
+        const EmberApp = require('ember-cli/lib/broccoli/ember-app');
+        const { maybeEmbroider } = require('@embroider/test-setup');
+
+        module.exports = function (defaults) {
+          let app = new EmberApp(defaults, {});
+          app.import('node_modules/third-party/third-party.css');
```
raycohen, line 179: add more of these so we get coverage on the order between app.import calls as well

raycohen:
I'm not sure what the CI failures are here...

PR #1868, opened 2024-04-09
Adjusting `@embroider/webpack` to use `@babel/preset-env` to avoid critical security audit

lupestro:
At its base, this is a straightforward change to one dependency in the package.json for @embroider/webpack. 

Running the full test suite locally, the last several release-watch-mode scenarios timed out, but otherwise it showed the same handful of errors that it showed when I ran the test suite before the change. I had seen one or two watch-mode tests fail earlier, so I'm not sure if that's test fragility or something to concern myself with about the change.

I'm also a little concerned about a number of the changes that happened in the `pnpm-lock.yaml` when I ran pnpm install after making the package.json change. I had only looked at the lock file after I'd run the tests with all those timeouts, though. Not sure if that made a difference.

This PR will at least open the conversation and I can take it from there with guidance and context.


kiwi-josh:
Chiming in here with some extra thoughts - it appears `babel-preset-env` is still being included in the `pnpm-lock.yaml`, because the [package.json within the tests](https://github.com/embroider-build/embroider/blob/2afc470d76f03f36021a25a76adf0a1921957f59/test-packages/support/package.json#L12) include it, and switch on its usage [here](https://github.com/embroider-build/embroider/blob/2afc470d76f03f36021a25a76adf0a1921957f59/test-packages/support/index.ts#L42).

I don't have enough context to understand why/what this is doing

mansona (approved):
This looks good to me. I did an audit of the difference between `babel-preset-env` and `@babel/preset-env` and I think the safest thing for us to do is to mark this as a breaking change on our side. There is a possibility of changes in behaviour because of some of the listed breaking changes: 

- https://github.com/babel/babel/pull/6452
- https://github.com/babel/babel/pull/6831
- https://github.com/babel/babel/pull/7417
- https://github.com/babel/babel/pull/7446
- https://github.com/babel/babel/pull/7358
- https://github.com/babel/babel/pull/8509

PR #1876, opened 2024-04-10
fix vite watch mode dep optimizer

patricklx:
- [x] all deps are optimized
- [x] initial dep scan creates all deps
- [x] all optimized chunks are used
- [x] new deps are optimized
- [x] different specifiers use same optimized id
- [ ] addon rebuilds are newly optimized

dep optimizer has multiple phases 
* Scanning, which uses the vite plugins and configured rollup plugins .
* pre bundle, does not use configured rollup plugins.

This is the order 
* inital scan 
* initial scan must provide correct import specifier to vite import analysis plugin, this is probably because we cannot use vite own alias handling 
* pre bundling 
* vite starts and gets requests
* vite checks if some need optimization 
* rules:
1. Specifier must be bare import 
2. Importer cannot be in node modules
3. Needs to be node_resolveable, so follow node resolution without additional plugin resolution 
* if a new dep is detected, vite schedules a new scan

Currently this is broken because of 2 things. app is in rewritten-app. It's actually easy to fix by changing to importer to the original app. Only needed when resolving bare imports.

The more difficult one is rewritten-packages. They must be node resolvable from original app. They are not currently.
I suggest to create a link from `@embroider/rewritten-packages` to the one inside . embroider and rewrite specifiers accordingly.

It currently looks like it's working, because the initial scan detects the deps. But any new dep added later will not be optimized.

fixes:
* rewrite rewritten-app to original app location 
* link rewritten-package to be node resolvable
* esbuild scan needs to pass renamed modules and appjsmatch to vite import analysis 
*

### DIFF
--- a/packages/core/src/virtual-entrypoint.ts
+++ b/packages/core/src/virtual-entrypoint.ts
@@ -148,7 +148,7 @@ export function renderEntrypoint(
 
   return {
     src: entryTemplate(params),
-    watches: [],
+    watches: [fromDir],
   };
 }
 
--- a/packages/vite/src/build.ts
+++ b/packages/vite/src/build.ts
@@ -42,6 +42,10 @@ export function emberBuild(command: string, mode: string, resolvableExtensions: 
   });
 }
 
+// we need a bare specifier for rewritten packages that is resolvable from root
+// we cannot use a bare specifier that starts with .embroider, that is not a bare import
+// according to vite
+// which is why we make this symlink
 function createSymlinkToRewrittenPackages(appRoot: string = process.cwd()) {
   const resolvableRewrittenPackages = resolve(
     locateEmbroiderWorkingDir(appRoot),
--- a/packages/vite/src/build.ts
+++ b/packages/vite/src/build.ts
@@ -1,5 +1,10 @@
 import { fork } from 'child_process';
 import type { Plugin } from 'vite';
+import { resolve } from 'path';
+import fs from 'fs-extra';
+const { ensureSymlinkSync, existsSync, writeFileSync } = fs;
+import { locateEmbroiderWorkingDir } from '@embroider/core';
+import * as process from 'process';
 
 export function emberBuild(command: string, mode: string, resolvableExtensions: string[] | undefined): Promise<void> {
   let env: Record<string, string> = {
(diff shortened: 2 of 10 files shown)


ef4:
> I suggest to create a link from @embroider/rewritten-packages to the one inside

This is how things used to work. Doing it correctly requires more than just links to the rewritten packages, in general it requires you to rewrite the entire node_modules graph in order to make every package see all the correct dependencies.

And I don't think it's necessary. I think that from the perspective of the rewritten-app, the rewritten-packages don't appear to be inside node_modules. But when we stop rewriting the app, they will.

patricklx:
> > I suggest to create a link from @embroider/rewritten-packages to the one inside
> 
> This is how things used to work. Doing it correctly requires more than just links to the rewritten packages, in general it requires you to rewrite the entire node_modules graph in order to make every package see all the correct dependencies.
> 
> And I don't think it's necessary. I think that from the perspective of the rewritten-app, the rewritten-packages don't appear to be inside node_modules. But when we stop rewriting the app, they will.

The problem is that vite needs a bare import and the importer cannot be in node modules. 

What we are currently for resolving rewritten-packages is doing a rehome inside the rewritten-packages, onto moved-target.js https://github.com/embroider-build/embroider/blob/main/packages/core/src/module-resolver.ts#L954

And here is how vite checks if it should optimise:
Specifier must be bare import
Importer cannot be in node modules
Needs to be node_resolveable, so follow node resolution without additional plugin resolution
https://github.com/vitejs/vite/blob/6c323d5b3ab3cdf81d21bbe965ed3c36aa7f0589/packages/vite/src/node/plugins/resolve.ts#L868



So it's not a problem rewritten-app.

patricklx:
It might be possible to make easier if we could force the addons into the dep optimizer.
Otherwise we really need to exclude them again

packages/core/src/module-resolver.ts
```diff
@@ -224,8 +225,8 @@ export class Resolver {
 
     switch (resolution.type) {
       case 'found':
-      case 'ignored':
         return resolution;
+      case 'ignored':
       case 'not_found':
```
patricklx, line 230: ignored should fallback resolve

packages/core/src/module-resolver.ts
```diff
@@ -482,10 +483,14 @@ export class Resolver {
     }
     let { type, rest } = match.groups!;
     let fromPkg = this.packageCache.ownerOfFile(request.fromFile);
-    if (!fromPkg?.isV2Ember()) {
+    if (!fromPkg) {
       return request;
     }
-    let engine = this.owningEngine(fromPkg);
+    if (!fromPkg?.isV2Ember() && !request.fromFile.includes('.vite/deps/@embroider_rewritten')) {
+      return request;
```
patricklx, line 490: workaround

packages/shared-internals/src/package.ts
```diff
@@ -25,7 +25,11 @@ export default class Package {
 
   @Memoize()
   protected get internalPackageJSON() {
-    return JSON.parse(readFileSync(join(this.root, 'package.json'), 'utf8'));
+    try {
+      return JSON.parse(readFileSync(join(this.root, 'package.json'), 'utf8'));
```
patricklx, line 29: workaround for fromFile pointint at /app/tests/package.json

mansona (changes requested):
I just discussed this on the Embroider Triage meeting with @patricklx and I'm going to do my best to summarise the situation here: 

Firstly we started to spec out a way to test if a dependency was optimised correctly in the office hours last week. A good first draft of that work can be seen in https://github.com/embroider-build/embroider/pull/1880 where we run vite in dev mode and try to "audit" the app across HTTP. This will allow us to make sure that the dependencies we want to check are coming from the right place (i.e. from the .vite/deps folder) 

This work is blocked on the refactor that myself and @ef4 are pairing on to allow audits to work across HTTP (and async) which hopefully will land this week.

Once the audit refactor lands we will make sure to provide a suite of tests that show the problems that @patricklx is identifying in this PR (and Discord discussions), we can then check those tests against other work that is going on (e.g. https://github.com/embroider-build/embroider/pull/1779 ) to make sure that they are still failing and we can fix accordingly 👍

mansona (changes requested):
This PR should be split up into individual PRs that address single issues that are verifiable with tests. Especially now that https://github.com/embroider-build/embroider/pull/2029 is merged we should be methodical with how we proceed

patricklx:
fixed in #2135

PR #1877, opened 2024-04-11
Command Watcher

BlueCutOfficial:
**Context**

When working with vite, running the dev server and generating the production build with Rollup are two different actions that both have a specific execution. For instance, when dealing with public assets provided by the app and addons: 
- when running the vite dev server, the `assets` plugin uses a middleware to catch the assets-related request, find the actual location of the asset, and serve it.
- when building the app (like for a production build), the `assets` plugin will emit the public assets in the `dist/` output, where they are supposed to be. 

This case highlights why we need to test some behaviors for both dev mode and build mode now the test suite is using vite.

**Testing dev mode easily**

This PR implements a `CommandWatcher` utility to ease running a vite dev server in tests. It relies on the former `EmberCLI` class that was used to run `ember serve`, but it makes it more generic (you can now choose the name of the command like `ember` or `vite`) and it makes it an importable helper so we can test the dev mode in multiple test files.

**Compat dummy app test**

The first use case for the new `CommandWatcher` is the `compat-dummy-app-test`. This PR prepares the field so the tests that check public assets presence now use vite and check both the dev mode and the build.

This preparation work is required to complete #1859, which will remove the part of the code that is now managed by the `assets` plugin: in #1859, `compat-dummy-app-test` will have to stop testing public assets are present in the rewritten app and instead test they are present in the `dist` (in build mode) or served correctly (in dev mode). 

Here, aside of the assertions, we do the following things: 
- We update the addon template used for the dummy app test scenarios. vite is now configured to run the `assets` plugin and we also update `ember-qunit` to fix an issue in the development build.
- The test waits for the dev server to be ready, and it knows when it's ready using a regexp matching the text that is supposed to be output on the server when it's started. This regexp is now slightly more flexible, and in `CommandWatcher` implementation, we strip the Ansi characters before analyzing the server output, because vite server colorize the text, which can mess-up the regex.

### DIFF
--- a/tests/scenarios/compat-dummy-app-test.ts
+++ b/tests/scenarios/compat-dummy-app-test.ts
@@ -88,7 +88,7 @@ dummyAppScenarios
       test('contains public assets from dummy app in dev mode', async function (assert) {
         const server = CommandWatcher.launch('vite', ['--clearScreen', 'false'], { cwd: app.dir });
         try {
-          const [, url] = await server.waitFor(/Local:\s+(http:\/\/127.0.0.1:\d+)\//);
+          const [, url] = await server.waitFor(/Local:\s+(https?:\/\/.*)\//g);
           let response = await fetch(`${url}/robots.txt`);
           let text = await response.text();
           assert.strictEqual(text, 'go away bots');
--- a/tests/scenarios/compat-dummy-app-test.ts
+++ b/tests/scenarios/compat-dummy-app-test.ts
@@ -21,6 +21,9 @@ dummyAppScenarios
           'example.hbs': `hello`,
         },
       },
+      public: {
+        'from-addon.txt': 'a public asset provided by the classic addon',
+      },
       tests: {
         dummy: {
           public: {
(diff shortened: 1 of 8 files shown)


tests/scenarios/package.json
```diff
@@ -81,6 +82,7 @@
     "ember-source-latest": "npm:ember-source@latest",
     "ember-truth-helpers": "^3.0.0",
     "execa": "^5.1.1",
+    "node-fetch": "2.7.0",
```
BlueCutOfficial, line 85: Using the latest version `3.3.2` triggers an error `Unknown file extension ".ts"`. When downgrading `node-fetch` to < 3, I don't have the issue and `fetch` runs fine.

tests/scenarios/helpers/command-watcher.ts
```diff
@@ -0,0 +1,143 @@
...
+  // implemented manually here to workaround issues with
+  // ts-nodes and ES modules imports
+  private stripAnsi(string: string): string {
+    const pattern = [
+      '[\\u001B\\u009B][[\\]()#;?]*(?:(?:(?:(?:;[-a-zA-Z\\d\\/#&.:=?%@~_]+)*|[a-zA-Z\\d]+(?:;[-a-zA-Z\\d\\/#&.:=?%@~_]*)*)?\\u0007)',
+      '(?:(?:\\d{1,4}(?:;\\d{0,4})*)?[\\dA-PR-TZcf-nq-uy=><~]))',
+    ].join('|');
+    let ansiRegExp = new RegExp(pattern, 'g');
+    return string.replace(ansiRegExp, '');
+  }
```
BlueCutOfficial, line 63: I don't like doing this 🙈 this is a little function that basically re-implements `strip-ansi` and `ansi-regex` packages. 

It's a very little piece of code so I did this after failing to import the `strip-ansi` module into `CommandWatcher`. I had the same error `Unknown file extension ".ts"` as I had when using `node-fetch` > 3, and after trying a few things it was rather:
```
Error [ERR_REQUIRE_ESM]: require() of ES Module 
.../embroider/node_modules/.pnpm/strip-ansi@7.1.0/node_modules/strip-ansi/index.js 
from 
.../embroider/tests/scenarios/helpers/command-watcher.js 
not supported.
Instead change the require of index.js in .../embroider/tests/scenarios/helpers/command-watcher.js 
to a dynamic import() which is available in all CommonJS modules.
at Object.require.extensions.<computed> [as .js] (.../embroider/node_modules/.pnpm/ts-node@10.9.2_typescript@5.2.2/node_modules/ts-node/dist/index.js:851:20)
```
☝️ I don't know how to debug these `ts-node`-related errors.

tests/scenarios/package.json
```diff
@@ -81,6 +82,8 @@
...
+    "strip-ansi": "^6.0.0",
```
BlueCutOfficial, line 86: I had to use `strip-ansi@6.0.0` because `> 7.x` is an ESM-only package. Importing this into the scenarios files triggers an `Unknown file extension ".ts"` error or an `ERR_REQUIRE_ESM` error.

tests/scenarios/compat-dummy-app-test.ts
```diff
@@ -59,18 +64,54 @@ dummyAppScenarios
...
+        let assets = expectFile('../../node_modules/.embroider/rewritten-app/package.json')
           .json()
-          .get('ember-addon.assets')
-          .includes('robots.txt');
+          .get('ember-addon.assets');
+        assets.includes('robots.txt');
+        assets.includes('./addon-template/from-addon.txt');
+      });
+
+      test('production build contains public assets from both addon and dummy app after a build', async function (assert) {
```
BlueCutOfficial, line 93: This test builds the dummy app the same way as the one checking the rewritten-app so I could have done all the checks in the same test. I preferred separated tests because I want to remove the rewritten-app one in PR #1859

mansona (approved):
Amazing work 🎉

PR #1880, opened 2024-04-15
Vite dep tests

mansona:
This PR is to demonstrate how we want to do vite dependency optimisation tests in the future. We don't want any/much vite-specific infrastructure in our tests and we want to rely on the audit system to make sure that the correct dependencies being built in vite dev mode

## TODO

- [ ] update the audit system to be able to load html (and references) across HTTP
- [ ] remove the "async file" implementation from this PR and make sure it uses the new audit system

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -68,3 +68,6 @@ package.json.ember-try
 #types added
 /types/**
 
+# scenario tester output
+/tests/scenarios/output/
+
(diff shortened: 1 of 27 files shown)


mansona:
The actual audit refactor is happening in https://github.com/embroider-build/embroider/tree/audit-refactor but there is no Pull request yet so I can't link to that 😂

mansona:
We are building a new test based on the work in https://github.com/embroider-build/embroider/pull/1882 which will be a new PR so I'm closing this now 👍

PR #1882, opened 2024-04-16
Splitting reusable module visitor out of Audit

ef4:
So we can use it to also traverse apps over HTTP when talking to vite, because we need to replace a bunch of stage2-style tests.


mansona (approved):
I think this is good to go 👍

PR #1885, opened 2024-04-18
Fix pre support in portable babel launcher

ef4:
The portable-babel-launcher that we use to deal with plugin serializability didn't correctly handle options in pre hooks. babel-plugin-ember-template-compilation just released with a change that needs to handle options in the pre hook.


mansona (approved):
LGTM 🎉 💪

PR #1887, opened 2024-04-22
fix compat namespace app test

patricklx:


tests/scenarios/compat-namespaced-app-test.ts
```diff
@@ -45,7 +45,10 @@ appScenarios
       type: 'addon',
       'implicit-modules': ['./my-implicit-module.js'],
     };
-    addon.files['my-implicit-module.js'] = 'globalThis.addonImplicitModulesWorked = true';
+    addon.files['my-implicit-module.js'] = `
+      globalThis.addonImplicitModulesWorked = true;
+      export default Object.assign({}, {});
```
mansona, line 50:
```suggestion
      export default {};
```

mansona: or even 
```suggestion
      export {};
```

?

patricklx: I didn't test it, but error was about default export 
https://github.com/embroider-build/embroider/actions/runs/8782825133/job/24097779002#step:4:67

mansona: that link didn't seem to bring me to the right line 🤔 also it seems like there are a lot of other known issues on that run so we can't isolate the issue 🤔

patricklx: Oops, that was something totally different somehow...
https://github.com/embroider-build/embroider/actions/runs/8783147806/job/24098748077#step:4:67

patricklx: And here how it's imported 
https://github.com/embroider-build/embroider/blob/ea895c2b1a3925318cb2bde89bbcabb78c4a41d9/packages/core/src/virtual-content.ts#L335

PR #1888, opened 2024-04-23
do not merge - test(vendor.js): investigate windows failure

BlueCutOfficial:
test vendor.js presence before virtualization so we can see if it fails on windows.

### DIFF
--- a/tests/scenarios/compat-addon-classic-features-test.ts
+++ b/tests/scenarios/compat-addon-classic-features-test.ts
@@ -116,6 +116,19 @@ appScenarios
         assert.true(content.includes('<p>Content for body</p>'));
         assert.true(!content.includes('<p>Content for custom</p>'));
       });
+
+      test('content-for are replaced: dev mode', async function (assert) {
+        const server = CommandWatcher.launch('vite', ['--clearScreen', 'false'], { cwd: app.dir });
+        try {
+          const [, url] = await server.waitFor(/Local:\s+(https?:\/\/.*)\//g);
+          let response = await fetch(`${url}/`);
+          let text = await response.text();
+          assert.true(text.includes('<p>Content for body</p>'));
+          assert.true(text.includes('<p>Content for custom</p>'));
+        } finally {
+          await server.shutdown();
+        }
+      });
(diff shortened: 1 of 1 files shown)


tests/scenarios/compat-addon-classic-features-test.ts
```diff
@@ -131,3 +131,30 @@ appScenarios
...
+  .map('compat-addon-classic-features-virtual-scripts', () => {})
+  .forEachScenario(scenario => {
+    Qmodule(`${scenario.name}`, function (hooks) {
+      let app: PreparedApp;
+      hooks.before(async () => {
+        app = await scenario.prepare();
+      });
+
+      test('vendor.js script is served', async function (assert) {
+        const server = CommandWatcher.launch('vite', ['--clearScreen', 'false'], { cwd: app.dir });
```
patricklx, line 145: if this is vite dev mode you will need this:
https://github.com/embroider-build/embroider/blob/e1142b1074720cc7a1c20f2f5f29a24686c7c301/tests/scenarios/vite-app-test.ts#L187

patricklx: maybe the first test only checked on the index.html? hmm,  are there any other vite dev mode tests that do more ?

this test now checks on vendor.js which now uses internal core/module-resolver. so it could be that it triggers the error

BlueCutOfficial: I don't understand why, but for now, the only thing with the 2 green commits I have is that the test file contains only one scenario using the `CommandWatcher`. Could it be some kind of cleanup that doesn't happen between several scenarios of the same file and a new server cannot be started, specifically on windows? 🤔

BlueCutOfficial: So the last commit pushes the branch back to a state that used to be red ❌ and it's now green ✅ so this might be just about performances and timeout 🙈

BlueCutOfficial:
Discussed the office hours: we will skip the failing test in PR #1801 and come back to it at a later time.

PR #1890, opened 2024-04-24
Add more tests virtual files

BlueCutOfficial:
**Context**
Recently, several PRs have been merged that virtualize `vendor.css` (#1805), `test-support.js` (#1807), and `test-support.css` (#1811) files. These changes have been tested manually, but back then, we didn't add any new tests to assert the virtual contents are correctly served or present in the build output.

This PR adds the missing tests. 

**About searching code in CSS files**
The CSS-related tests should have ideally relied on the `Audit` system to look for the pieces of styles that are expected to be present. But for now, this system presumes too many things about the location of the files we want to audit, it's not made to just take any given `dist/`. That's why `globby` is used to retrieve all the CSS files in the build output and we read their content to find the expected styles.


mansona (approved):
LGTM 🎉 💪

PR #1900, opened 2024-05-02
vite: change ext-es format 

patricklx:


packages/vite/src/resolver.ts
```diff
@@ -53,7 +53,10 @@ export function resolver(): Plugin {
     },
     load(id) {
       if (id.startsWith(virtualPrefix)) {
-        let { pathname } = new URL(id, 'http://example.com');
+        let { pathname, searchParams } = new URL(id, 'http://example.com');
+        if (pathname.includes('@embroider/ext-es')) {
+          pathname += '?exports=' + searchParams.get('exports')
+        }
```
BlueCutOfficial, line 59: (non-blocking, just a comment) I identified another query param we need to include, I think we will need a better place to list the params that are not bundler-specific and need to be passed to Embroider.

ef4:
This is not a good solution to the problem. I don't want to break the vite resolver every time core decides to add a search param. And what happens if core is using a searchParam but then vite decides to also start using the same name?
 
We probably don't need to use search params at all in the core. The structure of these URLs were fairly arbitrary decisions. I I think we can make them all part of the pathName, and decide to leave searchParams entirely free for use by the host environment (vite).

patricklx:
@ef4 I updated the pr, the failing checks look unrelated. more related to babel plugin template compilation

packages/vite/src/resolver.ts
patricklx: I think this is still important. does it also affect virtualized content? then i could include it in this pr

BlueCutOfficial: It's definitely the same kind of issue, but I think we should not include it in this PR.

To give you a bit more context, I am currently working on fixing the Vite build for applications that use Fastboot. There are a lot of different issues, and Fastboot relying on a query param is just one of them. It makes more sense to keep the fix in my PR in my opinion, because all of my changes are needed to get the Fastboot test green ✅ and validate that whatever will replace the query param works.

ef4:
Thanks, looking good.

Issue #1455, opened 2023-06-01
@embroider/router too much recursion error occurs in fire fox browser


NagarajNN:
too much recursion error occurs while navigating to another routes
stackTrace shows this @embroider/router/dist/index.ts file
I can't able put break point on that,
![image](https://github.com/embroider-build/embroider/assets/115421636/83622247-4ef3-4531-b915-427f7b0d4af0)
original error
![image](https://github.com/embroider-build/embroider/assets/115421636/cc408f54-9e63-47e4-a009-9efef94c1771)

void-mAlex:
hello
@NagarajNN a few things to check:
* does it happen with the normal ember router
* Which version of the embroider router is this
* is it on load or just in navigation and is it an engine and what type
* what is the URL causing the issue + what does the router. js look like to cause this

lastly a reproduction repo would be greatly appreciated

NagarajNN:
> 

it's doesn't happen in normal ember router
embroider router version - 2.0.0
occurs in navigating to another routes
http://localhost:766/#/dashboards/1/{dynamic-segment}
when i navigate to another dashboard [here only dynamic-segment will be changed ]
it occurs

void-mAlex:
are you able to create a minimal reproduction of what you're seeing, as that would greatly help in fixing this

NagarajNN:
any other option other than reproduction repo,
can you please suggest @void-mAlex 
i can't able to track error seems like firefox browser default recursion limit error.

void-mAlex:
I would ask you to try and narrow down where this is coming from, is the dynamic segment properly url encoded? I know FF/safari is less error tolerant than chrome in that respect
a reproduction of your route in a new repo/project set up as you have it will go a long way to figuring out what is causing that
if it is or not, possible to reproduce it that way, it will narrow it down to something else but at least will tell us more info

jembezmamy:
I am experiencing the same issue, here are my investigations:

![Zrzut ekranu 2024-05-6 o 12 20 37](https://github.com/embroider-build/embroider/assets/720354/7bb55354-69e7-4217-9bf2-8ee281a3ed0e)

The [`return original(name)` line](https://github.com/embroider-build/embroider/blob/e5d46e7537add8b80c07ea47f6a0c4d8afceaaf2/packages/router/src/index.ts#L80) is being called recursively 55 times for each `getRoute` call, and 1100 times per single transition. We see on Sentry that for some users on Firefox this causes the `too much recursion` error.

[...]

PR #1904, opened 2024-05-07
Fix getRoute override

jembezmamy:
This PR aims at resolving #1455 - `too much recursion` error.

The error is caused by `getRoute` being overridden multiple times: every time `setupRouter` is called, we create a new layer of recursion. This happens e.g. when we render a link, which calls `urlFor`, which calls `setupRouter`. So while the user clicks around the app, `getRoute` is overridden again and again, which can lead to `too much recursion` error at some point.

The proposed resolution is to leverage `isSetup` to make sure the `getRoute` is extended only once.


ef4:
Thanks.

I retargeted to stable and slightly adjusted the implementation. Now GitHub isn't running the actions...

PR #1906, opened 2024-05-07
`@embroider/macros` es-compat module is resolvable from type=module addons

NullVoxPopuli:
Extracted from https://github.com/embroider-build/embroider/pull/1572 (which has more issues to fix (wrong css path, for example)


NullVoxPopuli:
According to:
```
    require() of ES Module /tmp/tmp-31383035liZVSdehi7p/node_modules/v2-addon/addon-main.js from <.pnpm>/ember-cli@5.8.1/node_modules/ember-cli/lib/models/package-info-cache/package-info.js not supported.
    addon-main.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
    Instead rename addon-main.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /tmp/tmp-31383035liZVSdehi7p/node_modules/v2-addon/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).
```

It is required that addon-main.js be cjs.

NullVoxPopuli:
Current issue (and the one causing the two tests to hang):
image

```
Died on test #1: Could not find module `./side-effecting.js` imported from `(require)`
```

it seems that importSync isn't working on relative imports perhaps?

NullVoxPopuli:
Each time I push, I'm going to be cancelling the workflow so as to not hog CI

PR #1913, opened 2024-05-10
virtualise engine styles with their vendor styles

mansona:
For #1779 it would be a lot easier for the implementation of an engine's own styles to be handled with the same virtual handler that deals with its vendor files. This PR makes that change 👍

### DIFF
--- a/packages/compat/src/compat-app-builder.ts
+++ b/packages/compat/src/compat-app-builder.ts
@@ -27,7 +27,6 @@ import type { CompatResolverOptions } from './resolver-transform';
 import type { PackageRules } from './dependency-rules';
 import { activePackageRules } from './dependency-rules';
 import flatMap from 'lodash/flatMap';
-import sortBy from 'lodash/sortBy';
 import flatten from 'lodash/flatten';
 import partition from 'lodash/partition';
 import mergeWith from 'lodash/mergeWith';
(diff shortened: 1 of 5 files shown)


packages/core/src/virtual-compat-styles.ts
```diff
@@ -34,15 +34,22 @@ function getVendorStyles(owner: Package, resolver: Resolver): string {
...
 
-function generateVendorStyles(engine: Engine): string {
+function generateCompatStyles(engine: Engine): string {
   let result: string[] = impliedAddonVendorStyles(engine).map((sourcePath: string): string => {
     let source = readFileSync(sourcePath);
     return `${source}`;
   });
 
+  // add engine's own implicit styles after all vendor styles
+  let styles = getAddonImplicitStyles(engine.package as V2AddonPackage);
```
ef4, line 47: This is not what I had in mind. Consider the app: the app never has implicit styles, but it does have it's own styles (the stuff from `app/styles/app.css`). That's the stuff we're talking about incorporating here.

mansona: yes but during the v2 upgrade of an app don't their styles get applied to the engine's implicit styles? I would have expected a lot more tests to fail in this PR if the approach didn't work 🤔

mansona: I guess not 🙈 it turns out that the only test that checks for the app scripts right now is the namespaced app that has a qunit test `Acceptance | smoke-test: styles present` 🫠 

I'll take another look at it

mansona: So I've implemented this now in such a way that it goes to find the right file out of the `@embroider/synthesized-styles` package and read its contents, but I'm wondering if we need to even generate a synthesised styles package any more 🤔 

Looking at where it is generated https://github.com/embroider-build/embroider/blob/virtual-compat-styles/packages/compat/src/compat-app.ts#L505-L539 it seems to imply that there can be multiple files output other than just the `app-name.css` even though only one of the files was ever being consumed by the rewritten app. Am I missing some context here as to why it was written this way?

mansona:
I have now verified that this is working with the latest HEAD of #1779 👍

PR #1915, opened 2024-05-14
update ember engines

mansona:


SergeAstapov (dismissed):

mansona:
@SergeAstapov turns out we weren't testing engines at all on main 🫠 so we should probably make sure tests are passing before we update the dependency

PR #1918, opened 2024-05-15
New index.html format - vendor js

BlueCutOfficial:
## Context 

Lately, we have been working on the inversion of control for the Vite build, we replaced script assets like `vendor.js` with virtual content Vite can request to Embroider. 

Until now, we have never changed the `index.html` file of the initial Ember app to handle virtualization: in stage 2, the compat-app-builder generates the `index.html` of the rewritten-app and inserts the virtual entry points. This rewritten file is the one consumed by Vite.

Now that we have virtualized all the entry points, we are ready for the next step that will get us closer to the new blueprint: we are going to write the virtual entry points directly in the `index.html` of the initial Ember app so Vite will be able to consume it directly, without intermediate change.

This PR handles `vendor.js`.


packages/core/src/ember-html.ts
```diff
@@ -27,6 +27,11 @@ export interface EmberHTML {
 }
 
 class Placeholder {
+  static find(node: Node): Placeholder {
+    let placeholder = this.immediatelyAfter(node);
+    return placeholder;
+  }
```
BlueCutOfficial, line 33: This function is the same as `replacing`, except that it doesn't remove the `node`. 
Before, the `replacing` function was used to replace the script `assets/vendor.js` with `@embroider/core/vendor.js`. Now, it simply finds the script `@embroider/core/vendor.js`, it doesn't remove it, and it uses its position as a reference for `insertAfter`. 

The reason I need all of this is because Fastboot scripts are still inserted by the compat-app-builder at this point, and we need to have the location of the vendor script to insert them at the correct position:
```
// compat-app-builder.ts

if (this.fastbootConfig) {
      // any extra fastboot vendor files get inserted into our
      // html.implicitScripts section, after the regular implicit script
      // (vendor.js) have been inserted.
      for (let script of this.fastbootConfig.extraVendorFiles) {
        html.insertScriptTag(html.implicitScripts, script, { tag: 'fastboot-script' });
      }
}
```

mansona (approved):
Great work 🎉

PR #1919, opened 2024-05-16
improve audit tests

patricklx:


tests/scenarios/helpers/command-watcher.ts
```diff
@@ -108,7 +108,7 @@ export default class CommandWatcher {
       return;
     }
 
-    this.process.kill();
```
patricklx, line 111: vite needs to be killed with `SIGINT`, otherwise it leaves the server subprocess running

test-packages/support/audit-assertions.ts
```diff
@@ -38,15 +41,31 @@ export function setupAuditTest(hooks: NestedHooks, opts: () => AuditBuildOptions
...
+    await visit();
+  });
+
+  hooks.beforeEach(assert => {
+    installAuditAssertions(assert);
+    prepareResult(assert);
   });
 
   return {
+    async rerun() {
```
patricklx, line 56: for tests that change files while vite dev is running we need to be able to rerun the vistors

PR #1922, opened 2024-05-16
New `index.html` format - entrypoint

BlueCutOfficial:
## Context 

Lately, we have been working on the inversion of control for the Vite build, we replaced script assets like `vendor.js` with virtual content Vite can request to Embroider. 

Until now, we have never changed the `index.html` file of the initial Ember app to handle virtualization: in stage 2, the compat-app-builder generates the `index.html` of the rewritten-app and inserts the virtual entry points. This rewritten file is the one consumed by Vite.

Now that we have virtualized all the entry points, we are ready for the next step that will get us closer to the new blueprint: we are going to write the virtual entry points directly in the `index.html` of the initial Ember app so Vite will be able to consume it directly, without intermediate change.

This PR handles `entrypoint`.

### DIFF
--- a/packages/compat/src/compat-app-builder.ts
+++ b/packages/compat/src/compat-app-builder.ts
@@ -381,12 +381,6 @@ export class CompatAppBuilder {
       });
     }
 
-    // our tests entrypoint already includes a correct module dependency on the
-    // app, so we only insert the app when we're not inserting tests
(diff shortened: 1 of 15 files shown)


mansona (approved):
looks great! 🎉 :shipit:

PR #1924, opened 2024-05-16
virtualise test entrypoint

mansona:


packages/core/src/virtual-test-entrypoint.ts
```diff
@@ -0,0 +1,201 @@
...
+    const resolvedValues = await Promise.all(Object.values(fastbootModules));
+
+    Object.keys(fastbootModules).forEach((k, i) => {
+      d(k, function(){ return resolvedValues[i];});
+    })
+  }
+{{/if}}
+
+
+{{#if lazyRoutes}}
```
ef4, line 123: I think a test entrypoint would never use this.

packages/core/src/virtual-test-entrypoint.ts
```diff
@@ -0,0 +1,201 @@
...
+
+  for (let relativePath of appFiles.tests) {
+    amdModules.push(importPaths(resolver, appFiles, relativePath));
+  }
+
+  let src = entryTemplate({
+    amdModules,
+    testSuffix: true,
+    // this is a backward-compatibility feature: addons can force inclusion of test support modules.
+    defineModulesFrom: './-embroider-implicit-test-modules.js',
```
ef4, line 68: Should the tests entrypoint be doing this? Remember that it imports the app entrypoint, so anything the app entrypoint is doing shouldn't be repeated in here.

ef4 (commented):
I think this either needs to be radically reduced down to things that only happen in the tests entrypoint or (if we really do need a lot of this code) needs to share the implementation with the app entrypoint. I don't want to just duplicate the implementation.

packages/core/src/virtual-test-entrypoint.ts
mansona: removed 👍

packages/core/src/virtual-test-entrypoint.ts
mansona: you're right this shouldn't be needed, it's already importing the app entrypoint and that is handled there 👍 removed

ef4 (approved):
Much smaller!

PR #1925, opened 2024-05-17
New `index.html` format - app css

BlueCutOfficial:
## Context

To build an Ember app with Vite, we currently generate an intermediate app that Vide can consume: the rewritten-app. In stage 2, the compat-app-builder generates the `index.html` of the rewritten-app out of the `index.html` of the initial app. It re-inserts the entry points and replaces some of them with a different content.

Most of these replacements were about replacing script assets with virtual contents: #1918, #1920, #1921, #1922. But the CSS of the app remains an asset located at `dist/assets/app-name.css`.


mansona (approved):
This looks good enough to merge 👍 sorry for taking it out of draft 🙈 but if I did it too early we can fix-forward and add/change whatever we need on main 👍

PR #1931, opened 2024-05-20
update scenario-tester

mansona:
While working on https://github.com/embroider-build/embroider/pull/1930 I noticed some scenarios were falling even though I didn't expect them to be any different 🤔 

This PR is to update scenario-tester to rule out any issues in the upgrade 👍


BlueCutOfficial:
**lts_4_4-compat-app-script-attributes**

In some tests, the rewritten-packages contain multiple copies of `@ember-data/store`. Only one copy has content, the others are empty. During the build process, an import of the store resolves to a wrong empty copy.

At this point, I don't know why there are multiple copies nor how an empty one is accessed, I only know what makes them empty. It's [this hook in @ember-data/private-build-infra](https://github.com/emberjs/data/blob/lts-4-4/packages/private-build-infra/src/addon-build-config-for-data-package.js#L88). This exact file no longer exists in the latest EmberData (the release note for 5.3.0 mentions a deprecation removal). When`shouldIncludeChildAddon` hook isn't defined the multiple copies still exist, the build works because they are not empty: [this code is not executed](https://github.com/embroider-build/embroider/blob/main/packages/compat/src/build-compat-addon.ts#L33).

In the **release-compat-app-script-attributes** scenario, we have only single copies of each package in `rewritten-packages/@ember-data`.

PR #1945, opened 2024-05-28
avoid loading of lazy engines when generating linkto urls for routes

void-mAlex:
this came up when adding embroider scenarios to the ember-engines repo
there we test the same engine mounted at on different paths that can link between each other
this causes router_js implementation to generate url links from LinkTo component 
in the current implementation the queryparam hydration for non visited routes would cause the lazy engines to load eagerly upon encountering a link to that lazy engine even if it was never visited

it also has the sideeffect of firing off initializers and instance initializers out of order from classic builds

this change implements the same query param handling logic that the ember-engines repo was doing in its own router extension

engines repo has tests to verify that the qp's are correctly maintained with this implementation

### DIFF
--- a/packages/router/src/index.ts
+++ b/packages/router/src/index.ts
@@ -10,9 +10,6 @@ interface GlobalConfig {
   '@embroider/core'?: { active: boolean };
 }
 
-type EngineInfoByRoute = Record<string, { name: string }>;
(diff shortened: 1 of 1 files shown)


mansona (approved):
Looks good to me :shipit:

PR #1949, opened 2024-05-28
Fix ownerOfFile bug on windows

ef4:
https://github.com/embroider-build/embroider/pull/1762 only fixed the bug on unix. On windows the bug is still lurking.


ef4:
I went slightly mad here for a while until I realized CI won't show this passing until *after* we release it. Or use overrides.

Because we're testing the version of `@embroider/shared-internals` that's being brought in by ember-auto-import, not necessarily the version in the monorepo.

PR #1955, opened 2024-06-04
Remove template-only components compilation from Stage 2

BlueCutOfficial:
Fixes: #1958

During the build, the `hbs` Vite plugin and Babel work together so template-only components of the app are compiled as `.js` files. When the app requests the `.js` through an import, the `hbs` Vite plugin detects it's a template-only component and returns a consistent answer.

This is in theory. In practice, Stage 2 already compiles template-only components using the Broccoli plugin `SynthesizeTemplateOnlyComponents` and puts the resulting `.js` in the rewritten-app files. From start, the `.js` already exists in the rewritten-app consumed by Vite and the part of the `hbs` Vite plugin that is supposed to deal with template-only components has nothing to do.

As explained in #1958, this is a problem because it creates a difference between the initial Ember app and the rewritten-app, which prevents us from removing the rewrittent-app. This PR first removes this template-compilation step happening in Stage 2, then deals with all the side-effect this change raises. 

The main changes are the following: 
- When the `hbs` Vite plugin returns js code for a given request, it's a `load` operation rather than a `transform`
- When the virtual entry point lists all the app files, it used to add only `.js` components to the list (which was not a problem before because thanks to Stage 2 compilation step, every component had an associated `.js`). There's now a special handling for template-only components that need to be listed as well.
- A bunch of tests had stage-2-specific checks. These tests were reworked using the `CommandWatcher` so the new version focuses on what's in the Vite build output.

### DIFF
--- a/tests/scenarios/compat-template-colocation-test.ts
+++ b/tests/scenarios/compat-template-colocation-test.ts
@@ -125,8 +125,7 @@ scenarios
           });
       });
 
-      // TODO make sure that app files picks up the template-only-component
-      QUnit.skip(`app's template-only component JS is synthesized`, function (assert) {
+      test(`app's template-only component JS is synthesized`, function (assert) {
         expectAudit
           .module('./index.html')
           .resolves('/@embroider/core/entrypoint')
@@ -139,10 +138,10 @@ scenarios
               'imported template'
             );
             assert.ok(/import \{ setComponentTemplate \}/.test(contents), 'found setComponentTemplate');
-            assert.ok(/import templateOnlyComponent/.test(contents), 'found templateOnlyComponent');
+            assert.ok(/import templateOnly/.test(contents), 'found templateOnly');
 
             assert.ok(
-              /export default setComponentTemplate\(TEMPLATE, templateOnlyComponent\(\)\)/.test(contents),
+              /export default setComponentTemplate\(TEMPLATE, templateOnly\(\)\)/.test(contents),
               'default export is wrapped'
             );
             return true;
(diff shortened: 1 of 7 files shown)


BlueCutOfficial:
`splitAtRoutes-*` tests are failing because removing the synthesized .js from the rewritten-app messes up the results of the audit. It's the first time the audit system has to deal with `hbs` template-only components. The two options we discussed with @mansona are:
- An update of the audit system itself so it's able to synthesize the js. 
- Moving the `maybeSynthesizeComponentJS` logic from the `hbs` Vite plugin to the resolver.

tests/scenarios/compat-template-colocation-test.ts
```diff
@@ -84,60 +85,85 @@ scenarios
...
+            assert.ok(/import \{ setComponentTemplate \}/.test(contents), 'found setComponentTemplate');
+            assert.ok(/import templateOnly/.test(contents), 'found templateOnly');
 
-        assertFile.matches(
-          /export default setComponentTemplate\(TEMPLATE, templateOnlyComponent\(\)\)/,
-          'default export is wrapped'
-        );
+            assert.ok(
+              /export default setComponentTemplate\(TEMPLATE, templateOnly\(\)\)/.test(contents),
+              'default export is wrapped'
```
BlueCutOfficial, line 145: In this test, we used to look for `templateOnlyComponent`, and we now look for `templateOnly`. It's just a naming change. 

Before this PR, the component was compiled by the Broccoli plugin `SynthesizeTemplateOnlyComponents`, which includes the code:
```
import templateOnlyComponent from '@ember/component/template-only';
```

Now, the component is compiled by our Vite `hbs` plugin, which includes the code: 
```
import templateOnly from '@ember/component/template-only';
```

ef4 (approved):

ef4:
Fantastic work, thanks.

PR #1957, opened 2024-06-05
Move responsibility of booting the app from Embroider internals to the Ember app

BlueCutOfficial:
**The app-boot script is now part of the Ember app**

This PR implements the 1st part of #1960: 
> entrypoint.js should not have the responsibility to boot the app. We would prefer it to export the app-related modules, and the index.html should be where the app boot takes place.

- It removes the app-boot part from `virtual-entrypoint.ts` so the app-boot no longer happens in Embroider internals.
- It changes our different app templates (used for testing) to demonstrate the new way to boot the app (see the different `index.html` and `app/app.js` files).

**How it drives the decision to stop the support of content-for "app-boot"**

The app-boot script specifies instructions to boot the app. During the classic build, some Broccoli plugins and Ember libraries work together to generate the correct app-boot instructions, so it used to be something _internal_.

If you are an addon author and the way your addon works requires customizing the app-boot, you couldn't provide any configuration instructions to your users because they have no control over the app-boot anyway. But classic addons let you provide the required content yourself using the `contentFor` hook. There is an example [in ember-cli-fastboot](https://github.com/ember-fastboot/ember-cli-fastboot/blob/6904bea213a62e52701ba672a410e5794fc19b1a/packages/ember-cli-fastboot/index.js#L113).

With this PR, app developers now have control over the app-boot script: it's written directly in the `index.html`. Classic addons won't be able to impact this script, it will be the responsibility of the app.

A new error message has been implemented so if the app uses at least one classic addon providing custom content for app-boot, an explanation about how to move the instructions on the app side will be displayed.


mansona:
I think this PR should remove the support for `ember-app-boot.js` i.e. addons can't provide it any more

tests/scenarios/compat-exclude-dot-files-test.ts
```diff
@@ -74,12 +85,12 @@ appScenarios
 
         // but not be picked up in the entrypoint
         expectAudit
-          .module('./tmp/rewritten-app/index.html')
-          .resolves('/app-boot.js')
+          .module('./index.html')
+          .resolves(/\/index.html.*/) // in-html app-boot script
           .toModule()
-          .resolves('./app')
+          .resolves(/\/app\.[jt]s.*/)
```
mansona, line 91: do we really need the `[jt]` here? I really hope it would have been removed by this stage 🫠

tests/scenarios/compat-renaming-test.ts
```diff
@@ -199,98 +205,102 @@ appScenarios
...
       });
       test('renamed modules keep their classic runtime name when used as implicit-modules', function () {
         expectAudit
-          .module('./tmp/rewritten-app/index.html')
-          .resolves('/app-boot.js')
+          .module('./index.html')
+          .resolves(/\/index.html.*/) // in-html app-boot script
           .toModule()
-          .resolves('./app')
+          .resolves(/\/app\.[jt]s.*/)
```
mansona, line 264: same as above, I would love to be able to remove `ts` here 👍

tests/scenarios/compat-renaming-test.ts
```diff
@@ -199,98 +205,102 @@ appScenarios
...
+          .resolves(/.*\/-embroider-entrypoint\.js/)
           .toModule()
-          .resolves('./-embroider-implicit-modules.js')
+          .resolves(/.*\/-embroider-implicit-modules\.js/)
           .toModule()
           .withContents(contents => {
             const [, objectName] = /"somebody-elses-package\/environment": (own\d+),/.exec(contents) ?? [];
 
-            return contents.includes(
+            console.log('objectName', objectName);
```
mansona, line 273: delete me 🔪

tests/scenarios/compat-stage2-test.ts
```diff
@@ -491,105 +577,148 @@ stage2Scenarios
...
+            import { importSync } from '@embroider/macros';
+            export { default } from 'my-addon/components/hello-world';
+          `);
+      });
 
+      test('addon/hello-world.js', function () {
         expectAudit
-          .module('./node_modules/my-addon/_app_/components/hello-world.js')
-          .resolves('my-addon/components/hello-world')
-          .to('./node_modules/my-addon/components/hello-world.js', 'remapped to precise copy of my-addon');
```
BlueCutOfficial, line 592: I removed this assertion because I need to go through this resolve path anyway to test the content of `my-addon/components/hello-world.js` in another test. So I simply moved the message "remapped to precise copy of my-addon" to a comment in that other test.

tests/scenarios/compat-stage2-test.ts
```diff
@@ -491,134 +577,214 @@ stage2Scenarios
...
-        import layout from '../templates/components/hello-world';
-        import computed from '@ember/object/computed';
-        import somethingExternal from 'not-a-resolvable-package';
-        import { importSync } from "@embroider/macros";
-        export default Component.extend({
-          dynamicComponentName: computed('useDynamic', function () {
-            return this.useDynamic || 'default-dynamic';
-          }),
-          layout
-        });
```
BlueCutOfficial, line 576: Do we want to keep asserting all the imports are present, or do we just want enough proof that this is the right file that has been loaded?

ef4 (approved):

PR #1967, opened 2024-06-08
Address these issues in new apps (5.9):

NullVoxPopuli:
In a new 5.9 app, these deprecations occur

```
DEPRECATION: The parts property on path nodes is deprecated, use head and tail instead
DEPRECATION: The 'Program' visitor node is deprecated. Use 'Template' or 'Block' instead (node was 'Template')
DEPRECATION: The parts property on path nodes is deprecated, use head and tail instead
DEPRECATION: The 'Program' visitor node is deprecated. Use 'Template' or 'Block' instead (node was 'Template')
DEPRECATION: The 'Program' visitor node is deprecated. Use 'Template' or 'Block' instead (node was 'Template')
DEPRECATION: The 'Program' visitor node is deprecated. Use 'Template' or 'Block' instead (node was 'Template')
DEPRECATION: The 'Program' visitor node is deprecated. Use 'Template' or 'Block' instead (node was 'Template')
```

The goal of this PR is to get the deprecations down to 0.

The suggested fix is to change the `Program` visitor to either `Template` or `Block`.
and both of those have been around since glimmer-vm 0.39.0, which was: **2019, Jan 18**.
_after_ ember-source 3.7 :: **2019, Jan 7**
_before_ Octane Preview: ember-source 3.13 :: **2019, Sept 25**.
_but_, 0.39.0's changes weren't pulled in to ember-source until ember-source 3.17............ and using glimmer-vm 0.47.9 
- 3.17 using 0.47.9 @ **2020, Mar 4** https://github.com/emberjs/ember.js/blob/v3.17.0/package.json#L84
- 3.16 using 0.38.5-alpha.3 @ **2020, Jan 20** https://github.com/emberjs/ember.js/blob/v3.16.0/package.json#L82

`Program` is more closely analogous to `Template` which
- is introduced here: https://github.com/glimmerjs/glimmer-vm/blob/v0.39.0/packages/%40glimmer/syntax/lib/types/nodes.ts#L243
- doesn't exist in the version used by ember-source at the time: https://github.com/glimmerjs/glimmer-vm/blob/v0.38.5-alpha.3/packages/%40glimmer/syntax/lib/types/nodes.ts#L198

------------------------------------------

The node.path.parts vs node.path.head support is easy, because we have something that we can ask if its the new API or not.

### DIFF
--- a/packages/macros/src/glimmer/ast-transform.ts
+++ b/packages/macros/src/glimmer/ast-transform.ts
@@ -88,9 +88,13 @@ export function makeFirstTransform(opts: FirstTransformParams) {
           if (node.path.type !== 'PathExpression') {
             return;
           }
-          if (inScope(scopeStack, node.path.parts[0])) {
+
+          let head = 'head' in node.path ? node.path.head : node.path.parts[0];
+
+          if (inScope(scopeStack, head)) {
             return;
           }
+
           if (node.path.original === 'macroGetOwnConfig') {
             return literal(
               getConfig(node, opts.configs, opts.packageRoot, moduleName, true, packageCache),
@@ -120,7 +124,10 @@ export function makeFirstTransform(opts: FirstTransformParams) {
           if (node.path.type !== 'PathExpression') {
             return;
           }
-          if (inScope(scopeStack, node.path.parts[0])) {
+
+          let head = 'head' in node.path ? node.path.head : node.path.parts[0];
+
+          if (inScope(scopeStack, head)) {
             return;
           }
           if (node.path.original === 'macroGetOwnConfig') {
@@ -166,7 +173,7 @@ export function makeSecondTransform() {
       name: '@embroider/macros/second',
 
       visitor: {
-        Program: {
+        Block: {
           enter(node: any) {
             if (node.blockParams.length > 0) {
               scopeStack.push(node.blockParams);
@@ -182,7 +189,10 @@ export function makeSecondTransform() {
           if (node.path.type !== 'PathExpression') {
             return;
           }
-          if (inScope(scopeStack, node.path.parts[0])) {
+
+          let head = 'head' in node.path ? node.path.head : node.path.parts[0];
+
+          if (inScope(scopeStack, head)) {
             return;
           }
           if (node.path.original === 'if') {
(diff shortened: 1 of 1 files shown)


ef4:
I think the remaining failures here are now real ones from this branch.

ef4:
`Program` wasn't just the root. It also got called for every block, which updated the scopeStack.

NullVoxPopuli:
> Program wasn't just the root. It also got called for every block, which updated the scopeStack.

ah, so `Block` probably needs to be in the transform as well.

ef4:
Suggested solution: https://github.com/embroider-build/embroider/pull/1979

NullVoxPopuli:
beautiful

NullVoxPopuli:
Do you have a repro? in a fresh app I get 0 of those.
You may also want to verify that your lockfile has resolved to the newer `@embroider/macros` version for each of your dependencies that depends on it

johanrd:
Seems like upgrading @embroider/webpack removes the rest of the deprecation warnings for me for some weird reason. 

```diff
-  "@embroider/webpack": "4.0.3",
+ "@embroider/webpack": "4.0.2",
```

maybe related to that specific upgrade, but maybe some lockfile accident, as I've also recreated the lockfile and node_modules.

mkszepp:
@NullVoxPopuli @ef4 we still get this error with ember 5.9 (also 5.8) & latest embroider release.

mkszepp:
looks like there was any embroider cache in my case... now error doesn't appear anymore

shama:
We are still hitting some deprecations messages in our build with the latest dependencies. I pushed a repro app here: https://github.com/shama/embroider-deprecation-repro

I haven't had a chance to look where the deprecation messages originate from. It only happens on first the first build but not subsequent builds unless you clear out the embroider tmp folder.

In [this repo](https://github.com/shama/embroider-deprecation-repro), run `npm run clean` to remove the tmp folder and then `npm run build` to repro.

ef4:
It looks like all the remaining deprecations in there are coming from embroider's resolver transform:

https://github.com/embroider-build/embroider/blob/26f34c4c173ad222bbffc08e237f00ca78f316d4/packages/compat/src/resolver-transform.ts#L784

ef4:
It would be a helpful PR if somebody wants to go through that file (against the embroider stable branch) and update all of the `.parts` and .this` to the non-deprecated alternatives.

PR #1968, opened 2024-06-10
Update `fast-sourcemap-concat` to remove dependency `lodash.template` for consumer apps (security vulnerability)

mkszepp:
The package `fast-sourcemap-concat` v1 brings as dependency `sourcemap-validator` [see](https://github.com/ef4/fast-sourcemap-concat/blob/v1.4.0/package.json#L24).
`sourcemap-validator` brings as dependency `lodash.template` which has a security vulnerability [see](https://github.com/ben-ng/sourcemap-validator/blob/master/package.json#L26-L30)

`fast-sourcemap-concat` v2 doesn't bring anymore the dependency `sourcemap-validator`, so we don't have anymore the `lodash.template` dependency in consumer app and we solve the security vulnerability for them.

Internal in embroider addon the vulnerabilty is also present after this changes, because there is the devDependency `ember-cli-eslint`, which brings in deep still `fast-sourcemap-concat` v1.
`ember-cli-eslint` is a deprecated package and should be replaced with `eslint` [see](https://www.npmjs.com/package/ember-cli-eslint).

So we have some options to remove this vulnerability inside embroider test-app:
1. Replace `ember-cli-eslint` with `eslint`
2. Updating in `broccoli-lint-eslint` the dependency `broccoli-concat` from v3 to v4 (because `broccoli-concat` v3 brings `fast-sourcemap-concat` v1) [broccoli-lint-eslint package](https://github.com/ember-cli/broccoli-lint-eslint/blob/9cff904bad8cdafca797e23af145db807567bd3d/package.json#L37) - [broccoli-concat](https://github.com/broccolijs/broccoli-concat)

Let me know whats your idea. The owner of all packages is always ember

### DIFF
--- a/packages/compat/package.json
+++ b/packages/compat/package.json
@@ -50,7 +50,7 @@
     "chalk": "^4.1.1",
     "debug": "^4.3.2",
     "escape-string-regexp": "^4.0.0",
-    "fast-sourcemap-concat": "^1.4.0",
+    "fast-sourcemap-concat": "^2.1.1",
     "fs-extra": "^9.1.0",
     "fs-tree-diff": "^2.0.1",
     "jsdom": "^16.6.0",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -34,7 +34,7 @@
     "broccoli-source": "^3.0.1",
     "debug": "^4.3.2",
     "escape-string-regexp": "^4.0.0",
-    "fast-sourcemap-concat": "^1.4.0",
+    "fast-sourcemap-concat": "^2.1.1",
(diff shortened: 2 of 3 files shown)


mkszepp:
i have now removed `ember-cli-eslint` as it was the easier way

mansona:
Hey thanks for your contribution! 

You may want to rebase this change against the `stable` branch so that it could be released earlier. Main is preparing for a number of breaking changes so it won't be released for quite some time.

mkszepp:
ah thank you for info... created new PR for that #1969

PR #1971, opened 2024-06-11
Webpack: close the compiler

krasnoukhov:
We wanted to try webpack cache for our embroider-powered apps but it didn't work. After some investigation I realized that embroider does not close the compiler. Here's excerpt from the [doc about compiler usage](https://webpack.js.org/api/node/#run):

> Don't forget to close the compiler, so that low-priority work (like persistent caching) have the opportunity to complete.

### DIFF
--- a/packages/webpack/src/ember-webpack.ts
+++ b/packages/webpack/src/ember-webpack.ts
@@ -564,7 +564,16 @@ const Webpack: PackagerConstructor<Options> = class Webpack implements Packager 
               })
             );
           }
-          resolve(stats);
+          webpack.close(err => {
+            if (err) {
+              if (stats) {
+                this.consoleWrite(stats.toString());
+              }
+              throw err;
+            }
+
+            resolve(stats);
+          });
         } catch (e) {
           reject(e);
         }


mansona:
Hey 👋 you probably want to rebase and retarget this PR on top of stable if you are hoping that this goes out any times soon 🙈 main is breaking and won't be released for a while, and on top of that there is no/very little webpack code being executed on main right now so this change won't be exercised very well

krasnoukhov:
Hey @mansona, thanks for the tip, just made the change

mansona:
hey @krasnoukhov are you able to rebase this on stable? that should fix the remaining failing tests 👍

krasnoukhov:
Hey @mansona, unfortunately it seems it still fails

mansona:
yep sorry 🫠 other bug identified during office-hours today https://github.com/embroider-build/embroider/pull/1977

once that's merged we can rebase one more time and it should be fine 🙈

mansona:
It's merged now if you want to rebase ... Hopefully only one last time

ef4:
Thanks, merged via #1978

krasnoukhov:
Thanks for fast turnaround with the release folks 🙌 One less `patch-package` file 😁

PR #1973, opened 2024-06-11
Support corepack's packageManager field 

simonihmig:
For users that have [corepack](https://nodejs.org/api/corepack.html) enabled, this will auto-install the version of pnpm that we expect. The Github action will respect this field as well, so we have a single source of truth.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,5 @@
 name: CI
 
-env:
-  VOLTA_FEATURE_PNPM: 1
-
 on:
   workflow_dispatch:
   schedule:
--- a/package.json
+++ b/package.json
@@ -69,5 +69,6 @@
       "unlabeled": ":question: Unlabeled"
     },
     "wildcardLabel": "unlabeled"
-  }
+  },
+  "packageManager": "pnpm@8.15.8+sha256.691fe176eea9a8a80df20e4976f3dfb44a04841ceb885638fe2a26174f81e65e"
 }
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -1,8 +1,5 @@
 name: CI
 
-env:
-  VOLTA_FEATURE_PNPM: 1
```
simonihmig, line 4: I don't think that was used (anymore)!?

NullVoxPopuli (approved):

.github/workflows/ci.yml
mansona: correct, this was hanging around from before 👍

package.json
```diff
@@ -69,5 +69,6 @@
       "unlabeled": ":question: Unlabeled"
     },
     "wildcardLabel": "unlabeled"
-  }
-}
\ No newline at end of file
+  },
+  "packageManager": "pnpm@8.15.8+sha256.691fe176eea9a8a80df20e4976f3dfb44a04841ceb885638fe2a26174f81e65e"
```
mansona, line 73: does this need to have the +sha... suffix? 🤔

void-mAlex: also the action.yaml removed version 8 restriction is that correct when this 'pins' it to 8?

NullVoxPopuli: It works without the sha -- it's an integrity check to prevent installing of a MITM-injected version of the package manager

PR #1977, opened 2024-06-11
Fix modifier removal for "unless (macroCondition ...)"

ef4:
This bug was lurking since https://github.com/embroider-build/embroider/pull/1858 but it didn't actually cause a test failure until an updated glimmer-vm landed. Not entirely sure why that was, but the bug is definitely real and this is the appropriate fix.


mansona (approved):
:shipit:

PR #1981, opened 2024-06-12
Revert "Webpack: close the compiler"

krasnoukhov:
Revert of #1971 #1978 

Sorry for causing trouble with this, but that change broke rebuild. Since embroider reuses the compiler for rebuilds we can't close it because otherwise rebuilds won't work.

Maybe it would be possible to close it when process is about to exit, to save the cache if we're running the regular build, but I don't want to complicate this.


krasnoukhov:
cc @mansona @ef4

void-mAlex (approved):

PR #1984, opened 2024-06-13
narrow the audit exception for vite internals

ef4:
The binary expression vite's inserting is always marked by a special comment, so we can narrow this audit exception in order to not avoid other arbitrary binary expressions people might use.


mansona (approved):
:shipit:

PR #1985, opened 2024-06-13
Move rewritten app to ./tmp

ef4:
We're very close to eliminating the rewritten app, but vite's extremely loose `inNodeModules` behavior is breaking features that we need to use to actually get there. So we're moving the rewritten-app from `./node_modules/.embroider` to `./tmp`, since this path stops confusing vite.

The immediate motivation here was to get inline script tags working in index.html. If your index.html has the string `node_modules` anywhere in its path that vite feature breaks.


mansona (approved):
:shipit:

ef4:
Filed vite bug for the issue this PR works around: https://github.com/vitejs/vite/issues/17467

PR #1990, opened 2024-06-17
try gts/gjs resolution without checking for other files first

patricklx:
otherwise it can happen that it also matches app re-export files

### DIFF
--- a/packages/vite/src/template-tag.ts
+++ b/packages/vite/src/template-tag.ts
@@ -19,12 +19,6 @@ export function templateTag({ inline_source_map } = { inline_source_map: false }
       if (options.custom?.embroider?.isExtensionSearch) {
         return null;
       }
-      let resolution = await this.resolve(id, importer, {
-        skipSelf: true,
-      });
-      if (resolution) {
-        return resolution;
-      }
       for (let candidate of candidates(id)) {
         resolution = await this.resolve(candidate, importer, {
           skipSelf: true,
--- a/packages/vite/src/template-tag.ts
+++ b/packages/vite/src/template-tag.ts
@@ -44,7 +44,7 @@ export function templateTag({ inline_source_map } = { inline_source_map: false }
             };
           }
         } catch (e) {
-         // skip
+          // skip
         }
       }
     },
--- a/packages/vite/src/template-tag.ts
+++ b/packages/vite/src/template-tag.ts
@@ -20,7 +20,7 @@ export function templateTag({ inline_source_map } = { inline_source_map: false }
         return null;
       }
       for (let candidate of candidates(id)) {
-        resolution = await this.resolve(candidate, importer, {
+        let resolution = await this.resolve(candidate, importer, {
           skipSelf: true,
           custom: {
             embroider: {
--- a/packages/vite/src/template-tag.ts
+++ b/packages/vite/src/template-tag.ts
@@ -28,19 +28,23 @@ export function templateTag({ inline_source_map } = { inline_source_map: false }
         }
       }
       for (let candidate of candidates(id)) {
-        let resolution = await this.resolve(candidate, importer, {
-          skipSelf: true,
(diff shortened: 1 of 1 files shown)


packages/vite/src/template-tag.ts
```diff
@@ -19,26 +19,32 @@ export function templateTag({ inline_source_map } = { inline_source_map: false }
...
         if (resolution) {
-          return {
-            id: resolution.id,
-          };
+          return resolution;
+        }
+      }
+      for (let candidate of candidates(id)) {
+        try {
+          let resolution = await this.resolve(candidate, importer, {
```
patricklx, line 32: in try catch because it can fail during resolving: 
`[commonjs--resolver] Missing "./modulepreload-polyfill.gjs" specifier in "vite" package`

NullVoxPopuli:
I tried this patch: https://github.com/NullVoxPopuli/package-majors/pull/41
got:
```
Could not resolve "../components/intro-text" from "tmp/rewritten-app/templates/query.gts"
file: /home/nvp/Development/NullVoxPopuli/package-majors/tmp/rewritten-app/templates/query.gts
    at getRollupError (file:///home/nvp/Development/NullVoxPopuli/package-majors/node_modules/.pnpm/rollup@4.18.0/node_modules/rollup/dist/es/shared/parseAst.js:396:41)
    at error (file:///home/nvp/Development/NullVoxPopuli/package-majors/node_modules/.pnpm/rollup@4.18.0/node_modules/rollup/dist/es/shared/parseAst.js:392:42)
    at ModuleLoader.handleInvalidResolvedId (file:///home/nvp/Development/NullVoxPopuli/package-majors/node_modules/.pnpm/rollup@4.18.0/node_modules/rollup/dist/es/shared/node-entry.js:19102:24)
    at file:///home/nvp/Development/NullVoxPopuli/package-majors/node_modules/.pnpm/rollup@4.18.0/node_modules/rollup/dist/es/shared/node-entry.js:19062:26
 ELIFECYCLE  Command failed with exit code 1.
 ELIFECYCLE  Test failed. See above for more details.

```

so, I'm not sure this is the right fix either (now I can't resolve my own app components)

NullVoxPopuli (changes requested):
didn't work :sweat_smile:

packages/vite/src/template-tag.ts
```diff
@@ -19,26 +19,32 @@ export function templateTag({ inline_source_map } = { inline_source_map: false }
...
       }
-      let resolution = await this.resolve(id, importer, {
-        skipSelf: true,
-      });
-      if (resolution) {
-        return resolution;
-      }
-      for (let candidate of candidates(id)) {
-        resolution = await this.resolve(candidate, importer, {
+      if (id.endsWith('.gjs') || id.endsWith('.gts')) {
```
ef4, line 22: This condition doesn't make sense. If the request already has an explicit `.gjs` extension, the default resolving will handle it fine and this plugin will need no custom behavior in `resolveId`.

packages/vite/src/template-tag.ts
```diff
@@ -19,26 +19,32 @@ export function templateTag({ inline_source_map } = { inline_source_map: false }
...
+              },
+            },
+          });
+          if (resolution) {
+            return {
+              id: resolution.id,
+            };
+          }
+        } catch (e) {
+          // skip
```
ef4, line 47: This is absolutely not OK. It will hide bugs.

ef4 (changes requested):
I don't understand the point of this PR. It's very much intentional that the initial `this.resolve` might match app-reexports. It might match *anything* that the rest of the plugin stack is doing, and that is by design.

This plugin only does customized resolving after all other resolving has failed. If it succeeds, this plugin has nothing to do.

Please provide tests that illustrate the problem you are trying to solve.

PR #1991, opened 2024-06-18
Esbuild resolver/ handle template-only components correctly

### LINKED ISSUE #1988: `esbuild-resolver` throw errors when starting an app with template-only components
PR #1955 
1 - removed a compilation step that was used to generate `js` files for template-only components in the rewritten-app. 
2 - handle the side-effects this changes cause.

But I think one side-effect was forgotten: when the EsBuild resolver scans the dependencies and handles a template-only component by js, it no longer finds it because it tries to [`readFileSync` the rewritten-app/components/my-component.js](https://github.com/embroider-build/embroider/blob/main/packages/vite/src/esbuild-resolver.ts#L120).

BlueCutOfficial:
Fixes #1988

When Esbuild resolver tries to load a `js` or `ts` file that doesn't exist in the rewritten-app, it tries to load the `hbs` instead because it could be a template-only component.

### DIFF
--- a/packages/vite/src/esbuild-resolver.ts
+++ b/packages/vite/src/esbuild-resolver.ts
@@ -1,7 +1,7 @@
 import type { Plugin as EsBuildPlugin } from 'esbuild';
 import { type PluginItem, transform } from '@babel/core';
 import { ResolverLoader, virtualContent, locateEmbroiderWorkingDir, explicitRelative } from '@embroider/core';
-import { readFileSync, readJSONSync } from 'fs-extra';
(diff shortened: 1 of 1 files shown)


patricklx:
How can i reproduce this?
I tried to add a `app/components/template-only.hbs`
And used it in application.hbs template and i did not get an error.
Do i need to set some other config?

BlueCutOfficial:
@patricklx note that this PR should be closed at some point, Ed suggested a better approach during the last office hours, I guess the fix should be pushed at some point.

PR #1994, opened 2024-06-20
MacrosConfig should sync globalConfigs between copies

ef4:
It's a bug that most of the MacrosConfig state is synchronized between copies of @embroider/macros but the globalConfigs object is not.

### DIFF
--- a/packages/macros/src/macros-config.ts
+++ b/packages/macros/src/macros-config.ts
@@ -22,16 +22,16 @@ export type Merger = (
   }
 ) => object;
 
+interface GlobalSharedEntry {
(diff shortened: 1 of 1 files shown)


mansona (approved):
:shipit:

runspired:
love this, was a giant troll figuring out the `app` was a necessary key when creating `@warp-drive/build-config`, would love to drop both that and `__dirname` as args if possible.

ef4:
It was a bug that some state wasn't shared for a given key, but there's no plan to drop the key entirely. That would imply that it's never possible to talk about two apps in one node process.

PR #1996, opened 2024-06-20
Merge stable into main

mansona:


patricklx:
looks like you also need the windows fix from https://github.com/embroider-build/embroider/pull/1855/commits/5efaa0e3e67d992fc302ecd6f4aae90ae2f3f172

mansona:
@patricklx any idea why we might be hitting this now? what changed?

mansona:
I've added @patricklx 's change because I tested this locally and wasn't able to re-create the issue 🫠 

If this fixes CI we can just merge it and move on, Windows shouldn't be blocking our process like this 😭

PR #2002, opened 2024-06-21
test D drive for windows temp

mansona:
There are some issues reported with windows disks running slow so I'm just testing out moving our temp to `D:\`


.github/workflows/ci.yml
```diff
@@ -51,7 +51,7 @@ jobs:
           use_lockfile: false
       - name: Work around windows short path alias # https://github.com/actions/runner-images/issues/712
         if: runner.os == 'Windows'
-        run: echo "TEMP=$env:USERPROFILE\AppData\Local\Temp" >> $env:GITHUB_ENV
+        run: echo "TEMP=D:\temp" >> $env:GITHUB_ENV
```
patricklx, line 54: I guess you need mkdir as well `new-item d:\temp -ItemType Directory`

patricklx: maybe permission is limited to `D:\\a\\` ?

patricklx:
@mansona how about rerunning this? It should then pickup the new fixturify package, right?

.github/workflows/ci.yml
mansona: this comment is no longer relevant with the new implementation

PR #2004, opened 2024-06-21
use pnpm to publish unstable

mansona:
It looks like we're getting stung with https://github.com/npm/cli/issues/5007 when we're using npm to publish from this monorepo, so I'm going to swap it to pnpm for unstable 👍 

Edit: I tested this locally and I'm happy it's working as expected now 🎉

### DIFF
--- a/test-packages/unstable-release/publish.js
+++ b/test-packages/unstable-release/publish.js
@@ -10,7 +10,9 @@ async function publish() {
   for (let workspace of publicWorkspaces) {
     console.info(`Publishing ${workspace}`);
     try {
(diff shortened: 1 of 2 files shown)


ef4:
I think this implies we should delete the manual setting of dependency versions in update-release/version.js.

Because pnpm will handle the "workspace:" protocol automatically.

NullVoxPopuli:
@ef4 I don't think _I'd_ trust pnpm's behavior to use our specific unstable versions though

NullVoxPopuli (approved):

Issue #1889, opened 2024-04-23
`{{unique-id}}` helper is broken in 3.4.8


jknight12882:
Looks related to the changes made in `@embroider/compat@3.4.7`. Throws the following error when rendering a template with the `{{unique-id}}` helper

```
manager.js:394 Uncaught (in promise) Error: Attempted to use a value as a helper, but it was not an object or function. Helper definitions must be objects or functions with an associated helper manager. The value was: undefined
    at getInternalHelperManager (manager.js:394:1)
    at ConstantsImpl.helper (program.js:95:1)
    at resolveHelper (opcode-compiler.js:207:1)
    at encodeOp (opcode-compiler.js:1976:1)
    at pushOp (opcode-compiler.js:1911:1)
    at opcode-compiler.js:420:1
    at Compilers.compile (opcode-compiler.js:405:1)
    at expr (opcode-compiler.js:553:1)
    at CompilePositional (opcode-compiler.js:622:1)
    at opcode-compiler.js:1833:1
    at Compilers.compile (opcode-compiler.js:405:1)
    at compileStatements (opcode-compiler.js:1914:1)
    at maybeCompile (opcode-compiler.js:1896:1)
    at CompilableTemplateImpl.compile (opcode-compiler.js:1879:1)
    at VM.compile (runtime.js:4678:1)
    at Object.evaluate (runtime.js:1930:1)
    at AppendOpcodes.evaluate (runtime.js:1033:1)
    at LowLevelVM.evaluateSyscall (runtime.js:4109:1)
    at LowLevelVM.evaluateInner (runtime.js:4080:1)
    at LowLevelVM.evaluateOuter (runtime.js:4073:1)
    at VM.next (runtime.js:4871:1)
    at VM._execute (runtime.js:4858:1)
    at VM.execute (runtime.js:4833:1)
    at runtime.js:4916:1
    at runInTrackingTransaction (validator.js:133:1)
    at TemplateIteratorImpl.sync (runtime.js:4916:1)
    at index.js:4684:1
    at RootState.render (index.js:4653:1)
    at index.js:4943:1
    at inTransaction (runtime.js:3985:1)
    at Renderer._renderRoots (index.js:4925:1)
    at Renderer._renderRootsTransaction (index.js:4969:1)
    at Renderer._renderRoot (index.js:4914:1)
    at Renderer._appendDefinition (index.js:4839:1)
    at Renderer.appendOutletView (index.js:4829:1)
[...]

PR #2005, opened 2024-06-22
unique-id helper import based on ember-source version

void-mAlex:
fixes https://github.com/embroider-build/embroider/issues/1889

### DIFF
--- a/packages/compat/src/resolver-transform.ts
+++ b/packages/compat/src/resolver-transform.ts
@@ -18,6 +18,7 @@ import { Resolver, locateEmbroiderWorkingDir } from '@embroider/core';
 import type CompatOptions from './options';
 import type { AuditMessage, Loc } from './audit';
 import { camelCase, mergeWith } from 'lodash';
+import { satisfies } from 'semver';
 
 type Env = WithJSUtils<ASTPluginEnvironment> & {
   filename: string;
@@ -41,6 +42,7 @@ export interface CompatResolverOptions extends CoreResolverOptions {
 
 export interface Options {
   appRoot: string;
+  emberVersion: string;
 }
 
 type BuiltIn = {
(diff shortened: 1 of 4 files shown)


packages/compat/src/resolver-transform.ts
```diff
@@ -167,7 +175,11 @@ class TemplateResolver implements ASTPlugin {
 
   private moduleResolver: Resolver;
 
-  constructor(private env: Env, private config: CompatResolverOptions) {
+  constructor(
+    private env: Env,
+    private config: CompatResolverOptions,
+    private builtInsForEmberVersion: ReturnType<typeof builtInKeywords>
```
mansona, line 181: since you're converting it to a function, why would you pass the function to the constructor and not just make it part of the class? 🤔

void-mAlex: not passing the function to the class here, just the result of the function invocation
as it needs the result and don't want to invoke the function inside the class due to perf reasons of not evaluating the semver more than once

void-mAlex: you don't need memoize as it's only called once at the place I call it, memoize would mean having to pass in the function and execute it inside each time to get back the same result
your ember version doesn't change (hopefully during the execution of the build) so this would be fewer function invocations that need to look up memoized content

mansona (approved):
This looks great, the fact that we have a test covering the behaviour means I really don't think we should delay getting this merged 👍

packages/compat/src/resolver-transform.ts
mansona: all seems good to me 😂 thanks for humouring me and my questions 🎉

void-mAlex: that's what the process is for :)

PR #2006, opened 2024-06-25
Virtual entrypoint export modules

### LINKED ISSUE #1960: Work on the new app boot implementation
## Context
`entrypoint.js` is a virtual file produced by Embroider that corresponds to - literally - the "entry point" to the Ember app. Currently, if we look at the content of this file in the browser, here is what we see:

Image

1) It has an instruction to boot the app at the bottom.
2) It defines modules relying on the `define`

### 1) App boot
`entrypoint.js` should not have the responsibility to boot the app. We would prefer it to export the app-related modules, and the `index.html` should be where the app boot takes place.

Task:
- Removes app boot from the bottom of the virtual entrypoint
- Implement the app boot in `index.html`

```diff
<!-- index.html -->
- <script src="/@embroider/core/entrypoint" type="module"></script>
+ <script type="module">
+     import Application from './app.js';
+     import environment from './config/environment.js';
+     Application.create(environment.APP);
+ </script>
```

```diff
// my-app/app/app.js

import Application from '@ember/application';
import Resolver from 'ember-resolver';
import loadInitializers from 'ember-load-initializers';
import config from './config/environment';

+ import '@embroider/core/entrypoint';

export default class App extends Application {
  modulePrefix = config.modulePrefix;
  podModulePrefix = config.podModulePrefix;
  Resolver = Resolver;
}

loadInitializers(App, config.modulePrefix);
```

### 2) Modules definition
We don't want Embroider to rely on `define`. Using this method should rather be an implementation detail that belongs to the Ember app. 

- Instead of defining modules directly in the `entrypoint.js`, we would prefer this file to export modules (on the same model as implicit modules):

Image

- Then it's the Ember app that imports these modules and owns the `define` implementation:

```diff
// my-app/app/app.js

import Application from '@ember/application';
import Resolver from 'ember-resolver';
import loadInitializers from 'ember-load-initializers';
import config from './config/environment';

+ import allModules from '@embroider/core/modules';
+ 
+ for (const [name, module] of Object.entries(allModules)) {
+   d(name, function () {
+     return module;
+   });
+ }

export default class App extends Application {
  modulePrefix = config.modulePrefix;
  podModulePrefix = config.podModulePrefix;
  Resolver = Resolver;
}

loadInitializers(App, config.modulePrefix);
```

BlueCutOfficial:
This PR follows #1957 and fixes #1960. It implements the second part of the issue:

> We don't want Embroider to rely on `define`. Using this method should rather be an implementation detail that belongs to the Ember app. Instead of _defining_ modules directly in the `entrypoint.js`, we would prefer this file to export modules, then the Ember app imports these modules and owns the `define` implementation.

- It moves the module definition from Embroider internals to the Ember `app.js`
- All modules to define now constitute the virtual `entrypoint.js` default export.

### DIFF
--- a/packages/core/src/virtual-entrypoint.ts
+++ b/packages/core/src/virtual-entrypoint.ts
@@ -181,7 +181,6 @@ import environment from './config/environment';
 
 {{#each amdModules as |amdModule index| ~}}
   import * as amdModule{{index}} from "{{js-string-escape amdModule.buildtime}}"
-  d("{{js-string-escape amdModule.runtime}}", function(){ return amdModule{{index}}; });
 {{/each}}
 
 {{#if fastbootOnlyAmdModules}}
@@ -226,6 +225,12 @@ w._embroiderEngineBundles_ = [
   {{/each}}
 ]
 {{/if}}
+
+export default {
+  {{#each amdModules as |amdModule index| ~}}
+    "{{js-string-escape amdModule.runtime}}": amdModule{{index}},
+  {{/each}}
+};
 `) as (params: {
   amdModules: { runtime: string; buildtime: string }[];
   fastbootOnlyAmdModules?: { runtime: string; buildtime: string }[];
--- a/packages/core/src/virtual-entrypoint.ts
+++ b/packages/core/src/virtual-entrypoint.ts
@@ -168,12 +168,7 @@ import environment from './config/environment';
 
 {{#if defineModulesFrom ~}}
   import implicitModules from "{{js-string-escape defineModulesFrom}}";
-
-  for(const [name, module] of Object.entries(implicitModules)) {
-    d(name, function() { return module });
-  }
 {{/if}}
-
 
 {{#each eagerModules as |eagerModule| ~}}
   import "{{js-string-escape eagerModule}}";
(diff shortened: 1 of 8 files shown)


tests/ts-app-template/app/app.ts
```diff
@@ -1,9 +1,16 @@
 import Application from '@ember/application';
+import compatModules from '@embroider/core/entrypoint';
```
BlueCutOfficial, line 2: Locally I have a TypeScript error `Cannot find module '@embroider/core/entrypoint' or its corresponding type declarations.` but I haven't figured out yet how to declare a virtual module.

mansona: I wonder if it tries to resolve the `.d.ts` copy of the internal name 🤔 i.e. whatever the virtual content deals with `-embroider-entrypoint.js` is there a resolve request for `-embroider-entrypoint.d.ts` through the same mechanism? 

Does this make sense?

patricklx: i think you just need somewhere
```ts
declare module '@embroider/core/entrypoint' {
  export default ...whatever needs to be here;
}
```

patricklx: or make a `entrypoint.d.ts` in @embroider/core package?

patricklx: maybe it needs to be in global.d.ts ? only if a d.ts file has no imports it becomes a ambient module declarations

patricklx: oh, that also has an import. maybe create a `entry.d.ts` file then?

BlueCutOfficial: @patricklx that worked, thank you! 🙌

packages/core/src/virtual-entrypoint.ts
```diff
@@ -154,7 +154,6 @@ export function renderEntrypoint(
 const entryTemplate = compile(`
 import { importSync as i, macroCondition, getGlobalConfig } from '@embroider/macros';
 let w = window;
```
mansona, line 156: I wonder what the benefit of having window aliased like this is now that we've removed the `let d = w.define;` 🤔

mansona:
my 2 comments are super minor and I don't think we should block the PR on them (unless you would prefer to address them?)

PR #2007, opened 2024-06-26
Add just the necessary files to rollup watch mode

vstefanovic97:
I noticed when working on a v2 addon that it was considering all files inside `src` for watch mode.

This is not ideal, as in some cases we want to have exclusions, eg. if we have something (eg [postcss-d-ts](https://github.com/askirmas/postcss-d-ts)) that generated type declaration files for our css we don't want to have those files included in watch mode or else they will trigger infinite re-builds.

Made the change so we explicitly add only the files we need

### DIFF
--- a/packages/addon-dev/package.json
+++ b/packages/addon-dev/package.json
@@ -25,6 +25,14 @@
   "scripts": {
     "test": "jest"
   },
+  "peerDependencies": {
+    "rollup": "^4.6.0"
+  },
+  "peerDependenciesMeta": {
+    "rollup": {
+      "optional": true
+    }
+  },
   "dependencies": {
     "@embroider/core": "workspace:^",
     "@rollup/pluginutils": "^4.1.1",
--- a/packages/addon-dev/src/rollup-keep-assets.ts
+++ b/packages/addon-dev/src/rollup-keep-assets.ts
@@ -14,12 +14,6 @@ export default function keepAssets({
   return {
     name: 'copy-assets',
 
-    // Prior to https://github.com/rollup/rollup/pull/5270, we cannot call this
-    // from within `generateBundle`
-    buildStart() {
-      this.addWatchFile(from);
-    },
-
     // imports of assets should be left alone in the source code. This can cover
     // the case of .css as defined in the embroider v2 addon spec.
     async resolveId(source, importer, options) {
@@ -44,6 +38,8 @@ export default function keepAssets({
         globs: include,
         directories: false,
       })) {
+        this.addWatchFile(join(from, name));
+
         this.emitFile({
           type: 'asset',
           fileName: name,
--- a/packages/addon-dev/src/rollup-public-entrypoints.ts
+++ b/packages/addon-dev/src/rollup-public-entrypoints.ts
@@ -16,14 +16,14 @@ export default function publicEntrypoints(args: {
   return {
     name: 'addon-modules',
     async buildStart() {
-      this.addWatchFile(args.srcDir);
-
       let matches = walkSync(args.srcDir, {
         globs: [...args.include, '**/*.hbs', '**/*.ts', '**/*.gts', '**/*.gjs'],
         ignore: args.exclude,
       });
 
       for (let name of matches) {
+        this.addWatchFile(path.join(args.srcDir, name));
+
         // the matched file, but with the extension swapped with .js
         let normalizedName = normalizeFileExt(name);
 


packages/addon-dev/src/rollup-keep-assets.ts
```diff
@@ -14,12 +14,6 @@ export default function keepAssets({
   return {
     name: 'copy-assets',
 
-    // Prior to https://github.com/rollup/rollup/pull/5270, we cannot call this
```
vstefanovic97, line 17: From what I can see this issue got resolved

packages/addon-dev/src/rollup-public-entrypoints.ts
```diff
@@ -16,14 +16,14 @@ export default function publicEntrypoints(args: {
   return {
     name: 'addon-modules',
     async buildStart() {
-      this.addWatchFile(args.srcDir);
-
       let matches = walkSync(args.srcDir, {
         globs: [...args.include, '**/*.hbs', '**/*.ts', '**/*.gts', '**/*.gjs'],
         ignore: args.exclude,
```
vstefanovic97, line 21: Now it will take into account `exclude` option and exclude files from watch mode as well

NullVoxPopuli (approved):

packages/addon-dev/src/rollup-keep-assets.ts
mansona: They have a nifty bot that tells you what version of rollup includes the PR 🎉 apparently this was fixed in `rollup@4.6.0`.

I wonder if we have any peer dependency reference or anything that would allow us to say which versions of rollup are supported 🤔

mansona: also if we are changing the "rollup support matrix" this one change could technically make this a breaking change... maybe 🤔

vstefanovic97: @mansona that is a good point, I can add `rollup` as a peerDep to `@embroider/addon-dev`'s package.json file if you feel like that is the right approach?
And it might we a major release, if we take that into account?

packages/addon-dev/src/rollup-public-entrypoints.ts
```diff
@@ -16,14 +16,14 @@ export default function publicEntrypoints(args: {
...
       });
 
       for (let name of matches) {
+        this.addWatchFile(path.join(args.srcDir, name));
```
void-mAlex, line 25: what does this mean for css resources/images and paths that are not matched here?

vstefanovic97: @void-mAlex, if you just want to move some static assets to `dist` folder, then you should relly on the `rollup-keep-assets` file, but if you are talking about for example importing css or images inside of js files then it should be up to which ever rollup plugin you are using for that to add the css/image to the watched list.

For example, let's say you are doing
```
import styles from './foo.css'
```
And using [rollup-plugin-postcss](https://www.npmjs.com/package/rollup-plugin-postcss) to build this, in that case it is `rollup-plugin-postcss`'s job to include the css file to the watch list, which is does [here](https://github.com/egoist/rollup-plugin-postcss/blob/master/src/index.js#L124) I believe

packages/addon-dev/src/rollup-keep-assets.ts
mansona: so the fact that rollup does it really makes me feel like we should do the same thing. And we need to declare the minimum version of rollup that we support

if you want to go ahead and add that in this PR we can mark it as breaking and it will automatically get picked up and released correctly 👍

vstefanovic97: @mansona I added the change, take a look when you can

packages/addon-dev/package.json
```diff
@@ -25,6 +25,14 @@
   "scripts": {
     "test": "jest"
   },
+  "peerDependencies": {
+    "rollup": "^4.6.0"
+  },
+  "peerDependenciesMeta": {
+    "rollup": {
+      "optional": true
```
mansona, line 33: is this really true? 🤔 I mean there is no way that you can use this plugin in rollup without having rollup right?

vstefanovic97: @mansona as far as I saw, all the plugins from https://github.com/rollup/plugins/tree/master/packages have this, so thought it would be best to follow their convention

vstefanovic97: In theory maybe it's if someone build a plugin or wrapper on top of a plugin? IDK
But just left it their since I thought there must be a reason they have it

simonihmig (approved):
LGTM!

We do have addon-dev watch tests that would detect regressions, right? Or does this warrant some more manual testing?

mansona:
@simonihmig yea I'm happy this is covered 👍 LGTM

packages/addon-dev/package.json
mansona: I mean let's just go with what rollup does themselves 😂 👍

vstefanovic97:
Thank you @mansona, @void-mAlex,  @NullVoxPopuli , @simonihmig for the reviews and merging this so quickly!

simonihmig:
Thanks for your contribution! 👍🎉

patricklx:
I do not think this was correct. Now it does not detect new files added to the src folder

patricklx:
I think the right solution would have been to either setup watch exlcude or watch include
https://rollupjs.org/configuration-options/#watch-exclude

PR #2010, opened 2024-07-01
Fix v1 app self import

mansona:
This pulls the fixes out of https://github.com/embroider-build/embroider/pull/1997 that we know will fix a lot of people using vite. I'm not merging https://github.com/embroider-build/embroider/pull/1997 yet because it adds a test that will still fail because the work hasn't gone far enough

I don't want to block these fixes on solving the harder issues


mansona:
this was fixed in https://github.com/embroider-build/embroider/pull/2029

PR #2016, opened 2024-07-04
New babel config

### LINKED ISSUE #1962: implement the "correct solution" for babel config in vite
The current `babel.config.cjs` file imports something that is being built from the Embroider pre-build and re-exports it. We need to change this so that it is more "owned" by the app author. This means that we need to provide a function that will give you all the config necessary from classic (v1) addons that will be incorporated into a **real** babel config

Notes: 

- we should not rely on ember-cli-babel outputting any config, anything that ember-cli-babel adds should be part of the **real** babel config
- presets should always be in the top-level babel config and owned by the developer

A really basic design of this new babel config would look like: 

```js
const { loadLegacyBabelPlugins } = require('@embroider/compat/babel');

module.exports = {
  "plugins": [
    // the default set of plugins needed by an ember-app - currently provided by ember-cli-babel

    // this function will load all the other plugins that were provided by classic addons
    ...loadLegacyBabelPlugins(),
  ],
  "presets": [
    [
      "@babel/preset-env",
      {
        "modules": false,
        "targets": {
          "browsers": [
            "last 1 Chrome versions",
            "last 1 Firefox versions",
            "last 1 Safari versions"
          ]
        }
      }
    ]
  ]
}

```

BlueCutOfficial:
Fixes #1962 

TODO:
- [ ] If addons bring more than plugins (e.g. presets), use error messages to say it's not going to work. 
- [ ] Using test templates, check if we can remove the dependency on ember-cli-babel and write a dedicated test that still relies on it.
- [ ] Make sure we can use some `exclude` option with the new config that can achieve what `skipBabel` does.
- [ ] Merging app and legacy config: do we want to put effort in this case and let the app win? If so, how to identify the same plugin is defined on both side? (legacy json uses full path, so we would need to resolve app plugins to compare the paths?)
- [ ] Maybe rename `babel.ts` to `load-babel-config.ts` to be more explicit it's config loading and not some Babel plugin
- [ ] Check we can move `loadPluginDebugMacros` right after `loadLegacyPlugins` to simplify
- [ ] Create new issues to track the work to do on the upgrade guide


packages/compat/src/compat-app.ts
```diff
@@ -225,21 +225,8 @@ export default class CompatApp {
...
-    const babelConfig = babelAddon.buildBabelOptions({
-      'ember-cli-babel': {
-        ...this.legacyEmberAppInstance.options['ember-cli-babel'],
-        includeExternalHelpers: true,
-        compileModules: false,
-        disableDebugTooling: false,
-        disablePresetEnv: false,
-        disableEmberModulesAPIPolyfill: false,
-      },
-    });
```
BlueCutOfficial, line 239: Can I really remove this? I understand one goal of this task is to remove the dependency of any Ember app to ember-cli-babel, but I am not sure about this part.

In the present work-in-progress I debugged a build to identify what fields ember-cli-babel generates for the app-template and I moved the content from `.embroider/_babel_config_.json` to `babel.config.cjs`.

This 👆 is a set of "forced" options for ember-cli-babel, but it's not _all_ the [available options](https://www.npmjs.com/package/ember-cli-babel#options). What if more options are passed that impact the result list of plugins and presets? Should I keep `.embroider/_babel_config_.json` as it is even if duplicates some plugins with the blueprint? or is this point irrelevant and why?

BlueCutOfficial:
@ef4, @mansona,

About modifying this block 👉 https://github.com/embroider-build/embroider/blob/main/packages/compat/src/compat-app.ts#L244C1-L255C1 to help developers get how to update their app. I need some rough idea.

I understand the value of identifying app's options and addon's option to output good messages (e.g. addons' presets no longer working, move the app's options to `babel.config.cjs`...), but I don't think I can know that using only `legacyAppInstance.options.babel`. This field comes from ember-cli, which concat app's options and addon's option, so it's already an as-is parameter when we enter Embroider default pipeline. The only way I see to find what options come from where is to loop through addons, but if it's indeed the right thing to do, I wonder what would be the best place. Can I take advantage somehow of an existing loop, and create new information for `legacyAppInstance` to use?

ef4 (commented):
The only state that we need to capture during the prebuild is the plugins that were inserted by v1 addons. Most of our own plugins don't need to be in that legacy plugins list. They can be provided as directly importable utilities in the babel config.

Here is where we add many plugins: https://github.com/embroider-build/embroider/blob/93726590f5515de8e7194c2ffd9c4cd7b7e9fa89/packages/compat/src/compat-app-builder.ts#L227-L290

- syntax-dynamic-import may not be needed at all, I would be surprised if it's not enabled by default in new-enough babel. If it is needed, it can go directly in the user's config.
- rename-require-plugin is only needed by webpack, so it can be dropped entirely here
- babel-plugin-ember-template-compilation is important for users to be able to control in their own babel config. We will need to work on cleanly generating its options. For now, we could provide an importable utilities that builds the options for you.
- Macros is similar, users should be able to see it in their config, but right now it has a not-user-friendly API, so it could be an importable utility that wraps over that for now.
- template colocation could go directly into the users's config.
- babel-pluign-adjust-imports is a v1-addon rule-system-supporting plugin. I think we should consider renaming `loadLegacyPlugins()` to `v1AddonSupport()` and then it mean *both* the v1-addon-inserted plugins plus any of our own hacks like babel-plugin-adjust-imports that exist purely to support v1 addons. That way users either add or remove all v1 addon support, without needing to study a lot of details.

tests/addon-template/package.json
```diff
@@ -33,6 +33,10 @@
   },
   "devDependencies": {
     "@babel/core": "^7.19.3",
+    "@babel/plugin-proposal-decorators": "^7.22.5",
+    "@babel/plugin-transform-class-properties": "7.24.7",
+    "@babel/plugin-transform-private-methods": "7.24.7",
+    "@babel/plugin-transform-private-property-in-object": "^7.24.7",
```
ef4, line 39: This is a non-blocking comment, but it's probably safe for us to replace all four of these with https://github.com/ef4/decorator-transforms instead. It could be a separate PR.

BlueCutOfficial:
I could never finish this PR after the Embroider Initiative stopped, Ed implemented it in #2102.

PR #2028, opened 2024-07-05
Rename #embroider_compat/* to @embroider/virtual/*

### LINKED ISSUE #2026: Rename #embroider_compat/* to @embroider/virtual/*
Rename `#embroider_compat/(component|helper|modifier|ambiguous)/<path>` for invokable to `@embroider/virtual/(component|helper|modifier|ambiguous)/<path>` to stop confusing esbuild.

- Implement the rename by looking for references to `#embroider_compat`.
- `shared-internals/src/paths.ts`: pay attention to `const postfixRE = /[?#].*$/s;` that used to match `#embroider_compat` pattern: if we no longer have any request from a file starting with `#`, then there's no reason to separate `postfixRE` and `postfixREQueryParams` cases. `cleanUrl('@embroider/virtual/fancy?query="p"', false)` returns `@embroider/virtual/fancy` and we no longer need a special pattern to have a `#embroider` considered part of the path name.

BlueCutOfficial:
Fixes #2026

### DIFF
--- a/packages/compat/src/babel-plugin-adjust-imports.ts
+++ b/packages/compat/src/babel-plugin-adjust-imports.ts
@@ -119,7 +119,7 @@ function applyRules(
             t,
             adder,
             path,
-            `#embroider_compat/components/${dasherizedName}`,
+            `@embroider/virtual/components/${dasherizedName}`,
             `${lookup.owningEngine.packageName}/components/${dasherizedName}`
           )
         );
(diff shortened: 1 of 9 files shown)


ef4:
This is not going to be necessary. I have confirmed that we can do all this resolution from our esbuild resolver plugin before vite's depscan plugin sees them.

ef4:
(I'm not opposed to making the change for purely aesthetic reasons, if we want all our virtual stuff to match under one namespace. But we thought this would be necessary to get vite to respect these as potentially external deps, and that is not correct.)

BlueCutOfficial:
I have no strong opinion, we have this PR ready and it has the benefit of slightly simplifying the code in `shared-internals/src/paths.ts` 🤷‍♀️ As you like really, feel free to merge or close.

BlueCutOfficial:
Discussed with Chris: this PR brings value but we should wait for https://github.com/embroider-build/embroider/pull/2029 because it brings tests regarding dependencies making their way to Vite deps.

BlueCutOfficial:
@mansona rebased on `main` :)

PR #2032, opened 2024-07-07
add more vite dep tests

patricklx:


packages/vite/src/build.ts
```diff
@@ -47,7 +47,7 @@ export function compatPrebuild(): Plugin {
       viteCommand = command;
       viteMode = mode;
     },
-    async buildStart() {
+    async configResolved() {
```
patricklx, line 50: this is necessary, otherwise vite will not pick up changes to files

PR #2033, opened 2024-07-08
Remove deprecations warnings in resolver transform

### LINKED ISSUE #2031: Deprecations in TemplateResolver
When upgrading to Ember 5.9 we're seeing a huge number of deprecation messages in our build sourced in [glimmer-vm](https://github.com/glimmerjs/glimmer-vm/pull/1568/commits/02313376f21a606841b00a7789fd6128a1b292af) and triggered here in `packages/compat/src/resolver-transform.ts` (in several places) when `staticHelpers` is enabled.

```cli
DEPRECATION: The this property on path nodes is deprecated, use head.type instead
DEPRECATION: The parts property on path nodes is deprecated, use head and tail instead
DEPRECATION: The this property on path nodes is deprecated, use head.type instead
...
```

**Reproduction**:
```cli
ember new test-app --pnpm
cd test-app
pnpm add -D @embroider/core @embroider/compat @embroider/webpack webpack
```

Enable embroider and static helpers in `ember-cli-build.js`

```javascript
'use strict';

  const app = new EmberApp(defaults, {
    // Add options here
  });

    staticHelpers: true,
  });
};
```

```cli
pnpm build

> test-app@0.0.0 build
> ember build --environment=production

Environment: production
DEPRECATION: The this property on path nodes is deprecated, use head.type instead
DEPRECATION: The parts property on path nodes is deprecated, use head and tail instead
```

mkszepp:
Since ember v5.9 there are deprecation warnings while build

```
DEPRECATION: The this property on path nodes is deprecated, use head.type instead
DEPRECATION: The parts property on path nodes is deprecated, use head and tail instead
```

As described in https://github.com/embroider-build/embroider/pull/1967#issuecomment-2207031045, we should replace all `.parts` and `.this` with non deprecated code.

fix #2031

### DIFF
--- a/packages/compat/src/resolver-transform.ts
+++ b/packages/compat/src/resolver-transform.ts
@@ -777,7 +777,7 @@ class TemplateResolver implements ASTPlugin {
       if (node.path.type !== 'PathExpression') {
         return;
       }
-      let rootName = node.path.parts[0];
+      let rootName = headOf(node.path);
       if (this.scopeStack.inScope(rootName, path)) {
         return;
       }
@@ -819,7 +819,7 @@ class TemplateResolver implements ASTPlugin {
       if (node.path.this === true) {
         return;
       }
-      if (this.scopeStack.inScope(node.path.parts[0], path)) {
+      if (this.scopeStack.inScope(headOf(node.path), path)) {
         return;
       }
       if (node.path.original === 'component' && node.params.length > 0) {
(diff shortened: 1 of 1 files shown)


mkszepp:
@ef4 i hope everything its correct, otherwise let me know

jrjohnson:
We'd love to see this merged anything I can do to help?

NullVoxPopuli (approved):

bwbuchanan:
Does this need to get merged into `main` as well? I've been experimenting with running my app under vite and have seen these deprecation messages.

mansona:
we periodically merge stable back into main so it will get included in that 👍

PR #2038, opened 2024-07-10
Revert "Add just the necessary files to rollup watch mode"

patricklx:
Reverts embroider-build/embroider#2007

this was not correct, now its not detecting new added files. This means rollup will not include any new files in the entrypoint.
I think the right solution would have been to either setup watch exlcude or watch include
https://rollupjs.org/configuration-options/#watch-exclude

and was giving me a headache why #1855 was suddenly failing after merging stable...

### DIFF
--- a/packages/addon-dev/package.json
+++ b/packages/addon-dev/package.json
@@ -25,6 +25,14 @@
   "scripts": {
     "test": "jest"
   },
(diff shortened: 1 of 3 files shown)


patricklx:
@vstefanovic97 can you check if this would also fix your issue?

PR #2039, opened 2024-07-10
Fix(static-app-test deps)/ `@babel/helper-modules-imports` not found

BlueCutOfficial:
I encountered an issue with the `static-app-test` scenario:
- If I run the tests with `pnpm run test --filter canary-static-app`, the tests are passing ✅ 
- But if I `pnpm test:output --scenario canary-static-app` then `pnpm build` the output, I have a build error ❌ 

```
Build Error (broccoli-persistent-filter:Babel > [Babel: ember-bootstrap]) in ember-bootstrap/ember-bootstrap/components/bs-accordion/item/title.js

[BABEL]: Cannot find module '@babel/helper-module-imports'
Require stack:
- /Users/marine/Desktop/Repos/opensource/embroider-build/embroider/tests/scenarios/output/node_modules/ember-bootstrap/node_modules/ember-concurrency/lib/babel-plugin-transform-ember-concurrency-async-tasks.js
```
It seems that ember-concurrency has a reference to `require('@babel/helper-module-imports')`, but it's a dev dependency and when I build the app locally the package is not found until it's explicitly part of the static-app-test dependencies.

PR #2043, opened 2024-07-12
Fix/ static-app-test shoud actively use `isClassic` macro

BlueCutOfficial:
🔗 Depends on #2039 so `canary-static-app` can be built locally. 

## Problem 

While working on a different task I tried to figure out what tests rely on a classic build and I noticed something wrong with `static-app-test`. This scenario is supposed to rely on a macro `isClassic` to assert behaviors that may or may not happen in classic builds. But it seems that at some point, we have stopped using any classic build in any test of the file, so `isClassic` is never `true`:

Fixtures:
```js
// app component
'macros-example.hbs': `
  <h1 data-macro>
    Welcome to this {{#if (macroCondition (macroGetOwnConfig "isClassic"))}}classic{{else}}embroider{{/if}} app!
  </h1>
`,

// app acceptance test
await visit('/macros-example');
if (getOwnConfig().isClassic) {
  assert.dom('[data-macro]').hasText('Welcome to this classic app!'); // NEVER EXECUTED
} else {
  assert.dom('[data-macro]').hasText('Welcome to this embroider app!');
}
``` 

## This PR 

This PR creates an app template specifically for the classic build because it's now very different from the app template we use for Vite. This way, the `true` path of `isClassic` macro is executed again. 

⚠️ I couldn't have the tests pass because `loaded-components` and `loaded-helpers` found modules they shouldn't be finding, but I don't know if that's:
- a legitimate problem we introduced and didn't notice because `isClassic` was always `false`
- something wrong with the classic template I created


ef4:
Thanks, looks good.

PR #2046, opened 2024-07-14
vite fix custom base url

### LINKED ISSUE #2189: [Vite] @embroider/virtual/vendor.js is not covered properly by Vite asset handling
The `<script src="/@embroider/virtual/vendor.js"></script>` in the _source_ index.html is not properly converted in the _output_ `dist/index.jtml`. As other assets, in a production build it should get fingerprinted and take [base](https://vite.dev/config/shared-options.html#base) into account. Not doing so makes this break production deployments.

## Expected

With a `base` of `https://cdn.example.com/`:

```
<script src="https://cdn.example.com/@embroider/virtual/vendor-Dx01Zf34.js"></script>
```

## Actual

```
<script src="/@embroider/virtual/vendor.js"></script>
```

patricklx:
this is for vite build.
Vite does only prefix base path to known public files. So, they do have to exist. This only happens on build, during dev all urls are prefixed...
Which does not work for virtual vendor.js and others that are generated later in the process 
https://github.com/vitejs/vite/blob/fe30349d350ef08bccd56404ccc3e6d6e0a2e156/packages/vite/src/node/plugins/html.ts#L456

also, any custom middleware needs to handle this, since base path can be setup during dev

fixes #2189

### DIFF
--- a/packages/compat/src/compat-app.ts
+++ b/packages/compat/src/compat-app.ts
@@ -459,7 +459,7 @@ export default class CompatApp {
       // internal implementation detail, and respecting outputPaths here is
       // unnecessary complexity. The corresponding code that adjusts the HTML
       // <link> is in updateHTML in app.ts.
-      outputPaths: { app: `/assets/${this.name}.css` },
+      outputPaths: { app: `/@embroider/virtual/app.css` },
       registry: this.legacyEmberAppInstance.registry,
       minifyCSS: this.legacyEmberAppInstance.options.minifyCSS.options,
     };
@@ -473,10 +473,10 @@ export default class CompatApp {
         version: 2,
         'public-assets': {},
       };
-      let assetPath = join(outputPath, 'assets');
+      let assetPath = join(outputPath, '@embroider', 'virtual');
       if (pathExistsSync(assetPath)) {
         for (let file of walkSync(assetPath, { directories: false })) {
-          addonMeta['public-assets']![`./assets/${file}`] = `/assets/${file}`;
+          addonMeta['public-assets']![`./@embroider/virtual/${file}`] = `/@embroider/virtual/${file}`;
         }
       }
       let meta: PackageInfo = {
--- a/packages/vite/src/assets.ts
+++ b/packages/vite/src/assets.ts
@@ -43,8 +43,9 @@ export function assets(): Plugin {
     configureServer(server) {
       return () => {
         server.middlewares.use((req, res, next) => {
-          if (req.originalUrl && req.originalUrl.length > 1) {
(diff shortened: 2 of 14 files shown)


void-mAlex:
how about making all those URLs relative instead?

patricklx:
> how about making all those URLs relative instead?

that doesn't work, because website can open at some level deeper.
e.g my-website.com/my-app/current-page
and then that doesn't work anymore

patricklx:
Maybe that vendor.js could be removed anyway. And directly import loaderjs in entry point?
Other things can also all be imported directly? It would break backwards compatibility though

void-mAlex:
> my-website.com/my-app/current-page
is exactly why it should be relative
today, the no1 reason why you can't use ./ for root_url is because of the testing framework and how that assumes things to be available
but making everything relative is the way to go

patricklx:
But then it would access ./my-page/current-page/asset/some-asset.js ?
For a src=./asset/some-asset.js

void-mAlex:
> But then it would access ./my-page/current-page/asset/some-asset.js ? For a src=./asset/some-asset.js

why would it?
it'a all about build time relativity and relative to initial html

patricklx:
Vite keeps it like that in the output...

Alex ***@***.***> schrieb am So., 14. Juli 2024, 19:52:

patricklx:
It's not a module... That's why

Patrick Pircher ***@***.***> schrieb am So., 14. Juli 2024,
20:21:

void-mAlex:
> Vite keeps it like that in the output..

yes, have it relative to begin with and you don't have an issue ( if there is somewhere I can see this falling short in today's world is the router) but that has problems regardless so you lose nothing by having always relative urls which should be the default as it enables exactly the kind of deploy you mentioned of `my-website.com/my-app/`

ef4:
vendor.js has to remain a script, not a module. That's why it continues to exist as a separate thing. All the code in it expects to run in script context.

patricklx:
i found that this is already done automatically by vite. https://vitejs.dev/guide/build.html#public-base-path
not sure why it was not working for me

void-mAlex:
I would hate for us to re introduce this root url concept in vite world
between base html tag and using relative paths we have better tools to deal wiht this without adding back what I consider broken concepts

patricklx:
> I would hate for us to re introduce this root url concept in vite world between base html tag and using relative paths we have better tools to deal wiht this without adding back what I consider broken concepts

i had older code still there. have a look now

ef4:
The implementation changes look good. For the test, I'd rather not maintain a whole separate matrix full of tests for this case. Can we perhaps add an additional `basePath` argument to `buildViteInternalsTest()` and invoke it from within the existing matrix of tests?

PR #2061, opened 2024-08-09
Adjusting reverse-exports API

ef4:
I want to use `@embroider/reverse-exports` for a new case where it's not an error to discover that a file is internal. So I would rather that it didn't throw in this case.


ef4:
This should technically be breaking in `@embroider/reverse-exports` and only a patch in `@embroider/core`, but I don't think release-plan makes that easy to say.

PR #2062, opened 2024-08-09
Keep app dir

ef4:
This is the last major step before removing app rewriting.


packages/core/src/module-resolver.ts
```diff
@@ -466,8 +466,15 @@ export class Resolver {
...
-    return logTransition('entrypoint', request, request.virtualize(resolve(pkg.root, '-embroider-entrypoint.js')));
+    let matched = resolveExports(pkg.packageJSON, '-embroider-entrypoint.js', {
+      browser: true,
+      conditions: ['default', 'imports'],
+    });
+    return logTransition(
+      'entrypoint',
+      request,
+      request.virtualize(resolve(pkg.root, matched?.[0] ?? '-embroider-entrypoint.js'))
+    );
```
ef4, line 477: This was a first attempt at fixing the way the entrypoint works, since it may need to crawl a different directory than the package root, given what's in package.json exports.

But this is not really sufficient by itself. The `AppFiles` class assumes that app and tests have already been smashed together.

packages/core/src/module-resolver.ts
```diff
@@ -1494,16 +1509,10 @@ export class Resolver {
     if (engineConfig) {
       // we're directly inside an engine, so we're potentially resolvable as a
       // global component
-
-      // this kind of mapping is not true in general for all packages, but it
-      // *is* true for all classical engines (which includes apps) since they
-      // don't support package.json `exports`. As for a future v2 engine or app:
-      // this whole method is only relevant for implementing packageRules, which
-      // should only be for classic stuff. v2 packages should do the right
-      // things from the beginning and not need packageRules about themselves.
```
ef4, line 1503: This comment was claiming we don't care about making this code work in native v2 packages, but it honestly seems easier to just make it work than guarantee that nobody will try to ever use `reverseComponentLookup` in a v2 package.

packages/core/src/module-resolver.ts
ef4: Putting the entrypoint inside a path that is mapped via package.json exports seems like a good idea, so this code is probably helpful. It ensures that we're resolving things that are actually supposed to be resolvable from outside the package. But we can't use only this path to identify which dirs to crawl when building the entrypoint, since there is more than one dir to crawl.

Instead, it probably makes sense to make the crawling code attempt to apply `resolve.exports` to `the-package/tests/example` and `the-package/example` to identify the two directories we need to crawl to build AppFiles.

PR #2070, opened 2024-08-21
Update release-deps

NullVoxPopuli:


NullVoxPopuli:
Something seems convinced that `@ember/string` is a v1 addon.
that's _sometimes_ true, but not always. 🤔 

```
A module tried to resolve "@ember/string" and didn't find it (v1 package without auto-import)
```

PR #2071, opened 2024-08-22
change blueprint to not need a synthesized test entrypoint

ef4:
We no longer need to synthesize a dedicated test entrypoint, vite's built-in features (like eager import.meta.glob) are sufficient.

 - This will need test cleanup, as there are probably tests assuming the synthesized test entrypoint exists.
 - The other app templates need similar treatment to this one.


tests/ts-app-template/tests/index.html
```diff
@@ -30,9 +30,14 @@
...
+    <script type="module">import "ember-testing";</script>
+
+    <script type="module">
+      import { start } from './test-helper';
+      import.meta.glob("./**/*.{js,gjs,gts}", { eager: true });
+      start();
+    </script>
 
     {{content-for "body-footer"}}
-    {{content-for "test-body-footer"}}
```
patricklx, line 36: Should this be removed? It's still created in https://github.com/embroider-build/embroider/blob/b059849abbc8c9051192693f5b0fa70a77c836e3/packages/compat/src/content-for-config.ts#L20

PR #2075, opened 2024-08-23
Update ember standard modules to include @ember/renderer and @ember/-internals and ember-testing

NullVoxPopuli:
Ran in to this over on `@ember/test-helpers` v4 -- https://github.com/emberjs/ember-test-helpers/pull/1486/commits/0631b497dea8585f5d097a22d90508b9acff9d92

We have an escape hatch for addons to use, but these modules should probably be defined here.

Docs:
- https://api.emberjs.com/ember/release/modules/@ember%2Frenderer
  - (tho, these are incomplete, because renderSettled comes from here)

### DIFF
--- a/packages/shared-internals/src/ember-standard-modules.ts
+++ b/packages/shared-internals/src/ember-standard-modules.ts
@@ -45,6 +45,13 @@ emberVirtualPeerDeps.add('ember-source');
 // the modules-api-polyfill. Newer APIs need to be added here.
 emberVirtualPackages.add('@ember/owner');
 
+// Added in ember-source 4.5.0-beta.1
+emberVirtualPackages.add('@ember/renderer');
+
+// Not provided by rfc176-data, but is needed for special librarys
+// that know the dangers of importing private APIs
+emberVirtualPackages.add('@ember/-internals');
+
 // these are not public API but they're included in ember-source, so for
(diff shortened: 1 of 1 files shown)


mkszepp:
After adding in ember-power-select (and also ember-power-calendar) the peerDependency `ember/test-helpers` v4 we are running into the error:
Error: Could not find module `@ember/renderer` imported from `ember-power-select/test-support`

The error is only present in old ember versions (3.28 and 4.4...) everything else looks fine...

pipeline 3.28: https://github.com/cibernox/ember-power-select/actions/runs/10590872677/job/29347390528?pr=1848#step:6:115
pipeline 4.4: https://github.com/cibernox/ember-power-select/actions/runs/10590872677/job/29347390717?pr=1848#step:6:87

- For 3.28 i have set in ember-try `2.9.4` (as it the official last version which supports 3.x) =>  i don't understand why we can get now the error after updating test-helpers...
- For 4.4 i'm using latest test helper as its official supported...

The addons (and also the app) is forcing  `^4.0.2` of test-helpers

Maybe our issue is related and fixed with this PR.

I don't like to add `@ember/renderer`... in addon as external package as we don't use it directly, but this atm maybe the only fix...

Hope we can get this fix otherwise we can't allow test-helpers v4 for the addons

PR #2078, opened 2024-08-24
vite deps: correctly handle addon templates and addon app files

patricklx:
* makes external templates ignored
* optimised templates do not get a pairing
* hbs extension search during bundle 
* externalize app files during bundling 
  * since the request comes from a dep file we cannot know the original addon file. Which is why the from file will be encoded into the specifier.
  * This is required because the resolution can change during development 
* stage2 tests now correctly uses optmized deps

### DIFF
--- a/tests/scenarios/vite-internals-test.ts
+++ b/tests/scenarios/vite-internals-test.ts
@@ -90,11 +90,17 @@ appScenarios
         lib: {
           'app-lib-one.js': `
             globalThis.appLibOneLoaded = (globalThis.appLibOneLoaded ?? 0) + 1;
-            export default function() { return 'app-lib-one'; }
+            const localObject = {
+              message: 'app-lib-one'
+            };
+            export default function() { return localObject; }
           `,
           'app-lib-two.js': `
             globalThis.appLibTwoLoaded = (globalThis.appLibTwoLoaded ?? 0) + 1;
-            export default function() { return 'app-lib-two'; }
+            const localObject = {
+              message: 'app-lib-two'
+            };
+            export default function() { return localObject; }
           `,
         },
       },


tests/scenarios/compat-stage2-test.ts
```diff
@@ -537,7 +538,8 @@ stage2Scenarios
 
       hooks.before(async () => {
         app = await scenario.prepare();
-        server = CommandWatcher.launch('vite', ['--clearScreen', 'false'], { cwd: app.dir });
+        console.log(app.dir);
```
driesdl, line 541: Should the log be removed?

patricklx: Yes, thanks

packages/vite/src/esbuild-request.ts
```diff
@@ -146,27 +146,33 @@ export class EsBuildModuleRequest implements ModuleRequest {
...
 
-    let result = await this.context.resolve(request.specifier, {
-      importer: request.fromFile,
-      resolveDir: dirname(request.fromFile),
-      kind: this.kind,
-      pluginData: {
-        embroider: {
-          enableCustomResolver: false,
-          meta: request.meta,
+    const extensions = ['', '.hbs'];
```
patricklx, line 149: @ef4 I tried to move the extension resolving here to defaultResolve, but it does not work.
I think the issue is that when we resolve files in addons, we do use a relative specifier with a from file in the rewritten-addons. What esbuild then sees is something like
```js
{
   type: 'ignored',
   result: {
    external: true,
    path: './components/hello-world'
  }
}
```

so it has relative path as the externalized ID, I think this will confuse the system between esbuild and vite

patricklx: I did the same changes in https://github.com/embroider-build/embroider/pull/1876 and there it works

patricklx: I think we could do this as an intermediate step? I would change the tests to do the checks at runtime

patricklx:
this looks like its fixed, closing

PR #1972, opened 2024-06-11
Hide base path from public URL of rollup-public-assets

simonihmig:
This is bringing back the (breaking!) behaviour of https://github.com/embroider-build/embroider/pull/1697 to the `main` branch, after the change has been accidentally reverted. See https://github.com/embroider-build/embroider/pull/1697#issuecomment-2122890065

PR #2082, opened 2024-08-28
Hide base path from public URL of rollup-public-assets

simonihmig:
As discussed in yesterday's team meeting, after some previous back and forth and confusion around where breaking changes should go into (`main` vs. `stable`), this PR is applying the same (breaking!) change that landed in `main` (https://github.com/embroider-build/embroider/pull/1972) to stable, so we can get this published (as yet another new major unfortunately) without waiting for `main` to stabilize.

More context in https://github.com/embroider-build/embroider/pull/1697#issuecomment-2122890065...

### DIFF
--- a/packages/addon-dev/src/rollup-public-assets.ts
+++ b/packages/addon-dev/src/rollup-public-assets.ts
@@ -51,7 +51,7 @@ export default function publicAssets(
       });
       const publicAssets: Record<string, string> = filenames.reduce(
         (acc: Record<string, string>, v): Record<string, string> => {
-          const namespace = opts?.namespace ?? join(pkg.name, path);
+          const namespace = opts?.namespace ?? pkg.name;
 
           acc[`./${path}/${v}`] = resolve('/' + join(namespace, v));
           return acc;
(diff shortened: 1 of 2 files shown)


simonihmig:
@mansona does release-plan expect an additional/different label other than `breaking` to make this into the release/changelog? I guess we could tag it as a bug fix, since it "fixes" an unexpected behavior?

NullVoxPopuli (approved):

mansona:
@simonihmig no all you need is to define this as breaking and it will make it a major bump 👍

packages/addon-dev/src/rollup-public-assets.ts
```diff
@@ -51,7 +51,7 @@ export default function publicAssets(
       });
       const publicAssets: Record<string, string> = filenames.reduce(
         (acc: Record<string, string>, v): Record<string, string> => {
-          const namespace = opts?.namespace ?? join(pkg.name, path);
+          const namespace = opts?.namespace ?? pkg.name;
```
mansona, line 54: doesn't this mean that all the public assets will just be flattened 🤔 you can't have any structure at all any more

simonihmig: @mansone No, this is just the namespace, that is statically added to any file that is within the assets folder. Let's say the package name is `fancy-addon`, and its rollup config has `addon.publicAssets('src/assets/public')`, and it has a file in `src/assets/public/foo/bar.jpg`. Then before this change the plugin would add this to the package.json:

```
"public-assets": {
  "./src/assets/public/foo/bar.jpg": "fancy-addon/src/assets/public/foo/bar.jpg"
}
```

With this PR it would be:
```
"public-assets": {
  "./src/assets/public/foo/bar.jpg": "fancy-addon/foo/bar.jpg"
}
```

`path` in L54 would be `src/assets/public` in this example (as passed to the plugin), so that gets removed here.

Issue #2066, opened 2024-08-15
Projects on embroider don't see the "Non-colocated components" deprecation


kategengler:
A project on latest ember-source (v5.10.2) with many non-colocated component templates wasn't seeing any deprecations for those templates. After investigating, I realized `layoutFor` (the method with the deprecation in ember-source) was never called: The project is using embroider -- embroider is resolving the component's templates at build time. 

Embroider should add this deprecation. 

Relevant RFC advancement PR https://github.com/emberjs/rfcs/pull/1040

mansona:
For stable embroider the relevant place is https://github.com/embroider-build/embroider/blob/stable/packages/core/src/module-resolver.ts#L482-L526

Essentially I think we would either need to check your ember-source version and throw the deprecation if your ember-version matches the deprecation, or we could just decide that when you're using embroider this is just deprecated anyway regardless of your ember-source version 🤔 

I'm going to add this to the agenda for next week to discuss 👍

ef4:
We discussed this at tooling team weekly meeting and the tentative plan is to emit a deprecation inside the virtual pair component at runtime (https://github.com/embroider-build/embroider/blob/8e5763cbf19f400013932a4695c7c8b8ddcf3907/packages/core/src/virtual-content.ts#L67-L77) and also on Ember >= 6 do an actual build failure in the module-resolver where Chris linked above.

kategengler:
Does this need to be backported? Users can easily upgrade ember-source without upgrading embroider.

ef4:
Fixed in #2088.

PR #2088, opened 2024-08-30
Implement ember's component-template-resolving deprecation

ef4:
When using embroider with staticComponents enabled, users will never see Ember's [component-template-resolving](https://deprecations.emberjs.com/id/component-template-resolving/) deprecation, since Embroider takes care of implementing the whole feature at build time before Ember can see it at runtime.

This implements the same deprecation logic in Embroider apps so everybody can stay consistent.


ef4:
Fixes https://github.com/embroider-build/embroider/issues/2066

ef4:
This needs some test updates

PR #2089, opened 2024-08-30
add required fixes to keep-app-dir PR

mansona:
This PR is targeting https://github.com/embroider-build/embroider/pull/2062 and is intended to add the required changes to get tests passing again 👍 

In this PR I have: 

- merged main (which includes https://github.com/embroider-build/embroider/pull/2071 ) 
- updated a bunch of paths in tests (since they now contain `app/` in them)
- fixed the `isInComponents` check because it needs the `app/` folder in it now too
- reflected the same entrypoint changes to the route-entrypoint (for route splitting)
- add exports to the apps/tests that need it (since it's required now)
- fix import of app.js in typescript index.html (the actual file is a TS and we have an extension in the import)
- update a package rules path in a test (might need discussion - see comment)


packages/shared-internals/src/colocation.ts
```diff
@@ -40,9 +41,18 @@ function correspondingJSExists(id: string): boolean {
...
 
-function isInComponents(id: string, packageCache: Pick<PackageCache, 'ownerOfFile'>) {
+export function isInComponents(id: string, packageCache: Pick<PackageCache, 'ownerOfFile'>) {
   const pkg = packageCache.ownerOfFile(id);
-  return pkg?.isV2App() && id.slice(pkg?.root.length).split(sep).join('/').startsWith('/components');
+  if (!pkg?.isV2App()) {
+    return false;
+  }
+
+  let tryResolve = resolveExports(pkg.packageJSON, './components', {
```
ef4, line 50: We need to identify where `my-app/components/` has been mapped to via the package.json exports, instead of just assuming it will be `./app/components`.

mansona: nice one, great point! thanks for adding this 👍

tests/scenarios/static-app-test.ts
```diff
@@ -349,7 +349,7 @@ wideAppScenarios
               {
                 package: 'app-template',
                 appModules: {
-                  'components/fancy-box.js': {
+                  'app/components/fancy-box.js': {
```
mansona, line 352: I'm not sure if this is the right change 🤔 one side of me thinks it should be the full path (like this change) but on the other hand it would mean that people upgrading will have to update their packageRules.

It seems like a reasonable breaking change but I could be convinced either way 🤷

ef4: Yeah I think we should address this. The spelling of the packageRules shouldn't be changing.

PR #2091, opened 2024-08-31
Fix the template-tag codemod

NullVoxPopuli:
On main
- it wasn't running
- it errors when it does run

### DIFF
--- a/packages/compat/src/template-tag-codemod.ts
+++ b/packages/compat/src/template-tag-codemod.ts
@@ -68,12 +68,33 @@ class TemplateTagCodemodPlugin extends Plugin {
     const emberSourceEntrypoint = require.resolve('ember-source', { paths: [process.cwd()] });
     const emberVersion = JSON.parse(readFileSync(join(emberSourceEntrypoint, '../../package.json')).toString()).version;
 
-    const ember_template_compiler = await resolver.nodeResolve(
+    /**
+     * In embroider 3 we rewrite the app and place the template-compiler in this directory.
+     * However, in newer embroider-using projects, we leave ember-source in node_modules,
+     * so we need to check both places.
+     */
+
+    /**
+     * For embroider 3
+     */
+    const ember_template_compiler3 = await resolver.nodeResolve(
       'ember-source/vendor/ember/ember-template-compiler',
       resolve(locateEmbroiderWorkingDir(process.cwd()), 'rewritten-app', 'package.json')
     );
+
+    /**
+     * For embroider > 3
+     */
+    const ember_template_compiler4 = await resolver.nodeResolve(
+      'ember-source/dist/ember-template-compiler',
+      process.cwd()
+    );
+
+    const ember_template_compiler =
+      ember_template_compiler3.type === 'not_found' ? ember_template_compiler4 : ember_template_compiler3;
+
     if (ember_template_compiler.type === 'not_found') {
-      throw 'This will not ever be true';
+      throw `Could not find the ember template complier.`;
     }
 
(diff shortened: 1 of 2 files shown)


tests/scenarios/template-tag-codemod-test.ts
```diff
@@ -1,17 +1,18 @@
 import { readFileSync } from 'fs-extra';
-import { appScenarios } from './scenarios';
+import { wideAppScenarios } from './scenarios';
 import QUnit from 'qunit';
 import { join } from 'path';
 
 const { module: Qmodule, test } = QUnit;
 
-appScenarios
```
NullVoxPopuli, line 8: appScenarios no longer includes `release`

tests/scenarios/template-tag-codemod-test.ts
```diff
@@ -1,17 +1,18 @@
...
+wideAppScenarios
   .only('release')
   .map('template-tag-codemod', project => {
     project.mergeFiles({
       app: {
         components: {
           'face.hbs': `<h1> this is a gjs file</h1>`,
+          'daft.hbs': `<button {{on 'click' @onClick}}>click me</button>`,
```
NullVoxPopuli, line 15: I added a file that we'd expect to generate an import

packages/compat/src/template-tag-codemod.ts
```diff
@@ -68,12 +68,33 @@ class TemplateTagCodemodPlugin extends Plugin {
...
+    const ember_template_compiler4 = await resolver.nodeResolve(
+      'ember-source/dist/ember-template-compiler',
+      process.cwd()
+    );
+
+    const ember_template_compiler =
+      ember_template_compiler3.type === 'not_found' ? ember_template_compiler4 : ember_template_compiler3;
+
     if (ember_template_compiler.type === 'not_found') {
-      throw 'This will not ever be true';
```
NullVoxPopuli, line 76: This message was too confident, so I changed it

void-mAlex: was correct for the previous version :)

NullVoxPopuli: Maybe if someone always remembered to have ember-source installed.

An error message should at least say what went wrong :p

I've seen many errors logged that the original author thought could never happen.

_There are no absolute truths_.

(including this statement?)

:wink:

void-mAlex: I remember putting that in soley to satisfy TS, there is no case where that will be false and you'd use the template tag codemod.
because a build happens before, you'd never even reach this step if it was the case, but TS is bad to know that. so the statement stands to being always true in the 3.x context

NullVoxPopuli: > Soley to satisfy TS

TS knows what it's talking about -- because I ran in to this error, and it ended up being a helpful debugging tip. Type errors aren't amount knowing the "current state of your runtime", they help you with migrations, updates, etc -- provide more meaningful information, and give you a chance to provide more context in the failure cases (such as what happened here when I tried to run the codemod in main).

void-mAlex: there's an argument to be made that the code mod code should never really make it to main, my whole point was to get that to help people on stable and part of the upgrade to main would be run the code mod if you haven't already. so intent stays true to that should never be true

mansona:
Closing this since https://github.com/embroider-build/embroider/pull/2101 opted to not include the codemod in main (since it was only ever written to work in main)

PR #2093, opened 2024-09-02
Disabling app rewriting

ef4:


ef4:
At this point, vite is building the true app source, but the rewritten copy is still being emitted.

mansona (approved):
😍

PR #2097, opened 2024-09-07
Add test for library export coditions

NullVoxPopuli:
I'm trying to figure out why we're shipping the Dev version of the VM in production app builds, and this PR is exploring if we're losing export-condition compatibility somewhere.

See also:
- https://github.com/glimmerjs/glimmer-vm/issues/1605
- https://github.com/glimmerjs/glimmer-vm/issues/1590
- https://github.com/emberjs/ember.js/issues/20628

Of note, where I've been testing this is old embroider/webpack, so it's possible that, by using vite, all our problems will go away.


NullVoxPopuli:
ember-source now covers this situation

PR #2100, opened 2024-09-08
Esbuild fixes

ef4:
- expands our vite-internals so that they do full end-to-end app testing under *both* `vite dev` and `vite build`.
- prevents app code from getting sucked into vite deps (similar to the problem being solved in https://github.com/embroider-build/embroider/pull/2078, thank you @patricklx for the experimentation in this area).
- allows non-owned modules (meaning vite/deps) to resolve from the app's namespace, so that externalized references to the app can work
- allows non-owned modules (meaning vite/deps) to use the handleRenaming rules, so that virtual pair components that get synthesized "within" .vite/deps can resolve fake dependencies like `@ember/component`.
- changes the naming strategy for virtual pair components so there is no notional directory. This fixes the problem of esbuild not wanting to resolve from a notional directory.

Remamining steps:
- [x] implement the new stubbed tests to cover the rest of the backlink cases
- [x] rationalize all the resolvableExtensions to use the same config

### DIFF
--- a/packages/core/src/module-resolver.ts
+++ b/packages/core/src/module-resolver.ts
@@ -971,20 +971,13 @@ export class Resolver {
     }
 
     let pkg = this.packageCache.ownerOfFile(request.fromFile);
-    if (!pkg || !pkg.isV2Ember()) {
-      return request;
-    }
 
     // real deps take precedence over renaming rules. That is, a package like
     // ember-source might provide backburner via module renaming, but if you
     // have an explicit dependency on backburner you should still get that real
     // copy.
 
-    // if (pkg.root === this.options.engines[0].root && request.specifier === `${pkg.name}/environment/config`) {
-    //   return logTransition('legacy config location', request, request.alias(`${pkg.name}/app/environment/config`));
-    // }
-
-    if (!pkg.hasDependency(packageName)) {
+    if (!pkg?.hasDependency(packageName)) {
       for (let [candidate, replacement] of Object.entries(this.options.renameModules)) {
         if (candidate === request.specifier) {
           return logTransition(`renameModules`, request, request.alias(replacement));
@@ -1006,6 +999,10 @@ export class Resolver {
           request.alias(request.specifier.replace(packageName, this.options.renamePackages[packageName]))
         );
       }
+    }
+
+    if (!pkg || !pkg.isV2Ember()) {
+      return request;
     }
 
     if (pkg.name === packageName) {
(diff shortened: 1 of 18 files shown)


packages/core/src/module-resolver.ts
```diff
@@ -971,20 +971,13 @@ export class Resolver {
...
     // ember-source might provide backburner via module renaming, but if you
     // have an explicit dependency on backburner you should still get that real
     // copy.
 
-    // if (pkg.root === this.options.engines[0].root && request.specifier === `${pkg.name}/environment/config`) {
-    //   return logTransition('legacy config location', request, request.alias(`${pkg.name}/app/environment/config`));
-    // }
-
-    if (!pkg.hasDependency(packageName)) {
+    if (!pkg?.hasDependency(packageName)) {
```
ef4, line 980: This is where we're allowing a missing package to still match renamedModules.

packages/core/src/module-resolver.ts
```diff
@@ -1008,6 +1001,10 @@ export class Resolver {
       }
     }
 
+    if (!pkg || !pkg.isV2Ember()) {
+      return request;
```
ef4, line 1005: This used to happen before renameModules, etc, now it happens after.

packages/core/src/module-resolver.ts
```diff
@@ -1285,63 +1282,42 @@ export class Resolver {
...
-    let movedPkg = this.packageCache.maybeMoved(pkg);
-    if (movedPkg !== pkg) {
-      let originalFromFile = request.meta?.originalFromFile;
-      if (typeof originalFromFile !== 'string') {
-        throw new Error(`bug: embroider resolver's meta is not propagating`);
-      }
-      request = request.rehome(originalFromFile);
-      pkg = movedPkg;
+    if (pkg) {
+      ({ pkg, request } = this.restoreRehomedRequest(pkg, request));
```
ef4, line 1287: This is just an extraction because fallbackResolve was getting so long.

packages/core/src/module-resolver.ts
```diff
@@ -1285,63 +1282,42 @@ export class Resolver {
...
+      // to resolve from the app's namespace.
+      //
+      // This makes it possible for integrations like vite to leave references
+      // to the app in their pre-bundled dependencies, which will end up in an
+      // arbitrary cache that is not inside any particular package.
+      let description = pkg ? 'non-ember package' : 'unowned module';
+
+      let packageName = getPackageName(request.specifier);
+      if (packageName === this.options.modulePrefix) {
+        return logTransition(
```
ef4, line 1302: This is the new support for app resolving from non-owned or non-ember packages.

packages/core/src/module-resolver.ts
```diff
@@ -1285,63 +1282,42 @@ export class Resolver {
...
-        if (appJSMatch) {
-          return logTransition('fallbackResolve: relative appJsMatch', request, appJSMatch);
-        } else {
-          return logTransition('fallbackResolve: relative appJs search failure', request);
-        }
-      } else {
-        // nothing else to do for relative imports
-        return logTransition('fallbackResolve: relative failure', request);
-      }
+      return this.relativeFallbackResolve(pkg, request);
```
ef4, line 1314: This is another extraction to keep fallbackResolve shorter.

packages/core/src/virtual-content.ts
```diff
@@ -172,14 +172,14 @@ function decodeVirtualExternalCJSModule(filename: string) {
 }
 
 const pairComponentMarker = '-embroider-pair-component';
-const pairComponentPattern = /^(?<hbsModule>.*)\/(?<jsModule>[^\/]*)-embroider-pair-component$/;
+const pairComponentPattern = /^(?<hbsModule>.*)__vpc__(?<jsModule>[^\/]*)-embroider-pair-component$/;
```
ef4, line 175: This change in naming means that our virtual pair component is sitting inside a real directory, not an imaginary one. That makes esbuild resolve outbound requests from it correctly.

packages/vite/src/esbuild-request.ts
```diff
@@ -40,14 +52,24 @@ export class EsBuildModuleRequest implements ModuleRequest {
...
     readonly resolvedTo: Resolution<OnResolveResult, OnResolveResult> | undefined
-  ) {}
+  ) {
+    let plugins = (this.context.initialOptions.plugins ?? []).map(p => p.name);
+    if (plugins.includes('vite:dep-pre-bundle')) {
+      this.phase = 'bundling';
+    } else if (plugins.includes('vite:dep-scan')) {
+      this.phase = 'scanning';
+    } else {
+      throw new Error(`cannot identify what phase vite is in. Saw plugins: ${plugins.join(', ')}`);
```
ef4, line 62: This phase detection is extremely tied to vite internals. But we need to distinguish scanning from bundling because scanning already has the entire vite resolver configured whereas bundling is much more primitive and needs us to do more.

packages/vite/src/esbuild-request.ts
```diff
@@ -124,6 +151,24 @@ export class EsBuildModuleRequest implements ModuleRequest {
     ) as this;
   }
 
+  private phase: 'bundling' | 'scanning';
+
+  private *extensionSearch(specifier: string): Generator<string> {
```
ef4, line 156: Any integration with `@embroider/core` resolver must do extension search within its defaultResolve. esbuild is not configured with any during the bundling phase, so we have to do it ourselves.

PR #2101, opened 2024-09-09
Merge stable

ef4:
This does not accept template-tag-codemod, because it's not written to be compatible with main.


NullVoxPopuli (approved):

void-mAlex:
as far as I see this dropped the codemod which is perfect as it works differently on main
will pr it separately on main

PR #2103, opened 2024-09-11
Clean up old AMD externals support

ef4:
The older AMD externals system isn't really interoperable with true ES modules (and thus vite). We've kept it around this far, but the only way to actually use it is to manually register every single export that needs to be found in AMD, which is pretty clunkky.

It's also not really needed. In a typical app, zero imports (actual import statements. The ember resolver is a separate topic.) are actually getting handled by the AMD loader.

This PR drops the staticEmberSource option by forcing it to always be true, and drops the amdCompatibility option because it's no longer allowed.


ef4:
I think most of what remains is testing the `dependsOnComponents` and `dependsOnModules` rules, which have an AMD-based implementation. I don't think we can keep these rules working the same way. Both of them exist specifically to force a module or named component into the AMD loader, from an arbitrary place in the module graph.

We could possibly use the `Resolver.prototype.addModules` API just like lazy routes, but in general that API would not be accessible from an arbitrary module. 

I'm inclined to just remove them, I don't think they're heavily used. Looking through the built-in compat adapters, `dependsOnModules` is only used in ember-data < 4.11.0 (which is almost certainly not going to work in vite anyway. A new 4.x minor that sorts out the vite compatibility issues is coming soon). 

`dependsOnComponents` is used in five built-in compat adapters, but four out of the five have published v2 addons and the last one looks relatively abandoned.

mansona:
I think it's totally reasonable to remove those two features considering the size of the overall change that we're making with the vite support and the direction that we're going in 👍

PR #2105, opened 2024-09-11
add basic watch file support

mansona:
I'm not saying that this PR fixes all of our needs for watching app files, but since we are targeting real files on disk now and not a rewritten app this seems like it goes a decent way towards fixing things 🤷 If we're going to encourage people to try this out this week then maybe we can merge this and work on tests for it in our next office hours?

### DIFF
--- a/packages/core/src/virtual-entrypoint.ts
+++ b/packages/core/src/virtual-entrypoint.ts
@@ -147,7 +147,7 @@ export function renderEntrypoint(
 
   return {
     src: entryTemplate(params),
-    watches: [],
+    watches: ['app/**/*'],
   };
 }
 
@@ -330,7 +330,7 @@ function shouldSplitRoute(routeName: string, splitAtRoutes: (RegExp | string)[] 
 
 export function getAppFiles(appRoot: string): Set<string> {
   const files: string[] = walkSync(appRoot, {
-    ignore: ['_babel_filter_.js', 'app.js', 'assets', 'testem.js', 'node_modules'],
+    directories: false,
   });
   return new Set(files);
 }


packages/core/src/virtual-entrypoint.ts
```diff
@@ -147,7 +147,7 @@ export function renderEntrypoint(
 
   return {
     src: entryTemplate(params),
-    watches: [],
+    watches: ['app/**/*'],
```
mansona, line 150: I've tried this locally and it seems to be all we need to get it working 🤷

packages/core/src/virtual-entrypoint.ts
```diff
@@ -330,7 +330,7 @@ function shouldSplitRoute(routeName: string, splitAtRoutes: (RegExp | string)[]
 
 export function getAppFiles(appRoot: string): Set<string> {
   const files: string[] = walkSync(appRoot, {
-    ignore: ['_babel_filter_.js', 'app.js', 'assets', 'testem.js', 'node_modules'],
+    directories: false,
```
mansona, line 333: I removed the ignore because we now target the app folder as our appRoot so we don't need these ignores, and I added directories false because we only care about files 😂

packages/core/src/virtual-entrypoint.ts
ef4: This is another place where we need to be doing things as in d850898eb5c4ea3f58192e3746fcad0ea1180647

ef4: Instead of hard-coding the string "app".

patricklx:
im trying this in my vite watch mode pr, because currently its not detecting any new files. This is also not working, i think its missing changes in vite/resolver.ts

patricklx:
i think it should resolve to the full path to `/app` dir, without `*`

patricklx:
@mansona i think this could further be improved by limiting the updates to create/delete. Currently this solution would invalidate the file on every file change. But we only need to do it for create/delete. The file updates are handled by vite.
We should probably have an option to limit the watch to certain events.

patricklx:
maybe it would make sense to make the template only component virtual, similar to virtual pair component, instead of making it a fake file which can collide with the real one on disk?  maybe it would make it easier to make HMR work in vite. and making it virtual would also remove all the metadata code in hbs plugin.

ef4:
This got superseded, virtual-entry point now watched the whole app dir.

PR #2110, opened 2024-09-16
fix issue with paired component between app and addon

patricklx:
error: `ERROR: [plugin: embroider-esbuild-resolver] cannot identify what phase vite is in. Saw plugins: embroider-esbuild-resolver
`
not sure why this setup triggers this. But there is a vite esbuild mode where vite does not add any plugins here:
https://github.com/vitejs/vite/blob/ba56cf43b5480f8519349f7d7fe60718e9af5f1a/packages/vite/src/node/optimizer/index.ts#L1065

### DIFF
--- a/packages/vite/src/esbuild-request.ts
+++ b/packages/vite/src/esbuild-request.ts
@@ -11,7 +11,7 @@ type PackageCache = PublicAPI<_PackageCache>;
 export class EsBuildModuleRequest implements ModuleRequest {
   static from(
     packageCache: PackageCache,
-    phase: 'bundling' | 'scanning',
+    phase: 'bundling' | 'other',
     context: PluginBuild,
     kind: ImportKind,
     source: string,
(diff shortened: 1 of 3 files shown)


ef4:
Thanks, I'm going to run your updated tests without the fix to try to understand this case better.

patricklx:
@ef4 i'm hitting this in another project where im testing vite as well. its because it has .hbs as an entry point and vite wants to scan it for imports

patricklx:
i removed the length check so it also allows for other user plugins

ef4:
Using the tests here I see what vite is doing. It's trying to use esbuild to parse a dep's entrypoint when the dep isn't JS, so it needs to load all your custom config, so our resolver gets pulled in.

I think we can avoid this whole situation by moving a little bit of work into stage1 instead. V2 addons are already not allowed to use non-JS files, the only reason vite is encountering some is that we were intentionally lax with auto-upgraded v1 addons.

patricklx:
But that would probably also fix #2119 ?

PR #2118, opened 2024-09-19
was missing amdCompatibility

patricklx:


mansona:
We're in the process of removing all this 🤔 https://github.com/embroider-build/embroider/pull/2103

patricklx:
I guess I could just use my own import alias in vite

patricklx:
Well, it doesn't work by default, yes. But it can be made to work. Could also add a compat adapter maybe

patricklx:
mmm, i saw @ef4 had an alternative implementation idea for ember-fetch with embroider macros. maybe it's time to look at that

ef4:
ember-fetch could be simplified to keep working with its same public api without relying on AMD. All it needs is our renamed-modules feature in its ember-addon metadata.

ef4:
But yes. We’re about to drop all AMD support. There will be no AMD loader. I guess people could re-add it if they want to, but nothing in our core is going to use it.

PR #2120, opened 2024-09-20
fix gts in v1 addons

### LINKED ISSUE #2119: v1 addon with gts + ember-template-imports failing under Embroider
When a v1 addon has gts components and ember-template-imports, this is working under Classic, but failing under Embroider (stable). Imports of components used in `<template>` get stripped away (in the rewritten package), thereby leading to `Attempted to invoke a component that was not in scope in a strict mode template` errors.

Found some prior discussion around that on [Discord](https://discord.com/channels/480462759797063690/568935504288940056/1237704777211842600) between @Windvis and @NullVoxPopuli, where it was suggested to make the v1 babel typescript config same as for v2 addons. Which @Windvis did [here](https://github.com/appuniversum/ember-appuniversum/pull/492/files#r1599636132), and I can confirm in my own case that adding `onlyRemoveTypeImports` fixes the issue.

However, we should get this to work out of the box, right? It was suggested to add this config to ember-cli-babel by default, but not sure if that is viable? Couldn't this break stuff when imports are only used for types, but not declared explicitly with e.g. `import type`?

Also puzzling to me is the fact that this only causes errors under Embroider? (which is the reason I opened this issue here, even when the fix might go somewhere else)

patricklx:
fixes #2119

### DIFF
--- a/packages/compat/src/v1-addon.ts
+++ b/packages/compat/src/v1-addon.ts
@@ -116,9 +116,10 @@ export default class V1Addon {
   @Memoize()
   private get templateCompilerBabelPlugin(): PluginItem | undefined {
     let plugins = loadAstPlugins(this.addonInstance.registry);
+    let hasTemplateTag = this.addonInstance.addons.find((a: any) => a.name === 'ember-template-imports');
     // our macros don't run here in stage1
     plugins = plugins.filter((p: any) => !isEmbroiderMacrosPlugin(p));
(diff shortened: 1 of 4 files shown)


ef4:
This could really use a test.

ef4:
Also, if you can update the comments in the code to align with my explanation above it will help us in the future.

simonihmig (approved):
LGTM, thanks for picking this up!

PR #2122, opened 2024-09-20
speedup windows ci on stable

patricklx:


packages/shared-internals/src/paths.ts
```diff
@@ -12,7 +12,7 @@ export function explicitRelative(fromDir: string, toFile: string) {
   if (!isAbsolute(result) && !result.startsWith('.')) {
     result = './' + result;
   }
-  if (isAbsolute(toFile) && result.endsWith(toFile)) {
+  if (isAbsolute(toFile) && result.split(sep).join('/').endsWith(toFile)) {
```
ef4, line 15: I think if this is necessary the real bug is elsewhere. Callers are supposed to be passing `toFile` in system format. If they're passing it in unix format on windows, that is the bug.

patricklx: i copied this from main. https://github.com/embroider-build/embroider/pull/2048

ef4: I see, ok.

PR #2126, opened 2024-09-24
re-enable watch tests

mansona:


patricklx:
Maybe using something like virtual pair component for template only would help to reduce code in the hbs plugin and then it would probably be easier to support hmr.

ef4:
@mansona I fixed the particular tests that we were looking at. There are other tests that need cleanup now. For example "Scenario 4" is assuming that it will see a fresh app, but the command watcher is shared between tests, so it's initial condition fails because hello-world.js already exists from another test.

packages/vite/src/hbs.ts
```diff
@@ -59,17 +68,34 @@ export function hbs(): Plugin {
...
+        } else {
+          let correspondingHBS = syntheticJStoHBS(resolution.id);
+          if (correspondingHBS) {
+            depTracker?.trackDeps(resolution.id, [correspondingHBS]);
+          }
+        }
       }
+
+      // we should be able to clear any earlier meta by returning
+      // resolution.meta here, but vite breaks that rollup feature.
```
patricklx, line 92: Just use own cache then?
MetaCache = new Map();
And set the value for the id in the resolveId

ef4: We could but this is working now. I checked and the rollup docs say it's OK to mutate the meta object that comes back from getModuleInfo().meta.

patricklx:
@ef4 I looked at the tests and I think its missing the case `deleting a co-located js`

PR #2128, opened 2024-09-25
Re-add local/dev volta for folks not using corepack

NullVoxPopuli:
_unless_ folks know of a way to only enable corepack for certain projects?

it's been problematic on my work repo, so I have it globally disabled.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,5 +68,9 @@
       "unlabeled": ":question: Unlabeled"
     },
     "wildcardLabel": "unlabeled"
+  },
+  "volta": {
+    "node": "20.14.0",
+    "pnpm": "8.15.9"
   }
 }


package.json
```diff
@@ -68,5 +68,9 @@
       "unlabeled": ":question: Unlabeled"
     },
     "wildcardLabel": "unlabeled"
+  },
+  "volta": {
+    "node": "20.14.0",
+    "pnpm": "8.15.9"
```
NullVoxPopuli, line 74: matches `packageManager`

package.json
```diff
@@ -68,5 +68,9 @@
       "unlabeled": ":question: Unlabeled"
     },
     "wildcardLabel": "unlabeled"
+  },
+  "volta": {
+    "node": "20.14.0",
```
NullVoxPopuli, line 73: matches `.npmrc`

package.json
NullVoxPopuli: tbh, volta should probably just read packageManager and .npmrc, and only _optionally_ have the volta package.json thing

NullVoxPopuli:
Closing, because _the solution here_ is to find "the one tool" to rule them all.

No clue what that is yet, but we'll see...

PR #2131, opened 2024-10-01
Rename @embroider/core/entrypoint to @embroider/virtual/compat-modules

### LINKED ISSUE #2025: Rename `@embroider/core/entrypoint` to `@embroider/virtual/compat-modules`
## Context

`@embroider/core/entrypoint` used to be literally the entry point to the Vite-built Ember app. See an example [in an old commit](https://github.com/embroider-build/embroider/blob/e61adc42483c8aff25e4f375f55402d41d68bcf8/tests/app-template/app/index.html#L20): the script `@embroider/core/entrypoint` was present in the `index.html`.

But with the latest way to boot apps, `@embroider/core/entrypoint` is not longer the "entry point", it's [imported in `app/app.js`](https://github.com/embroider-build/embroider/pull/1957/files#diff-84a17ec3977e75d36e2d35954ffbcf8fc51de3934c94a82c7dcaa0411d19ff98). We need a better name for the module, that represents what it really does: importing the virtual modules that maintain compatibility with classic addons. The new name should be `@embroider/virtual/compat-modules`.

NullVoxPopuli:
Resolves #2025


packages/virtual/src/compat-modules.d.ts
```diff
@@ -0,0 +1,7 @@
+interface UnknownModule {
+  [exportName: string]: unknown;
+}
+
+declare const compatModules: Record<string, UnknownModule>;
+
+export default compatModules;
```
NullVoxPopuli, line 7: types!

packages/util/tests/dummy/app/index.html
```diff
@@ -17,7 +17,7 @@
     {{content-for "body"}}
 
     <script src="/@embroider/core/vendor.js"></script>
-    <script src="/@embroider/core/entrypoint" type="module"></script>
+    <script src="/@embroider/virtual/compat-modules" type="module"></script>
```
mansona, line 20: this will literally do nothing 🤔 are these tests even run?

test-packages/sample-transforms/tests/dummy/app/index.html
```diff
@@ -18,7 +18,7 @@
     {{content-for "body"}}
 
     <script src="/@embroider/core/vendor.js"></script>
-    <script src="/@embroider/core/entrypoint" type="module"></script>
+    <script src="/@embroider/virtual/compat-modules" type="module"></script>
```
mansona, line 21: again this will do nothing, we should check if we can just delete it

test-packages/sample-transforms/tests/index.html
```diff
@@ -31,7 +31,7 @@
     <script src="/testem.js" integrity=""></script>
     <script src="/@embroider/core/vendor.js"></script>
     <script src="/@embroider/core/test-support.js"></script>
-    <script src="/@embroider/core/entrypoint" type="module"></script>
+    <script src="/@embroider/virtual/compat-modules" type="module"></script>
```
mansona, line 34: does nothing 🫠

mansona (commented):
I'm inclined to approve this since it does what it says on the tin and solves a serious type issue that we had. I wonder though if there could be more discussion about the approach and see if we can't do something a little smarter rather than having another package to deploy 🙈 (see my question in a comment)

packages/util/tests/dummy/app/index.html
NullVoxPopuli: Better to be updated than to lie?

🤷🙈🙃

tests/app-template/package.json
```diff
@@ -41,6 +41,7 @@
     "@embroider/core": "workspace:*",
     "@embroider/router": "workspace:*",
     "@embroider/test-setup": "workspace:*",
+    "@embroider/virtual": "workspace:*",
```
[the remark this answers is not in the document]
NullVoxPopuli: I think build needs it to, not just types. The errors i was running in to yesterday were because I didn't have a ./* Entry in the exports.

That's probably a separate issue tbh. Like.. why's a real package.json get involved with virtual files?

> How does ember-source solve this similar problem? 🤔

ember-source does a bunch of declare module, and requires a specific config. That could take the form of an import 'ember-source/types', or what we've moved to in the blueprints, where you specify ember-source/types in tsconfig#compilerOptions#types

We could do the same trick, but resolving the exports lookup thing would need to happen first.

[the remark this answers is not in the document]
NullVoxPopuli: That said, idk if possible here, i'd like us to not have so many diffs between js and usages.

Some linting would also complain about importing from a package that is not declared.

ef4:
I think we should do the same thing vite does for extra types (`import.meta.glob`, for example, needs a type declaration).

They tell you to put this in your tsconfig:

```
"compilerOptions": {
    "types": ["vite/client"]
  }
```

For us that could be:

```
"compilerOptions": {
    "types": ["vite/client", "@embroider/core"]
  }
```

And I think then we just need some `declare module` in `core/src/index.ts`.

ef4:
Link to their docs: https://vitejs.dev/guide/features.html#client-types

ef4:
Possibly it would be safer for us to pick a subpath like `@embroider/core/client`, since it's not clear that it would be good for frontend code to need to see the types for all of our build-time things.

mansona:
@ef4 what about @embroider/core/virtual so that it makes a reference to the fact that we're typing the `@embroider/virtual` things?

ef4:
I'm not sure if everything we'll ever want to put in there is under `@embroider/virtual`.  We might want to put in `import.meta.resolve` for example.

NullVoxPopuli:
As I was poking around in here , I noticed that test-support and vendor are also virtual files -- should those be under the same prefix fake-package? (maybe in a separate PR, once we work out the mechanics here?)

mansona:
everything that is virtual should be renamed as such, and yes they should be separate PRs 👍

NullVoxPopuli:
> I think we should do the same thing vite does 

this is also the same thing we do :P :tada: 
(for ember-source types)

tests/ts-app-template/types/ts-app-template/index.d.ts
```diff
@@ -1,3 +1,4 @@
+/// <reference types="@embroider/core/virtual" />
```
NullVoxPopuli, line 1: this is the only way I could get the types present when type-checking.

The (imo) nicer way, via `compilerOptions.types` doesn't seem to work for @-prefixed package names.

tests/ts-app-template-classic/types/ts-app-template/index.d.ts
```diff
@@ -1,3 +1,4 @@
+/// <reference types="@embroider/core/virtual" />
```
mansona, line 1: I don't think `@embroider/virtual` is used in the ts-app-template-classic 🤔 do we need this line?

packages/util/tests/dummy/app/index.html
mansona: I mean if you delete the line no test will fail, if you delete the file no test will probably fail 🤷 

there is no side-effect of importing `/@embroider/virtual/compat-modules` any more so just adding a script will do nothing to any app ever

NullVoxPopuli: should we delete the whole dummy folder?

I can't leave @embroider/core/entrypoint because it doesn't exist anymore

tests/ts-app-template-classic/types/ts-app-template/index.d.ts
NullVoxPopuli: right you are! removed

mansona (approved):

PR #2132, opened 2024-10-01
Better error for module-resolver, maybe?

NullVoxPopuli:
Saw a todo when I was poking around in https://github.com/embroider-build/embroider/pull/2131

### DIFF
--- a/packages/core/src/module-resolver.ts
+++ b/packages/core/src/module-resolver.ts
@@ -415,11 +415,11 @@ export class Resolver {
       return request;
     }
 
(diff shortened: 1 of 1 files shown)


patricklx:
i think it would help if all errors include the specifier & fromfile.
recently had the error `an addon tried to import from "@embroider/virtual", but it,s not one of its explicit dependencies`
not hepful.
it was power-select. which has peer deps, when those are not installed, that error appears

NullVoxPopuli:
Gonna close this as its way out of date now, and the errors we have with vite are pretty good

PR #2135, opened 2024-10-02
Capture new optimized deps

ef4:
Builds on the tests and findings in #1876, with a retry-based implementation.

The strategy here is that a newly added dep is a pretty rare occurrence, so it's better to let most resolutions run normally and only check them after the fact to see if they *could* have been new un-optimized dependencies, then do a second resolve for them that's adjusted so to conform to vite's rules about what is allowed to be optimized, as explained by @patricklx in #1876. 

We really only need to hide any `node_modules` in the `fromFile`. The rest (like always having a non-relative specifier) is already handled in core module-resolver.

### DIFF
--- a/packages/vite/src/resolver.ts
+++ b/packages/vite/src/resolver.ts
@@ -42,7 +42,13 @@ export function resolver(): Plugin {
         return await observeDepScan(this, source, importer, options);
       }
 
-      let request = RollupModuleRequest.from(this, source, importer, options.custom);
+      let request = RollupModuleRequest.from(
+        resolverLoader.resolver.packageCache,
+        this,
+        source,
+        importer,
+        options.custom
+      );
       if (!request) {
         // fallthrough to other rollup plugins
         return null;
--- a/packages/vite/src/resolver.ts
+++ b/packages/vite/src/resolver.ts
@@ -1,19 +1,25 @@
 import type { Plugin, ViteDevServer } from 'vite';
-import core from '@embroider/core';
-const { virtualContent, ResolverLoader } = core;
+import core, { type Resolver } from '@embroider/core';
+const { virtualContent, ResolverLoader, explicitRelative, cleanUrl, tmpdir } = core;
 import { RollupModuleRequest, virtualPrefix } from './request.js';
 import { assertNever } from 'assert-never';
 import makeDebug from 'debug';
-import { resolve } from 'path';
+import { resolve, join } from 'path';
 import { writeStatus } from './esbuild-request.js';
-import type { PluginContext } from 'rollup';
+import type { PluginContext, ResolveIdResult } from 'rollup';
+import { externalName } from '@embroider/reverse-exports';
+import fs from 'fs-extra';
+import { createHash } from 'crypto';
+
+const { ensureSymlinkSync, outputJSONSync } = fs;
 
 const debug = makeDebug('embroider:vite');
 
 export function resolver(): Plugin {
-  let resolverLoader = new ResolverLoader(process.cwd());
+  const resolverLoader = new ResolverLoader(process.cwd());
   let server: ViteDevServer;
-  let virtualDeps: Map<string, string[]> = new Map();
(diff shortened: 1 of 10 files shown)


packages/vite/src/resolver.ts
```diff
@@ -108,3 +119,75 @@ async function observeDepScan(context: PluginContext, source: string, importer:
...
+  }
+
+  if (notViteDeps.has(foundFile)) {
+    debug('maybeCaptureNewOptimizedDep: already attmpted %s', foundFile);
+    return result;
+  }
+
+  debug('maybeCaptureNewOptimizedDep: doing re-resolve for %s ', foundFile);
+
+  let jumpRoot = join(tmpdir, 'embroider-vite-jump', hashed(pkg.root));
```
patricklx, line 170: Ah, nice trick 
There was an advantage in the previous implementation that it was doing it only for rewritten packages,  so auto-upgraded.

It might be that now this will also optimise other deps.

If a dep is excluded in vite, the default behaviour is to also exclude the dependenciea of it.
If this really does optimise deps of excluded deps i think it would be nice and definitely a candidate for another vite plugin :)

ef4: > There was an advantage in the previous implementation that it was doing it only for rewritten packages, so auto-upgraded.

Yeah, I think that probably left a bug for native v2 addons, since they also need this behavior but won't get rewritten. Consider your app depends on `addon-one` which depends on `addon-two` and both are native v2 addons so neither is rewritten.

During dev if you add a new usage of a component from `addon-two`, module resolving will look like:

1. defaultResolve runs for `your-app/components/the-new-one` and doesn't find it.
2. fallbackResolve finds it in the mergeMap and retries with `request.alias('addon-two/_app_/components/the-new-one.js').rehome('/path/to/your-app/node_modules/addon-one/package.json')`. The rehome here is chosen to be in a package that can definitely resolve `addon-two`. 
3. defaultResolve finds the file, and the specifier is a bare import, but the fromFile includes `node_modules`, so you get an un-optimized dependency.

> If a dep is excluded in vite, the default behaviour is to also exclude the dependencies of it. Because it's in node_modules.

That's a good point, and another reason why vite's isInNodeModules check is silly. It should really be crawling the package graph to decide where the boundary between not-optimized and optimized packages is.

Yeah, a plugin could fix that problem for everyone, although IMO it would be better to fix it inside vite. I think their package info caches are equivalently powerful to the packageCache in embroider, so could solve the problem without a lot of extra infrastructure.

ef4:
@patricklx I don't understand yet why the new tests are timing out on windows. It looks like we're failing to shutdown the server between tests, since the next run logs that it had to try multiple ports before finding a free one.

ef4:
Running locally on windows I think the problem is that there's a module cycle in the tests, and that cycle apparently evaluates in a different order on windows.

The cycle is that the service in `my-services-addon` imports `app/app.js`, but `app/app.js` imports the core entrypoint which imports the service.

I think if we change the test to use a different example file in app rather than app.js it will address the cycle.

ef4:
The cycle problem was real and I could reliably reproduce it and it's fixed.

I've seen the windows tests passing locally. I'm also seeing inconsistent behaviors. It looks like sometimes the prebuild hasn't really finished successfully, because you get failures during dep optimization that can only be explained by missing prebuild output.

On CI it's so slow I've never seen it finish. So I'm skipping them on windows.

patricklx:
> The cycle problem was real and I could reliably reproduce it and it's fixed.
> 
> I've seen the windows tests passing locally. I'm also seeing inconsistent behaviors. It looks like sometimes the prebuild hasn't really finished successfully, because you get failures during dep optimization that can only be explained by missing prebuild output.
> 
> On CI it's so slow I've never seen it finish. So I'm skipping them on windows.

Yeah, i saw it pass sometimes, took like 16min...
Great to see this fixed finally 👍🚀🚀

ef4:
@patricklx thanks for your work on this

PR #2136, opened 2024-10-02
Fix addon-dev gjs plugin order

simonihmig:


tests/scenarios/v2-addon-dev-watch-test.ts
```diff
@@ -51,14 +51,19 @@ Scenarios.fromProject(() => baseV2Addon())
...
             addon.appReexports(['components/**/*.js']),
-            addon.gjs(),
-            addon.hbs(),
+
             addon.dependencies(),
-            addon.publicAssets('custom-public'),
 
             babel({ babelHelpers: 'bundled' }),
 
+            addon.hbs(),
```
patricklx, line 61: I think babel must be the last. whats the issue with the order?

patricklx: ah, I see, thats how the blueprint is.
But the recent PR chanhed hbs & gjs plugins to use transform hook. So the order of the blueprint must be changed.
the tests in embroider were already at correct order before the change... I wonder why it wasnt the same order in the blueprint

patricklx: looks like the PR i did was more of a breaking change then?

patricklx: well, the blueprint should really have the babel plugin as last.

simonihmig: @patricklx ready to go now, see the PR description! 

> well, the blueprint should really have the babel plugin as last.

We can still decide to make that change in the blueprint if we feel this is the safer path forward. But the change in this PR makes this optional, which I think is better than forcing this on users by running into build errors...

simonihmig:
Weird, `beta` tests (Ember 6) are failing here with `Components with separately resolved templates were removed at Ember 6.0. Migrate to either co-located js/ts + hbs files or to gjs/gts.`, while they are passing for `canary`, and last [scheduled job](https://github.com/embroider-build/embroider/actions/runs/11139302582) is also passing. But the changes here cannot have caused this (these are app tests), right? 🤔

ef4:
Canary is not really the current canary. https://github.com/embroider-build/embroider/pull/2073

simonihmig:
> while they are passing for canary, and last [scheduled job](https://github.com/embroider-build/embroider/actions/runs/11139302582) is also passing

nvm, the passing scheduled job was for `main`. There was a previous [failing build](https://github.com/embroider-build/embroider/actions/runs/11128890120) on `stable` before, so this is an issue we have on `stable`, and not related to this PR!

ef4 (approved):

PR #2138, opened 2024-10-02
Delete unneeded util files

NullVoxPopuli:
Would slightly reduce the diff (and changelog) for: https://github.com/embroider-build/embroider/pull/2131


packages/util/tests/dummy/app/index.html
NullVoxPopuli, line 1: this file is needed even though we don't explicitly use it.
We run tests for util, but those use tests/index.html

packages/util/tests/dummy/app/index.html
```diff
@@ -1,24 +1 @@
...
-  <body>
-    {{content-for "body"}}
-
-    <script src="/@embroider/core/vendor.js"></script>
-    <script src="/@embroider/core/entrypoint" type="module"></script>
-
-    {{content-for "body-footer"}}
-  </body>
-</html>
+<!-- not used, but required by ember-cli -->
```
NullVoxPopuli, line 1: this is another app/styles/app.css :see_no_evil:

mansona (approved):
:shipit:

PR #2143, opened 2024-10-08
Vite hbs plugin should return resolutions

ef4:
The lack of the return is probably doubling the amount of resoultion work in downstream plugins.

Also the extra fallback here should not be needed anymore given that the default resolve is supposed to check for hbs.


mansona (approved):
looks good to me 👍

PR #2148, opened 2024-10-09
Fix broken default export when required from CJS

simonihmig:
When this package was converted to TS, `module.exports = ` was converted to `export default`, without taking into account that this would make TS emit this as a named export of "default" (`exports.default = sideWatch`), so CJS consumers would have to call it as `require('@embroider/broccoli-side-watch').default()` (aka "double default" problem). As we are only targeting CJS consumers, this changes makes the compiled js have a single main export as `module.exports = sideWatch`.

Tests weren't able to catch this, as TS was providing the interoperability, while node.js does not.

### DIFF
--- a/packages/broccoli-side-watch/src/index.ts
+++ b/packages/broccoli-side-watch/src/index.ts
@@ -27,7 +27,7 @@ interface SideWatchOptions {
   dependencies that you're actively developing. For example, right now
   @embroider/webpack doesn't rebuild itself when non-ember libraries change.
 */
-export default function sideWatch(actualTree: InputNode, opts: SideWatchOptions = {}) {
+function sideWatch(actualTree: InputNode, opts: SideWatchOptions = {}) {
   const cwd = opts.cwd ?? process.cwd();
 
   if (!opts.watching || !Array.isArray(opts.watching)) {
@@ -65,3 +65,6 @@ export default function sideWatch(actualTree: InputNode, opts: SideWatchOptions 
       }),
   ]);
 }
+
+// We expose this as CJS, so make sure this transpiles to module.exports = sideWatch
+export = sideWatch;


packages/broccoli-side-watch/src/index.ts
```diff
@@ -27,7 +27,7 @@ interface SideWatchOptions {
   dependencies that you're actively developing. For example, right now
   @embroider/webpack doesn't rebuild itself when non-ember libraries change.
 */
-export default function sideWatch(actualTree: InputNode, opts: SideWatchOptions = {}) {
+function sideWatch(actualTree: InputNode, opts: SideWatchOptions = {}) {
```
void-mAlex, line 30: I believe it's better to author as proper modules and expose a compiled dist instead of source than to ship cjs compatible code by default

simonihmig: Not sure what exactly you mean here? How could I write code or set up TS so that `export default` compiles to something that would work in CJS as `const sideWatch = require('@embroider/broccoli-side-watch')`?

FWIW, we have a similar case [here](https://github.com/embroider-build/embroider/blob/main/packages/macros/src/ember-addon-main.ts#L8).

The alternative would be to not using a default export, but require users to `const { sideWatch } = require('@embroider/broccoli-side-watch')`, which wouldn't suffer from that interoperability problem.

void-mAlex: personally I don't see why `import { sideWatch } ...` is frowned upon
I would prefer that to writting the similar case you mentioned before

while this will not really be needed in vite land and so on
it sets precedence which could be followed when adding other packages that may be used from module only places

and I'm not sure if we're not going to have issues with macros in the future because of the choice you mentioned there in systems that use modules exclusively

are we in a new enough node environment where we could say use top level import? would that work?

as for how to make the current syntax work, I'm talking ab out shipping two separate bundles one for cjs and one for modules with entry points specific to each declared in package json
all of that is a much more convoluted setup and would require more of a pre build step than we're doing now

PR #2156, opened 2024-10-21
Support other plugins in keepAssets

ef4:
This extends the keepAssets feature in `@embroider/addon-dev` so that it composes nicely with other plugins.

For example, if you use a plugin that synthesizes CSS imports, it's nice for keepAssets to consume those and keep them as a real CSS files on disk when you build your addon.


SergeAstapov:
@ef4 sorry if being silly here, is this smth we could document in readme.md of addon-dev package?

ef4:
Yes, good idea. The default way that this plugin is used in the v2-addon-blueprint doesn't change its public API at all, but I did also add a new optional argument. And we should probably explain this whole thing better.

SergeAstapov (approved):

PR #2157, opened 2024-10-23
error instead of warn for missing exports

amk221:
See https://github.com/webpack/webpack/discussions/14659


amk221:
Anyone?

amk221:
Yeh that works, switch from `strictExportPresence` to `exportsPresence`

NullVoxPopuli:
How should this be framed in the changelog?
- generally throwing errors where we weren't previously is a breaking change
- it could be argued that the errors were always there, but now moved from runtime to build time

amk221:
I don't really know. There's some info [here](https://github.com/webpack/webpack/discussions/14659), like that it's already true in mjs.

I can see why it would be a breaking change, but, then again if it breaks for somebody then something was already broken they just didn't know it. If they were in runtime, and you weren't actively monitoring in (Sentry for example) you'd never know.

Yeh anyway, I don't necessarily need this to go in, just as a developer it feels like no brainer. Can I leave it to you all to decide whether you want it or not?

Also, what does Vite do?

NullVoxPopuli:
> Also, what does Vite do?

Vite doesn't let you import things that don't exist 🎉

PR #2159, opened 2024-10-31
pinning @types/qunit to fix ci

ef4:
@types/qunit 2.19.11 is incompatible with ember-qunit right now.


ef4:
Oh, that's good.

But also I see our release scenarios are explicitly using ember-qunit 7 instead of 8. We'll need to sort out why and whether we're ready to only test 8.

SergeAstapov:
good point! we may want to release patch for ember-qunit v7 then

PR #2162, opened 2024-11-02
Fix sourcemaps for GJS

NullVoxPopuli:
Using `@embroider/addon-dev@6.0.1`, I noticed this message:
```
(!) Broken sourcemap
https://rollupjs.org/troubleshooting/#warning-sourcemap-is-likely-to-be-incorrect
Plugins that transform code (such as "rollup-gjs-plugin") should generate accompanying sourcemaps.
```

This PR fixes that

This PR enables viewing actual gts and gjs in the browser, as original source

image

image
image

and most importantly, breakpoints line up:
image


packages/addon-dev/src/rollup-gjs-plugin.ts
```diff
@@ -19,12 +22,17 @@ export default function rollupGjsPlugin(
...
       let input = readFileSync(id, 'utf8');
-      let code = processor.process(input, {
+      let codeWithInlineMap = processor.process(input, {
         filename: id,
         inline_source_map,
       });
+
+      let map = convert.fromSource(codeWithInlineMap).toJSON();
+      let code = codeWithInlineMap.replace(/\/\/# sourceMappingURL=\/.+$/, '');
+
```
patricklx, line 32: Best to start with `^` too in the regex.
Still would not protect against this if this text is inside a a template literal.

NullVoxPopuli: true -- it _is_ supposed to be the last line, so I could do a `.lastIndexOf` sort of thing

NullVoxPopuli:
Superseded by https://github.com/embroider-build/embroider/pull/2166

PR #2164, opened 2024-11-05
Support v2 ember-source

ef4:
These minor compatibility tweaks are needed to support an upcoming ember-source release that declares itself as a v2 addon.


bartocc:
Upgrading from 3.6.2 to 3.7.0 breaks the build of our app because of the following error in an .hbs template

```hbs

PR #2166, opened 2024-11-08
Fix gjs/gts sourcemaps -- we accidentally unlocked really good DX

NullVoxPopuli:
Alternative to:
- https://github.com/embroider-build/embroider/pull/2162
  (and see here for context)

Requires
- https://github.com/embroider-build/content-tag/pull/86
  I will need to upgrade content-tag in this (#2166) PR once released

Something that is quite surprising (in a good way), is that we can inspect live values _in templates_ (which makes sense, and really shows the power of the content-tag strategy we're using.

For example:
image
image
image

### DIFF
--- a/packages/addon-dev/src/rollup.ts
+++ b/packages/addon-dev/src/rollup.ts
@@ -59,8 +59,8 @@ export class Addon {
     return hbs(options);
   }
 
-  gjs(options?: { inline_source_map: boolean }) {
-    return gjs(options);
+  gjs() {
+    return gjs();
   }
 
   // this does incremental updates to the dist files and also deletes files that are not part of the generated bundle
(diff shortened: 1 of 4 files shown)


packages/addon-dev/src/rollup.ts
```diff
@@ -59,8 +59,8 @@ export class Addon {
     return hbs(options);
   }
 
-  gjs(options?: { inline_source_map: boolean }) {
```
NullVoxPopuli, line 62: removing this makes this PR a breaking change... _but_ inline source map is likely never something anyone wants as it hasn't been compatible with rollup.

So... this PR "could" be a bugfix, I suppose.

SergeAstapov: @NullVoxPopuli IMO it;s change in public API - so breaking regardless if ppl prefer/use inline source maps

SergeAstapov (approved):

packages/addon-dev/src/rollup.ts
mansona: @SergeAstapov agreed, and integers are cheap 👍

mansona (approved):
:shipit:

Issue #2168, opened 2024-11-12
`Outdated Optimize Dep` error when consuming 'normal' (direct) npm packages with vite – deps not ending up in `node_modules/.vite/deps`


johanrd:
## Steps to reproduce
1) Install the npm library `sanitize-html`, or clone this repositiory for reproduction: https://github.com/johanrd/ember-vite-npm

<details>
<summary>Detailed steps</summary>

1.1) create a new ember vite app with @NullVoxPopuli's great little ember-vite-blueprint `npx ember-cli@latest new glimmer-apollo-vite --blueprint @nullvoxpopuli/ember-vite-app --pnpm --typescript`

1.2) pnpm install any npm library, e.g. 'pnpm install sanitize-html`

1.3) import the pnpm library into wherever, e.g. application.gts

1.4) run `pnpm start`
</details>

2) See that the console errors with `Failed to load resource: the server responded with a status of 504 (Outdated Optimize Dep)`

<img width="1020" alt="Screenshot 2024-11-12 at 09 07 09" src="https://github.com/user-attachments/assets/bccb6602-bdba-4bd7-b07d-6ba0eeca1255">

Npm packages most often do not end up in `node_modules/.vite/deps`. Possibly related to https://github.com/embroider-build/embroider/issues/1583. 

Is there a config I am missing? I have tried to add `include` in optimizeDeps(), but that stopped the vite server completely.

patricklx:
can you try running the command `vite optimize`? it is failing when I test it on your repo. But maybe that is related to my node_modules.

patricklx:
right, thats what i am seeing as well `let terminalHighlight = require('./terminal-highlight');`
but why is it importing postcss into the app itself?

patricklx:
oh, its used by sanitize-html.
just tried it on a repo without embroider and there the optimization of sanitize-html  works

mansona:
I've had to fix this with sanatize-html before too. Unfortunately this isn't an embroider problem, I imagine you would have the exact same issue if you used it in a vanilla vite package too

I've fixed this in the past by patching postcss to add the extensions that it expects to be able to resolve that file: https://github.com/ember-learn/ember-api-docs/pull/920/files#diff-218116475a824ffaa3d495332d9bcf7dcd105c41f131ed71626cf1e7a66ac7d8

patricklx:
@mansona it works in a vanilla vite app.
[...]

PR #2170, opened 2024-11-12
Add the 'browser' condition for conditions when resolving

NullVoxPopuli:
A better alternative to: 
- https://github.com/embroider-build/content-tag/pull/84
  (which I couldn't really get working for all the use cases (thanks attw & publint!))


packages/core/src/module-resolver.ts
```diff
@@ -1016,7 +1016,7 @@ export class Resolver {
       if (pkg.packageJSON.exports) {
         let found = resolveExports(pkg.packageJSON, request.specifier, {
           browser: true,
-          conditions: ['default', 'imports'],
+          conditions: ['default', 'imports', 'browser'],
```
NullVoxPopuli, line 1019: todo: check if vite exports a list of conditions we can use (so we don't need to maintain this)

NullVoxPopuli: TODO: order probably matters, do `import, module, browser, default`

ef4:
This is not related to #2168. That bug is about the top-level `browser` field in package.json. This bug is about `browser` *inside* package.json `exports`.

tests/scenarios/v2-addon-test.ts
NullVoxPopuli, line 1: a failing test!

My fix was insufficient!

NullVoxPopuli:
potentially blocked on https://github.com/vitejs/vite/pull/17837 (unless I use a different dependency for testing -- which I probably should, as the problems I'm having here are vite-related, not embroider related)

PR #2171, opened 2024-11-13
fix ci on main

patricklx:


ef4:
Merged via https://github.com/embroider-build/embroider/pull/2173

PR #2177, opened 2024-11-19
Update @embroider/router to use new ember-resolver instead of AMD

ef4:
Extracted from https://github.com/embroider-build/embroider/pull/2103

This is breaking in @embroider/resolver because
 - minimum supported ember-resolver is 13.1.0
 - minimum supported ember-source is now 4.10


ef4:
This landed as part of 2103 after all.

PR #2183, opened 2024-11-22
Simplify Vite Plugin

evoactivity:
This will allow the vite config to just be

```js
import { defineConfig } from 'vite';
import ember from '@embroider/vite';

export default defineConfig({
  plugins: [ember()],
});
```


packages/vite/src/plugin.ts
```diff
@@ -0,0 +1,57 @@
...
+    },
+    hbs(),
+    templateTag(),
+    scripts(),
+    resolver(),
+    compatPrebuild(),
+    assets(),
+    contentFor(),
+
+    babel({
```
NullVoxPopuli, line 48: I liked your earlier suggestion of having babel present in the userland vite config.

folks may want to change this and it isn't "our" plugin

packages/vite/src/plugin.ts
```diff
@@ -0,0 +1,57 @@
...
+  return [
+    {
+      name: 'vite-plugin-ember',
+      enforce: 'pre',
+      config(_config: UserConfig, env: ConfigEnv) {
+        return {
+          resolve: {
+            extensions,
+          },
+          optimizeDeps: optimizeDeps(),
```
NullVoxPopuli, line 24: if I specify my own optimizeDeps in the userland config, do these two objects get merged together? or should `_config` be spread into `optimizeDeps` here, for every object / array emitted by `optimizeDeps()`

packages/vite/src/plugin.ts
```diff
@@ -0,0 +1,57 @@
...
+export const extensions = ['.mjs', '.gjs', '.js', '.mts', '.gts', '.ts', '.hbs', '.json'];
+
+export default function ember() {
+  return [
+    {
+      name: 'vite-plugin-ember',
+      enforce: 'pre',
+      config(_config: UserConfig, env: ConfigEnv) {
+        return {
+          resolve: {
```
NullVoxPopuli, line 21: what happens if the userland config adds `resolve: { alias: {} }`?

packages/vite/src/plugin.ts
```diff
@@ -0,0 +1,57 @@
...
+            extensions,
+          },
+          optimizeDeps: optimizeDeps(),
+          server: {
+            port: 4200,
+          },
+          build: {
```
NullVoxPopuli, line 28: I think build, and the choice to configure `tests` should be left to the user.

right now there some unease around `/tests/` and the escape hatch we have in the existing vite config is to encourage changing `/tests/` to `/tests__whatever_i_want/`

NullVoxPopuli (commented):
excited for shrinking the "default" config, but i left some comments/questions/concerns

NullVoxPopuli (changes requested):
ope

packages/vite/src/plugin.ts
evoactivity: babel would be configured with the babel config file, not in the vite config. 

See react and vue vite plugins, they also setup babel but it's ultimately configured by the user with a babel config file.
https://github.com/vitejs/vite-plugin-react/blob/main/packages/plugin-react/src/index.ts
https://github.com/vitejs/vite-plugin-vue/blob/599f81369816aca918314e86f5b0a6e32367bb38/packages/plugin-vue-jsx/src/index.ts#L4

packages/vite/src/plugin.ts
NullVoxPopuli: for this one, specifically, I think it's _better_ to expose always, for discoverability

packages/vite/src/plugin.ts
NullVoxPopuli: legit

packages/vite/src/plugin.ts
evoactivity: what if you could just do and we have this in the userland vite config? 

```js
ember({
  testUrl: '/tests/`
})
```

NullVoxPopuli: why invent and maintain a new API tho?

NullVoxPopuli: (i'll defer to others on this if they disagree with me, ofc)

evoactivity: Actually maybe I need to understand this unease and escape hatch.

How would the user configure this, they would also need to rename the tests folder wouldn't they?

evoactivity:
Anyone have an idea why the CI bails out saying the lockfile needs to be updated? (I've ran pnpm and lockfile is not changing between runs)

packages/vite/src/plugin.ts
evoactivity: I'll leave this one open to see if anyone else has an opinion. I would vote for keeping this internal though, as users should only need to worry about `babel.config.js`, not how it actually is run on the files.

NullVoxPopuli:
does your tooling change your pnpm version based on the `packageManager` field? odds are you're using newer pnpm, and this repo is on pnpm 8

evoactivity:
I'm using volta with the pnpm flag enabled, so I think it should be, but it's installing 9 instead of 8, so I guess it's not.

NullVoxPopuli:
yeah, we don't have a volta config in the package.jsons

evoactivity:
Ah ok, this would suggest otherwise
https://github.com/embroider-build/embroider/blob/main/CONTRIBUTING.md

evoactivity:
Downgraded to pnpm 8.15.8 using corepack and now getting a strange typescript issue when running pnpm install, showing in CI as well.

```
prepare$ tsc -b && pnpm build-v2-addons
│ packages/macros/tests/babel/helpers.ts(85,15): error TS2454: Variable 'mode' is used before being assigned.
```

but it's not used before being assigned

https://github.com/embroider-build/embroider/blob/0aa8fc682d56a99d0305534ca7a00d165ad27019/packages/macros/tests/babel/helpers.ts#L82-L85

evoactivity:
Looks like typescript updated. Should that be pinned instead since stuff breaks often with typescript minor updates?

I don't understand why `let mode` is broken and `const mode` is fine, but 🤷

PR #2185, opened 2024-11-25
this-property-fallback is not a possibility on ember >=4

ef4:


packages/compat/src/resolver-transform.ts
```diff
@@ -986,7 +991,13 @@ export default function makeResolverTransform({ appRoot, emberVersion, externalN
...
       };
     }
-    return new TemplateResolver(env, config, builtInKeywords(emberVersion), externalNameHint);
+    return new TemplateResolver(
+      env,
+      config,
+      builtInKeywords(emberVersion),
+      supportsThisFallback(emberVersion),
+      externalNameHint
+    );
```
balinterdi, line 1000: Can the TemplateResolver constructor also take in 5 positional parameters?

mansona: that was added in this same PR: https://github.com/embroider-build/embroider/pull/2185/files#diff-a5cbba44c120a6bcfc99a5a7fe32bc73be243236913f8060889d56a02e4a79ffR191

mansona: I understand that the comment was from a long time ago but I thought I would answer it for archival sake 😂

mansona (approved):
looks good to me 👍

packages/compat/src/resolver-transform.ts
balinterdi: Sure, thank you!

PR #2187, opened 2024-11-26
bundle up vite plugins in a nicer way

mansona:
This comes from the suggestion of @evoactivity on the v2 app blueprint RFC

We discussed the structure at the Embroider office hours and meeting and agreed that this is pretty close to what we're hoping to ship with

Closes https://github.com/embroider-build/embroider/pull/2183

### DIFF
--- a/packages/vite/src/ember.ts
+++ b/packages/vite/src/ember.ts
@@ -1,0 +1,50 @@
+import { templateTag } from './template-tag.js';
+import { resolver } from './resolver.js';
+import { mergeConfig, type UserConfig, type ConfigEnv } from 'vite';
+import { esBuildResolver } from './esbuild-resolver.js';
+
+export let extensions = ['.mjs', '.gjs', '.js', '.mts', '.gts', '.ts', '.hbs', '.hbs.js', '.json'];
+
+export function ember() {
+  return [
+    templateTag(),
+    resolver(),
+    {
+      name: 'vite-plugin-ember-config',
+      async config(config: UserConfig, env: ConfigEnv) {
+        return mergeConfig(
+          {
+            resolve: {
+              extensions,
+            },
+
+            optimizeDeps: {
+              exclude: ['@embroider/macros'],
+              extensions: ['.hbs', '.gjs', '.gts'],
+              esbuildOptions: {
+                plugins: [esBuildResolver()],
+              },
+            },
+
+            build: {
+              rollupOptions: {
+                input: {
+                  main: 'index.html',
+                  ...(shouldBuildTests(env.mode) ? { tests: 'tests/index.html' } : undefined),
+                },
+              },
+            },
+            server: {
+              port: 4200,
+            },
+          },
+          config
+        );
+      },
+    },
+  ];
+}
+
+function shouldBuildTests(mode: string) {
+  return mode !== 'production' || process.env.FORCE_BUILD_TESTS;
+}
--- a/tests/app-template/vite.config.mjs
+++ b/tests/app-template/vite.config.mjs
@@ -1,70 +1,15 @@
 import { defineConfig } from "vite";
-import {
-  resolver,
-  hbs,
-  scripts,
-  templateTag,
-  optimizeDeps,
-  compatPrebuild,
-  assets,
-  contentFor,
-} from "@embroider/vite";
-import { resolve } from "path";
+import { extensions, classicEmberSupport, ember } from "@embroider/vite";
 import { babel } from "@rollup/plugin-babel";
 
-export default defineConfig(({ mode }) => {
-  return {
-    cacheDir: resolve("node_modules", ".vite"),
-    resolve: {
-      extensions: [
-        ".mjs",
-        ".gjs",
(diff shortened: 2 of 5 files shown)


packages/vite/src/ember.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export let extensions = ['.mjs', '.gjs', '.js', '.mts', '.gts', '.ts', '.hbs', '.hbs.js', '.json'];
+
+export function ember() {
+  return [
+    templateTag(),
+    resolver(),
+    {
+      name: 'vite-plugin-ember-config',
+      async config(config: UserConfig, env: { mode: string; command: string }) {
```
evoactivity, line 14: a type of ConfigEnv exists in vite, can use that for the env argument.

```ts
import { type ConfigEnv } from 'vite';
```

packages/vite/src/ember.ts
```diff
@@ -0,0 +1,52 @@
...
+        if (config?.resolve?.extensions) {
+          if (extensions.some(ext => !config?.resolve?.extensions?.includes(ext))) {
+            throw new Error(
+              `You are customising your resolve.extensions but you are missign some that are required for Ember. Make sure you include at least all of the following ${extensions}`
+            );
+          }
+        }
+
+        return mergeConfig(config, {
```
evoactivity, line 23: this should be the other way around, otherwise our config will override the user config.

packages/vite/src/ember.ts
```diff
@@ -0,0 +1,52 @@
...
+  return [
+    templateTag(),
+    resolver(),
+    {
+      name: 'vite-plugin-ember-config',
+      async config(config: UserConfig, env: { mode: string; command: string }) {
+        if (config?.resolve?.extensions) {
+          if (extensions.some(ext => !config?.resolve?.extensions?.includes(ext))) {
+            throw new Error(
+              `You are customising your resolve.extensions but you are missign some that are required for Ember. Make sure you include at least all of the following ${extensions}`
```
evoactivity, line 18: Could explain here they can import the extensions from `@embroider/vite` to spread into their array?

tests/app-template/vite.config.mjs
```diff
@@ -1,70 +1,21 @@
...
-      templateTag(),
-      scripts(),
-      resolver(),
-      compatPrebuild(),
-      assets(),
-      contentFor(),
+      classicEmberSupport(),
+      ember(),
 
       babel({
```
evoactivity, line 11: Not disagreeing with this choice, but I am curious what the reasoning is behind keeping it a user land plugin?

evoactivity:
Generally looks good, I left a few comment and a question :)

packages/vite/src/ember.ts
mansona: nice one thanks!

tests/app-template/vite.config.mjs
```diff
@@ -1,70 +1,21 @@
...
-  compatPrebuild,
-  assets,
-  contentFor,
-} from "@embroider/vite";
-import { resolve } from "path";
+import { extensions, classicEmberSupport, ember } from "@embroider/vite";
 import { babel } from "@rollup/plugin-babel";
 
-export default defineConfig(({ mode }) => {
+export default defineConfig(() => {
```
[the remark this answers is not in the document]
mansona: good catch! I fixed it

packages/vite/src/ember.ts
mansona: 🤦 I read the docs like 20 times and convinced myself I did it the right way round... but you are of course correct, a simple test confirmed it 🫠 I've fixed it now

packages/vite/src/ember.ts
mansona: so my thinking was that if someone was customising their extensions it might seem strange/magic to just add a bunch to it 🤔 but I convinced myself that it was probably ok when I was thinking about your comment, especially because things will just stop working if you don't 🤷 

I'll leave this comment unresolved in case someone else has an opinion but I have swapped it to just add to the extensions if they are configured 👍

tests/app-template/vite.config.mjs
mansona: so this one felt a bridge too far for me. Sure we have ember stuff in the babel config but ember doesn't need to be in charge of the babel plugin (I think)

I don't have a very strong feeling on it so if we think it should just go in the ember plugin I'm happy, but I didn't want to hide everything 😂

packages/vite/src/ember.ts
mansona: we simplified this a bit more again so that people would be able to **exclude** extensions if they wanted to 👍

tests/app-template/vite.config.mjs
mansona: we discussed this again today. We've decided it is best to leave it in the config so that you can customise the babel plugin, it's not really possible to customise config going to other plugins from our plugin 🙈

ef4 (approved):

evoactivity (approved):

PR #2190, opened 2024-12-02
Fix: tests, resolve importSync module resolution issue

MehulKChaudhari:
This PR addresses and resolves the failing tests related to `importSync` and the V2 addon module type. The changes ensure compatibility with ES module imports and correct handling of `.js` extensions.

Changes made: 
- Updated test `import-sync.test.ts` file to include `.js` extensions.
- Added correct import paths in `v2-addon-type-module-test.ts` file for ES module compatibility.

### DIFF
--- a/packages/macros/tests/babel/import-sync.test.ts
+++ b/packages/macros/tests/babel/import-sync.test.ts
@@ -11,7 +11,7 @@ describe('importSync', function () {
       import { importSync } from '@embroider/macros';
       importSync('foo');
       `);
-      expect(code).toMatch(/import esc from "\.\.\/\.\.\/src\/addon\/es-compat2"/);
+      expect(code).toMatch(/import esc from "\.\.\/\.\.\/src\/addon\/es-compat2\.js"/);
       expect(code).toMatch(/esc\(require\(['"]foo['"]\)\)/);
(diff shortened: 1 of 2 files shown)


tests/scenarios/v2-addon-type-module-test.ts
```diff
@@ -32,8 +33,7 @@ appScenarios
          */
         'demo.js': `
           import { importSync } from '@embroider/macros';
-
-          importSync('./side-effecting.js');
```
NullVoxPopuli, line 36: this should remain relative, ya?

packages/macros/tests/babel/import-sync.test.ts
```diff
@@ -31,7 +31,7 @@ describe('importSync', function () {
       import { importSync as i } from '@embroider/macros';
       i('foo');
       `);
-      expect(code).toMatch(/require\(['"]foo['"]\)/);
```
NullVoxPopuli, line 34: why this extra `\)`?

MehulKChaudhari: its a mistake, it shouldn't here. 

we want `require("foo")` not `require("foo"))`

PR #2191, opened 2024-12-02
fix tests url

### LINKED ISSUE #2150: localhost:4200/tests (without trailing slash) not working with Vite
Just a minor DX issue, but I think worth tackling: when you open `localhost:4200/tests` Vite will serve the main `index.html`, and the Ember app will error with an `UnrecognizedURLEerror`. However, with a trailing slash, it will serve `tests/index.html` as expected.

I wonder if we should provide a vite plugin with a server middleware to do a redirect for this very common case, since this works ootb with Ember CLI!?

patricklx:
fixes #2150

### DIFF
--- a/packages/vite/src/resolver.ts
+++ b/packages/vite/src/resolver.ts
@@ -41,6 +41,20 @@ export function resolver(): Plugin {
           }
         }
       });
+      return () => {
+        server.middlewares.use((req, _res, next) => {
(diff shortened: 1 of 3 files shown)


ef4:
This behavior is good, it should probably go in a different place though. The assets plugin is specifically for supporting v1 addon assets, and it won't be included if people drop classic ember support.

PR #2192, opened 2024-12-03
Run compatPrebuild in a separate output directory

simonihmig:
I was running into an issue where `ember build` being invoked as part of the compatPrebuild plugin and vite itself would both run on the same `dist` folder in conflicting ways. I believe `ember build` is cleaning up its output directory (`dist` by default) before running the build. Which already is causing some conflicts, as I think Vite should own that directory, and it may or may not do that cleanup itself depending on its `emptyOutDir` config. 

I my case, the conflict was even larger as we were building to different targets inside of `dist`, like `dist/test` and `dist/production` (custom `outDir` vite config). These outputs were meant to persist, but when running the build in test mode first and then it production, the production build would make ember-cli clear `dist/**`, so after the build only `dist/production` is left.

### DIFF
--- a/packages/vite/src/build.ts
+++ b/packages/vite/src/build.ts
@@ -13,15 +13,23 @@ export function emberBuild(command: string, mode: string, resolvableExtensions: 
 
   if (command === 'build') {
     return new Promise((resolve, reject) => {
-      const child = fork('./node_modules/ember-cli/bin/ember', ['build', '--environment', mode], { env });
+      const child = fork(
+        './node_modules/ember-cli/bin/ember',
+        ['build', '--environment', mode, '-o', 'tmp/compat-prebuild', '--suppress-sizes'],
+        { env }
+      );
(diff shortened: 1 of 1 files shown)


packages/vite/src/build.ts
```diff
@@ -13,15 +13,23 @@ export function emberBuild(command: string, mode: string, resolvableExtensions:
 
   if (command === 'build') {
     return new Promise((resolve, reject) => {
-      const child = fork('./node_modules/ember-cli/bin/ember', ['build', '--environment', mode], { env });
+      const child = fork(
+        './node_modules/ember-cli/bin/ember',
+        ['build', '--environment', mode, '-o', 'tmp/compat-prebuild', '--suppress-sizes'],
```
simonihmig, line 18: `--suppress-sizes` is unrelated, but fixes getting the following red error message on production builds:
```
No asset files found in the path provided: dist/
```

simonihmig:
Failing tests are also on `main`!

PR #2195, opened 2024-12-05
read your browser targets from config/targets.js automatically

mansona:
We discussed this during the weekly meeting as an alternative to https://github.com/embroider-build/app-blueprint/pull/122

This implementation works, but we do have a problem. The set of supported targets has now changed. I tested out `ie 11` (just for laughs 😭) and I saw a lot of the following errors in the output: 

```
Transforming class syntax to the configured target environment ("chrome131", "firefox132", "ie11", "safari18.1" + 2 overrides) is not supported yet
```

I'm not exactly sure this is a blocker but it's probably something that we should mention in the RFC 🤔 thoughts? One way to work around it I think would be to setup preset-env yourself in babel which makes me think this isn't a dealbreaker

### DIFF
--- a/packages/vite/package.json
+++ b/packages/vite/package.json
@@ -21,6 +21,7 @@
     "@embroider/reverse-exports": "workspace:*",
     "@rollup/pluginutils": "^5.1.0",
     "assert-never": "^1.2.1",
+    "browserslist": "*",
     "browserslist-to-esbuild": "^2.1.1",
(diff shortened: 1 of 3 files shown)


ef4:
I think it would be preferable to put this in the legacy vite plugin instead of the default ember one.

NullVoxPopuli (approved):
:shipit:

PR #2198, opened 2024-12-06
Virtual response namespace unification

ef4:
At present, it's not possible to generate a virtual response whose path is in the same namespace as the real modules. Integrations like vite are allowed to keep their virtual responses in a different namespace.

But I think we can rely on other forms of metadata propagation in all supported build systems, so that whether a response is virtual doesn't need to depend solely on its path.

This would make it possible to emit virtual modules that live in the same namespace as all the real files, which would also make things like component template colocation possible to implement entirely within the core module resolver.


mansona:
it looks like the `addon-dev-watch windows` flakey test might be fixed by https://github.com/embroider-build/embroider/pull/2208 🤔 it's worth reviewing (I haven't had a chance to check the code yet)

ef4:
All green on the core change here. Now we can clean up the parameter passing to the virtual content renderers.

mansona:
I assume we could solve that in a follow up PR now that this is green?

PR #2200, opened 2024-12-09
Add rollup declarations plugin

simonihmig:
This invokes glint as a rollup plugin, and contains a workaround for broken gts imports, see https://github.com/typed-ember/glint/issues/628.

This is an attempt to fix this long standing issue, which Glint v2 seems to fix. But we want this to work now, so v2 addons have a blessed way to support gts files. It's influenced by previous attempts (https://github.com/embroider-build/addon-blueprint/pull/255, https://github.com/embroider-build/addon-blueprint/pull/292), but takes a slightly different approach:
* packages the plugin in `@embroider/addon-dev` so people don't have inlined plugin code in their rollup config
* https://github.com/NullVoxPopuli/fix-bad-declaration-output uses `jscodeshift` to rewrite broken imports (strip the `.gts` extension), but I was hesitant to bring in that large dependency for that simple task. Rewriting import paths using a simple regex seems safe enough to me, especially since this is operating only on Glint-emitted code (no custom formatting expected)
* optimizes the concurrency of running the `glint` executable

The long-term idea is that once Glint v2 is out, we can use that for the plugin, removing the workaround and ideally invoking programmatic Glint APIs instead of spawning it as a new process. But all of that without causing much churn for users, as they should be able to keep their rollup config unchanged.

Here is how this would be eventually adopted in the blueprint: https://github.com/embroider-build/addon-blueprint/pull/315

### DIFF
--- a/packages/addon-dev/jest.config.js
+++ b/packages/addon-dev/jest.config.js
@@ -1,0 +1,6 @@
+module.exports = {
+  testEnvironment: 'node',
+  testMatch: [
+    '<rootDir>/tests/**/*.test.js',
+  ],
+};
--- a/packages/addon-dev/package.json
+++ b/packages/addon-dev/package.json
@@ -37,6 +37,7 @@
     "@embroider/core": "workspace:^",
     "@rollup/pluginutils": "^4.1.1",
     "content-tag": "^3.0.0",
+    "execa": "^5.1.1",
     "fs-extra": "^10.0.0",
     "minimatch": "^3.0.4",
     "rollup-plugin-copy-assets": "^2.0.3",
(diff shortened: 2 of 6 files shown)


NullVoxPopuli:
> https://github.com/NullVoxPopuli/fix-bad-declaration-output

There is also the fix for `/// <reference` stuff, which TS has said they won't fix, because they don't support what we're doing, where we want peer-provided types (using the consumer's deps).  :cry:

simonihmig:
I haven't really looked deeply into that issue, just picked [here](https://github.com/embroider-build/addon-blueprint/pull/292#discussion_r1639308115) that this issue is not anymore. Did I get this wrong?

And btw, fully broken CI here is unrelated, `stable` itself is broken. Trying to get this fixed in #2201

NullVoxPopuli:
>  this issue is not anymore. 

it's been a while for me, but I _remember_ TS 5.5 did not fix the issue, though i did not try setting explicitly `preserve: false` -- I could also be misremembering at this point, so it's probably good to re-evaluate -- based on this emberobserver search: https://emberobserver.com/code-search?codeQuery=fix-bad-declaration-output, it's possible _I'm_ out of date! haha

ef4:
We discussed at tooling team, this seems worth landing.

simonihmig:
Green now and ready for review!

SergeAstapov (approved):

PR #2210, opened 2024-12-14
Deprecate staticHelpers, staticModifiers, and staticComponents in favour of staticInvoakables 

mansona:


packages/core/src/options.ts
```diff
@@ -124,11 +152,13 @@ export default interface Options {
...
+export type CoreOptionsType = Required<
+  Omit<Options, 'staticHelpers' | 'staticModifiers' | 'staticComponents' | 'staticInvokables'>
+> &
+  Pick<Options, 'staticHelpers' | 'staticModifiers' | 'staticComponents' | 'staticInvokables'>;
+
+export function optionsWithDefaults(options?: Options): CoreOptionsType {
   let defaults = {
-    staticHelpers: false,
-    staticModifiers: false,
-    staticComponents: false,
```
void-mAlex, line 131: shouldn't this add staticInvokables default?

mansona: so the defaults are there to make sure the types are always right. Essentially we can't set a default for staticInvokables because we would need to set it only if you haven't defined the other settings. I asked Ed about it and he just said to make everything optional and we can rationalise it in the constructor 👍

miguelcobain:
In my v2 addon, I get the following error under the `embroider-optimized` scenario:
```
You cannot set `staticHelpers`, `staticComponents`, or `staticModifiers` if you have set `staticInvokables`. Delete these configs to continue.
```
However, I'm not defining these configs anywhere in my codebase.
I believe this is related to this PR.

Here is the failing CI run: https://github.com/miguelcobain/ember-yeti-table/actions/runs/12661940772/job/35286056334

Pinning the previous `@embroider/compat` version with 
```json
"resolutions": {
  "@embroider/compat": "3.7.1"
}
```
in the root `package.json` fixed the issue. Not sure what in my build was setting `staticInvokables` or anything related. 🤔

gvdp:
Seeing the same failure in our CI run here: https://github.com/Bagaar/ember-graphql-mocking/actions/runs/13159870096/job/36725656239

Is there a good way to circumvent this? Maybe by setting these props manually in the `embroiderOptimized()` scenario?

mansona:
the best way that I've seen to fix this is to re-roll your lock file. ember-try really shouldn't be using a lock file for any of this but making sure that you have the latest `@embroider/core` in your lock file fixes this problem 👍

gvdp:
> the best way that I've seen to fix this is to re-roll your lock file. ember-try really shouldn't be using a lock file for any of this but making sure that you have the latest `@embroider/core` in your lock file fixes this problem 👍

That worked, thanks!

PR #2211, opened 2024-12-15
Require Vite 6

NullVoxPopuli:
I didn't want to see this message in the terminal anymore:
```bash
 WARN  Issues with peer dependencies found
# ...
└─┬ @embroider/vite 0.2.1-unstable.c51ca2b
  └── ✕ unmet peer vite@^5.2.0: found 6.0.3

```

This is a breaking change, but it's no biggy, really.

### DIFF
--- a/packages/vite/package.json
+++ b/packages/vite/package.json
@@ -10,7 +10,7 @@
   },
   "peerDependencies": {
     "@embroider/core": "workspace:^",
-    "vite": "^5.2.0"
+    "vite": "^6.0.3"
   },
   "scripts": {
     "test": "jest"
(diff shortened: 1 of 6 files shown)


packages/vite/package.json
```diff
@@ -10,7 +10,7 @@
   },
   "peerDependencies": {
     "@embroider/core": "workspace:^",
-    "vite": "^5.2.0"
+    "vite": "^6.0.3"
```
simonihmig, line 13: Shouldn't we allow both versions?

mansona: counter point: is there any need to drop support for 5 🤷 We should be expanding the range because we don't have any functionality that requires v6

mansona: no

ef4:
Current `main` now has test coverage for both vite 5 and 6.

PR #2215, opened 2024-12-18
fix optional chaining undefined for non macros

### LINKED ISSUE #1812: Invalid code output when using `@embroider/macros` (ember app)
I have an issue of `@embroider/macros` installation messing up the output of my Ember app code.
From what I'm guessing, it's got something to do with (incorrect) optional chaining operator implementation, and appears only in production builds.

**Reproduce steps:**
1. Create a fresh EmberJS app
2. Add the following code somewhere
```
const test = {
    items: [],
};
[1, 2].forEach((i) => {
    test.items.push(i);
});
return String(test.items?.[0]);
```

3. `ember serve --environment=production`
4. The following code outputs `"1"`. <-- this is the expected output (first array item)

5. install embroider/macros package
`npm install @embroider/macros --save-dev`
6. `ember serve --environment=production`
7. The same code outputs `"undefined"` <-- not good

Not only the result is undefined, but if you look at the assets file, it's actually compiled as such: `String(void 0)`

There is also a minimum viable reproduce here: https://github.com/nulle/testapp/pull/1 

Note: the only reason I'm adding `@embroider/macros` is for [ember-data randomUUID polyfill](https://github.com/emberjs/data?tab=readme-ov-file#randomuuid-polyfill).

### LINKED ISSUE #2212: @embroider/macros babel plugin: Optional chaining expression gets aggressively optimized to `undefined`
### Summary

While upgrading to Embroider in our main app, I noticed that the babel plugin for `@embroider/macros` transpiles code like `aKnownValue?.foo` to `undefined` prematurely.

### Steps to reproduce

I created a repo here which showcases the issue: https://github.com/ArnaudWeyts/embroider-macros-reproduction

What happens is the following:

```js
const aKnownValue = {};
aKnownValue.foo = true;
console.log(aKnownValue?.foo);
```

gets transpiled into

```js
const aKnownValue = {};
aKnownValue.foo = true;
console.log(undefined);
```

We have several places in our codebase where this optional chaining is applied, and it's currently blocking us from upgrading to Embroider since it seems like Embroider makes this babel plugin run on all the source code while it didn't before.

I tried to pinpoint why this was happening but it seems like this is currently happening by design, as I noticed there's a test that makes sure this works: https://github.com/embroider-build/embroider/blob/e55b05521885772abb056351744a607c397e17b5/packages/macros/tests/babel/eval.test.ts#L35-L38

Other pointers that I discovered:

- Exact place where the code is replaced: https://github.com/embroider-build/embroider/blob/acb8c6ff1749304ce3bde8a7723617405e2d4f85/packages/macros/src/babel/macros-babel-plugin.ts#L250
- The evaluator that returns a "confident" undefined: https://github.com/embroider-build/embroider/blob/f41229b274828f4e92ab147a1bddc8d1b603b868/packages/macros/src/babel/evaluate-json.ts#L156-L169

If anyone could help me give some pointers as what exactly needs to happen, I can look into creating a PR, but I'm not really sure where this issue should be fixed.

Thanks for working on this project!

patricklx:
fixes #2212
fixes #1812

### DIFF
--- a/packages/macros/tests/babel/get-config.test.ts
+++ b/packages/macros/tests/babel/get-config.test.ts
@@ -152,7 +152,7 @@ describe(`getConfig`, function () {
             aKnownValue.foo = true;
             result = aKnownValue?.foo;
           `);
-          expect(code).toMatch(`result = aKnownValue?.foo`);
+          expect(code).toMatch(`result = aKnownValue`);
         });
 
         runTimeTest(`runtime getConfig is still present in runtime mode when using optional chaining`, () => {
--- a/packages/macros/tests/babel/get-config.test.ts
+++ b/packages/macros/tests/babel/get-config.test.ts
@@ -146,6 +146,15 @@ describe(`getConfig`, function () {
           expect(code).toMatch(/doSomething\(undefined\)/);
         });
 
+        buildTimeTest(`does not collapse nullish coalescing for non embroider macros, nullish case`, () => {
+          let code = transform(`
+            const aKnownValue = {};
+            aKnownValue.foo = true;
+            result = aKnownValue?.foo;
+          `);
+          expect(code).toMatch(`result = aKnownValue?.foo`);
+        });
+
         runTimeTest(`runtime getConfig is still present in runtime mode when using optional chaining`, () => {
           let code = transform(`
             import { getConfig } from '@embroider/macros';
--- a/packages/macros/src/babel/macros-babel-plugin.ts
+++ b/packages/macros/src/babel/macros-babel-plugin.ts
@@ -205,9 +205,32 @@ export default function main(context: typeof Babel): unknown {
     // our getConfig and getOwnConfig macros are supposed to be able to absorb
     // optional chaining. To make that work we need to see the optional chaining
     // before preset-env compiles them away.
+
+    function fromGetConfig(path: NodePath<t.MemberExpression | t.OptionalMemberExpression> | null) {
+      while (path) {
+        const obj = path.get('object');
+        if (obj.isCallExpression()) {
+          const callee = obj.get('callee');
+          if (
+            callee.referencesImport('@embroider/macros', 'getOwnConfig') ||
+            callee.referencesImport('@embroider/macros', 'getGlobalConfig') ||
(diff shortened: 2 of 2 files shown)


packages/macros/tests/babel/get-config.test.ts
```diff
@@ -146,6 +146,15 @@ describe(`getConfig`, function () {
           expect(code).toMatch(/doSomething\(undefined\)/);
         });
 
+        buildTimeTest(`does not collapse nullish coalescing for non embroider macros, nullish case`, () => {
+          let code = transform(`
+            const aKnownValue = {};
+            aKnownValue.foo = true;
+            result = aKnownValue?.foo;
+          `);
+          expect(code).toMatch(`result = aKnownValue`);
```
NullVoxPopuli, line 155: Is it still ?.foo?

patricklx: There are 2 runs with the same test with and without preset env. So one converts it to manual null check, cannot make one version that matches both results

simonihmig (approved):
Interesting one!

It seems the root cause of this is our `Evaluator` returning a confident result for that expression, which itself delegates this to babel's own `NodePath#evaluate()` which again returns a confident result. Seems basically the same as what @ef4 reported here already: https://github.com/babel/babel/issues/14197

Even when that bug got fixed, I think the change here is good, as we only want to take the cost of that evaluation in the context of some get*Config call, which is basically what this PR is doing, right? So I wouldn't be surprised if this has also some performance benefits!

patricklx:
thats, right. only if its inside, or rather "part of" a get*Config call

ArnaudWeyts:
I just patched the dependency with this PR and it does indeed fix our issues :) Thanks a lot for looking into it so quickly!

ArnaudWeyts:
Anything that still needs to be done for this?  Looks like one check is failing due to a browser timeout but seems unlikely that this change caused that failure 🤔

patricklx:
So, need to wait on https://github.com/babel/babel/issues/14197

ef4:
In the process of explaining how to do this it turned I just did it: https://github.com/embroider-build/embroider/pull/2286

PR #2219, opened 2024-12-20
Set MiniCssExtractPlugin ignoreOrder to true for default config

lfloyd117:
From this discussion in the Ember Discord: https://discord.com/channels/480462759797063690/1308128146058514474/1319687395024830507 with @ef4 

The warnings don't particularly mean much when handling CSS via modules, so it should be safe to make the base config ignore them.

### DIFF
--- a/packages/webpack/src/ember-webpack.ts
+++ b/packages/webpack/src/ember-webpack.ts
@@ -606,6 +606,8 @@ const Webpack: PackagerConstructor<Options> = class Webpack implements Packager 
             // but in fastboot, we need to disable that in favor of doing our
             // own insertion of `<link>` tags in the HTML
             runtime: variant.runtime === 'browser',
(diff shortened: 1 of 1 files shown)


lfloyd117:
What other steps need to be done in this PR (for versioning or other tasks)? I did not see further references in the contributing doc

NullVoxPopuli (approved):
Makes sense to me

For others, here are the docs on the option:
https://webpack.js.org/plugins/mini-css-extract-plugin/#remove-order-warnings

ef4:
This should target `stable`, which means you will want to rebase your commit on stable as well so you're not pulling in all of `main`.

NullVoxPopuli:
@ef4 there is a separate PR targeting stable

NullVoxPopuli:
Closing in favor of #2220

PR #2232, opened 2025-01-21
move @embroider/vite typescript build to a project reference

mansona:
This moves all the built files into a `build/` folder and also removes the need to hand-roll the index.d.ts file 👍


packages/vite/package.json
```diff
@@ -4,8 +4,8 @@
   "type": "module",
   "exports": {
     ".": {
-      "default": "./index.mjs",
-      "types": "./index.d.ts"
+      "default": "./build/index.js",
```
void-mAlex, line 7: could we go with the more standard dist?

mansona: sure 🤷 fine either way 👍

mansona: fixed 🎉

PR #2239, opened 2025-01-29
Adding test coverage for vite 6

ef4:


tests/scenarios/vite-internals-test.ts
```diff
@@ -321,7 +321,8 @@ function runViteInternalsTest(scenario: Scenario) {
     });
 
     Qmodule('vite optimize', function () {
-      test('vite optimize should succeed', async function (assert) {
+      // skipped due to https://github.com/vitejs/vite/issues/19316
+      QUnit.skip('vite optimize should succeed', async function (assert) {
```
ef4, line 325: This is the only unpleasant surprise so far.  https://github.com/vitejs/vite/issues/19316

mansona: This has now been fixed upstream and will be released in vite@6.1.0 soon, I have updated our CI to target the corresponding beta

mansona: correction: the fix didn't get all the way to where we needed to 😞 Keeping an eye on upstream

patricklx:
I see that this PR is actually testing the beta now. not necessary 6, https://pkg.pr.new/vite@main points to beta.
might rename vite-6 to vite-beta. I think its good to also have beta in the matrix.
I see that it was not necessary to update lock file? so scenario tester will get the latest?

mansona:
The reason this is currently tracking main is because our CI is hitting a bug in Vite@6 https://github.com/vitejs/vite/issues/19316 

The first attempt of fixing this was released in beta-1 (that's why we were originally targeting it to see if it would make CI pass) but the real/full fix hasn't been released yet in any beta or release version: https://github.com/vitejs/vite/pull/19356

Personally I don't think we should merge this PR until it's at least in a beta and we can target that as the minimum version in CI (and update it to the release version once it's out)

tests/scenarios/package.json
```diff
@@ -91,23 +91,25 @@
...
-    "ember-template-imports": "^4.1.2",
     "execa": "^5.1.1",
     "node-fetch": "2.7.0",
     "popper.js": "^1.16.1",
     "strip-ansi": "^6.0.0",
     "testem": "^3.10.1",
     "tslib": "^2.6.0",
     "typescript": "^5.4.5",
+    "vite-5": "npm:vite@^5.0.0",
+    "vite-6": "https://pkg.pr.new/vite@main",
```
mansona, line 112: We should probably wait for a beta to be released rather than keep this floating 🤔

mansona (dismissed):
I'm just flagging this incase anyone feels the urge to merge this too early 😂

packages/vite/src/build.ts
```diff
@@ -11,7 +11,7 @@ export function emberBuild(command: string, mode: string, resolvableExtensions:
     env['EMBROIDER_RESOLVABLE_EXTENSIONS'] = resolvableExtensions?.join(',');
   }
 
-  if (command === 'build') {
+  if (command === 'build' || process.env.FORCE_EMBER_CLI_EXIT) {
```
ef4, line 14: We can keep this but let's use something like `EMBROIDER_VITE_PREBUILD=false`

Issue #2207, opened 2024-12-10
Remaining issues before we cut the first alpha of Embroider@4


mansona:
While https://github.com/embroider-build/embroider/issues/1860 is a tracking issue for the next Stable release from the main branch, we discussed the changes that we would like to see on main before we cut the first alpha release. This will effectively allow us to give some semver functionality to people using the new blueprint (with obvious alpha caveats before it is actually stable 😂 ) 

## Todo

- [x] Update the implementation for ember-cli-build to match what is now said in the RFC https://github.com/NullVoxPopuli/rfcs/blob/v2-app-spec/text/0977-v2-app-format.md#ember-pre-build-config---ember-cli-buildjs
- [x] implement setting `staticInvokables` that is a combination of `staticComponents`, `staticModifiers`, and `staticHelpers`
  - [x] implement this on stable with a deprecation
  - [x] Throw an error on main https://github.com/embroider-build/embroider/pull/2246
- [x] throw an error if anyone sets the `skipBabel` setting https://github.com/embroider-build/embroider/pull/2247
- [x] throw an error if anyone sets `staticAddonTrees` or `staticAddonTestSupportTrees` https://github.com/embroider-build/embroider/pull/2256
  - [ ]  optionally document how people would workaround this issue by manually putting things in the requirejs loader
  - [x] implement on stable with a deprecation https://github.com/embroider-build/embroider/pull/2244
- [x] stop nullifying the loader.js library so people can use it if they want to (even if we don't) - https://github.com/embroider-build/embroider/pull/2242
- [x] throw an error if anyone sets `staticEmberSource` to false https://github.com/embroider-build/embroider/pull/2255
  - [x] implement on stable with a deprecation https://github.com/embroider-build/embroider/pull/2245
- [x] ensure the simplified `ember()` and `classicEmberSupport()` vite plugins allow the appropriate level of overriding their results. Right now it doesn't appear possible to unset things like the rollup build options. https://github.com/embroider-build/embroider/pull/2260
- [x] configure release plan to publish the alpha correctly https://github.com/embroider-build/embroider/pull/2259

mansona:
Alpha is out 💪

PR #2242, opened 2025-02-01
stop nullifying loader.js

mansona:

We agreed that we would stop nullifying loader.js even though we don't use it any more for anything. This is to stop us from preventing other people using it for custom things.

Note: this won't mean that there are any of ember's entities in requirejs.entries, it just means that if you have custom code that is using this system it's not blocked from working any more.

Note: I considered adding tests for this but that would suggest that we support loader.js and could imply that we recommend using it. This PR just stops preventing people from using it which I think is ok without a test? thoughts?

### DIFF
--- a/packages/compat/src/compat-app.ts
+++ b/packages/compat/src/compat-app.ts
@@ -675,7 +675,5 @@ interface Preprocessors {
 function forbiddenVendorPath(path: string) {
   // ember-source does not go in vendor under embroider (we always use the
   // separate ES modules)
-  //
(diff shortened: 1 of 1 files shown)


NullVoxPopuli (approved):
That argument makes sense to me

void-mAlex (approved):

ef4:
A note about this: due to this change it's actually pretty important that loader.js gets removed from the blueprints. It's an addon and will put itself into vendor.js if it's present, and that will mislead people because `require` will be present but not functional.

PR #2243, opened 2025-02-01
add a simple deprecation for the skipBabel setting

mansona:

This PR just adds a simple deprecation for the `skipBabel` setting. I didn't exactly know what to put in the message since people are going to need to fix this manually **after** upgrading rather than getting rid of the "deprecation" before upgrading 🤔


NullVoxPopuli (approved):

mansona:
We decided against doing this since there is no real way for people to get rid of the deprecation before upgrading

PR #2246, opened 2025-02-02
remove static components, helpers, modifiers settings

mansona:

This removes the staticComponents, staticHelpers, and staticModifiers settings and throws an error if they are used


NullVoxPopuli (approved):

tests/scenarios/compat-resolver-test.ts
```diff
@@ -145,7 +143,7 @@ Scenarios.fromProject(() => new Project())
           'templates/application.hbs': `{{hello-world arg=1}}`,
         });
 
-        await configure({ staticComponents: true });
+        await configure({ staticInvokables: true });
 
         expectTranspiled('templates/application.hbs').failsToTransform(
```
mansona, line 148: So this case doesn't throw an error any more 🤔 I don't know if that is expected

packages/compat/src/resolver-transform.ts
```diff
@@ -323,15 +320,15 @@ class TemplateResolver implements ASTPlugin {
   }
 
   private get staticComponentsEnabled(): boolean {
```
ef4, line 322: All three of these private getters can get replaced with a single staticInvokablesEnabled

ef4 (approved):

PR #2247, opened 2025-02-02
throw an error if anyone sets skipBabel

mansona:

This PR adds an error whenever someone still has the `skipBabel` config set. This setting has already been removed in https://github.com/embroider-build/embroider/pull/2102 and this error is just informing people that they should remove the setting since it no longer does anything.

### DIFF
--- a/packages/compat/src/options.ts
+++ b/packages/compat/src/options.ts
@@ -141,6 +141,12 @@ export function optionsWithDefaults(options?: Options): CompatOptionsType {
     );
   }
 
+  if ((options as any)?.skipBabel !== undefined) {
+    throw new Error(
+      `You have set 'skipBabel' on your Embroider options. This setting has been removed and you can now configure your babel ignores directly in the babel config in your repo https://babeljs.io/docs/options#ignore`
+    );
+  }
+
   return Object.assign({}, defaults, options);
 }
 
--- a/tests/addon-template/ember-cli-build.js
+++ b/tests/addon-template/ember-cli-build.js
@@ -15,11 +15,5 @@ module.exports = function (defaults) {
   */
 
   const { maybeEmbroider } = require('@embroider/test-setup');
-  return maybeEmbroider(app, {
-    skipBabel: [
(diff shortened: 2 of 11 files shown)


NullVoxPopuli (approved):

packages/core/src/options.ts
```diff
@@ -107,6 +107,11 @@ export type CoreOptionsType = Required<
   Pick<Options, 'staticHelpers' | 'staticModifiers' | 'staticComponents' | 'staticInvokables'>;
 
 export function optionsWithDefaults(options?: Options): CoreOptionsType {
+  if ((options as any)?.skipBabel) {
```
SergeAstapov, line 110: Just theoretically possible someone make skipBabel: false and this won’t detect the case so maybe could be
```suggestion
  if (‘skipBabel’ in (options as any) || {}) {
```

SergeAstapov (approved):

packages/core/src/options.ts
mansona: oh that's nice thanks! Although this PR isn't ready because I was expecting CI to error because we have some places where this is being set 🙈

mansona: oh turns out this doesn't work? 🤔 I just tried it with `staticEmberSource` and it's always reporting true:

```
  if ('staticEmberSource' in (options as any) || {}) {
    console.log({ options });
  }
```

That logs:

```
{
  options: {
    staticAddonTrees: true,
    staticAddonTestSupportTrees: true,
    allowUnsafeDynamicComponents: false,
    staticInvokables: true,
    amdCompatibility: { es: [] }
  }
}
```

mansona: I've tested the new version and I'm happy that it's right now 👍

NullVoxPopuli:
needs rebase?

mansona:
@NullVoxPopuli no this is the error I was looking for, there are some places in our CI that are setting this and we've turned it into an error

PR #2205, opened 2024-12-10
Add minimal app template

NullVoxPopuli:
1. first commit: copy the app-template
2. remove everything related to compat
3. ???
4. tests pass?

### ember-qunit pulls on ember-cli-test-loader
```
✘ [ERROR] Could not resolve "ember-cli-test-loader/test-support/index.json" [plugin embroider-esbuild-resolver]

    ../../node_modules/.pnpm/ember-qunit@8.1.1_@ember+test-helpers@4.0.4_ember-source@6.1.0-beta.1_qunit@2.22.0/node_modules/ember-qunit/dist/test-loader.js:2:60:
      2 │ import AbstractTestLoader, { addModuleIncludeMatcher } from 'ember-cli-test-loader/test-support/index';
        ╵                                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### importing v2 addons fails in esbuild - [repo](https://github.com/NullVoxPopuli/ember-vite-minimal-testing/pull/1/files#diff-ef7badc94a5facb6ca713fcf3162b6f4edf3a147849b05dfffc3a6d25a88ef61)
```
Error: Build failed with 1 error:
  
node_modules/.pnpm/typescript-memoize@1.1.1/node_modules/typescript-memoize/dist/memoize-decorator.js:132:51: 
  ERROR: [plugin: embroider-esbuild-resolver] 
    <.pnpm>/ember-welcome-page@7.0.2/node_modules/ember-welcome-page/dist/components/welcome-page.js: 
      this.config.activePackageRules is not iterable
    at failureErrorWithLog (<.pnpm>/esbuild@0.21.5/node_modules/esbuild/lib/main.js:1472:15)
    at <.pnpm>/esbuild@0.21.5/node_modules/esbuild/lib/main.js:945:25
    at <repo>/node_modules/.pnpm/esbuild@0.21.5/node_modules/esbuild/lib/main.js:1353:9
```

<details><summary>full error, regarding `@ember/test-helpers` (a v2 addon)</summary>

```
  ➜  press h + enter to show help
✘ [ERROR] <.pnpm>/@ember+test-helpers@4.0.4_@babel+core@7.26.0_ember-source@6.1.0-beta.1/node_modules/@ember/test-helpers/dist/setup-rendering-context.js: this.config.activePackageRules is not iterable [plugin embroider-esbuild-resolver]

[...]

PR #2249, opened 2025-02-04
Expose macros config in babel

NullVoxPopuli:
@embroider/macros were not configurable in main/vite before this change.

Proposed way of configuring macros is similar to how we've been doing so in ember-cli-build.js

<details><summary>removed from this PR</summary>

## If we do end up needing this in the future,

we should have compat-macros output their combined options to compat-app-builder -> addBabelCompat

rather than the output,

and then have the babel plugin from compat read that in and it could then merge the options.

----------

For example, in the babel config:
```js
const {
  babelCompatSupport,
  templateCompatSupport,
} = require("@embroider/compat/babel");

module.exports = {
  plugins: [
   // ... 
    ...babelCompatSupport({
      '@embroider/macros': {
        setOwnConfig: {},
        setConfig: {
          'library-name': {}
        },
      }
    }),
  ],
  // ...
};
```

</details>

However, this doesn't solve the problem of configuring macros for the minimal app over in https://github.com/embroider-build/embroider/pull/2205

For the minimal app blueprint, we'd have folks do this in their babel config:

```js
const {
  babelCompatSupport,
  templateCompatSupport,
} = require("@embroider/compat/babel");

const { MacrosConfig } = require('@embroider/macros/src/macros-config');

let root = process.cwd();
let macros = MacrosConfig.for({}, root);
let { plugins: templateMacros, setConfig } = MacrosConfig.transforms();

setConfig(macros);

macros.setOwnConfig(root, { /* ... */ });
macros.setConfig(root, 'library-a', { /* ... */ });
macros.setConfig(root, 'library-b', { /* ... */ });

if (process.env.NODE_ENV === 'development') {
  macros.enablePackageDevelopment(process.cwd());
  macros.enableRuntimeMode();
}

macros.finalize();


module.exports = {
  plugins: [
    [
      "babel-plugin-ember-template-compilation",
      {
        compilerPath: "ember-source/dist/ember-template-compiler.js",
        transforms: [...templateMacros],
      },
    ],

   // ... 
   macros.babelPluginConfig();
  ],
  // ...
};
```
which.... could be better...

So.... I've done added to macros to allow for:
```js
const { buildMacros } = require('@embroider/macros/babel'); 

const macros = buildMacros({
  // this is how you configure your own package
  setOwnConfig: {
    // your config goes here
  },
  // this is how you can optionally send configuration into your
  // dependencies, if those dependencies choose to use
  // @embroider/macros configs.
  setConfig: {
    'some-dependency': {
      // config for some-dependency
    },
  },
});

module.exports = {
  plugins: [
   // ... 
    [
      "babel-plugin-ember-template-compilation",
      {
        compilerPath: "ember-source/dist/ember-template-compiler.js",
        transforms: [...macros.templateMacros],
      },
    ],
    ...macros.babelMacros,
  ],
  // ...
};

```

Which should be nice for the minimal blueprint in https://github.com/embroider-build/embroider/pull/2205

### DIFF
--- a/packages/macros/package.json
+++ b/packages/macros/package.json
@@ -17,7 +17,7 @@
   "exports": {
     ".": "./src/index.js",
     "./babel": "./src/babel.js",
-    "./src/*": "./src/*"
+    "./src/*": "./src/*.js"
   },
   "files": [
     "src/**/*.js",
--- a/packages/macros/package.json
+++ b/packages/macros/package.json
@@ -14,6 +14,11 @@
   "license": "MIT",
   "author": "Edward Faulkner",
   "main": "src/index.js",
+  "exports": {
+    ".": "./src/index.js",
+    "./babel": "./src/babel.js",
+    "./src/*": "./src/*"
+  },
   "files": [
     "src/**/*.js",
     "src/**/*.d.ts",
--- a/packages/compat/src/babel.ts
+++ b/packages/compat/src/babel.ts
@@ -18,7 +18,41 @@ export interface CompatBabelState {
   templateMacros: Transform[];
 }
 
-function loadCompatConfig(): CompatBabelState {
+interface CompatOptions {
+  /**
+   * Options for @embroider/macros
+   */
+  '@embroider/macros': {
+    /**
+     * How you configure your own package / app
+     */
+    setOwnConfig?: object;
+    /**
+     * This is how you can optionally send configuration into
+     * your dependencies, if those dependencies choose to use
+     * @embroider/macros configs.
+     *
+     * @example
+     * ```js
+     * setConfig: {
+     *   'some-dependency': {
+     *      // config for some-dependency
+     *   }
+     * }
+     * ```
+     */
+    setConfig?: Record<string, object>;
+
+    /**
+     * Callback for further manipulation of the macros' configuration instance.
+     *
+     * Useful for libraries to provide their own config with defaults shared between sub-dependencies of those libraries.
+     */
+    configure?: (macrosInstance: MacrosConfig) => void;
+  };
+}
+
+function loadCompatConfig(options?: CompatOptions): CompatBabelState {
   let compatFile = join(locateEmbroiderWorkingDir(process.cwd()), '_babel_compat_.js');
   if (existsSync(compatFile)) {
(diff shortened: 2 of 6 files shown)


ef4:
Notes from discussion at office hours:
 - we should make the argument to `babelCompatSupport` work even if you're using a prebuild, and we think that doesn't work now because when there's a prebuild the entire macros config gets built into _babel_compat_.js ahead of time and your options are ignored.

NullVoxPopuli:
I think in order to do that we need to have the classic macros config, in ember-cli-build.js, load the babel config.

It's too intense to re-apply the options post-macros-creation.

NullVoxPopuli:
err no, it'd just be the same problem, but in reverse

NullVoxPopuli:
Update: I removed the changes to `babelCompatSupport()`, and updated the description with notes from our meeting today for if/when we want to add in the merging

packages/macros/package.json
```diff
@@ -14,6 +14,11 @@
   "license": "MIT",
   "author": "Edward Faulkner",
   "main": "src/index.js",
+  "exports": {
+    ".": "./src/index.js",
+    "./babel": "./src/babel.js",
+    "./src/*": "./src/*.js"
```
simonihmig, line 20: Do we want people to import everything from `src`? Or is this for backwards compatibility, because people were doing so before?

NullVoxPopuli: its backwards compatibility because _we_ were doing it before.
I didn't audit what all we're importing, but it probably would be good to restrict at some point.

mansona: if we're going to restrict it now is the time. We won't be able to reduce it once this is landed because we really don't want a new major. Can we just remove it and see what fails?

NullVoxPopuli: I'm more than happy to do a follow-up PR which removes `src`.
I don't really want to make this PR do 2 things (1. current focus, 2. change available modules)

simonihmig: I agree with both of you: let's remove it, but in a follow-up PR :)

simonihmig (approved):
LGTM, but I'll leave the final call to more knowledgeable people! 🙃

PR #2251, opened 2025-02-05
test vite beta

patricklx:


patricklx:
oh, I see, that PR is actually testing the beta now. not necessary 6, https://pkg.pr.new/vite@main points to beta.
might rename `vite-6` to `vite-beta`
I see that it was not necessary to update lock file? so scenario tester will get the latest?

mansona:
CI always runs in no-lockfile mode, so if you have a dependency correctly specified that it will float in that context then yes it will work 👍 

When we get the other PR merged you should just add beta to the matrix and it will get tested along with 5.x and 6.x

PR #2253, opened 2025-02-06
Add `disableCssProcessing` option for webpack

evoactivity:
When an app provides it's own webpack css pipeline there is no reason for the default webpack config to also process css files, you can end up with multiple versions of the same file output into the build (eg. renaming the css files by configuring MiniCssExtractPlugin yourself).

I'm not sure how to add a test for this, some guidance here would be helpful. Do I add a new scenario? Do I just add a test to an existing scenario?

Some documentation should be added to explain this option.

### DIFF
--- a/packages/webpack/src/ember-webpack.ts
+++ b/packages/webpack/src/ember-webpack.ts
@@ -111,6 +111,7 @@ const Webpack: PackagerConstructor<Options> = class Webpack implements Packager 
   private extraThreadLoaderOptions: object | false | undefined;
   private extraBabelLoaderOptions: BabelLoaderOptions | undefined;
   private extraCssLoaderOptions: object | undefined;
+  private disableCssProcessing: boolean | undefined;
   private extraStyleLoaderOptions: object | undefined;
   private _bundleSummary: BundleSummary | undefined;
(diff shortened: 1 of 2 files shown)


mansona:
I just noticed that you're targeting main here 🙈 main Webpack support doesn't work yet and it won't work this way any more, it will have its own webpack config when we eventually release it

evoactivity:
Yeah, I switched to stable and there were 388 files in this PR 🙈. So I changed back until I had time to merge stable back in locally and push up a correct PR.

evoactivity:
@mansona I've got this targeting the stable branch now.
What would be the best way to test this? I don't see any current tests for things like `styleLoaderOptions`, or `cssLoaderOptions`.

NullVoxPopuli:
Is this still needed, given we have https://github.com/embroider-build/embroider/pull/2299 now?

evoactivity:
I think so, you still end up with two rules processing css files when you use a custom pipeline, doubling the workload.

evoactivity:
I've added documentation to the readme, the failing tests I'm pretty sure are just flaky windows issues.

evoactivity:
@mansona Anything else for me to do on this?

PR #2263, opened 2025-02-12
template-tag-codemod test coverage and robustness fixes

ef4:
- we were picking up ambient babel configs unintentionally
- we would fail to exit due to broccoli-babel-transpiler in the prebuild (sigh)


mansona:
the code looks great, it looks like it's not compatible with windows 😭 I can't instantly see from the code where it's trying to execute `.` as a command 🤔

mansona (approved):

PR #2264, opened 2025-02-12
Improved codemod scope management

ef4:
This refactors template-tag-codemod so that all cases can handle arbitrary existing scope collisions.

### DIFF
--- a/packages/compat/src/resolver-transform.ts
+++ b/packages/compat/src/resolver-transform.ts
@@ -40,10 +40,13 @@ export interface ExternalNameHint {
   (path: string): string | null;
 }
 
+export type ExternalResolver = (module: string) => string;
+
 export interface Options {
   appRoot: string;
   emberVersion: string;
   externalNameHint?: ExternalNameHint;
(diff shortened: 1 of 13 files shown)


ef4:
I still have one more bug to investigate in this area. I have an app where the codemod fails on the first run (where a prebuild is done) and succeeds the second time (when reusing prebuild).

mansona:
@ef4 I've been working on something this weekend and I don't think that problem is isolated to just the codemod 🤔 I'm seeing the same behaviour when converting an app to Vite and it essentially fails the first run and then gets farther on the second (haven't fixed all the issues yet so can't say it succeeds on the second run 😂)

ef4:
Yeah, I agree. I think the codemod is just a good example to exercise the bug.

I'm going to land this. I regret not hitting merge on Friday because it would have avoided confusion and conflicts with https://github.com/embroider-build/embroider/pull/2271

mansona:
sorry that was my bad 🙈 I looked for this PR to see if it was merged before merging that, I just missed it 🙃

PR #2270, opened 2025-02-14
Introduce `EMBROIDER_WORKING_DIRECTORY` env var for concurrent builds

### LINKED ISSUE #2086: Allow concurrent builds
Opening this to explore if we can support running production builds (for the same app) concurrently. Use case is to build the same app, but with different config, e.g. for different environments like changing CDN host names.

Obviously, the webpack/vite output directory would need to be different. But that should be easy to do, like using an environment variable that gets read from in the webpack/vite config.

The part that seems to require changes here would be avoiding conflicts for the stuff that Embroider writes into `node_modules/.embroider`. 

Any concerns for trying to make this folder more dynamic, like making this have a random part like `node_modules/.embroider-<random-hash-like-string>`? And having that by default, would that cause performance regressions for example, like does Embroider reuse (parts of) that output between builds (which wouldn't work when it was randomized)?

On the implementation side, seem there is already a [central place](https://github.com/embroider-build/embroider/blob/d9fc688f7abd5663045e182da9f28a44c84cfc64/packages/shared-internals/src/working-dir.ts) where this folder location is determined, would it be as simple as adding the randomization there, or are there more things to be aware of?

simonihmig:
Currently Embroider will write rewritten packages and other build meta data into `node_modules/.embroider`. That location is hardcoded. Usually not a problem, but when running more than one build for the same app concurrently, builds will step on each others toes by trying to write into the same dir. The optional environment variable introduced here will allow to specify that directory for each build to isolate them.

Fixes #2086

### DIFF
--- a/packages/shared-internals/src/working-dir.ts
+++ b/packages/shared-internals/src/working-dir.ts
@@ -9,7 +9,10 @@ export function locateEmbroiderWorkingDir(appRoot: string): string {
   if (cache.has(appRoot)) {
     return cache.get(appRoot);
   }
-  if (existsSync(resolve(appRoot, 'package.json'))) {
+  if (process.env.EMBROIDER_WORKING_DIRECTORY) {
+    let path = resolve(appRoot, process.env.EMBROIDER_WORKING_DIRECTORY);
+    return path;
+  } else if (existsSync(resolve(appRoot, 'package.json'))) {
     // the normal case
     let path = resolve(appRoot, 'node_modules', '.embroider');
     cache.set(appRoot, path);


simonihmig:
Do we want to document this somewhere, or is this more like a hidden feature? t seems no one than us had a need for this before...

Also no tests. I could add some unit tests for the `locateEmbroiderWorkingDir`, which does not have any. Testing the actual "feature" of concurrent builds in CI seems like too much overhead I think? I tested this locally, and it was solving our problem.

SergeAstapov:
+1 for documenting this.

as for question "where to document this", maybe "Options" section in readme.md would work

SergeAstapov (approved):

mansona:
I feel like this probably should have a test for it 🙈 it seems like a sensible change if we can guarantee that **everything** is going through this function that relies on that location, but verifying would make me feel more comfortable about that. Also it means that we would have a regression test to avoid breaking this in future

PR #2271, opened 2025-02-15
Template tag codemod stability fixes

void-mAlex:
documents codemod usage in readme
fixes case when codemod is run against stable build of an app
make codemod work oob without failing on self resolvability

### DIFF
--- a/packages/template-tag-codemod/src/index.ts
+++ b/packages/template-tag-codemod/src/index.ts
@@ -162,8 +162,16 @@ export async function inspectContents(
               {
                 appRoot: process.cwd(),
                 emberVersion: resolverLoader.resolver.options.emberVersion,
-                externalNameHint(name: string) {
-                  return name;
+                externalNameHint(path: string) {
+                  return path
+                    .split('::')
+                    .map(part =>
+                      part
+                        .split('-')
+                        .map(innerPart => innerPart.charAt(0).toUpperCase() + innerPart.slice(1))
+                        .join('')
+                    )
+                    .join('_');
                 },
               } satisfies ResolverTransformOptions,
             ],
--- a/packages/template-tag-codemod/src/index.ts
+++ b/packages/template-tag-codemod/src/index.ts
@@ -101,11 +101,15 @@ export async function ensureAppSetup() {
   try {
     pkg = resolverLoader.resolver.packageCache.get(process.cwd());
   } catch (err) {
-    console.error(`Run template-tag-codemod inside a Ember app.`);
+    console.error(`Run template-tag-codemod inside a Ember app or run an Embroider build of the app first`);
     process.exit(-1);
   }
   if (!pkg.packageJSON.exports) {
-    throw new Error(`must use package.json exports for self-resolvability. Plase add this to package.json:
+    pkg.packageJSON.exports = {
+      './tests/*': './tests/*',
+      './*': './app/*',
+    };
+    console.info(`Using package.json exports for self-resolvability. Please add this to package.json to silence this message:
 
  "exports": {
     "./tests/*": "./tests/*",
--- a/packages/template-tag-codemod/src/index.ts
+++ b/packages/template-tag-codemod/src/index.ts
@@ -106,7 +106,7 @@ export async function ensureAppSetup() {
   try {
     pkg = resolverLoader.resolver.packageCache.get(process.cwd());
   } catch (err) {
-    console.error(`Run template-tag-codemod inside a Ember app or run an Embroider build of the app first`);
+    console.error(`Run template-tag-codemod inside a Ember app.`);
     process.exit(-1);
   }
   if (!pkg.packageJSON.exports) {
--- a/packages/template-tag-codemod/src/index.ts
+++ b/packages/template-tag-codemod/src/index.ts
@@ -1,4 +1,4 @@
-import { existsSync, readFileSync, unlinkSync, writeFileSync } from 'fs';
+import { existsSync, readFileSync, unlinkSync, writeFileSync, rmSync } from 'fs';
 import { globSync } from 'glob';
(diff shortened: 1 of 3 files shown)


packages/template-tag-codemod/src/index.ts
```diff
@@ -101,11 +106,15 @@ export async function ensureAppSetup() {
   try {
     pkg = resolverLoader.resolver.packageCache.get(process.cwd());
   } catch (err) {
-    console.error(`Run template-tag-codemod inside a Ember app.`);
+    console.error(`Run template-tag-codemod inside a Ember app or run an Embroider build of the app first`);
```
mansona, line 109: isn't the "embroider prebuild" the "Embroider build" you're talking about here? I feel like this could error for a million different reasons and maybe we should actually share `err.message` or something?

packages/template-tag-codemod/src/prebuild.ts
```diff
@@ -31,7 +31,7 @@ export async function prebuild() {
 
   let project = new Project(root, pkg, ui, cli);
   let app = new EmberApp({ project });
-  let builder = new broccoli.Builder(Compat.prebuild(app));
+  let builder = new broccoli.Builder(Compat.prebuild(app, { staticEmberSource: true }));
```
mansona, line 34: This will console.log a warning that it's not needed any more, it defaults to true on main

mansona (approved):
Looks like some good fixes, I've just added some micro comments that I'm fine with us fixing after this is merged since it's such a benefit to update the instructions

packages/template-tag-codemod/src/prebuild.ts
void-mAlex: last time I tried it it spilled warning on the console which I are counter productive to running the codemod as there is no way to cler them
if you're saying that it not going to happen with the next release I can remove this

mansona: so I'm saying that that shouldn't be the case on alpha right now 🤔

void-mAlex:
I would imagine this is a bugfix rather than an enhancement but I defer to you @mansona

mansona:
@void-mAlex your changes to `externalNameHint` I think warrant  the enhancement label but it doesn't really matter since everything is being released as an alpha at the moment 😂

ef4:
We discussed making this configurable. 

I think the new default behavior here is debatable. I don't like it because namespacing on component names is an old habit based on having all components in one big global pool. Now that they're imported, it's not great style.

By analogy, which of these do you do?

```js
import { uniq } from 'lodash';
import { uniq as lodash_uniq } from 'lodash';
```

The default here is equivalent to the second one of those.

A better default for namespaced stuff would probably be dropping everything before the last `::`. 

```gjs
// before
<Widgets::Big />

// after
import Big from './widgets/big';
<template><Big /></template>
```

packages/template-tag-codemod/src/index.ts
```diff
@@ -86,7 +86,12 @@ export function optionsWithDefaults(options?: Options): OptionsWithDefaults {
 type OptionsWithDefaults = Required<Options>;
 
 export async function ensurePrebuild() {
-  if (!existsSync('node_modules/.embroider')) {
+  const stablePrebuildExists = existsSync('node_modules/.embroider/rewritten-app');
```
ef4, line 89: This is not good. It leaves the bug behind to bite us again anytime there's version skew between the embroider that wrote `node_modules/.embroider` and the embroider that's running. We should use a version marker instead.

packages/template-tag-codemod/src/index.ts
```diff
@@ -105,7 +110,11 @@ export async function ensureAppSetup() {
     process.exit(-1);
   }
   if (!pkg.packageJSON.exports) {
-    throw new Error(`must use package.json exports for self-resolvability. Plase add this to package.json:
+    pkg.packageJSON.exports = {
```
ef4, line 113: This is mutating the package cache. It might happen to work, but it changes the meaning of the package cache from an optimization to a correctness requirement.

I'm in favor of auto-fixing this problem for users, but we should do it by writing their real package.json. Not trying to maintain an inconsistency in the package cache.

mansona:
I reopened this PR here https://github.com/embroider-build/embroider/pull/2280 sorry for the confusion everyone 🙈

void-mAlex:
> We discussed making this configurable.
> 
> I think the new default behavior here is debatable. I don't like it because namespacing on component names is an old habit based on having all components in one big global pool. Now that they're imported, it's not great style.
> 
> By analogy, which of these do you do?
> 
> ```js
> import { uniq } from 'lodash';
> import { uniq as lodash_uniq } from 'lodash';
> ```
> 
> The default here is equivalent to the second one of those.
> 
> A better default for namespaced stuff would probably be dropping everything before the last `::`.
> 
> ```gjs
> // before
> <Widgets::Big />
> 
> // after
> import Big from './widgets/big';
> <template><Big /></template>
> ```

namespecing is actually more important now in the new world order where we have 
```js
import { hash } from '@ember/helper';
import { hash } from 'rsvp';
```
and these are far from the only examples and they will be pushed to the extereme once route templates gain traction

namepsacing is also a closer to existing behaviour that people are used to with the framework
ask too much change at once and it will reflect negatively in the experience they have with the new format
I agree that having the option for shorter names should be presented but I strongly believe the ootb experience should the the one closer to what people are already familiar with to make the transition smoother

while you may have your opinions on what may or may not make sense in a gjs world fact is if you're using this piece of code to do transformations chances are you have an older codebase with namespacing folders in place and there is a high likelyhoold of multiple Cards/Card/Slideup in skeleton/provider/loader etc variants and having that be preserved is much more valuable than haing a poorly mangled Card/Card0/Card1 name

PR #2272, opened 2025-02-16
Cleanup v2 addon rollup warnings, update virtual parkages list

chancancode:
Context: https://github.com/embroider-build/addon-blueprint/issues/327

image

image

1. standard modules is missing the new template compiler API, causing an unresolved import warning

2. gjs publish format uses `eval()` which rollup understandably doesn't like, but we know what we are doing™, so we can hide that warning for our users

The eval warning suppression is perhaps a little broad, since it suppresses the warning for other uses of evals in any gjs/gts files, but it would be challenging to be more precise than this.

Realistically, if this isn't already covered by your eslint rules, you have bigger problems than this.


NullVoxPopuli (approved):
Thank you!

PR #2274, opened 2025-02-16
remove duplicate options default

void-mAlex:


mansona (approved):
whoops! must have been a copy pasta error 🍝

PR #2276, opened 2025-02-17
make sure all packages stay within pre-release

mansona:
The `semverIncrementAs` is a bit finicky to work with and we need to make sure that things are correct on the Prepare Release PR before merging 👍


mansona:
I cancelled CI since this change has no effect on the tests 👍

PR #2279, opened 2025-02-17
fix failure on first-run and auto-apply package.json updates

ef4:
This fixes https://github.com/embroider-build/embroider/pull/2264#issuecomment-2660582278 and also addresses automatically adding `exports` to package.json, since they're both in the same function.

The problem with first-run failures was that Resolver expects to get constructed after any prebuilding is done. We were constructing it in ensureAppSetup before the prebuild.

This class of bug is harder to see now that prebuilds are optional. ResolverLoader will happily construct a Resolver with no prebuilt output present. But the resulting Resolver won't know anything about any v1 addons that were present.

### DIFF
--- a/packages/template-tag-codemod/src/index.ts
+++ b/packages/template-tag-codemod/src/index.ts
@@ -1,13 +1,13 @@
 import { existsSync, readFileSync, unlinkSync, writeFileSync } from 'fs';
 import { globSync } from 'glob';
-import core, { type Package } from '@embroider/core';
+import core from '@embroider/core';
 import { traverse, parseAsync, type types, transformFromAstAsync } from '@babel/core';
 import * as babel from '@babel/core';
 import templateCompilation, { type Options as EtcOptions } from 'babel-plugin-ember-template-compilation';
 import { createRequire } from 'module';
 import { extractTemplates, locateTemplates } from './extract-meta.js';
 import reverseExports from '@embroider/reverse-exports';
-import { dirname } from 'path';
+import { dirname, resolve } from 'path';
 import type { ResolverTransformOptions } from '@embroider/compat';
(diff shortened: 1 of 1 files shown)


mansona (approved):
Looks good 👍

packages/template-tag-codemod/src/index.ts
```diff
@@ -97,22 +97,21 @@ export async function ensurePrebuild() {
...
-  },
-
-`);
+  let json = JSON.parse(content);
+  if (!json.exports) {
+    json.exports = {
+      './tests/*': './tests/*',
+      './*': './app/*',
+    };
+    writeFileSync(filename, JSON.stringify(json, null, 2));
```
void-mAlex, line 114: this is touching files I would never expect a codemod to touch and messing with my indentation at the same time

packages/template-tag-codemod/src/index.ts
```diff
@@ -97,22 +97,21 @@ export async function ensurePrebuild() {
 }
 
 export async function ensureAppSetup() {
-  let pkg: Package;
+  let filename = resolve(process.cwd(), 'package.json');
+  let content: string;
   try {
-    pkg = resolverLoader.resolver.packageCache.get(process.cwd());
+    content = readFileSync(filename, 'utf8');
```
void-mAlex, line 103: this no longer serves the purpose of ensuring an ember app
I had fun digging into what it was actually doing because it was buried very deep inside resolver code but it actually ensures ember-source is in package json which may not be the case for a v2 addon but I wonder if that was the intention here

mansona: That sounds like a good case to add to the test suite. I'm assuming we want to encourage people to be able to use this to update a v2 addon?

PR #2281, opened 2025-02-17
Include version in prebuild output

ef4:
Normal builds don't get skew here because the embroider working dir gets deleted when you start a fresh build. But template-tag-codemod is attempting to reuse the prebuild to make it easier to iterate on codemod runs. This ensures that the prebuild was made on the exact same @embroider/core version.


mansona (approved):
👍

PR #2285, opened 2025-02-18
Introduce EMBROIDER_WORKING_DIRECTORY env var for concurrent builds

simonihmig:
Backporting #2270 to `stable`.

### DIFF
--- a/packages/shared-internals/src/working-dir.ts
+++ b/packages/shared-internals/src/working-dir.ts
@@ -9,7 +9,10 @@ export function locateEmbroiderWorkingDir(appRoot: string): string {
   if (cache.has(appRoot)) {
     return cache.get(appRoot);
   }
-  if (existsSync(resolve(appRoot, 'package.json'))) {
+  if (process.env.EMBROIDER_WORKING_DIRECTORY) {
+    let path = resolve(appRoot, process.env.EMBROIDER_WORKING_DIRECTORY);
+    return path;
+  } else if (existsSync(resolve(appRoot, 'package.json'))) {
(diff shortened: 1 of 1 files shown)


simonihmig:
Failing tests seem to be a general issue on `stable`!?

mansona:
@simonihmig yes that's true, since Ember 6.1 we can't rely on the auditing system because we now rely on `exports` in package.json 😞 We discussed this with @ef4 and we have a very soft "let's not change stable" since we're so close to getting the new release out, the reason being that it would be a significant effort to fix the audit system on stable 

I don't know how we want to proceed on this 😞

simonihmig:
For my understanding, what is the "auditing system"? 😅 

So after having migrated hundreds of addons to v2 and Embroider compatibility, we are only now in the process to migrate production apps to "stable" Embroider + webpack. It might take us quite a while to be able to adopt Embroider/Vite (for existing apps), pending Ember upgrades but also some other roadblocks potentially (left over AMD shenanigans). 

So I do expect us to be on Embroider/webpack for a while. I don't expect new development there of course, but I/we would appreciate if we could still have it in some kind of maintenance mode, to accept an occasional bug fix or land some low hanging fruits like this one.

> I don't know how we want to proceed on this

If it is only for Ember 6.1+, can we limit the test matrix to only run on earlier versions?

ef4:
I don't think it's only a bug in the audit system. On `stable`, `nodeResolve` is used for more than just audits. It's used for resolving components for example. It's only on `main` where we've been able to unify all resolving through the prevailing environment.

PR #2286, opened 2025-02-18
Workaround babel evaluation bug

### LINKED ISSUE #1812: Invalid code output when using `@embroider/macros` (ember app)

### LINKED ISSUE #2212: @embroider/macros babel plugin: Optional chaining expression gets aggressively optimized to `undefined`

ef4:
Fix #2212
Fix #1812
Supersedes #2215

We were already doing most of the evaluation ourselves. Of all the cases we test, the only spot we were relying on babel's `evaluate` was for the undefined keyword, which was easy to add to our own Evaluator.

### DIFF
--- a/packages/macros/src/babel/evaluate-json.ts
+++ b/packages/macros/src/babel/evaluate-json.ts
@@ -113,12 +113,6 @@ export interface UnknownResult {
 
 export type EvaluateResult = ConfidentResult | UnknownResult;
 
-// this is needed to make our strict types work when inter-operating with
-// babel's own built-in evaluator
-function isConfidentResult(result: { confident: boolean; value: any }): result is ConfidentResult {
-  return result.confident;
-}
-
 export interface EvaluationEnv {
   knownPaths?: Map<NodePath, EvaluateResult>;
   locals?: { [localVar: string]: any };
@@ -193,11 +187,6 @@ export class Evaluator {
   }
 
   private realEvaluate(path: NodePath): EvaluateResult {
-    let builtIn = path.evaluate();
-    if (isConfidentResult(builtIn)) {
-      return { ...builtIn, hasRuntimeImplementation: false };
-    }
-
     if (path.isMemberExpression()) {
       return this.evaluateMember(path, false);
     }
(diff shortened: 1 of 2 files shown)


mansona:
If we're targeting stable on this do we need to invest in fixing the audit system? 🙈

PR #2295, opened 2025-02-19
Ember canary fixes

ef4:
We're getting type errors against canary, I think updating the dependency is required.


ef4:
There's an actual failure remaining in our canary-typescript-app scenario. The component authored in GTS is working correctly, the component authored in TS looks like its decorators are getting transpiled wrong.

PR #2299, opened 2025-02-21
Add ability to configure MiniCssExtractPlugin

lfloyd117:
Often times developers need the ability to configure the MiniCssExtractPlugin and we haven't been able to. Adding a simple option to pass into Embroider's webpack config so we can configure it now. 
Can change the name if we'd rather it be something different.


NullVoxPopuli (approved):
seems safe

NullVoxPopuli:
Looks like the core of the problem that requires the need for this is that we use different CSS tools for dev and prod:

image

evoactivity:
> Looks like the core of the problem that requires the need for this is that we use different CSS tools for dev and prod:

We absolutely should be.

lfloyd117:
@NullVoxPopuli  Yeah, the issue for us has been that Embroider has its own MiniCssExtractPlugin setup without configuration for production builds only.

PR #2301, opened 2025-02-21
Make app-template-minimal more minimal

NullVoxPopuli:
Make the minimal template _super_ minimal.

Learnings along the way:
- we get to use type=module
- we get to use package.json#imports
- we can use `src` instead of `app`
- we don't need a config directory
- we still have to have ember-cli (for running tests)
- ember-cli does not auto-detect a testem.cjs file, so we must use
  --config-file

Things that need to happen, but haven't yet, because I think we should make a better API in macros for this:
```js
import { getGlobalConfig } from '@embroider/macros/src/addon/runtime';

const ENV = {
  modulePrefix: 'limber',
  environment: import.meta.env.DEV ? 'development' : 'production',
  rootURL: '/',
  locationType: 'history',
  EmberENV: {},
  APP: {},
};

export default ENV;

export function enterTestMode() {
  ENV.locationType = 'none';
  ENV.APP.rootElement = '#ember-testing';
  ENV.APP.autoboot = false;
  getGlobalConfig()['@embroider/macros'].isTesting = true;
}
```

otherwise `macroCondition(isTesting())` always hits the else branch


NullVoxPopuli:
Well, this is unexpected:
```
Rollup failed to resolve import "app-template-minimal/components/fancy-component" from "/tmp/tmp-2080nydRmxXxC16i/tests/integration/components/fancy-component-test.gjs".
```

tests/app-template-minimal/package.json
```diff
@@ -6,41 +6,47 @@
...
+    "#components/*": "./src/components/*"
+  },
   "exports": {
     "./tests/*": "./tests/*",
-    "./*": "./app/*"
+    "./*": "./src/*"
   },
   "scripts": {
     "build": "vite build",
+    "build:test": "NODE_ENV=development vite build --mode test",
```
ef4, line 21: cross-env?

ef4 (approved):

tests/app-template-minimal/package.json
ef4: I realized there's a better solution than this env var. The vite docs say you can create `.env.[mode]` and set `NODE_ENV` inside there. So if we make `.env.test` with `NODE_ENV=development` that should do it.

PR #2305, opened 2025-02-22
Fix vendor ember-template-compiler.js on v2 ember-source

ef4:
This fixes a typo pointed out by @chancancode.

This code exists to allow people to `app.import('vendor/ember-template-compiler.js')` after upgrading to ember-source >= 6.2.

It's not clear whether anybody actually *does* that, and the forward-looking plan for this area is that v2 apps can't do `app.import`, rendering the question moot.

### DIFF
--- a/packages/compat/src/compat-app.ts
+++ b/packages/compat/src/compat-app.ts
@@ -362,7 +362,7 @@ export default class CompatApp {
       trees.push(writeFile('vendor/ember/ember.js', () => ''));
       trees.push(writeFile('vendor/ember/ember-testing.js', () => ''));
       const templateCompilerSrc = readFileSync(join(emberSource.root, 'dist/ember-template-compiler.js'), 'utf8');
-      trees.push(writeFile('vendor/ember/ember-testing.js', () => templateCompilerSrc));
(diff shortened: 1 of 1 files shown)


chancancode:
@ef4 you mentioned in the ember.js thread that the other something else uses the other dist files, so figured I would confirm that you intentionally left the other ones (a few line above this) blank.

Also, to fully fix the source maps warning, either the source maps needs to be copied over (slightly complicated by the fact that it’s missing in older versions) or the reference comment needs to be scrubbed. Doesn’t have to be in this PR though.

ef4:
The empty files here are intentional. If they weren't empty the code would be duplicated. In the case we're looking at, all of ember is consumed as separate modules instead.

ef4:
Also I think the source map warning you're seeing depends on this bug. ember-testing.js is actually consumed (and was supposed to be empty). That's why the build would care about it containing a broken source map. 

Whereas the ember-template-compiler.js would only be consumed if somebody is intentionally loading the runtime template compiler in the old way.

ef4:
(We can still fix the broke link of course. Just explaining why I think this PR alone silences the warning.)

chancancode (approved):

PR #2319, opened 2025-02-27
template-tag-codemod: only reuse ember prebuild if you explicitly pass the option to do so

mansona:
This was discussed during the office hours and weekly meeting

essentially while developing the template-tag-codemod it was inconvenient to have to wait for the Ember prebuild every time the script was run. But when people are using this in real apps it is usually better that they don't re-use the prebuild, especially if they are likely to be making changes that would change the contents of the prebuild

This PR removes the reuse of the prebuild entirely 👍

### DIFF
--- a/packages/template-tag-codemod/src/index.ts
+++ b/packages/template-tag-codemod/src/index.ts
@@ -78,6 +78,14 @@ export interface Options {
   // need to install `@embroider/template-tag-codemod` as a dependency to access
   // it in this way)
   renamingRules?: 'string';
+
+  /**
+   * This allows you to reuse the ember-prebuild if you are running the codemod multiple times.
+   *
+   * You should only set this setting if you know that any changes you're making will not influence
+   * the prebuild in any way. Most people should leave this set to its default `false` value
+   */
+  reusePrebuild?: boolean;
 }
 
 export function optionsWithDefaults(options?: Options): OptionsWithDefaults {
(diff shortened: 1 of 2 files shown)


ef4:
This is part of what we discussed but it's missing a key step: we need to never use the same working dir as the regular embroider build. Otherwise people will get trolled if they have a build running in some other shell, and it starts messing with the working dir again.

I think it's sufficient to set `process.env.EMBROIDER_WORKING_DIRECTORY` to something like `node_modules/.template-tag-codemod`.

Also, please leave some way to explicitly ask to skip the prebuild, for when we're working on the codemod itself. A flag or env var.

mansona:
@ef4 done 👍

packages/template-tag-codemod/src/index.ts
```diff
@@ -682,9 +693,11 @@ function applyEdits(source: string, edits: { start: number; end: number; replace
 }
 
 export async function run(partialOpts: Options) {
+  // this makes sure that we don't fight with other versions of embroider currently running on the system
+  process.env.EMBROIDER_WORKING_DIRECTORY = 'node_modules/.template-tag-codemod';
```
ef4, line 697: This is too late. Run the code and you'll see that it's not taking effect.

ef4 (changes requested):

packages/template-tag-codemod/src/index.ts
mansona: yikes you're right, I was driving myself mad trying to figure out why it wasn't working and then I realised it's being ready in **module scope** so there is no way to get it to the "start of execution" 🙈 

I've updated it now and set it just before it is read and confirmed locally that it's working 👍

PR #2320, opened 2025-02-27
convert from yargs to commander

mansona:
This is a draft PR because it does have in influence on the public API of the codemod and we should discuss if we want to move in this direction

I don't know if it's just a preference thing but I find commander a lot easier to use, apart from the way that they handle boolean options 🙈 hence the change in the API

I'll add it to the agenda to discuss on Tuesday 👍


ef4:
I will point out that the general ambiguity of how boolean args are handled currently is load-bearing infrastructure. 😂

`--components`, `--renderTests`, and `--routeTemplates` are string-valued only, but you can treat them like booleans and make the expected thing happen. `--components false` happens to work because none of your components will match the glob "false".

PR #2321, opened 2025-02-27
move ember-cli to peerDependency

mansona:
Most of the time this won't be a problem for people since ember-cli can usually be upgraded independently of your app, but this breaks for anyone trying to use the codemod with ember-source@3.28 because of the bowerDependencies deprecation and removal in ember-cli@5

The issue I have here is that we kinda want a peer, but we don't actually install the codemod 🙈 does `npx` consider the cwd when thinking about peers? What is the best way forward here? should I be doing an explicit resolve from cwd and dropping the dependency altogether? Thoughts?


ef4:
Yeah, I doubt you can rely on a peer when you're executing via `npx` or `dlx`. 

Can we just downgrade our dependency to 4.x?

mansona:
I guess that's probably ok 🤔 I don't think there is anything since 4.x that people would rely on for the prebuild. Let me try that 👍

mansona:
@ef4 I've tested this locally with my 3.28 app and it's working amazingly 🎉

PR #2322, opened 2025-02-27
add a failing test for Component.extend()

mansona:
testing this out on an old and out of date package threw up another place where people might have issues 🙈 This adds a test for cases where components are exported using `export default Component.extend({})` 

I couldn't find the code in the rest of embroider that handles this 🤔 but I figure we must be handling this somewhere else. Alternatively if we don't want to support this path we could easily add a case that instructs people to update that file to a `export default Class` without the `.extend({})` call, maybe pointing them at a different codemod? 

thoughts?

### DIFF
--- a/tests/scenarios/template-tag-codemod-test.ts
+++ b/tests/scenarios/template-tag-codemod-test.ts
@@ -180,6 +180,63 @@ tsAppScenarios
         });
       });
 
+      test('basic js backing component', async function (assert) {
+        await assert.codeMod({
+          from: {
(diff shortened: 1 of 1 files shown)


ef4:
Just found this case myself in an app and I'm working on better handling.

I don't think we'll attempt to codemod this case, but we can handle it more gracefully. 

 - the codemod should skip over files that it can't handle and report them all at the end, rather than stop.
 - we should detect this known case and explain it, rather than throwing `unimplemented declaration: CallExpression`

I think the thing we can point people at is https://www.npmjs.com/package/ember-native-class-codemod

PR #2323, opened 2025-02-27
Add Vite support for Ember@5.8 

mansona:
I'm starting to work on the auto-patching of older ember-source versions and I noticed that in my empty generated app for ember-source@5.8 tests were passing 🤔 I wanted to verify in our CI that we still need to patch ember-source@5.8 and hopefully move the patching to be in a compat adapter 👍 

~Edit: turns out something has changed and we don't need to patch Ember 5.8 any more 🤔 that... is surprising 🫠~

~I'm going to keep going with older versions to see where the line is but I guess we can merge this now?~

Edit: as @ef4 points out the module cycle with the `deprecate` function was only really in dev mode so we needed to make sure we're doing our Vite dev mode style test to expose the problem. This PR now shows the issue in one of the commits and then fixes it in the final commit 👍


ef4:
I think nothing is exercising 5.8 Add 5.8 to the matrix in vite-internals-test and I bet the `vite dev` part fails.

mansona:
I took a slightly different approach because `vite-internals-test` uses the tsAppScenarios and we haven't set up the wideTSAppScenarios yet. We really need to audit our tests soon 🙈 

I made sure to submit this as a failing commit with the expected failure and a new commit with everything passing so you can see that it is behaving correctly 👍

ef4 (approved):

PR #2333, opened 2025-03-01
Ember 5 4

mansona:
This is a bit of a hacky implementation of fixing the test loading, I will probably want to convert this to be a real babel-based implementation like #2323 

This essentially "backports" a fix for loading tests similar to the implementation in this commit of ember-source https://github.com/emberjs/ember.js/commit/f679f035d1318829652da714e13031092e2f6f8e

We're obviously not making as many changes, I was aiming on getting tests loading and leaving it at that and I succeeded 🎉 (at least locally)


ef4 (approved):

MrChocolatine:
Will this addition be documented somewhere? Or is it already maybe?

PR #2342, opened 2025-03-05
Avoid crashing the rollup watchmode build when glint hits a syntax error

### LINKED ISSUE #2341: [addon-dev@7.1.1] Syntax errors in watch mode crash glint, breaking hot-reloading in the test-app
When developing, it's quite common to accidentally save files with syntax errors, especially if you use some kind of autosaving mechanism. 

Currently, in a v2 typescript addon, this breaks the watch-mode of the addon build, which means that even when you fix the syntax error, your test-app no longer reloads.
This is frustrating, as you may not notice this happened, and are left wondering why your code changes don't seem to make it to the test-app.

Steps to reproduce:
- `npx ember-cli@latest addon repro -b @embroider/addon-blueprint --pnpm --typescript`
- `pnpm start`
- make a syntax error in the addon code
- notice the build breaks (as expected, of course)
- fix the syntax error
- notice the addon doesn't rebuild

I can provide a minimal repro repo if necessary, but I think I've already found the fix and am making a PR.

abeforgit:
As described in the issue, you don't want the build watching script to completely exit when it hits a syntax error. This means you have to restart `pnpm start` once you've fixed the error, which is annoying.

fixes #2341

### DIFF
--- a/packages/addon-dev/src/rollup-declarations.ts
+++ b/packages/addon-dev/src/rollup-declarations.ts
@@ -12,6 +12,7 @@ export default function rollupDeclarationsPlugin(declarationsDir: string) {
         await execa('glint', ['--declaration'], {
           stdio: 'inherit',
           preferLocal: true,
+          reject: false
         });
 
         await fixDeclarationsInMatchingFiles(declarationsDir);
--- a/packages/addon-dev/src/rollup-declarations.ts
+++ b/packages/addon-dev/src/rollup-declarations.ts
@@ -1,18 +1,21 @@
+import type { Plugin } from 'rollup';
 import execa from 'execa';
 import walkSync from 'walk-sync';
 import { readFile, writeFile } from 'fs/promises';
 
-export default function rollupDeclarationsPlugin(declarationsDir: string) {
+export default function rollupDeclarationsPlugin(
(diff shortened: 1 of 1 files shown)


NullVoxPopuli (dismissed):

packages/addon-dev/src/rollup-declarations.ts
```diff
@@ -1,17 +1,21 @@
...
-export default function rollupDeclarationsPlugin(declarationsDir: string) {
+export default function rollupDeclarationsPlugin(
+  declarationsDir: string
+): Plugin {
   let glintPromise: Promise<void>;
 
   return {
     name: 'glint-dts',
-    buildStart: () => {
+    buildStart() {
```
abeforgit, line 13: had to change this to method syntax to get access to `this`, unsure of the impact that might have. Other plugins seem to do this as well

NullVoxPopuli (approved):
Looks good. Will still fail normal builds which is what we want

Thank you!

abeforgit:
Don't immediately see why these tests would fail, let me know if there's anything I can do

NullVoxPopuli:
looks like it was just windows being flaky (I reran)

PR #2344, opened 2025-03-05
move most of the implementation of staticInvokables to core

mansona:
I didn't realise in #2210 what the implications of version drift would be splitting the deprecation and the default across two packages. This fixes that slight mess that I created 🙈 this effectively reverts the changes to `@embroider/compat` that provides any hint of the deprecation and handles it all in `@embroider/core`. This means that you won't get strange errors any more when you have a miss-match of versions 👍 

Edit: I moved all the cases where it would throw an error to core so they are now contained in one package but I still needed to accomidate the functionality in compat because of the implementation of `optionsWithDefaults()`. I also bumped the minimum peer version of compat->core so that should help with the version skew 👍

### DIFF
--- a/packages/compat/src/compat-app-builder.ts
+++ b/packages/compat/src/compat-app-builder.ts
@@ -68,9 +68,6 @@ import type { CompatOptionsType } from './options';
 export class CompatAppBuilder {
   // for each relativePath, an Asset we have already emitted
   private assets: Map<string, InternalAsset> = new Map();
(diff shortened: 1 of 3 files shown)


NullVoxPopuli (approved):
Seems good -- had to compare with https://github.com/embroider-build/embroider/pull/2210/files to see that a good chunk of this PR is a revert of #2210

ef4 (approved):

PR #2348, opened 2025-03-06
Support named const exports in template-tag-codemod

CvX:
This adds a `--addNameToTemplateOnly` flag to the codemod to optionally make template-only components use named consts:

```gjs
const MyComponent = <template>…</template>;
export default MyComponent;
```

That helps with autocompletion in IDEs.

Ideally, editors/plugins/language server would recognize components that don't have an explicit export/a name, but currently this is an okay workaround.


ef4:
Thanks, this seems fine as an optional thing. 

It should probably compose correctly with `--defaultFormat gts` though. Right now if you use `--defaultFormat gts`, then `--addNameToTemplateOnly` is silently ignored.

mansona (approved):
It looks like the requested changes have been made 👍 thanks for the contribution 🎉

PR #2354, opened 2025-03-08
Fix addon-dev's declarations plugin when glint fails

NullVoxPopuli:
### Successes

image
image

### Type errors retain color

image

(and there is a summary at the bottom that reports which plugin errored)

### The downsides, some types of errors are hard to understand:

(this error is from Volar)
#### Watch

image

#### Build

image

### DIFF
--- a/packages/addon-dev/src/rollup-declarations.ts
+++ b/packages/addon-dev/src/rollup-declarations.ts
@@ -2,6 +2,7 @@ import type { Plugin } from 'rollup';
 import execa from 'execa';
 import walkSync from 'walk-sync';
 import { readFile, writeFile } from 'fs/promises';
+import { existsSync } from 'fs';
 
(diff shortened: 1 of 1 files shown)


packages/addon-dev/src/rollup-declarations.ts
```diff
@@ -12,12 +13,16 @@ export default function rollupDeclarationsPlugin(
...
-        await execa('glint', ['--declaration'], {
+        let { exitCode } = await execa('glint', ['--declaration'], {
           stdio: 'inherit',
           preferLocal: true,
-          reject: this.meta.watchMode,
+          reject: !this.meta.watchMode,
         });
 
+        if (this.meta.watchMode && exitCode > 0) {
+          this.warn(`Failed to generate declarations`);
```
NullVoxPopuli, line 23: we can't use `this.error` because it halts the watch

PR #2358, opened 2025-03-10
fix gitignore for switching between stable and main

mansona:
This is purely to make our lives better while working on the embroider repo 😂 

Essentially when you run a build on the `main` branch and then switch to stable there are a bunch of build artefacts that are seen as new files. This PR updates the `.gitignore` file to ignore files that are currently being ignored on main 👍


mansona:
I cancelled the workflow because this just changes the `.gitignore` file and we don't need to run tests against it 👍

PR #2361, opened 2025-03-11
add support for maintaining git history for both sides of the codemodded files

void-mAlex:
create a branch for renaming both parts of the combined file to cause a conflict and show git which parts come from which file


ef4:
Yes, we discussed that to really land this is should be a standalone command closer to that.

That history looks nice on GitHub. I didn't realize that [GitHub supports .git-blame-ignore-revs by default](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view).

PR #2363, opened 2025-03-11
Add "merge-history" subcommand to codemod

ef4:
Based on ideas from both @void-mAlex's https://github.com/embroider-build/embroider/pull/2361 and @davidtaylorhq's https://gist.github.com/davidtaylorhq/484505beccd5eeac1166bd75002b3e5a

This one does all the work in a git ~subtree~ worktree, so it doesn't actually touch your checked out code at all. The only visible change should be the creation of its outputBranch.


davidtaylorhq:
This works great, thanks @ef4! Love the git subtree technique.

I ran it on [this sizable Discourse PR](https://github.com/discourse/discourse/pull/31746) and everything worked. The only tiny nitpick would be that it took more than 30s, so some kind of simple progress output would be nice to have.

PR #2365, opened 2025-03-12
Short circuit the reverseSearchAppTree when possible to reduce time

raycohen:
In a large app with lots of addons, this change dropped time spent in `reverseSearchAppTree` from 2.2s to 600ms.

I think `inAddonName` will always have `/_app_/` as part of it making this change safe.

### Before

Screenshot 2025-03-12 at 9 50 33 AM

### After

Screenshot 2025-03-12 at 10 04 54 AM

### DIFF
--- a/packages/core/src/module-resolver.ts
+++ b/packages/core/src/module-resolver.ts
@@ -1371,9 +1371,14 @@ export class Resolver {
     // really be understood as a request for a module in the containing engine
     if (owningPackage.isV2Addon()) {
       let sections = [owningPackage.meta['app-js'], owningPackage.meta['fastboot-js']];
+      let fromPackageRelativePath = explicitRelative(owningPackage.root, fromFile);
+      if (!fromPackageRelativePath.includes('/_app_/')) {
+        // the inAddonName values always contain /_app_/, so we can exit early without it
+        return;
+      }
+
       for (let section of sections) {
         if (section) {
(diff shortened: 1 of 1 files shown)


packages/core/src/module-resolver.ts
```diff
@@ -1371,9 +1371,14 @@ export class Resolver {
     // really be understood as a request for a module in the containing engine
     if (owningPackage.isV2Addon()) {
       let sections = [owningPackage.meta['app-js'], owningPackage.meta['fastboot-js']];
+      let fromPackageRelativePath = explicitRelative(owningPackage.root, fromFile);
```
raycohen, line 1374: I don't know if it is common for the sections to both be `undefined`, if so this change would hit `explicitRelative` more often which I could guard against

ef4:
It doesn't really need to include `_app_`, that's a convention, not a hard rule. But it's helpful that you've identified this spot as a bottleneck. 

We already build an optimized map for "forward" searches (from logical names in the app to actual files in the addons). We can store the reverse map too. I'm attempting that now in https://github.com/embroider-build/embroider/pull/2368

ef4:
If you're able to measure the perf of the alternate fix in https://github.com/embroider-build/embroider/pull/2368, that would be helpful.

raycohen:
Thanks, that looks pretty likely to have a similar or better effect.

For measuring, my app is currently a bit stuck on an older version of Embroider so I don't know if I can get a successful build with those changes though.

PR #2370, opened 2025-03-12
Handle name collisions when using `--addNameToTemplateOnly`

CvX:
A follow-up to #2348 that fixes cases where a template name would clash with one of the import names.

@mansona, let me know if you think it would be better to use JSUtils from https://github.com/emberjs/babel-plugin-ember-template-compilation (and know how 😉)


mansona:
So I was thinking about using babel-import-util but I guess it is not quite the same since you're not introducing a new import, you're introducing a new literal 🤔 @ef4 do you think this is ok? I'm personally ok with it since it adds a test to verify the behaviour

ef4 (dismissed):

PR #2372, opened 2025-03-13
always use source maps for gjs in vite

NullVoxPopuli:
Matches:
https://github.com/embroider-build/embroider/blob/main/packages/addon-dev/src/rollup-gjs-plugin.ts#L21

### DIFF
--- a/packages/vite/src/esbuild-resolver.ts
+++ b/packages/vite/src/esbuild-resolver.ts
@@ -42,7 +42,8 @@ export function esBuildResolver(): EsBuildPlugin {
     if (path.endsWith('.hbs')) {
       src = hbsToJS(src);
     } else if (['.gjs', '.gts'].some(ext => path.endsWith(ext))) {
-      src = preprocessor.process(src, { filename: path });
+      let { code /*,  map */ } = preprocessor.process(src, { filename: path });
+      src = code;
     }
     if (['.hbs', '.gjs', '.gts', '.js', '.ts'].some(ext => path.endsWith(ext))) {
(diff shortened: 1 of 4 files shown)


patricklx:
you also need to upgrade content-tag to 3.0?

packages/vite/src/esbuild-resolver.ts
```diff
@@ -42,7 +42,8 @@ export function esBuildResolver(): EsBuildPlugin {
     if (path.endsWith('.hbs')) {
       src = hbsToJS(src);
     } else if (['.gjs', '.gts'].some(ext => path.endsWith(ext))) {
-      src = preprocessor.process(src, { filename: path });
+      let { code /*,  map */ } = preprocessor.process(src, { filename: path });
+      src = code;
```
patricklx, line 46: I guess for this case we would actually need inline source map?

NullVoxPopuli: probably, ya. added!

mansona: I mean it probably won't make much difference here since this is mostly just used for dependency discovery in the esbuild phase 🤔

patricklx (approved):
:)

packages/vite/src/esbuild-resolver.ts
patricklx: Oh, right. Addons already pre compiled the gjs

mansona (approved):
Seems good to me 👍

PR #2383, opened 2025-03-14
[wip] Sketching new resolver

ef4:
This is a spike to demonstrate how I think a new minimalist Resolver implementation could work.

It illustrates one thing that we should add in embroider itself, which is that we need an import similar to `@embroider/virtual/compat-modules` that represents **only** addon-provided things (like the PageTitleService) without any of your own app things (which you're going to control instead with import.meta.glob). The format of the filenames in that import should be like I'm using here, with no modulePrefix.

I think the actual new `Resolver` class would ideally get added to ember-source, so there's not an additional package to depend on. That seems like the sensible place to put a runtime feature.


ef4:
If we put the new resolver into ember-source itself, we can also change its relationship to Application if we want to. The only reason it exists as a class was to customize resolving, but we don't ever want to support customized resolving again.

```js
class App extends Application {
  modules = import.meta.glob("./{services,routes,templates}/*.{gjs,gts,js,ts}")
}
```

PR #2385, opened 2025-03-14
Add Vite support for Ember 4.12

mansona:
You can see that this works here https://github.com/mainmatter/ember-vite-codemod/pull/53

Edit: ember-data@4.12 has issues so I don't think we can support it. I tried ember-data@4.13-alpha but it also has issues so we can update the 4.x test to use 4.13 once we iron out those problems 👍

### DIFF
--- a/packages/compat/src/compat-adapters/ember-source.ts
+++ b/packages/compat/src/compat-adapters/ember-source.ts
@@ -172,6 +172,52 @@ class ReplaceRequire extends Plugin {
     updateFileWithTransform(this, 'ember/index.js', function (babel: typeof Babel) {
       const { types: t } = babel;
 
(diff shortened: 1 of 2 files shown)


NullVoxPopuli (approved):

MrChocolatine:
Will this addition be documented somewhere? Or is it already maybe?
I already understand, from the PR description, we would need to execute this codemod: https://github.com/mainmatter/ember-vite-codemod

mansona:
@MrChocolatine yes you would need to run the codemod to upgrade to vite, and now it will work if you're still on ember-source@4.12 👍

mansona:
as for documentation, there is going to be a bunch of communication once the work is done ;)

PR #2396, opened 2025-03-18
use ember-data@4.13 for ember-source@4.12 test

mansona:


ef4:
Note from our debugging on this: 
 - babel-plugin-debug-macros is copying the predicate expression as a prefix before deprecate(), the predicate expression is itself a use of embroider macros, the interaction of the two results in an output macro that's not processed.
 - but all of this is fixed by fixing the bug that `@embroider/compat` always configures babel-plugin-debug-macros in debug mode. Once it goes into production mode, the deprecations get stripped anyway and there's no problem.
 - we'll want to use NODE_ENV, and we should update the blueprints with .env files to help keep NODE_ENV aligned with vite mode.

robbytx:
I was able to workaround this issue with the following change to `babel.config.cjs`:
```patch
--- a/babel.config.cjs
+++ b/babel.config.cjs
@@ -3,6 +3,29 @@ const {
   templateCompatSupport,
 } = require('@embroider/compat/babel');
 
+/**
+ * Workaround for https://github.com/embroider-build/embroider/pull/2396:
+ * @embroider/compat always configures babel-plugin-debug-macros in debug mode
+ */
+function fixBabelCompatSupport() {
+  const support = babelCompatSupport();
+
+  // Set debug/CI flags as in https://github.com/emberjs/ember-cli-babel/blob/v7.26.11/lib/babel-options-util.js#L58
+  const isProduction = process.env.NODE_ENV === "production";
+  const isDebug = !isProduction;
+
+  for (const [ , options, ] of support.filter(plugin => plugin[0].includes('/babel-plugin-debug-macros/'))) {
+    options.debugTools.isDebug = isDebug;
+    const glimmerOptions = options.flags?.find(({ source }) => source === '@glimmer/env');
+    if (glimmerOptions) {
+      glimmerOptions.flags.DEBUG = isDebug;
+      glimmerOptions.flags.CI = !!process.env.CI;
+    }
+  }
+
+  return support;
+}
+
 module.exports = {
   plugins: [
     [
@@ -33,7 +56,7 @@ module.exports = {
         regenerator: false,
       },
     ],
-    ...babelCompatSupport(),
+    ...fixBabelCompatSupport(),
   ],
 
   generatorOpts: {
```

robbytx:
I've created https://github.com/embroider-build/embroider/pull/2472 to address this issue in `@embroider/compat`.

PR #2401, opened 2025-03-19
Handle a couple more common use cases in template tag codemod

bendemboski:
This allows the template tag codemod to handle a couple more common formats for `.js`/`.ts` component files.


ef4:
I just added a commit adjusting how the typescript interface case gets handled.  Checking for `parentPath.node.type !== 'Program'` is not really sufficient. People can have `export default interface` at the top-level, and we need to ignore that case too.

bendemboski:
Ah, I didn't realize you could separately export a default value and a default type!

ef4:
Thanks, nice work.

PR #2408, opened 2025-03-20
Support `templateOnlyComponent()` usages in template tag codemod

bendemboski:
Add support to the template tag codemod for glint's [recommended](https://typed-ember.gitbook.io/glint/environments/ember/template-only-components) way of typing template-only components.

### DIFF
--- a/tests/scenarios/template-tag-codemod-test.ts
+++ b/tests/scenarios/template-tag-codemod-test.ts
@@ -353,6 +353,84 @@ tsAppScenarios
         });
       });
 
+      test('template-only ts component', async function (assert) {
+        await assert.codeMod({
+          from: {
+            'app/components/example.hbs': `Hello world`,
+            'app/components/example.ts': `
+              import templateOnlyComponent from '@ember/component/template-only';
+              interface FooSignature {
+                Args: {
+                  value: string;
+                };
+              }
+              export default templateOnlyComponent<FooSignature>();
+            `,
+          },
+          to: {
+            'app/components/example.gts': `
+              import type { TemplateOnlyComponent } from '@ember/component/template-only';
+              interface FooSignature {
+                Args: {
+                  value: string;
+                };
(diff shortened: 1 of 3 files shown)


tests/scenarios/template-tag-codemod-test.ts
```diff
@@ -353,6 +353,84 @@ tsAppScenarios
...
+          },
+          to: {
+            'app/components/example.gts': `
+              import type { TemplateOnlyComponent } from '@ember/component/template-only';
+              interface FooSignature {
+                Args: {
+                  value: string;
+                };
+              }
+              export default <template>Hello world</template> satisfies TemplateOnlyComponent<FooSignature>;
```
Windvis, line 378: Technically the `export default` is not needed in the .gts version, should we output it without that instead?

```suggestion
              <template>Hello world</template> satisfies TemplateOnlyComponent<FooSignature>;
```

The Glint docs still show it as `export default <template>` [due to a bug](https://github.com/emberjs/ember.js/pull/20862#discussion_r1997212481) that has [since been fixed](https://github.com/embroider-build/content-tag/pull/97). (Which reminds me we should PR a docs update as well :smile:).

bendemboski: The codemod already generates `export default` for template-only use cases ([here](https://github.com/embroider-build/embroider/blob/1218cbefc1ae0be5cb787369abd6de2211eded4f/tests/scenarios/template-tag-codemod-test.ts#L166) and [here](https://github.com/embroider-build/embroider/blob/1218cbefc1ae0be5cb787369abd6de2211eded4f/tests/scenarios/template-tag-codemod-test.ts#L410)). So I would propose that we keep this as-is in this PR, and then in a separate PR address removing the `export default` in all template-only cases.

ef4 (approved):
Thanks this looks good.

PR #2409, opened 2025-03-21
Document --templateInsertion option to codemod

kategengler:


kategengler:
I suspect CI failed for another reason since this was only a README change...

kategengler:
I agree. I'm not sure it will occur to everybody to run `--help` for something typically run via `npx`. I opened this PR after being asked how to do this.

jrjohnson:
I was about to open a PR to add this as a feature, thrilled it's already here!  

`Additional options are available via interactive --help` is obvious in hindsight, but because many of the options are documented I assumed they all were.

PR #2419, opened 2025-04-01
Add `--custom-resolver` option to template-tag-codemod

cvx:
It allows to provide custom import resolving rules.


CvX:
added a `resolve` callback support

tests/scenarios/template-tag-codemod-test.ts
```diff
@@ -34,6 +34,21 @@ tsAppScenarios
...
+              return 'bar/really-exists/foo';
+            }
+          }
+        `,
+        'another-resolver.mjs': `
+          export default async function(path, filename, resolve) {
+            if (path === '@embroider/virtual/components/phone-booth') {
+              path = '@embroider/virtual/components/message-box';
+            }
+            return await resolve(path, filename);
```
mansona, line 49: So micro point on this test, the intention of this API is to allow you resolve **first** and if it gives you an answer that you don't agree with then you can change it before returning. This doesn't really matter because the test clearly shows that the resolve function works 👍

mansona (approved):
🎉

PR #2422, opened 2025-04-02
set addon-shim to release a pre-minor

mansona:


mansona:
again cancelled to save the effort 👍

PR #2423, opened 2025-04-02
update the readme to more accurately reflect the current system

mansona:
This should be the final blocker for release 👍 😱


NullVoxPopuli (approved):

README.md
```diff
@@ -5,148 +5,51 @@
...
-* `nameHint` - optional function control the import name and template replacement values - valid JS identifier required or it will be coerced into one
-* `dryRun` - option can be used to obtain a summary of the changed the build would perform and which files it would act upon
-
-### Limitations
-
-* App templates only
-* `@embroider/compat` >= 3.6.0
+- `EMBROIDER_WORKING_DIRECTORY`: by default Embroider writes internal build-time artifacts like rewritten packages to `node_modules/.embroider`. In the case of running multiple builds concurrently (e.g. building for production and test in parallel) this would cause conflicts when concurrent processes try to write into the same directory. For this case you can point each Embroider process to a different directory using this environment variable. It can be an absolute file path, or relative to the application root directory.
 
 ## Compatibility
```
kategengler, line 54: GH won't let me comment on the untouched lines but below there is a 'With Classic Builds' header that seems off

mansona: yea you're right, we actually don't yet have an answer for engines so we can just remove the whole section. Thanks for the catch 🎉

ef4 (approved):

PR #2425, opened 2025-04-02
Adding  a compat adapter for ember-fetch

ef4:
This adapts ember-fetch so that it participates as an ES modules, instead of directly trying to do AMD.

It's necessary on ember-source > 6.1, where ember-source became a real V2 addon and you can't find it in AMD.

### DIFF
--- a/packages/compat/src/compat-adapters/ember-fetch.ts
+++ b/packages/compat/src/compat-adapters/ember-fetch.ts
@@ -1,0 +1,49 @@
+import { readFileSync } from 'fs-extra';
+import AddToTree from '../add-to-tree';
+import V1Addon from '../v1-addon';
+import { resolve } from 'path';
+import { writeFileSync } from 'fs';
+
+export default class extends V1Addon {
+  get packageMeta() {
+    let result = super.packageMeta;
+    result['renamed-modules'] = {
+      fetch: 'ember-fetch/index.js',
+    };
+    return result;
+  }
+  get v2Tree() {
+    // ember-fetch attempts to emits an AMD define into vendor.js. We instead
+    // adapt it into an ES module, so it can continue accessing ember even once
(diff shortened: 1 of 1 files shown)


NullVoxPopuli (approved):

packages/compat/src/compat-adapters/ember-fetch.ts
```diff
@@ -0,0 +1,49 @@
...
+    return result;
+  }
+  get v2Tree() {
+    // ember-fetch attempts to emits an AMD define into vendor.js. We instead
+    // adapt it into an ES module, so it can continue accessing ember even once
+    // ember is not in AMD.
+    return new AddToTree(super.v2Tree, (outputPath: string) => {
+      let vendorFile = resolve(outputPath, 'vendor/ember-fetch.js');
+      let src = readFileSync(vendorFile, 'utf8');
+      writeFileSync(vendorFile, '');
```
NullVoxPopuli, line 22: empty file? do we need this and line 20?

ef4: Yes, we need the empty file, because the addon has already done `app.import('vendor/embber-fetch.js')` and that is hard to reverse. If we just delete the file, the build will crash.

NullVoxPopuli: feel free to ignore, but I'd like a comment here 🙈 
This is a great example of when "why" comments are super useful. ymmv, ofc
I just really like when absolutely-not-obvious behaviors are documented around what would otherwise be pretty easy to understand code.

It's the tricks'n'such that we hold as tribal knowledge that make maintenance / fixes scary at times. <3

ef4: Fair point, comment added.

PR #2426, opened 2025-04-02
fix types on stable

ef4:
This is adjusting our own dev types, no public api change.


NullVoxPopuli (approved):
:see_no_evil: VM types

ef4:
Yeah I think we're getting the newer types via babel-plugin-ember-template-compilation and they fight with the olde ones we had here.

PR #2436, opened 2025-04-04
Fix resolving EmberCLI in vite prebuild

simonihmig:
It assumed the `ember` bin would exist in `./node_modules`, which might be true for most modern package managers, but in codebases using e.g. yarn hoisting (yes, I know, shame on us) this does not hold true. Using node module resolution here to find the executable.


mansona (approved):
Looks good to me 👍

mansona (approved):

PR #2449, opened 2025-04-11
Major speedup of reverse-exports

simonihmig:
When running Embroider/Vite the first time on one of our larger test-apps, I was seeing unexpectedly long build times for the initial load. Time to first paint was about 5 minutes. Network tab showed some requests like `app.css` taking 4 minutes to complete, but I don't think this was related to that particular request being processed, rather Vite was busy processing all kinds of other things in the background.

I took a [CPU profile](https://vite.dev/guide/troubleshooting.html#performance-bottlenecks) and was able to identify where this is coming from: the `@embroider/reverse-exports` package:

image

A few interesting things:
* the `mergeMaps` getter took 220s, not on its own but for the APIs it calls, mostly `externalName` from that package (221s in total)
* the biggest offenders under "Self time" are the utilities in that package, like `_prepareStringForRegex` and some arrow functions, e.g. the matcher that is creating and testing the Regex
* running the  `[/\-\\^$*+?.()|[\]{}]` RegEx replace alone (used in `_prepareStringForRegex`) is taking 15s
* lots of GC (14s)

For context: this is a typical YMMV situation. We were very affected by this performance bottleneck because we were historically using the "index-layout " for components and other modules (e.g. `<FancyButton>` lives in `components/fancy-button/index.ts` but gets imported from `fancy-addon/fancy-button`), which for v2 addons requires each and every index module to have its own package.json exports entry (see https://github.com/embroider-build/addon-blueprint/pull/229). So we have _a lot_! 😅 

This PR is fixing that by optimizing these utility functions that are in the hot path. Best reviewed by commit. The major changes are:
* use memoization: I was able to see these functions being called multiple times with the same args
* avoid using `.split('*')` for processing wildcards, as this creates intermediary arrays that need to get GC'ed
* avoid using RegEx when the exports path has no wildcard. A simple string match is enough.

With these changes, the browser has its first paint after 70s (instead of 300s) and my CPU profile now looks like this:

image

The `memoized` function take less than a second, but in return we get `mergeMaps` running in 3.7s (instead of 220s, see above), and no `reverse-exports` calls showing up anywhere. Also much less GC and no RegEx time (in the top of the list).

One more thing to point out. In the [first commit](https://github.com/embroider-build/embroider/pull/2449/commits/f1845336bb77642198384842ac81fa90ae843f65) I was changing the tests a bit, because I think they actually had wrong assumptions about how package.json exports work. I am a bit puzzled that it seems there is actually no detailed spec AFAICT, but the most authoritative resource I know of is the [node.js docs on exports](https://nodejs.org/api/packages.html#package-entry-points). Correct me if I'm wrong, but I don't think any of the following things that we had in tests are actually supported:
* use of multiple `*` wildcards. Node docs and every example I have ever seen only shows a single wildcard, also `resolve.exports` assumes that [here](https://github.com/lukeed/resolve.exports/blob/master/src/utils.ts#L43)
* use of globs like `**/*`
* some assumption that a path ending with a slash would cover deeper paths. Like `'./prefix/deep/': './other-directory/'` would map `my-addon/prefix/deep/foo/bar.js` to `./other-directory/foo/bar.js`. That is not the case, that is what you use wildcards for.

### DIFF
--- a/packages/reverse-exports/package.json
+++ b/packages/reverse-exports/package.json
@@ -15,6 +15,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "mem": "^8.0.0",
     "resolve.exports": "^2.0.2"
   }
 }
--- a/packages/reverse-exports/src/index.ts
+++ b/packages/reverse-exports/src/index.ts
@@ -2,7 +2,7 @@ import { posix } from 'path';
 import { exports as resolveExports } from 'resolve.exports';
 import memoize from 'mem';
 
-type PkgJSON = { name: string; exports?: Exports };
+type PkgJSON = { name: string; version: string; exports?: Exports };
 type Exports = string | string[] | { [key: string]: Exports };
 
 /**
@@ -81,7 +81,7 @@ export function _findPathRecursively(
   Returns undefined for a relativePath that is forbidden to be accessed from the
   outside.
 */
-export function externalName(pkg: PkgJSON, relativePath: string): string | undefined {
+function _externalName(pkg: PkgJSON, relativePath: string): string | undefined {
   let { exports } = pkg;
   if (!exports) {
     return posix.join(pkg.name, relativePath);
@@ -111,6 +111,10 @@ export function externalName(pkg: PkgJSON, relativePath: string): string | undef
 
   return posix.join(pkg.name, resolvedPath);
 }
+
+export const externalName = memoize(_externalName, {
+  cacheKey: ([pkg, relativePath]) => `${pkg.name}::${pkg.version}::${relativePath}`,
+});
 
 function regexEscape(input: string): string {
   return input.replace(/[/\-\\^$*+?.()|[\]{}]/g, '\\$&');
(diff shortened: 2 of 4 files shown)


packages/reverse-exports/package.json
```diff
@@ -15,6 +15,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "mem": "^8.0.0",
```
simonihmig, line 18: New versions are ESM only, which we cannot use it seems, since we are still compiling to CJS. 😟

NullVoxPopuli (approved):

mansona:
Now that we have https://github.com/embroider-build/embroider/pull/2450 you should be able to rebase and change the tsconfig target accordingly to have this package support the ESM that you need

packages/reverse-exports/package.json
NullVoxPopuli: is switching this to memoize the only thing that "needs" to be done post rebase? 

or, does this even need to change?

simonihmig:
> to have this package support the ESM

Trying [here](https://github.com/embroider-build/embroider/pull/2454), but not successful so far. Any help would be appreciated.

NullVoxPopuli:
Gonna merge, since the above esm efforts are kind of a side quest, and this improvement is amazing! Thank you @simonihmig !

PR #2455, opened 2025-04-18
Virtual fix for ember-inspector

### LINKED ISSUE #2484: Vite compatibility for ember-inspector
As part of the Ember Initiative, @mansona and I are trying to get ember-inspector to work for Vite apps. Two draft PRs have been opened in [`ember.js`](https://github.com/emberjs/ember.js/pull/20892) and [`ember-inspector`](https://github.com/emberjs/ember-inspector/pull/2625) as proof of concept of a new protocol that would allow the inspector to communicate with the Ember Vite app on the page.

Changes in `ember.js` will require going through an RFC process though, and meanwhile, Vite users are stuck with a non-functional inspector.

This issue consists in proposing a temporary solution to have ember-inspector work for Vite apps.

BlueCutOfficial:
Fixes #2484

## Context

As part of the Ember Initiative, @mansona and I are trying to get ember-inspector to work for Vite apps. Two draft PRs have been opened in [`ember.js`](https://github.com/emberjs/ember.js/pull/20892) and [`ember-inspector`](https://github.com/emberjs/ember-inspector/pull/2625) as proof of concept of a new protocol that would allow the inspector to communicate with the Ember Vite app on the page.

Changes in `ember.js` will require going through an RFC process though, and meanwhile, Vite users are stuck with a non-functional inspector.

## Changes

This PR proposes a temporary solution to have ember-inspector work for Vite apps: 
- a vite plugin creates the file that should be ultimately provided by `ember.js`: a `globalThis.emberInspectorLoader.load()` that returns everything the inspector needs to know to show information about the page.
- a compat-adapter for ember-source < 4.8 returns more or less the same thing but for these versions `@ember/enumerable/mutable` didn't exist yet, and `EnumerableMutable` must be imported from another place.

## How to test

- Serve the `app-template` on `http://localhost:4200/`
- Clone `mansona/ember-inspector` and switch to [vite-support branch](https://github.com/emberjs/ember-inspector/pull/2625)
- Follow the section of `ember-inspector`'s README about `pnpm serve:bookmarklet` to run the local inspector against `http://localhost:4200/`

Capture d’écran 2025-04-24 à 18 46 18


mansona (commented):
So looking at this PR I figured it looked strange that a virtual file would have a namespace that really exists in a ember-source 🤔 and also one that we expect to make a real file in some future version of ember-source

I know I kept saying "Virtual file" when we talked about this but with the fact that it's something that could change with different Ember versions and exists in a real namespace I think we should add this to the ember-source compat adapter 

I'm pretty sure the `replaceFile()` function that I created should be able to create new files 🤔 https://github.com/embroider-build/embroider/blob/main/packages/compat/src/compat-adapters/ember-source.ts#L320C5-L320C16

packages/compat/src/compat-adapters/ember-source.ts
```diff
@@ -115,6 +115,13 @@ export default class extends V1Addon {
       trees.push(new FixCycleImports([packages]));
     }
 
+    // As long as the inspector support is not implemented in an actual
+    // ember-source release, we can't set a range. This fix should be
+    // included for any version that has Vite support starting 3.28.
+    if (satisfies(this.packageJSON.version, '>=3.28')) {
```
mansona, line 121: turns out that semver isn't considering beta or canary as greater than 3.28 🫠 I figure we should remove this check but leave a comment that when we add support upstream we should wrap the tree in a condition 👍

BlueCutOfficial: Ah Ok 🫠 I pushed this at the end of my work day, I didn't have a look yet. Let's remove that condition then ^^

patricklx:
hi, I saw that this work started 👍 
also consider possible race conditions for the loading part.
especially when a window is reloaded and inspector was already opended, the order of script loading is not guaranteed .
so inspector debug inject vs ember app scripts.
in previous ember source there was a dispatched event `ember`, which inspector uses

mansona:
@patricklx yes that was at the bottom of the barrel file which is now deprecated, likely won't be included if you're using vite and will be removed in Ember 7 👍 so we can't rely on that for our architecture

patricklx:
Here is also a similar implementation directly in ember.js.
It contains the full list of all used classes and functions required by inspector.
https://github.com/emberjs/ember.js/pull/20580/files

packages/vite/src/inspector-support.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    async resolveId(source) {
+      if (source === '@embroider/virtual/compat-inspector-support') {
+        return '-compat-inspector-support.js';
+      }
+    },
+
+    load(id) {
+      if (id === '-compat-inspector-support.js') {
+        return `// This script exposes the modules used by Ember Inspector so an app that
```
ef4: Let's move the contents of the virtual file to their own real file and read that file here, just so it's easier to maintain.

ef4 (approved):

packages/vite/src/inspector-support.ts
```diff
@@ -0,0 +1,159 @@
...
+      } else {
+       modules = {
+          ...modules,
+          EnumerableMutable: await import('@ember/enumerable/mutable'),
+        };
+      }
+      return modules;
+    },
+  });
+  window.dispatchEvent(new Event('ember'));
```
patricklx: should be with upper case `Ember`

patricklx:
```suggestion
  window.dispatchEvent(new Event('Ember'));
```

packages/vite/src/inspector-support.ts
BlueCutOfficial: @ef4 I think you also mentioned removing the macro condition during the meeting. My understanding is that you recommend something like that:

```js
async load(id) {
      if (id === '-compat-inspector-support.js') {
         return `
           // Small script to load the modules required by Ember Inspector
           // it contains the instruction...
           await import('@embroider/virtual/compat-inspector-modules');
           // ...which resolves to -compat-inspector-modules.js
      }
     if (id === '-compat-inspector-modules.js') {
       if (ember-source < 4.8) {
         return // 3.28+ re-exports virtual file
       }
       return // 4.8+ re-exports virtual file
     }
}
```
But in the context of a Vite plugin, I am not sure what's the way to write the conditional about the ember-source version. Did you have a specific approach in mind? 

So far I was able to extract a set of re-exports as we mentioned during the meeting, but I still have the macro in the virtual load function to choose what set to import..

tests/app-template/app/app.js
```diff
@@ -4,6 +4,9 @@ import Resolver from 'ember-resolver';
 import loadInitializers from 'ember-load-initializers';
 import config from './config/environment';
 
+import setupInpector from '@embroider/virtual/compat-inspector-support';
```
BlueCutOfficial: @ef4 I didn't catch what we said about declaring `@embroider/virtual/compat-inspector-support` for the TypeScript app 🙈 Should I do something equivalent to what `@embroider/core` does for `@embroider/virtual/compat-modules`, but in the `@embroider/vite` package? Or was a different solution proposed?

BlueCutOfficial: We removed the exported function for app auto-naming so that should no longer be relevant.

packages/vite/src/inspector-support.ts
BlueCutOfficial: I'll try a method using `semver` and `resolve-package-path`.

PR #2456, opened 2025-04-19
Ember engines compatibility

void-mAlex:


packages/core/src/module-resolver.ts
```diff
@@ -1020,6 +1022,19 @@ export class Resolver {
...
+        } else if (pkg.isEngine()) {
+          let targetingEngine = this.engineConfig(packageName);
+          if (targetingEngine) {
+            let appJSMatch = await this.searchAppTree(
+              request,
+              targetingEngine,
+              request.specifier.replace(packageName, '.')
+            );
+            if (appJSMatch) {
+              console.log('fallbackResolve: non-relative appJsMatch', request.specifier);
```
mansona, line 1034: I would probably change this message to say something about engines since it's hyper engine specific 🤔 actually we should remove the console log and update the one below now that I look at it

ef4:
Notes from office hours: the newly added commits here are good and should be landed, but the first part of this PR should get replaced with the alternate implementation in https://github.com/embroider-build/embroider/pull/2465 (which needs some investigation to finish it).

Issue #2214, opened 2024-12-17
Using regexp for `splitAtRoutes` causes error


davidtaylorhq:
`splitAtRoutes` has handling for [regexp values](https://github.com/embroider-build/embroider/blob/528a2f40952940e2c65e9487244f160ac2f684bf/packages/core/src/virtual-entrypoint.ts#L324C17-L328C6). However, when this config is serialized to/from JSON, the regexp values are lost.

`ember-cli-build.js`:
```
splitAtRoutes: [/.*/]
```

Builds to `node_modules/.embroider/resolver.json`:
```
"splitAtRoutes": [
    {}
  ],
```

Resulting in this error:

```
[vite] Internal server error: pattern.test is not a function
at ... /@embroider/core/src/virtual-entrypoint.js:274:28
```

---

Running all `@embroider/*` packages on unstable b89b741

mkszepp:
i was also running into this isssue with now stable vite... we have used `new RegExp(/top-level\.([a-zA-Z-]+)/g)`.

It was working in embroider 3.x

Error occures in this line:
https://github.com/embroider-build/embroider/blob/4574bb78d25a3d2bffc602b1484e089b0f5c4af1/packages/core/src/virtual-entrypoint.ts#L309-L320

mansona:
I got a fix for this working https://github.com/embroider-build/embroider/pull/2458 🎉 there are a few open questions on the PR so I don't know how long it might take to get it landed

PR #2458, opened 2025-04-23
Fix splitAtRoutes when they are defined as a regex

mansona:
Fixes https://github.com/embroider-build/embroider/issues/2214

### DIFF
--- a/packages/compat/src/compat-app-builder.ts
+++ b/packages/compat/src/compat-app-builder.ts
@@ -143,7 +143,26 @@ export class CompatAppBuilder {
   }
 
   private addResolverConfig(config: CompatResolverOptions) {
-    outputJSONSync(join(locateEmbroiderWorkingDir(this.compatApp.root), 'resolver.json'), config, { spaces: 2 });
+    outputJSONSync(
+      join(locateEmbroiderWorkingDir(this.compatApp.root), 'resolver.json'),
+      {
+        ...config,
+        splitAtRoutes: config.splitAtRoutes?.map(s => {
+          if (typeof s === 'string') {
+            return {
+              type: 'string',
+              value: s,
+            };
+          } else {
+            return {
+              type: 'regex',
+              value: s.toString(),
+            };
+          }
+        }),
+      },
+      { spaces: 2 }
+    );
   }
 
   private addContentForConfig(contentForConfig: any) {
--- a/packages/core/src/resolver-loader.ts
+++ b/packages/core/src/resolver-loader.ts
@@ -6,6 +6,8 @@ import { join } from 'path';
 import type { FSWatcher } from 'fs';
 import { watch as fsWatch } from 'fs';
 
+type SplitRouteConfigType = { type: 'string'; value: string } | { type: 'regex'; value: string };
+
 export class ResolverLoader {
   #resolver: Resolver | undefined;
   #configFile: string;
@@ -28,7 +30,21 @@ export class ResolverLoader {
     if (!this.#resolver) {
       let config: Options;
       if (existsSync(this.#configFile)) {
-        config = readJSONSync(this.#configFile);
+        let rawConfig = readJSONSync(this.#configFile);
+
+        rawConfig.splitAtRoutes = rawConfig.splitAtRoutes?.map((splitRouteConfig: SplitRouteConfigType) => {
+          if (splitRouteConfig.type === 'regex') {
+            const fragments = splitRouteConfig.value.match(/^\/(.*)\/([gimsuy]*)$/);
+            if (!fragments) {
+              throw new Error(`Unable to parse splitAtRoutes pattern ${splitRouteConfig.value}`);
+            }
+            const [, parsedPattern, parsedFlags] = fragments;
+            return new RegExp(parsedPattern, parsedFlags);
+          }
+          return splitRouteConfig.value;
+        });
+
+        config = rawConfig;
       } else {
         config = buildResolverOptions({});
       }


packages/core/src/virtual-entrypoint.ts
```diff
@@ -312,9 +312,18 @@ function shouldSplitRoute(routeName: string, splitAtRoutes: (RegExp | string)[]
...
+        if (!fragments) {
+          throw new Error(`Unable to parse splitAtRoutes pattern ${pattern}`);
+        }
+        let regex = new RegExp(fragments[1], fragments[2] || '');
+        return regex.test(routeName);
+      }
       return pattern === routeName;
     } else {
-      return pattern.test(routeName);
+      throw new Error('Invalid splitAtRoutes type found');
```
mansona, line 326: I'm not sure this is useful or valid 🤔 I was toying with just removing the surrounding `if (typeof pattern === 'string') {` since they will always be serialised into a string so the check is pointless. 

Any thoughts?

packages/compat/src/compat-app-builder.ts
```diff
@@ -143,7 +143,11 @@ export class CompatAppBuilder {
   }
 
   private addResolverConfig(config: CompatResolverOptions) {
-    outputJSONSync(join(locateEmbroiderWorkingDir(this.compatApp.root), 'resolver.json'), config, { spaces: 2 });
+    outputJSONSync(
+      join(locateEmbroiderWorkingDir(this.compatApp.root), 'resolver.json'),
+      { ...config, splitAtRoutes: config.splitAtRoutes?.map(s => s.toString()) },
```
mansona, line 148: This is obviously a quick and dirty way to force every regex to a string, but this is essentially relying on regex detection by checking for it starting and ending with `/` when it's being deserialised. I don't know if there is a better way forward for this since we're (currently) relying on this being serialised into JSON 🙈 

It also means that there is currently no way to add this config for people not using a prebuild. Is this the time that we implement some sort of config file outside of the ember-cli-build? Thoughts?

packages/core/src/virtual-entrypoint.ts
```diff
@@ -312,9 +312,18 @@ function shouldSplitRoute(routeName: string, splitAtRoutes: (RegExp | string)[]
   }
   return splitAtRoutes.find(pattern => {
     if (typeof pattern === 'string') {
+      if (pattern.startsWith('/') && pattern.endsWith('/')) {
+        let fragments = pattern.match(/\/(.*?)\/([a-z]*)?$/i);
```
mansona, line 316: this was taken from a few places on stack-overflow and other blog posts on the topic. I'm sure this is probably overkill but it seems to work fine 🤷

tests/scenarios/compat-route-split-test.ts
```diff
@@ -21,7 +21,9 @@ let splitScenarios = appScenarios.map('compat-splitAtRoutes', app => {
         let app = new EmberApp(defaults, {});
         return maybeEmbroider(app, {
           staticInvokables: true,
-          splitAtRoutes: ['people'],
+          // this is quite complicated because we don't want to match 'people.show' just 'people' to
+          // keep the behaviou to what it was before
+          splitAtRoutes: [/peo([^.\s])+$/],
```
mansona, line 26: I didn't want to create a whole new setup to test the regex so I updated the existing one. It probably makes it overly complicated a test and removes the simple test from our CI 🤔 

Do we think it's worth it writing a new scenario or test setup for just regexes?

packages/core/src/virtual-entrypoint.ts
```diff
@@ -312,9 +312,18 @@ function shouldSplitRoute(routeName: string, splitAtRoutes: (RegExp | string)[]
   }
   return splitAtRoutes.find(pattern => {
     if (typeof pattern === 'string') {
+      if (pattern.startsWith('/') && pattern.endsWith('/')) {
```
mkszepp, line 315: this doesn't work when you are using regex like `/main-route\.([a-zA-Z-]+)/g`

we should something like `/^\/.*\/[gimsuy]*$/.test(pattern)`

packages/core/src/virtual-entrypoint.ts
```diff
@@ -312,9 +312,18 @@ function shouldSplitRoute(routeName: string, splitAtRoutes: (RegExp | string)[]
...
+        let fragments = pattern.match(/\/(.*?)\/([a-z]*)?$/i);
+
+        if (!fragments) {
+          throw new Error(`Unable to parse splitAtRoutes pattern ${pattern}`);
+        }
+        let regex = new RegExp(fragments[1], fragments[2] || '');
+        return regex.test(routeName);
+      }
```
mkszepp, line 323:
```suggestion
      if (/^\/.*\/[gimsuy]*$/.test(pattern)) {
          const fragments = pattern.match(/^\/(.*)\/([gimsuy]*)$/);

          if (!fragments) {
            throw new Error(`Unable to parse splitAtRoutes pattern ${pattern}`);
          }

          let regex = new RegExp(fragments[1], fragments[2]);

          return regex.test(routeName);
        }
```

mkszepp (changes requested):

packages/core/src/virtual-entrypoint.ts
mkszepp: the logic here works fine... but i think we have an other issue... when passing after the `/g`, shouldSplitRoute brings `true`, but it doesn't make route splitting...

packages/core/src/virtual-entrypoint.ts
mansona: yea so the reason I did it like this was because I know that regex can be quite expensive, but I did some logging of this function and it's not really a hot-path. It gets run twice for each route in your app (I don't know it's run twice and not once but that's a problem for another day)

packages/core/src/virtual-entrypoint.ts
mansona: I moved the fix down to the deserialisation layer so this code didn't need to change any more

packages/compat/src/compat-app-builder.ts
mansona: I have enhanced the check to see if something is a regex with the help of @mkszepp 's comments, but we still need to decide if this is the right kind of fix 👍

tests/scenarios/compat-route-split-test.ts
mansona: so my thoughts are that actually we should write a better split route test that covers all the cases we want to cover, this is a quick and dirty smoke test just for serialisation and deserialisation 👍

packages/core/src/virtual-entrypoint.ts
mansona: I've moved this and put it where the resolver config is being loaded so we have a better chance that it will only be called once now 👍 I used your code below in the new place

packages/core/src/virtual-entrypoint.ts
mansona: I used your code for the new implementation, thanks for the help 👍

packages/core/src/resolver-loader.ts
```diff
@@ -28,7 +28,21 @@ export class ResolverLoader {
     if (!this.#resolver) {
       let config: Options;
       if (existsSync(this.#configFile)) {
-        config = readJSONSync(this.#configFile);
```
mansona, line 31: This type inference here is a complete lie. The right hand side is a JSON with a bunch of `any` types and the config has the ability to have RegEx objects 🙈 i don't know what the right thing to do here is to deserialise this properly but I deferred the type inference while I added the checks I needed

ef4:
Instead of detecting regex strings, let's change the config file format so it always annotates whether the pattern is a string or a regex. We have full control over that format's serialization and deserialization.

PR #2464, opened 2025-04-30
Don't run terser when --mode=test

NullVoxPopuli:


ef4 (approved):

packages/vite/src/ember.ts
```diff
@@ -112,3 +95,31 @@ export function ember() {
...
+
+  /**
+   * Outside of test, the only other time "build" is used,
+   * is production
+   */
+  config.build ||= {};
+
+  if (!config.build.minify) {
+    config.build.minify = 'terser';
+  }
```
Techn1x, line 112: Rather, should this be checking that `config.build.minify` is undefined? Since it is allowed to be a boolean, if someone sets this `minify: false` in their config (for whatever reason), then this will ignore that

https://vite.dev/config/build-options#build-minify

NullVoxPopuli: That is a good point -- this should be updated to check undefined specifically

PR #2466, opened 2025-05-03
Fix bug with `{{content-for` when parsed by JSDOM

mfeckie:
Following a discussion here https://discord.com/channels/480462759797063690/1367859915179425927

I noticed that when JSDOM encounters HTML with `{{content-for "head"}}` etc, it moves the contents into `<body>`

I _suspect_ it's related to this https://github.com/jsdom/jsdom/issues/2221

This behaviour isn't ideal.

I've take a few liberties here in adding a test for this case using `vitest` and fixing the problem by switching `JSDOM` for Cloudflare's `htmlrewriter`.

I understand this may not be exactly the kind of solution you folks might like, so happy to adjust things.

Based on the CI matrix, I couldn't figure out how to run the unit tests, but they pass locally.

I've also provided a reproduction of the problem and more explanation here https://github.com/mfeckie/ember-test-app-vite

### DIFF
--- a/packages/vite/package.json
+++ b/packages/vite/package.json
@@ -32,7 +32,7 @@
     "debug": "^4.3.2",
     "fast-glob": "^3.3.2",
     "fs-extra": "^10.0.0",
-    "jsdom": "^25.0.0",
+    "htmlrewriter": "^0.0.12",
     "send": "^0.18.0",
     "source-map-url": "^0.4.1",
     "terser": "^5.7.0"
--- a/packages/vite/package.json
+++ b/packages/vite/package.json
@@ -18,7 +18,7 @@
     "vite": ">= 5.2.0"
   },
   "scripts": {
-    "test": "jest"
+    "test": "vitest"
   },
   "dependencies": {
     "@babel/core": "^7.22.9",
@@ -46,7 +46,8 @@
     "@types/send": "^0.17.4",
     "esbuild": "^0.25.0",
     "rollup": "^4.18.0",
-    "vite": "^6.2.0"
+    "vite": "^6.2.0",
+    "vitest": "^3.1.2"
   },
   "files": [
     "dist"
--- a/packages/vite/src/scripts.ts
+++ b/packages/vite/src/scripts.ts
@@ -1,6 +1,6 @@
 import type { Plugin } from 'vite';
 import type { EmittedFile } from 'rollup';
-import { JSDOM } from 'jsdom';
+import { HTMLRewriter } from 'htmlrewriter';
 import fs from 'fs-extra';
 const { readFileSync, readJSONSync, existsSync } = fs;
 import { dirname, posix, resolve } from 'path';
@@ -46,23 +46,22 @@ export function scripts(params?: { include?: string[]; exclude?: string[] }): Pl
       }
     },
 
-    transformIndexHtml(htmlIn, context) {
+    async transformIndexHtml(htmlIn, context) {
       // we don't do anything in `vite dev`, we only need to work in `vite
       // build`
       if (!context.server) {
-        return optimizer.transformHTML(htmlIn, config.base);
+        return await optimizer.transformHTML(htmlIn, config.base);
       }
     },
   };
 }
 
-class ScriptOptimizer {
+export class ScriptOptimizer {
   private emitted = new Map<string, string>();
   private transformState:
     | {
         htmlIn: string;
         htmlOut: string;
-        parsed: JSDOM;
       }
     | undefined;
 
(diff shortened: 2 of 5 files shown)


ef4:
Thanks for taking a stab at this fix.

I think we can address this a different way. If we run the `contentFor` plugin before the `scripts` plugin, the content-for will be gone before we try to parse the HTML.

I agree with the comments in the jsdom issue that jsdom is following the spec, and if htmlwriter fixes the bug then htmlwrite itself has a bug for not following the spec.

In a quick test, we can reorder the plugins by making `contentFor` have `enforce: 'pre'` and changing their relative order in `classic-ember-support.ts`. We should confirm if `scripts` really needs to be `pre` though.

mfeckie:
> I agree with the comments in the jsdom issue that jsdom is following the spec, and if htmlwriter fixes the bug then htmlwrite itself has a bug for not following the spec.

I'm not sure I agree here. htmlwriter isn't aiming to be a spec compliant DOM parser, it's a tool for rewriting file in a programmatic way.

> We should confirm if scripts really needs to be pre though.

Can you help me understand further what you mean here?

I'm happy to rework this PR, or put up a different one, not sure how to proceed now.

Techn1x:
@ef4 I gave your suggestion a quick test, it seems to work. The index.html file contains the meta and stylesheets in the head

```diff
export function classicEmberSupport() {
    return [
        hbs(),
+        contentFor(),
        scripts(),
        compatPrebuild(),
        assets(),
-        contentFor(),
```

```diff
export function contentFor() {
    return {
        name: 'embroider-content-for',
+        enforce: 'pre',
        transformIndexHtml(html, { path }) {
```

> We should confirm if scripts really needs to be pre though.

Also wasn't sure what you meant by this, so left it alone

Techn1x:
before (cleaned up some whitespace to make it readable)
```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <title>My App</title>
    <meta name="description" content="">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <link rel="preconnect" href="https://fonts.googleapis.com">
    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin="">
  </head>
  <body>
    <meta name="my-appconfig/environment" content="..." />

    <link integrity="" rel="stylesheet" href="/@embroider/virtual/app.css">

    <script type="module" crossorigin="" src="/assets/main-JjVMGTOU.js"></script>
    <link rel="modulepreload" crossorigin="" href="/assets/lodash-DYsN8HHh.js">
    <link rel="modulepreload" crossorigin="" href="/assets/contentful-C1v7W5-K.js">
    <link rel="modulepreload" crossorigin="" href="/assets/date-fns-CyyNkqhO.js">
    <link rel="modulepreload" crossorigin="" href="/assets/downloadjs-DHit6peV.js">
    <link rel="modulepreload" crossorigin="" href="/assets/posthog-js-UBRHEnfd.js">
    <link rel="modulepreload" crossorigin="" href="/assets/rollbar-DHqNRXQQ.js">
    <link rel="modulepreload" crossorigin="" href="/assets/highcharts-DJYf3eJu.js">
    <link rel="modulepreload" crossorigin="" href="/assets/highcharts/highcharts-more-DD4v6nKY.js">
    <link rel="modulepreload" crossorigin="" href="/assets/papaparse-YfDu9Pjw.js">
    <link rel="modulepreload" crossorigin="" href="/assets/nouislider-m3ZEbeDM.js">
    <link rel="modulepreload" crossorigin="" href="/assets/app-CPaEjFaA.js">
    <link rel="stylesheet" crossorigin="" href="/assets/app-4uSwbUPB.css">

    <script src="/@embroider/virtual/vendor.js"></script>
  </body>
</html>
```

after (cleaned up some whitespace to make it readable)
```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <title>My App</title>
    <meta name="description" content="">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <link rel="preconnect" href="https://fonts.googleapis.com">
    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin="">

    <meta name="my-app/config/environment" content="...">

    <link integrity="" rel="stylesheet" href="/@embroider/virtual/app.css">

    <script type="module" crossorigin="" src="/assets/main-JjVMGTOU.js"></script>
    <link rel="modulepreload" crossorigin="" href="/assets/lodash-DYsN8HHh.js">
    <link rel="modulepreload" crossorigin="" href="/assets/contentful-C1v7W5-K.js">
    <link rel="modulepreload" crossorigin="" href="/assets/date-fns-CyyNkqhO.js">
    <link rel="modulepreload" crossorigin="" href="/assets/downloadjs-DHit6peV.js">
    <link rel="modulepreload" crossorigin="" href="/assets/posthog-js-UBRHEnfd.js">
    <link rel="modulepreload" crossorigin="" href="/assets/rollbar-DHqNRXQQ.js">
    <link rel="modulepreload" crossorigin="" href="/assets/highcharts-DJYf3eJu.js">
    <link rel="modulepreload" crossorigin="" href="/assets/highcharts/highcharts-more-DD4v6nKY.js">
    <link rel="modulepreload" crossorigin="" href="/assets/papaparse-YfDu9Pjw.js">
    <link rel="modulepreload" crossorigin="" href="/assets/nouislider-m3ZEbeDM.js">
    <link rel="modulepreload" crossorigin="" href="/assets/app-CPaEjFaA.js">
    <link rel="stylesheet" crossorigin="" href="/assets/app-4uSwbUPB.css">
  </head>
  <body>
    <script src="/@embroider/virtual/vendor.js"></script>
  </body>
</html>
```

PR #2469, opened 2025-05-08
fix: ensure minify can be disabled

Techn1x:
Allows for users to disable `minify` whenever they wish. Currently if you set it to `false`, embroider/vite will turn it back on again.

### DIFF
--- a/packages/vite/src/ember.ts
+++ b/packages/vite/src/ember.ts
@@ -107,7 +107,7 @@ function minification(config: UserConfig, mode: string) {
    */
   config.build ||= {};
 
-  if (config.build.minify === undefined) {
+  if (config.build.minify === undefined || config.build.minify === true) {
     config.build.minify = 'terser';
   }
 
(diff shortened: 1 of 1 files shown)


packages/vite/src/ember.ts
```diff
@@ -107,7 +107,7 @@ function minification(config: UserConfig, mode: string) {
    */
   config.build ||= {};
 
-  if (!config.build.minify) {
+  if (config.build.minify === undefined || config.build.minify === true) {
     config.build.minify = 'terser';
```
Techn1x, line 111: Need to make sure that `minify: false` isn't overwritten with `'terser'`

The option technically allows boolean so I've catered for `minify: true` as well, to apply 'terser'
https://vite.dev/config/build-options#build-minify

Everything else will be handed off to vite (false, 'esbuild')

NullVoxPopuli (approved):

PR #2472, opened 2025-05-08
Ensure macroCondition expressions passed through babel-plugin-debug-macros are stripped in production builds

robbytx:
## Scenario

 * Vite application
 * Ember 4.12
 * Ember Data 4.13[.0-canary.8] - v4 Special Release w/vite support

## Problem

`pnpm build` produces an application that fails with: `Uncaught ReferenceError: getGlobalConfig is not defined`

This error comes from the following expression found at `node_modules/ember-data/node_modules/@ember-data/store/dist/-private-uEHANwvx.js`:
```js
macroCondition(getGlobalConfig().WarpDrive.deprecations.DEPRECATE_STORE_EXTENDS_EMBER_OBJECT)
```
where `macroCondition` and `getGlobalConfig` are imported from `@embroider/macros`.

When this file is processed through Babel, `isDebug: true` is currently set for `babel-plugin-debug-macros`, so macroCondition expressions are retained *incorrectly* in the bundled source code for evaluation at runtime.

## Solution

Utilize `NODE_ENV` (as opposed to `EMBER_ENV`) to determine whether debug macros should be retained in the transpiled source code, thereby mirroring the current behavior of [ember-cli-babel](https://github.com/emberjs/ember-cli-babel/blob/v8.2.0/lib/babel-options-util.js#L59).

## Additional Consideration

This change seems appropriate for Ember applications.  However, I'm concerned that the original behavior (of retaining macroCondition and other macros) may have been desirable/necessary when bundling addons.  

It appears that [tests/addon-template/babel.config.cjs] utilizes the `babelCompatSupport` function that is being modified by these changes, and I wonder if the logic needs to be more sophisticated to handle use of this function during addon packaging.

### DIFF
--- a/packages/compat/src/babel.ts
+++ b/packages/compat/src/babel.ts
@@ -68,6 +68,10 @@ export function babelMacros() {
 
 export function oldDebugMacros(): PluginItem[] {
   let debugMacros = require.resolve('babel-plugin-debug-macros');
+
+  const isProduction = process.env.NODE_ENV === 'production';
+  const isDebug = !isProduction;
+
   return [
     [
       debugMacros,
@@ -76,13 +80,13 @@ export function oldDebugMacros(): PluginItem[] {
           {
             source: '@glimmer/env',
             flags: {
-              DEBUG: true,
-              CI: false,
+              DEBUG: isDebug,
+              CI: !!process.env.CI,
             },
           },
         ],
         debugTools: {
-          isDebug: true,
+          isDebug,
           source: '@ember/debug',
           assertPredicateIndex: 1,
         },
@@ -99,7 +103,7 @@ export function oldDebugMacros(): PluginItem[] {
           module: '@ember/application/deprecations',
         },
         debugTools: {
-          isDebug: true,
+          isDebug,
           source: '@ember/application/deprecations',
           assertPredicateIndex: 1,
         },
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -2045,6 +2045,9 @@ importers:
       ember-data-4.12:
         specifier: npm:ember-data@~4.12.0
         version: /ember-data@4.12.8(@babel/core@7.27.1)(@ember/string@3.1.1)(ember-source@6.6.0-alpha.1)(webpack@5.99.8)
+      ember-data-4.13:
+        specifier: npm:ember-data@~4.13.0-alpha.8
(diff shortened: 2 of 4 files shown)


robbytx:
A couple other thoughts:

1. Should there be a more specific test to verify that macroConditions or other debug macros are stripped during production builds?  Why didn't passing `isDebug: true` yield failures in any of the existing test scenarios?

2. Why is this problem specific to Ember Data v4.13, when Ember Data 5.3 has the same line of code?  It seems like the difference is that `ember-data-5.3` and `ember-data-latest` are pinned to `npm:ember-data@5.3.0`, which is a [v1 addon](https://github.com/emberjs/data/blob/v5.3.0/packages/-ember-data/package.json#L97), while ember-data@5.3.4 and beyond are [v2 addons](https://github.com/emberjs/data/blob/v5.3.4/packages/-ember-data/package.json#L23), like v4.13.

For (2), it seems like we should at least update the test scenarios to verify compatibility with more recent Ember Data versions that are packaged as v2 addons.  In that case, I think the same bug would surface and we could eliminate the "lts_4_12_data_4_13" scenario, unless it is worth retaining in the support matrix.

Please let me know how you suggest we update the test scenarios, and I'll review the existing failures.

robbytx:
I did not see test failures locally, but that was apparently because I had output the scenario into a subdirectory of the embroider/tests/scenarios, and `@ember/test-waiters` was being resolved to version 3.1.0 from there.  

When I direct the output to a separate /tmp/folder, then I see the failure, so I will investigate.

robbytx:
https://github.com/embroider-build/embroider/pull/2473 updates Ember Data to 5.3.13 and resolves the underlying issue with unresolved `@ember/test-waiters` import.  

After this update, it is unclear why the `Uncaught ReferenceError: getGlobalConfig is not defined` is not an issue, as it seems to be in Ember Data 4.13.  In the production build output for `lts_4_12-static-app` (using ED 5.3), I do see code relating to deprecation notices ("The Store class extending from EmberObject is deprecated..."), but the surrounding `getGlobalConfig` reference was resolved to a function and minified, which must not be occurring with ED 4.13 for some reason.

So currently it appears that debug/deprecation code is being retained in the production output as expected in this ticket, which is obviously not ideal, but other factors are at play that cause the code to be a no-op rather than a ReferenceError.

robbytx:
Based on my updated understanding, I've replaced this PR with #2501.

robbytx:
@machty this issue should be resolved now with @embroider/compat 4.1.2

PR #2474, opened 2025-05-11
Allow declarations plugin to have its command customized

NullVoxPopuli:
Use cases:
- passing `--project` (e.g.: `--project ./config/tsconfig.publish.json`)
- using `tsc` (or any other declaration generating tool)

Default is unchanged (`glint --declaration)`

Demo'd here: https://github.com/NullVoxPopuli/nullui/pull/39/files


ef4:
This is fine as a low-level piece, the addon blueprint should eventually get cleaned up to show how to use this so people don't need to discover it. Specifically, we need to offer development-only tsconfig that is distinct from the published one.

PR #2477, opened 2025-05-13
Prevent excessive recursion of findActiveAddons

simonihmig:
Adding a dependency to an addon had the devastating effect of a massive increase in build times. Capturing a CPU profile showed this:

Bildschirmfoto 2025-05-13 um 10 23 22

We have a pretty complex dependency graph, and `console.log`'ing all calls to `findActiveAddons` was creating a log of 77M lines, and as the profile shows the recursive calls contributes to several minutes of "build" time (not really building anything though).

Not sure if we loose anything by cutting corners here?


ef4:
This looks good, let's also try to move the sorting out of findActiveAddons because it's not guarded by the cache.

NullVoxPopuli:
works on main https://github.com/embroider-build/embroider/pull/2479

NullVoxPopuli:
@simonihmig does this need rebasing?

ef4 (approved):

simonihmig:
> does this need rebasing?

No? This is targeting `stable`, and there hasn't been any changes happening on that branch...

Do we want to merge this despite the heavily red build?

PR #2480, opened 2025-05-14
Remove unneeded peerDependency declaration from @embroider/util

NullVoxPopuli:


void-mAlex:
we sure this would still work?
https://github.com/embroider-build/embroider/blob/main/packages/util/addon/ember-private-api.js#L11

NullVoxPopuli:
I'm mostly certain:
1. tests pass
2. if ember-source doesn't exist on disk between `@embroider/util` and the consuming app, it'll just work
   (when this isn't the case you get build errors today anyway)

mansona:
so tests are passing in a monorepo where we can't exactly be sure of the way things are being laid out (also our pnpm version is out of date so there are a lot of trust issues around the monorepo here)

secondly: I'm pretty sure we have had things in the past where dependencySatisfies relies on peers to be able to check the currently installed version of a dep 🤔 we need this to be able to check the right version of ember-source that it needs to check i.e. the one from the app, and the only way to be able to do that is to use a peer unless I'm mistaken

NullVoxPopuli:
> n peers to be able to check the currently installed version of a dep

right, but not from package.json -- the package.json declaration of peers don't matter at all.
the algo for resolving peers is the same tho

void-mAlex:
> > n peers to be able to check the currently installed version of a dep
> 
> right, but not from package.json -- the package.json declaration of peers don't matter at all. the algo for resolving peers is the same tho

pretty sure that we read package json when we build and use the resolvability rules

I asked if that would still work exactly because of fact that the declaration there makes things like I_HAVE_BAD_PEERS_AND_I_WANT_A_BROKEN_BUILD possible to check by embroider

NullVoxPopuli:
this PR is not needed, due to https://github.com/embroider-build/embroider/pull/2485

PR #2482, opened 2025-05-15
Propagate other plugin's custom metadata

ef4:
This silences the warning:

> [plugin commonjs--resolver] It appears a plugin has implemented a "resolveId" hook that uses "this.resolve" without forwarding the third "options" parameter of "resolveId". This is problematic as it can lead to wrong module resolutions especially for the node-resolve plugin and in certain cases cause early exit errors for the commonjs plugin.
In rare cases, this warning can appear if the same file is both imported and required from the same mixed ES/CommonJS module, in which case it can be ignored.


NullVoxPopuli (approved):

void-mAlex (approved):
:shipit:

PR #2492, opened 2025-05-23
Update porting-addons-to-v2.md

x-m-el:
explain how to support .gjs files

Babel asked to also add "@babel/plugin-transform-class-static-block". Unsure if this is also a requirement when using gjs.


NullVoxPopuli:
> Babel asked to also add "@babel/plugin-transform-class-static-block". Unsure if this is also a requirement when using gjs.

it's not -- odds are your targets or something is out of date, because static blocks have been in browsers for a while now!

it should not be part of the addon build

NullVoxPopuli (approved):
This re-arranging of the order of steps makes sense to me. thank you!

PR #2501, opened 2025-06-12
Ensure @ember/debug methods behave correctly in development vs production builds

robbytx:
## Problem 1

The change in #2323 was an incomplete translation of the patch it replaced, as it translates to `let currentDeprecate = noop;`, and `noop` is non-nullish so all calls to `deprecate` are passed to the noop function rather than the actual deprecation handling infrastructure.

### Solution

e9fa326 fixes the transpiler for `@ember/debug` applied to ember-source <5.12.

## Problem 2

With ember-source <5.10, calls to functions defined in the `@ember/debug` package (e.g. assert, deprecate, runInDebug) rely on the `DEBUG` flag to be processed by the [babel-plugin-debug-macros](https://github.com/ember-cli/babel-plugin-debug-macros) plugin so that the code is retained during development builds and stripped from production builds.  As identified in #2396, `vite build` is not stripping such code from production builds. 

In my tests in this PR, ember-source >= 5.12 appears to be unaffected by this problem because its build infrastructure (since https://github.com/emberjs/ember.js/pull/20675) substitutes the `isDevelopingApp()` function in place of the `DEBUG` flag, as opposed to relying on babel-plugin-debug-macros during the app build process.  However, it may still be the case that calls to `@ember/debug` functions were currently being retained in the production builds -- I have not investigated this.

### **Final** Solution

Described in https://github.com/embroider-build/embroider/pull/2501#issuecomment-2968900611 below.

---

The following observations were from the original PR description when the final solution was not yet apparent.

### (Attempted) Solution

c383228 attempts to fix this issue by utilizing the `NODE_ENV` variable as suggested in #2396.  

While this solution works correctly for both standard development (`pnpm start` => `vite`) and production (`pnpm build` => `vite build`), it fails when running tests (`pnpm test` => `vite build --mode test`) because the [Vite mode](https://vite.dev/guide/env-and-mode.html#modes) **test** has no effect on the `NODE_ENV` variable, which gets initialized separately to `'production'` as a result of using `vite build`.

Because of this, the development version of the static-app test scenarios are currently failing across all versions, since the current solution is incorrectly setting the `DEBUG` flag to false in this context.

### Observations related to a final solution

* The [compatPrebuild plugin](https://github.com/embroider-build/embroider/blob/e78fa39/packages/vite/src/build.ts#L66) uses the Vite-specific [`config` hook](https://vite.dev/guide/api-plugin.html#config) to obtain the Vite mode and pass value through to ember-cli as the environment to use in various scenarios, including resolving `isDevelopingApp` conditions.
  * This build emits its Babel configuration to `node_modules/.embroider/_babel_compat_.js`, which is subsequently [loaded](https://github.com/embroider-build/embroider/blob/b628891/packages/compat/src/babel.ts#L25) as part of Babel initialization during the second stage of the build process, ensuring that `isDevelopingApp` and other macros are resolved equivalently in both stages.
  * **Possibility:** The `oldDebugMacros` function to be modified is currently ignorant of `_babel_compat_.js`, but it could potentially be modified to receive the `babelMacros` configuration to determine whether `isDevelopingApp` considered true, and set the `DEBUG` flag accordingly.  It is unclear if such coupling is acceptable or undesirable.
* If the compatPrebuild plugin is NOT used (and therefore `_babel_compat_.js` is not available), then the [`buildMacros` function](https://github.com/embroider-build/embroider/blob/3db3825/packages/macros/src/babel.ts#L72) currently uses `NODE_ENV` to determine whether `isDevelopingApp` should be true.  it appears that `isDevelopingApp` will be treated as false with `vite build --mode test` for the same reason that my attempted solution fails.  
  * If this is a supported application mode, then the only solution appears to be setting `NODE_ENV` or some other alternate environment variable (e.g. [BABEL_ENV](https://babeljs.io/docs/options#envname)).  
  * **Possibility:** One clear solution would be to update `package.json` so that `pnpm test` sets it explicitly as in `NODE_ENV=development vite build --mode test`.  This would clearly resolve the issue, but it would require an update to the Embroider application blueprint, and all applications that have been created from it.
  * A more seamless solution would involve updating the @embroider/vite in some way such that the Vite mode is detected and made available as an alternate variable (EMBER_ENV?) that can inform the configuration of the babel-plugin-debug-macros plugin.
  * Another suggestion (see below) is to introduce `.env` files in the blueprint so that Vite mode drives a consistent `NODE_ENV`.


packages/compat/src/babel.ts
```diff
@@ -68,6 +68,10 @@ export function babelMacros() {
 
 export function oldDebugMacros(): PluginItem[] {
   let debugMacros = require.resolve('babel-plugin-debug-macros');
+
+  const isProduction = process.env.NODE_ENV === 'production';
```
robbytx: `process.env.NODE_ENV` is initialized to `development` with `vite` and to `production` with `vite build`, but the latter occurs regardless of Vite mode, so this condition is true even when running `vite build --mode test` (as in the current `pnpm test` script).

NullVoxPopuli: so we should add `NODE_ENV=development` to our `vite build --mode test` in the blueprint?

robbytx: That's definitely one possibility that I mentioned above.  I've also seen past suggestions to use `.env` files to align the Vite mode with `NODE_ENV`.

https://github.com/embroider-build/embroider/pull/2396#issuecomment-2752001494 said:
> we'll want to use NODE_ENV, and we should update the blueprints with .env files to help keep NODE_ENV aligned with vite mode.

Then https://github.com/mainmatter/ember-vite-codemod/issues/77 was created as a placeholder for this.

NullVoxPopuli: perfect! thank you <3

NullVoxPopuli:
thanks for working on this! this is exciting!
Looks like there are some failures tho

NullVoxPopuli:
> A more seamless solution would involve updating the @embroider/vite in some way such that the Vite mode is detected and made available as an alternate variable (EMBER_ENV?) that can inform the configuration of the babel-plugin-debug-macros plugin.

What would set EMBER_ENV?

since we invoke Vite directly, we have to play by its rules

packages/compat/src/babel.ts
NullVoxPopuli: setting NODE_ENV is what I've been doing in my projects

robbytx: So, I see a couple other possibilities now:

1. We could set `isDebug = '@embroider/macros'` instead, taking advantage of https://github.com/ember-cli/babel-plugin-debug-macros/pull/96 to translate it to `macroCondition(isDevelopingApp())` instead of a static boolean, and then we would need to move this plugin to execute *before* the macros-babel-plugin in order to translate that to the correct true/false value based on the received Vite mode.  
    * This works so long as the Vite mode is passed through to ember cli to generate `_babel_compat_.js`.  As mentioned in the PR description above, if this does not occur, then we're back to relying on the [buildMacros function](https://github.com/embroider-build/embroider/blob/3db3825/packages/macros/src/babel.ts#L72), which currently reads `NODE_ENV`. 😞 
2. Alternatively, we could update babel-plugin-debug-macros so that we can translate debug flags/calls to use `import.meta.env.MODE !== 'production'` instead.  Vite/Rollup apparently inlines this during the build process as I can observe in the output test javascript for the test I added in this PR.
    * This defers resolution of the mode until later in the build process, and resolves the immediate problem without requiring `NODE_ENV` alignment.
    * Fixing the aforementioned buildMacros condition isn't as simple, since it currently wants to make a configuration-time decision, but a similar strategy could be adopted even there.

My perspective is that the second solution is optimal for now, given the complexities in aligning NODE_ENV, but I'd like agreement before moving ahead with that, since it expands the scope of this change to another package/repo.

robbytx: After further reflection, I now feel like the first solution is the most compatible at the moment -- tying `{ DEBUG } from '@glimmer/env'` and methods from `@ember/debug` to the same condition as `isDevelopingApp()`.  This would be consistent with the behavior of [ember-source itself](https://github.com/emberjs/ember.js/blob/76b4a2e/rollup.config.mjs#L52).

Therefore, as long as `isDevelopingApp()` is correctly bound to the Vite mode, then the debug flags/calls will follow.

However, in attempting to implement this, it appears that there are some issues with ensuring that the macros plugin properly translates the macros condition generated by the debug tools plugin, since Babel apparently processes the file in AST order, rather than with a full pass through each plugin sequentially.

robbytx:
### Solution (for Problem 2)

(Mentioned as option 1 from https://github.com/embroider-build/embroider/pull/2501#discussion_r2143537583 above)

Because `NODE_ENV` does not currently align to the Vite mode (i.e. development/test/production), this PR instead changes the `babel-plugin-debug-macros` usages to translate the `DEBUG` flag and calls to `@ember/debug` methods to use the `isDevelopingApp()` condition via the plugin's existing functionality.  This aligns with how Ember itself [translates the @ember packages](https://github.com/emberjs/ember.js/blob/76b4a2e/rollup.config.mjs#L52) it ships within ember-source, so we will have consistency with Ember itself.

In order for the substituted `isDevelopingApp()` condition to be resolved statically, two additional changes were necessary:
1. It was necessary to reverse the order of Babel plugins so that the debug tools plugin runs before the macros plugin.
2. A functional change was required in https://github.com/ember-cli/babel-plugin-debug-macros/pull/103 so that the debug tools plugin will translate `assert(...);` and similar statements in place during the traversal, so that the subsequent macros plugin has the opportunity to replace the `isDevelopingApp()` condition in each statement with a boolean literal.

### Outstanding work

This solution resolves all known problems with `@ember/debug` processing so long as Embroider's compatBuild/Prebuild is used.

If the compatPrebuild plugin is NOT used (and therefore _babel_compat_.js is not available), then the [buildMacros function](https://github.com/embroider-build/embroider/blob/3db3825/packages/macros/src/babel.ts#L72) currently uses NODE_ENV to determine whether isDevelopingApp should be true, and this will be false with `vite build --mode test` for the same reason that my originally attempted solution failed.  I consider this a separate problem (with a much wider impact!) that will need to be addressed in due time.

robbytx:
Test failures:
 * core-resolver-test is failing because `isDevelopingApp()` was not translated to `true`. 😕
 * canary-minimal-app windows is failing with error: "Browser failed to connect within 120s. testem.js not loaded?" -- is it a canary problem?
 * lts_5_4-vite-app-basics windows is failing with error: "Global error: Uncaught Error: The `@glimmer/validator` library has been included twice in this application." -- unclear how this is related.
 * Other two failures were caused by npm download issue: `GET https://registry.npmjs.org/ember-source/-/ember-source-6.7.0-alpha.1.tgz error (ERR_PNPM_ENOENT). `

tests/scenarios/core-resolver-test.ts
```diff
@@ -845,12 +858,15 @@ Scenarios.fromProject(() => new Project())
 
           let switcherModule = expectAudit.module('./app.js').resolves('my-app/hello-world').toModule();
           switcherModule.codeEquals(`
-            import { macroCondition, getGlobalConfig, importSync } from '@embroider/macros';
+            import { macroCondition as macroCondition0, getGlobalConfig as getGlobalConfig0 } from "../node_modules/@embroider/compat/node_modules/@embroider/macros/src/addon/runtime";
+            import esc from "../node_modules/@embroider/compat/node_modules/@embroider/macros/src/addon/es-compat2";
+            import * as _importSync20 from "./fastboot";
+            import * as _importSync40 from "./browser";
```
robbytx: Promotion of importSync to the top-level outside of the macroCondition is apparently the current and expected beahvior, even though it can lead to problematic double imports like this.  (The test was previously asserting against the code produced by the resolver, without it passing through the @embroider/macros babel plugin.)

The [solution](https://discord.com/channels/480462759797063690/568935504288940056/1383089548870815804) per @ef4 is to use `await import(...)` in place of [importSync](https://github.com/embroider-build/embroider/blob/91b124a/packages/core/src/virtual-content.ts#L132).  That is an unrelated problem that already exists, and it makes sense to address it (and [other `importSync` usages](https://github.com/search?q=repo%3Aembroider-build%2Fembroider+importSync+path%3A%2F%5Epackages%5C%2Fcompat%5C%2Fsrc%5C%2F%2F&type=code)) separately.

robbytx: Per today's meeting, this test should not actually be transpiling the macros, so I'm backing this out.

pnpm-lock.yaml
```diff
@@ -4436,7 +4436,7 @@ packages:
       '@ember-data/private-build-infra': 4.12.8
       '@ember-data/store': 4.12.8(@babel/core@7.28.0)(@ember-data/graph@4.12.8)(@ember-data/json-api@4.12.8)(@ember-data/legacy-compat@4.12.8)(@ember-data/model@4.12.8)(@ember-data/tracking@4.12.8)(@ember/string@3.1.1)(ember-source@6.8.0-alpha.1)
       '@ember/string': 3.1.1
-      '@embroider/macros': 1.16.13
+      '@embroider/macros': 1.18.1(@glint/template@1.5.2)
```
robbytx: `pnpm install` wants to do this, no matter how I modify package.json.

mansona (approved):
LGTM 🎉

PR #2507, opened 2025-06-18
Remove auto-import from the app-template

NullVoxPopuli:
This is the repro that @ef4 asked for in https://github.com/embroider-build/ember-auto-import/pull/659


.github/workflows/ci.yml
```diff
@@ -24,8 +24,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup
-      - name: lint
-        run: pnpm lint
+      # - name: lint
```
NullVoxPopuli, line 27: temporary -- do not allow merge without uncommenting

PR #2520, opened 2025-07-08
Inspector support for Vite 

### LINKED ISSUE #2484: Vite compatibility for ember-inspector

BlueCutOfficial:
Fixes #2484

## Context

As part of the Ember Initiative, @mansona and I are trying to get ember-inspector to work for Vite apps. In the big lines, this support will imply that `ember.js` exposes a new Vite-compatible API for tools like the inspector, and the inspector should use this API instead of importing globally exposed things as it does now. However, Vite is available from Ember 3.28 to the latest, and we need the inspector to work for these versions. 

## Changes

We want to bring Vite compatibility to the pieces of the ecosystem that weren't designed for it, and this matches the job of `@embroider/compat`. This PR proposes adding a script to `@embroider/compat` that can be imported into an Ember+Vite app to get the inspector support. (This script adds up with changes required in ember-inspector itself.)

## How to test

- On Embroider side, serve the `app-template` on `http://localhost:4200/`
- On ember inspector side, clone the branch [`mainmatter:vite-support`](https://github.com/emberjs/ember-inspector/pull/2625)
- Follow the section of `ember-inspector`'s README about `pnpm serve:bookmarklet` to run the local inspector against `http://localhost:4200/`

👀 The inspector should load and display the component tree.


packages/compat/src/inspector-support.mts
```diff
@@ -0,0 +1,120 @@
+// This script exposes the modules used by Ember Inspector so an app that
+// builds with Vite can be inspected.
+
+// Disable TypeScript because this is a static script we want to import as is.
+// @ts-nocheck
+
+import { macroCondition, dependencySatisfies } from '@embroider/macros';
+
+globalThis.emberInspectorLoader = {
```
patricklx, line 9: this is probably not the final solution, but it any case:
there where some discussions in the past around reducing globals. see https://github.com/emberjs/rfcs/issues/781
which is basically to support multiple ember apps in one page. Not sure how relevant that still is. anyway, ember_debug is far away for supporting multiple apps on one page. but the new approach might bring us nearer. currently that only worjs with iframes. 

Also, there can be race conditions between page scripts loading and ember_debug scripts loading.
after this is setup, a `window.dispatchEvent(new Event("ember"))` is required. This is what is currently being used on ember_debug side.

BlueCutOfficial: Honnestly I don't think we will get rid of the global in this context. It's just a compatibility piece that we are working on at the moment, something that makes bridge between the old world and Vite.

> there can be race conditions between page scripts loading and ember_debug scripts loading.
after this is setup, a window.dispatchEvent(new Event("ember")) is required. This is what is currently being used on ember_debug side.

I don't understand what I should conclude from that. Who is calling `window.dispatchEvent(new Event("ember"))` in the classic world?

patricklx: What i mean is that ember_debug can load before the global emberInspectorLoader was setup. Which then leads to inspector not seeing the app.
The ember/index or now ember/barrel file was dispatching that event after it finished loading.

PR #2522, opened 2025-07-22
Add compat ember-inspector support

BlueCutOfficial:
(Edit from Chris) This PR has gone through a number of iterations, the current plan was developed during the Embroider Office hours on the 16th September 25

Essentially we are opting **not** to make the inspector support automatic, you have to import a function and run it on your app (you can see an example of this in the readme that we've added to the new `@embroider/legacy-inspector-support` package.

I've also added a basic implementation to try and provide types for the function 👍 I tested it in the `ts-app-template` and it seems to work fine


mansona (dismissed):
do not merge this until tests are running again. See https://github.com/embroider-build/embroider/pull/2562

PR #2525, opened 2025-07-22
App tree merging should be insensitive to file extensions

### LINKED ISSUE #2074: embroider resolver favours v2 addon template over app template for routes
given a v2 addon with a following appReeports [
 ... ,
'routes/**/*.js',
'templates/**/*.js',
]
with a route `./template/test.gjs` (hbs works as well because they both get compiles to js files)

and an app that has its own `./template/test.hbs`

you will get the addon template resolved instead of the app one even though the expectation is that the app should win

ef4:
Fixes #2074

### DIFF
--- a/packages/core/src/app-files.ts
+++ b/packages/core/src/app-files.ts
@@ -31,16 +31,45 @@ export class AppFiles {
     let otherAppFiles: string[] = [];
     this.perRoute = { children: new Map() };
 
-    let combinedFiles = new Set<string>();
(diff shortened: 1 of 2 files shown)


TSenter:
Will these changes be ported to `@embroider/core@^3`? We have some things that aren't ready for Vite yet that this would improve for us (namely the issue that [kdagnan reported](https://github.com/embroider-build/embroider/issues/2074#issuecomment-2406139660)).

PR #2539, opened 2025-08-14
POC: Provide explicit overloads for the specific uses of importSync mentioned in deprecation guides

johanrd:
Provide explicit overloads or the specific uses of [importSync mentioned in deprecation guides.](https://deprecations.emberjs.com/id/deprecate-import-get-owner-from-ember)

This would make every addon that aims for both supporting ember-source < 4.11 and ember-source >= 7.0 avoid the type gymnastics seen in e.g. https://github.com/NullVoxPopuli/ember-resources/blob/main/ember-resources/src/ember-compat.ts and https://github.com/nickschot/ember-mobile-menu/pull/1256/commits/ccc21e1b3cb897445ee204620d504e5e887a2456

If we are also willing/able to change the [compat example in the deprecation guides](https://deprecations.emberjs.com/id/deprecate-import-get-owner-from-ember/) to use assignment of getOwner through ternaries, there would be no typescript gymnastics needed at all, as the javascript and typescript solution would be exactly the same:

```js
import {
  macroCondition,
  dependencySatisfies,
  importSync
} from '@embroider/macros';

const getOwner = macroCondition(dependencySatisfies('ember-source', '>= 4.11'))
  ? importSync('@ember/owner').getOwner
  : importSync('@ember/application').getOwner; 
 ```

### DIFF
--- a/packages/macros/package.json
+++ b/packages/macros/package.json
@@ -49,6 +49,8 @@
     "@types/babel__generator": "^7.6.2",
     "@types/babel__template": "^7.4.0",
     "@types/babel__traverse": "^7.18.5",
+    "@types/ember__application": "^4.0.0",
+    "@types/ember__owner": "^4.0.0",
     "@types/lodash": "^4.14.170",
     "@types/node": "^22.9.3",
     "@types/resolve": "^1.20.0",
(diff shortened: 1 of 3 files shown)


johanrd:
Re-created here https://github.com/embroider-build/embroider/pull/2565 to fix merge mess

Issue #1358, opened 2023-02-14
Styles disappear in embroider production build when there are css layers in app.css


candunaj:
When you specify CSS layers at the top of `app.css` then imports below it disappear in the embroider production build. The development build works as expected (nothing is removed from app.css).

So when you specify the following `app.css` file

```css
/* app.css */
@layer base, components;
@import 'global.css' layer(base);

h1{
  color: red;
}
```

when you run `ember serve --environment=production` then you get the following `app.css` in the `dist/assets` directory:

```css
/* app.css */
@layer base, components;
h1{
  color: red;
}
```
> Import disappeared!

When you move defining layers under the imports then everything works as expected and nothing is removed from `app.css`

[I have put together a repo with the issue here.](https://github.com/candunaj/css-layer-embroider-bug)

NullVoxPopuli:
Are you using postcss/tailwind? If so, imports are only supported at the top of the file (non embroider issue).

See: https://tailwindcss.com/docs/using-with-preprocessors#using-post-css-as-your-preprocessor

mansona:
@NullVoxPopuli the reproduction is pretty vanilla so no it's not using any of those things

NullVoxPopuli:
I've just never seen layer and stuff without additional config / tools 🤷

ef4:
Discussed this at embroider office hours, the next action suggestion is to try reproducing this in a plain webpack project first to rule that out.  You can find the style configuration we're using here:

https://github.com/embroider-build/embroider/blob/4f583c9b87a6da95b59ecde0939d4177dfd15d3b/packages/webpack/src/ember-webpack.ts#L569

https://github.com/embroider-build/embroider/blob/4f583c9b87a6da95b59ecde0939d4177dfd15d3b/packages/webpack/src/ember-webpack.ts#L208

https://github.com/embroider-build/embroider/blob/4f583c9b87a6da95b59ecde0939d4177dfd15d3b/packages/webpack/src/ember-webpack.ts#L257-L260

candunaj:
Thank you, I will try to reproduce it in a plain webpack project.

candunaj:
## 1. setupStyleConfig is used only for imported CSS into a javascript file
I have found, that `stylePlugins` and `styleLoaders` are used only for CSS files imported in a javascript file. In our case, `app.css` is not imported to any javascript file, so `stylePlugins` and `styleLoaders` don't have any effect on `app.css`.
[...]

PR #2541, opened 2025-08-16
fix: replace `csso`  with `lightningcss`

aklkv:
fixes: #1358 

## Summary
Replaced the deprecated `csso` dependency with `lightningcss` in the Webpack packager.  
The `writeStyle` method now uses `lightningcss.transform({ minify: true })` for production builds, keeping behavior identical to before (pure minification only, no target-based transforms).

## Motivation
- **Media Queries Issue:** While investigating support for Media Queries Level 4 range syntax (e.g. `@media (height >= 480px)`), we found that `csso` does not handle these correctly, causing breakages in optimized builds.  
  `lightningcss` natively supports these newer constructs, resolving this problem.  
- **Tracking issue:** This directly addresses #1358, where developers reported CSS minification issues due to outdated tooling.  
- **Deprecation:** `csso` is effectively in maintenance mode, while `lightningcss` is modern, actively maintained, and significantly faster.

## Changes
- Removed `csso` and `@types/csso`  
- Added `lightningcss`  
- Updated `writeStyle` to:
  - Read CSS as a Buffer  
  - Minify with Lightning CSS  
  - Fingerprint and write the output  

## Benefits
- Resolves media query parsing/minification issues (e.g. range context syntax)  
- Modern, actively maintained minifier with better performance  
- Future-proof: supports draft CSS features and integrates with browserslist if needed

### DIFF
--- a/packages/webpack/src/ember-webpack.ts
+++ b/packages/webpack/src/ember-webpack.ts
@@ -365,6 +365,7 @@ const Webpack: PackagerConstructor<Options> = class Webpack implements Packager 
       code: cssBuffer,
       filename: cssPath,
       minify: true,
+      errorRecovery: true,
     });
     const minifiedCss = code.toString();
 


kategengler:
This would likely be a breaking change, which may be tricky as `stable` is a branch for embroider@3.x and and embroider@4.x is vite-based.

NullVoxPopuli:
how do you know it would be a breaking change?

packages/webpack/src/ember-webpack.ts
```diff
@@ -358,9 +358,16 @@ const Webpack: PackagerConstructor<Options> = class Webpack implements Packager
...
-    const cssContent = readFileSync(join(this.pathToVanillaApp, style), 'utf8');
-    const minifiedCss = csso.minify(cssContent).css;
+    const { transform } = await import('lightningcss');
+    const cssPath = join(this.pathToVanillaApp, style);
+    const cssBuffer = readFileSync(cssPath);
+    const { code } = transform({
+      code: cssBuffer,
+      filename: cssPath,
+      minify: true,
+      errorRecovery: true,
```
aklkv, line 368: this is inline how `csso` treats invalid syntax it just ignores it and drops the whole block from final minified output, a bit crazy I know

But given that error might be a little cryptic in embroider context (I ran into so while testing) and emberoidr/webpack is not longer suggested way it "ok" to fallback to bad behaviours. As a result I create `rollup-plugin-lightningcss-validator` 😆

mansona:
Closing for now, let me know if we want to reopen this discussion and I can reopen the PR 👍

PR #2550, opened 2025-09-05
new Macro: `appEmberSatisfies(range)`

NullVoxPopuli:
Why?:
- while trying to implement:
  https://github.com/tracked-tools/tracked-built-ins/pull/440
  I ran in to an issue where `dependencySatisfies` was always returning false (because the peerDep was not declared). We _can't_ declare the peerDep because then we force all consumers of `tracked-built-ins` to _forward_ the peer. We've already learned that we can't successfully manage a fully correct package.json-declared dep graph. It's been best to just _not_ declare peers, and rely on the resolve algo, as well as the assumption that the host app _must_ provide ember-source.

### DIFF
--- a/packages/macros/src/babel/app-ember-satisfies.ts
+++ b/packages/macros/src/babel/app-ember-satisfies.ts
@@ -14,7 +14,7 @@ export default function appEmberSatisfies(path: NodePath<t.CallExpression>, stat
   const [range] = path.node.arguments;
   if (range.type !== 'StringLiteral') {
     throw error(
-      assertArray(path.get('arguments'))[1],
+      assertArray(path.get('arguments'))[0],
       `the only argument to appEmberSatisfies must be a string literal`
     );
   }
--- a/packages/macros/src/babel/app-ember-satisfies.ts
+++ b/packages/macros/src/babel/app-ember-satisfies.ts
@@ -9,7 +9,7 @@ const packageName = 'ember-source';
 
 export default function appEmberSatisfies(path: NodePath<t.CallExpression>, state: State): boolean {
   if (path.node.arguments.length !== 1) {
-    throw error(path, `dependencySatisfies takes exactly one argument, you passed ${path.node.arguments.length}`);
+    throw error(path, `appEmberSatisfies takes exactly one argument, you passed ${path.node.arguments.length}`);
   }
   const [range] = path.node.arguments;
   if (range.type !== 'StringLiteral') {
--- a/packages/macros/src/babel/app-ember-satisfies.ts
+++ b/packages/macros/src/babel/app-ember-satisfies.ts
@@ -1,0 +1,40 @@
+import type { NodePath } from '@babel/traverse';
+import type { types as t } from '@babel/core';
+import type State from './state';
+import { satisfies } from 'semver';
+import error from './error';
+import { assertArray } from './evaluate-json';
+
+const packageName = 'ember-source';
+
+export default function appEmberSatisfies(path: NodePath<t.CallExpression>, state: State): boolean {
+  if (path.node.arguments.length !== 1) {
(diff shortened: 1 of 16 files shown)


packages/macros/src/babel/app-ember-satisfies.ts
```diff
@@ -0,0 +1,40 @@
...
+import type State from './state';
+import { satisfies } from 'semver';
+import error from './error';
+import { assertArray } from './evaluate-json';
+
+const packageName = 'ember-source';
+
+export default function appEmberSatisfies(path: NodePath<t.CallExpression>, state: State): boolean {
+  if (path.node.arguments.length !== 1) {
+    throw error(path, `dependencySatisfies takes exactly one argument, you passed ${path.node.arguments.length}`);
```
mansona:
```suggestion
    throw error(path, `appEmberSatisfies takes exactly one argument, you passed ${path.node.arguments.length}`);
```

packages/macros/src/babel/app-ember-satisfies.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+  const [range] = path.node.arguments;
+  if (range.type !== 'StringLiteral') {
+    throw error(
+      assertArray(path.get('arguments'))[1],
+      `the second argument to dependencySatisfies must be a string literal`
```
mansona:
```suggestion
      `the only argument to appEmberSatisfies must be a string literal`
```

packages/macros/src/babel/app-ember-satisfies.ts
```diff
@@ -0,0 +1,40 @@
...
+const packageName = 'ember-source';
+
+export default function appEmberSatisfies(path: NodePath<t.CallExpression>, state: State): boolean {
+  if (path.node.arguments.length !== 1) {
+    throw error(path, `dependencySatisfies takes exactly one argument, you passed ${path.node.arguments.length}`);
+  }
+  const [range] = path.node.arguments;
+  if (range.type !== 'StringLiteral') {
+    throw error(
+      assertArray(path.get('arguments'))[1],
```
mansona:
```suggestion
      assertArray(path.get('arguments'))[0],
```

Isn't there only one argument to this now? I don't know if this suggestion is correct 🤔

packages/macros/src/babel/app-ember-satisfies.ts
```diff
@@ -0,0 +1,40 @@
...
+      `the second argument to dependencySatisfies must be a string literal`
+    );
+  }
+  try {
+    let root = state.packageCache.get(state.packageCache.appRoot);
+
+    if (!root?.hasDependency(packageName)) {
```
mansona: I thought this was a copyPasta error 🍝 because the packageName is in module scope.. . it's only used in this try block so we could just make it a const in here 🤔

mansona (dismissed):
I noticed a few things when I had a quick look. I know it's draft but I thought I would share 😂

packages/macros/src/babel/app-ember-satisfies.ts
NullVoxPopuli: why spend the cost of re-defining a variable every function evaluation?
it makes more sense for consts to remain in module space

mansona: I mean that's premature optimisation right there, v8 will 100% inline it if it's a const 🤷

NullVoxPopuli: yea, i mean, the activation energy to change it is worth more than anything the computer would do.
(and what energy is spent discussing such changes).

Additionally to my prior point, I use consts like this to avoid typos in strings.
I _hate_ debugging string typos. So if I only write the word once, I can't have mismatches. :tada:

PR #2560, opened 2025-09-23
Update tests/app-template with latest @ember/app-blueprint

pichfl:
Part of https://github.com/ember-cli/ember-cli/issues/10814

---

This work is supported by the [Mainmatter Ember Initiative](https://mainmatter.com/ember-initiative/).

### DIFF
--- a/tests/app-template/package.json
+++ b/tests/app-template/package.json
@@ -50,7 +50,7 @@
     "babel-plugin-ember-template-compilation": "^3.0.0",
     "concurrently": "^9.2.1",
     "decorator-transforms": "^2.3.0",
-    "ember-auto-import": "^2.10.0",
+    "ember-auto-import": "^2.11.1",
     "ember-cli": "^6.6.0",
     "ember-cli-babel": "^8.2.0",
     "ember-cli-deprecation-workflow": "^3.4.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1524,8 +1524,8 @@ importers:
         specifier: ^2.3.0
         version: 2.3.0(@babel/core@7.28.4)
       ember-auto-import:
-        specifier: ^2.10.0
-        version: 2.10.1(@glint/template@1.5.2)(webpack@5.101.3)
+        specifier: ^2.11.1
+        version: 2.11.1(@glint/template@1.5.2)(webpack@5.101.3)
       ember-cli:
         specifier: ^6.6.0
         version: 6.7.0(@types/node@22.18.6)(babel-core@6.26.3)(handlebars@4.7.8)(underscore@1.13.7)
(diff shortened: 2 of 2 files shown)


tests/app-template/.github/workflows/ci.yml
```diff
@@ -0,0 +1,47 @@
...
+      - uses: actions/checkout@v3
+      - name: Install Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          cache: npm
+      - name: Install Dependencies
+        run: npm ci
+      - name: Run Tests
+        run: npm test
```
mansona: don't need this file since it will never be run

tests/app-template/app/app.js
```diff
@@ -2,7 +2,12 @@ import Application from '@ember/application';
 import compatModules from '@embroider/virtual/compat-modules';
 import Resolver from 'ember-resolver';
 import loadInitializers from 'ember-load-initializers';
-import config from './config/environment';
+import config from 'app-template/config/environment';
+import { importSync, isDevelopingApp, macroCondition } from '@embroider/macros';
+
+if (macroCondition(isDevelopingApp())) {
+  importSync('./deprecation-workflow');
```
mansona: the app-template is intended to be a minimal implementation, while we want it to be closer to the current blueprint we don't want it to be exact. We don't need to have deprecation-workflow in the setup at all, can you remove this and the macroCondition?

tests/app-template/app/deprecation-workflow.js
```diff
@@ -0,0 +1,24 @@
...
+     * and then run in the browser console:
+     *
+     *    deprecationWorkflow.flushDeprecations()
+     *
+     * And copy the handlers here
+     */
+    /* example: */
+    /* { handler: 'silence', matchId: 'template-action' }, */
+  ],
+});
```
mansona: we can delete this

tests/app-template/app/router.js
```diff
@@ -1,4 +1,4 @@
-import EmberRouter from '@ember/routing/router';
+import EmberRouter from '@embroider/router';
```
mansona: can we double check to see if there are any tests that add the embroider router to the template app 🤔 I'm wondering if we are actually testing both cases?

tests/app-template/app/templates/application.gjs
```diff
@@ -0,0 +1,12 @@
+import { pageTitle } from 'ember-page-title';
+import { WelcomePage } from 'ember-welcome-page';
+
+<template>
+  {{pageTitle "AppTemplate"}}
+
+  {{outlet}}
+
+  {{! The following component displays Ember's default welcome message. }}
+  <WelcomePage />
```
mansona: the app-template needs to be more minimal, an we remove pageTitle and Welcome page from this?

mansona: also can we check to see if there are any tests that are targeting application.hbs and they will probably need to be updated

tests/app-template/package.json
```diff
@@ -10,85 +10,80 @@
...
+    "@babel/eslint-parser": "^7.28.0",
+    "@ember/optional-features": "^2.2.0",
+    "@ember/string": "^4.0.1",
+    "@ember/test-helpers": "^5.2.2",
+    "@ember/test-waiters": "^4.1.1",
+    "@embroider/macros": "^1.18.1",
+    "@embroider/core": "^4.1.3",
+    "@embroider/vite": "^1.2.0",
+    "@embroider/compat": "^4.1.3",
+    "@embroider/router": "^3.0.3",
```
mansona: these will need to be converted back to `workspace:` deps or we will have a bad time 😂 also double check if there are any more that used to have it

tests/app-template/testem.cjs
```diff
@@ -3,6 +3,7 @@
 if (typeof module !== 'undefined') {
   module.exports = {
     test_page: 'tests/index.html?hidepassed',
+    cwd: 'dist',
```
mansona: I'm wondering if this is going to cause some issues 😬

PR #2561, opened 2025-09-24
`@embroider/compat`: Remove `ember-source` from dependencies

mkszepp:
Today `@embroider/compat` brings as dependency `"ember-source": "6.1.0-beta.1"` to apps, which looks incorrect

It was added here in this PR https://github.com/embroider-build/embroider/pull/2103/files#diff-ca4ac111fea0534801dbc2d0b872ca3ceeec61b41feb93a4b7187b6300f047f3R56


NullVoxPopuli (approved):
tests pass, so this is great!

PR #2566, opened 2025-09-30
Fix CI for upstream rolldown and webpack changes

ef4:
- a type change in webpack broke CI
 - latest rolldown is getting a failure that looks like timing of the embroider-vite-jump feature, this attempts to fix.

### DIFF
--- a/packages/vite/src/resolver.ts
+++ b/packages/vite/src/resolver.ts
@@ -12,7 +12,7 @@ import fs from 'fs-extra';
 import { createHash } from 'crypto';
 import { BackChannel } from './backchannel.js';
 
-const { ensureSymlinkSync, writeFileSync, mkdirpSync } = fs;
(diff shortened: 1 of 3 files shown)


mansona (approved):
Looks good to me!

packages/vite/src/resolver.ts
```diff
@@ -243,9 +243,16 @@ async function maybeCaptureNewOptimizedDep(
...
+
+  mkdirpSync(dirname(fromFile));
+  writeFileSync(
+    fromFile + '.tmp',
+    JSON.stringify({
+      name: 'jump-root',
+    }),
+    { flush: true }
+  );
+  renameSync(fromFile + '.tmp', fromFile);
```
void-mAlex, line 255: @mansona how does this change affect the upstream breakage?
I had to rollback patch this change as it broke all CI

PR #2568, opened 2025-09-30
Add explicit type overloads for the specific uses of `importSync` mentioned in deprecation guides  

johanrd:
We should provide explicit overloads for the specific uses of importSync that are mentioned in official guides, such as deprecation guides. See e.g. https://deprecations.emberjs.com/id/deprecate-import-get-owner-from-ember

This would make every addon that aims for both supporting ember-source < 4.11 and ember-source >= 7.0 avoid the type gymnastics seen in e.g. https://github.com/NullVoxPopuli/ember-resources/blob/main/ember-resources/src/ember-compat.ts and https://github.com/nickschot/ember-mobile-menu/pull/1256/commits/ccc21e1b3cb897445ee204620d504e5e887a2456

If we are also willing/able to change the [compat example in the deprecation guides](https://deprecations.emberjs.com/id/deprecate-import-get-owner-from-ember/) to use assignment of getOwner through ternaries, there would be no typescript gymnastics needed at all, as the javascript and typescript solution would be exactly the same:

```js
import {
  macroCondition,
  dependencySatisfies,
  importSync
} from '@embroider/macros';

const getOwner = macroCondition(dependencySatisfies('ember-source', '>= 4.11'))
  ? importSync('@ember/owner').getOwner
  : importSync('@ember/application').getOwner;
  ```

### DIFF
--- a/packages/macros/package.json
+++ b/packages/macros/package.json
@@ -50,6 +50,8 @@
     "@types/babel__generator": "^7.6.2",
     "@types/babel__template": "^7.4.0",
     "@types/babel__traverse": "^7.18.5",
+    "@types/ember__application": "^4.0.0",
+    "@types/ember__owner": "^4.0.0",
     "@types/lodash": "^4.14.170",
     "@types/node": "^22.9.3",
     "@types/resolve": "^1.20.0",
--- a/packages/macros/src/index.ts
+++ b/packages/macros/src/index.ts
@@ -35,6 +35,16 @@ export function each<T>(array: T[]): T[] {
 // We would prefer to write:
 //   export function importSync<T extends string>(specifier: T): typeof import(T) {
 // but TS doesn't seem to support that at present.
+//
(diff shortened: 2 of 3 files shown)


johanrd:
@ef4 okay, thanks, yes, maybe. How can we use the native types in a good way here, and still achieve the same thing?

ef4:
I think the type gymnastics are better left in addons. There are opinions involved. For example, the `ember-mobile-menu` example you linked works differently than this PR. It chooses to always take the type from one of the implementations, instead of union-ing them together. As types evolve going forward, that can matter.

Perhaps we should instead PR the deprecation guide so that it shows addon authors who are using typescript exactly what to do. Any time you're using `importSync` in typescript it's pretty much expected that you'll need to provide an explicit type, and having special-cases for two cases still leaves that cliff for people to fall off. It's probably best to always teach the explicit type form.

Issue #2570, opened 2025-10-01
Problem with running projects that uses node < 16 after the latest release


ixxvivxxi:
After #2550 I can't start any legacy project and get this error:
```
Build Error (broccoli-persistent-filter:Babel > [Babel: @embroider/macros]) in @embroider/macros/runtime.js

[BABEL]: Cannot find module 'node:url'
Require stack:
- /home/anton/projects/ams-java-api/booking-app/rerum/node_modules/ember-flatpickr/node_modules/@embroider/macros/src/babel/app-ember-satisfies.js
- /home/anton/projects/ams-java-api/booking-app/rerum/node_modules/ember-flatpickr/node_modules/@embroider/macros/src/babel/evaluate-json.js
- /home/anton/projects/ams-java-api/booking-app/rerum/node_modules/ember-flatpickr/node_modules/@embroider/macros/src/babel/get-config.js
- /home/anton/projects/ams-java-api/booking-app/rerum/node_modules/ember-flatpickr/node_modules/@embroider/macros/src/babel/macros-babel-plugin.js
- /home/anton/projects/ams-java-api/booking-app/rerum/node_modules/@babel/core/lib/config/files/module-types.js
- /home/anton/projects/ams-java-api/booking-app/rerum/node_modules/@babel/core/lib/config/files/configuration.js
- /home/anton/projects/ams-java-api/booking-app/rerum/node_modules/@babel/core/lib/config/files/index.js
- /home/anton/projects/ams-java-api/booking-app/rerum/node_modules/@babel/core/lib/index.js
- /home/anton/projects/ams-java-api/booking-app/rerum/node_modules/@ember/test-helpers/node_modules/broccoli-babel-transpiler/lib/worker.js (While processing: /home/anton/projects/ams-java-api/booking-app/rerum/node_modules/ember-flatpickr/node_modules/@embroider/macros/src/babel/macros-babel-plugin.js)Error: [BABEL]: Cannot find module 'node:url'
```

if I downgrade ember-flatpickr version, another library shows this error

node: 12
ember-cli: 3.16
ember: 3.16

NullVoxPopuli:
We only test against node 20+

seems like it's time to upgrade! <3 

please join us in the the community discord if you want help / guidance with the upgrade https://discord.gg/emberjs

NullVoxPopuli:
[...]

PR #2571, opened 2025-10-01
Drop `node:` prefix on imports in app-ember-satisfies to accomodate node < 16

NullVoxPopuli:
Fixes: #2570

The PR that added appEmberSatisfies broke folks using super old node versions.
- https://github.com/embroider-build/embroider/pull/2550

The features that the PR is using:
- `node:url`
- `node:module`
- `node:path`

However, 
- https://nodejs.org/api/esm.html#esm_node_imports -- the `node:` prefix has been supported since node @ 16 -- and that's not old enough -- as `@embroider/macros` declares support for node 12+

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,7 @@ jobs:
           node-version: 22 # this is only used for the pnpm install, the user-node-version above will be used in tests
           cache: pnpm
       - run: pnpm install --no-lockfile
+      - run: pnpm node -v
       - run: pnpm --filter "@embroider/macros" test
 
   preflight:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,15 +48,16 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - name: run test
-        run: echo "auto-install-peers=false\nuse-node-version=${{matrix.node}}" > .npmrc
       - uses: pnpm/action-setup@v4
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
NullVoxPopuli, line 1: Turns out we don't prettier ignore this

packages/macros/src/babel/app-ember-satisfies.ts
```diff
@@ -4,9 +4,9 @@ import type State from './state';
 import { satisfies, coerce } from 'semver';
 import error from './error';
 import { assertArray } from './evaluate-json';
-import { pathToFileURL } from 'node:url';
```
NullVoxPopuli, line 7: So simple. But ugh

.github/workflows/ci.yml
```diff
@@ -22,8 +22,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: ['20', '22', '24']
-        os: [ubuntu-latest, 'macos-latest'] # TODO add windows here
+        node: ["20", "22", "24"]
+        os: [ubuntu-latest, "macos-latest"] # TODO add windows here
```
driesdl, line 26: should ubuntu-latest be between quotes like macos-latest?

NullVoxPopuli: Idk, it appears to not matter somehow

driesdl (approved):

ef4 (approved):

PR #2576, opened 2025-10-02
don't break SSR with legacy-inspector-support

mansona:
- update `window` reference to `globalThis` since we're a module
- only call `globalThis.dispatchEvent` if the function is available

I have tested this with a patch locally on an SSR'd Ember app 👍

### DIFF
--- a/packages/legacy-inspector-support/src/common-implementation.js
+++ b/packages/legacy-inspector-support/src/common-implementation.js
@@ -9,5 +9,8 @@ export default function setupGlobal(app, importCallback) {
     loadCompatInspector: importCallback,
   });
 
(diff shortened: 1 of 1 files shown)


packages/legacy-inspector-support/src/common-implementation.js
```diff
@@ -9,5 +9,8 @@ export default function setupGlobal(app, importCallback) {
     loadCompatInspector: importCallback,
   });
 
-  window.dispatchEvent(new Event('Ember'));
+  // Only dispatch an event if window.dispatchEvent is available i.e. we are not in SSR mode
```
NullVoxPopuli, line 12: we are not in an SSR environment that defined dispatchEvent :p

PR #2578, opened 2025-10-07
Add installation section to `@embroider/legacy-inspector-support` README.md

johanrd:
Added installation instructions for the package. (It took me a minute to understand that this wasn't installed in embroider by default: I had forgot there is a package split.)

(not sure if we should keep `ember install` or move to `pnpm intall`, but `ember install` seems more aligned with the other package readmes)


NullVoxPopuli (approved):
Tyty!

PR #2585, opened 2025-10-13
Split windows tests

mansona:
This means that we're not limited to 256 tests on both Windows and Linux combined 👍

### DIFF
--- a/test-packages/support/suite-setup-util.ts
+++ b/test-packages/support/suite-setup-util.ts
@@ -33,7 +33,7 @@ async function githubMatrix() {
 
   let include = [
     ...suites.map(s => ({
-      name: s.name,
+      name: `${s.name}${process.argv.includes('--windows') ? '-windows' : ''}`,
       command: s.command,
     })),
   ];


.github/workflows/ci.yml
```diff
@@ -77,16 +77,58 @@ jobs:
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup
+      - name: lint
+        run: pnpm lint
```
NullVoxPopuli, line 115: why run lint again?

.github/workflows/ci.yml
```diff
@@ -77,16 +77,58 @@ jobs:
...
+
+  preflight-windows:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup
```
NullVoxPopuli, line 113: if we don't run lint in these preflight workflows, do we even need to install?

can we just run node ./suite-setup-util? or did we use dependencies? 🤔

NullVoxPopuli (approved):
Image

good improvement

PR #2595, opened 2025-10-22
Add support for disabling watching and reusing the prebuild in the classicEmberSupport vite plugin

NullVoxPopuli:
Originally this was a separate PR from #2596, but the capabilities are closely related.

This adds two new options to the vite plugin: `classicEmberSupport(...)`

## `watch`

I need this because I (and a few people on my team) ran out of FS watchers.

Usage:
```js
// in vite config:
plugins: [
  classicEmberSupport({
    watch: false,
  })
]
```

## `reusePrebuild`

Similar to the template-tag codemod, and uses the same logic that vite uses for determining when prebuild should happen again. Also respects `--force`.

Usage
```js
// in vite config:
plugins: [
  classicEmberSupport({
    reusePrebuild: true,
  })
]
```

----------

When enabling reusePrebuild, watch is _always_ `false`.

While two booleans is normally 4 states, here we have 3:
- watch: true (default), reusePrebuild: false (default)
- watch: false (set by user), reusePrebuild: false (default)
- reusePrebuild: true (set by user), watch is hard-coded to false

### DIFF
--- a/packages/vite/README.md
+++ b/packages/vite/README.md
@@ -1,0 +1,72 @@
+# `@embroider/vite`
+
+
+## Install
+
+```bash
+npm add --save-dev @embroider/vite
+```
+
+## Usage
+
+```js
+// vite.config.js or vite.config.mjs
+import { defineConfig } from "vite";
+import { classicEmberSupport, ember } from "@embroider/vite";
+
+export default defineConfig({
+  plugins: [
+    classicEmberSupport(),
+    ember(),
+    
+    // ...
+  ],
+});
+```
+
+And then start your app with `vite` or `npm exec vite`
+
+## exports
+
+
+### `ember`
+
+
+The `ember()` plugin is responsible for:
+- gjs / gts support
+- resolving imports, addons, dealing with app-tree-merging
+- configuring 
+  - the dev server
+  - the `index.html` and `tests/index.html` as entrypoints (depending on build environment)
+  - production builds
+
+This plugin has no options.
+
+### `classicEmberSupport`
+
+The `classicEmberSupport()` plugin is need if you need (or have):
+
+- hbs and content-for support
+- the compatibility prebuild (running ember-cli / broccoli)
+- classic asset pipelines
+- any v1 addons (direct or indirect)
+  You can check which v1 addons you may have via [this tool](https://github.com/IgnaceMaes/ember-addon-v2-scanner)
+  ```bash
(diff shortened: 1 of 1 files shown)


packages/shared-internals/src/get-lockfile-hash.ts
NullVoxPopuli, line 1: This file is copied from vite

packages/shared-internals/src/index.ts
```diff
@@ -29,6 +29,7 @@ export {
   version as cacheBustingPluginVersion,
 } from './babel-plugin-cache-busting';
 export { locateEmbroiderWorkingDir } from './working-dir';
+export { getLockfileHash } from './get-lockfile-hash';
```
NullVoxPopuli, line 32: so we can use it in the `@embroider/vite` via `@embroider/core`s `export *` of shared-internals

packages/vite/README.md
```diff
@@ -0,0 +1,72 @@
...
+    ember(),
+    
+    // ...
+  ],
+});
+```
+
+And then start your app with `vite` or `npm exec vite`
+
+## exports
```
NullVoxPopuli, line 29: I will not be documenting all the (unknown?) exports in this PR -- but can in a followup -- the fewer things to talk about in this PR the higher the chance of merge, ya?

packages/shared-internals/src/get-lockfile-hash.ts
```diff
@@ -0,0 +1,139 @@
...
+  },
+].sort((_, { manager }) => {
+  return process.env.npm_config_user_agent?.startsWith(manager) ? 1 : -1;
+});
+const lockfilePaths = lockfileFormats.map(l => l.path);
+
+/**
+ * https://github.com/vitejs/vite/blob/29cdb390374689e4dec9017b21fefe88b6ce4203/packages/vite/src/node/utils.ts#L410
+ */
+function lookupFile(dir: string, fileNames: string[]): string | undefined {
```
NullVoxPopuli, line 73: @ef4 fwiw, I left these copied to _reduce_ maintenance.


This function, for example, is just `findUpSync` -- which we already have probably somewhere in the monorepo already -- but I think trying to implement this logic ourselves is a greater mistake than copying -- now... if we don't care about being the exact same as vite... then fine, we can trim this file down a ton --- the key thing that I don't want to maintain is the lockfile formats above

PR #2599, opened 2025-10-29
Fix runtime implementation of negated macroCondition

NullVoxPopuli:


packages/macros/tests/babel/macro-condition.test.ts
```diff
@@ -403,6 +403,21 @@ describe('macroCondition', function () {
         expect(code).not.toMatch(/alpha/);
       });
 
+      test('tolerates negation between IfStatement and macroCondition CallExpression (with isTesting))', () => {
+        let code = transform(`
+      import { macroCondition, isTesting } from '@embroider/macros';
+      export default function() {
+        if (!macroCondition(isTesting())) {
```
ijlee2: Is this syntax correct? I think we always had to use `macroCondition(!isTesting())`.

ijlee2: Got my answer in https://discord.com/channels/480462759797063690/1433113082623824045/1433131598861307982.

NullVoxPopuli: Ye, here is the pr that added the behavior 
https://github.com/embroider-build/embroider/pull/2590

PR #2600, opened 2025-11-03
Use Vite's resolved cacheDir to determine cache location

mogstad:
Replaces the hard coded path to the cache with the resolved directory for config. This will allow Storybook to function correctly.

There is probably an unwanted performace implication, as the file path will always needs to be normalized.

I'm unsure where to best add tests for this, any suggestions are appreciated


mogstad:
Was afraid that could happen, do you think we can skip the normalizing as we're no longer using join?

PR #2601, opened 2025-11-04
Support `content-for` with single quotes as well

bertdeblock:
~~Apparently, for whatever reason, our `index.html` file was using single quotes for `{{content-for}}`.~~
The blueprints (`app` and `addon`) used to use single quotes for these, and it's possible not everyone updated to using double quotes. This isn't flagged or transformed when running the Vite codemod, and you end up with a screen like this:

Screenshot from 2025-11-04 11-55-53

### DIFF
--- a/packages/vite/src/content-for.ts
+++ b/packages/vite/src/content-for.ts
@@ -15,6 +15,7 @@ export function contentFor(): Plugin {
       let contentsForConfig = config[path];
       for (const [contentType, htmlContent] of Object.entries(contentsForConfig)) {
         html = html.replace(`{{content-for "${contentType}"}}`, `${htmlContent}`);
+        html = html.replace(`{{content-for '${contentType}'}}`, `${htmlContent}`);
       }
       return html;
     },
--- a/tests/scenarios/compat-addon-classic-features-test.ts
+++ b/tests/scenarios/compat-addon-classic-features-test.ts
@@ -25,8 +25,12 @@ appScenarios
             switch (type) {
               case 'body':
                 return '<p>Content for body</p>';
+              case 'body-single-quotes':
+                return '<p>Content for body-single-quotes</p>';
               case 'custom':
                 return '<p>Content for custom</p>';
+              case 'custom-single-quotes':
+                return '<p>Content for custom-single-quotes</p>';
               default:
                 return '';
             }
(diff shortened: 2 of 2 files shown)


packages/vite/src/content-for.ts
```diff
@@ -14,7 +14,7 @@ export function contentFor(): Plugin {
       let config: any = readJSONSync(join(locateEmbroiderWorkingDir(process.cwd()), 'content-for.json'));
       let contentsForConfig = config[path];
       for (const [contentType, htmlContent] of Object.entries(contentsForConfig)) {
-        html = html.replace(`{{content-for "${contentType}"}}`, `${htmlContent}`);
+        html = html.replace(new RegExp(`{{content-for ['"]${contentType}['"]}}`), `${htmlContent}`);
```
void-mAlex:
```suggestion
        html = html.replace(new RegExp(`{{content-for ['"]${RegExp.escape(contentType)}['"]}}`), `${htmlContent}`);
```
otherwise you could get trolled depending on what your contentType is

simpler version I think would be to just do two replace calls

packages/vite/src/content-for.ts
```diff
@@ -14,7 +14,7 @@ export function contentFor(): Plugin {
       let config: any = readJSONSync(join(locateEmbroiderWorkingDir(process.cwd()), 'content-for.json'));
       let contentsForConfig = config[path];
       for (const [contentType, htmlContent] of Object.entries(contentsForConfig)) {
-        html = html.replace(new RegExp(`{{content-for ['"]${contentType}['"]}}`), `${htmlContent}`);
+        html = html.replace(new RegExp(`{{content-for ['"]${RegExp.escape(contentType)}['"]}}`), `${htmlContent}`);
```
NullVoxPopuli: escape is only supported on node 24

Image

NullVoxPopuli: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/escape

void-mAlex: think that's why I was suggesting two replaces instead of RegExp

bertdeblock: Updated to use two replaces, and also updated the corresponding test.

PR #2604, opened 2025-11-07
Drop the need for ember-auto-import dependency on app for @embroider/addon-shim

mansona:
This PR should be the last thing that we need for the blueprint to drop ember-auto-import 🎉 https://github.com/ember-cli/ember-app-blueprint/pull/133

I pushed the removal of ember-auto-import from the app-template as a separate commit so that we can see CI failing first, and prove that adding the new functionality actually fixes the problem.

@ef4 I know you're uncomfortable using `EMBROIDER_PREBUILD` to detect that we're in a Vite build so I'm wondering if you have an idea of how to make this more lazy 🤔 we can't rely on the ember-auto-import FakeLeader because we need to also cater for the case where there is literally no ember-auto-import dependency anywhere in the addon tree. 

Edit: we had a good signal that this change still correctly throws an error because we have a classic app that we're testing macros against. So that classic scenario needs to add ember-auto-import back for the test to pass 😂

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1910,10 +1910,10 @@ importers:
         version: ember-data@4.8.8(@babel/core@7.28.4)(@glimmer/tracking@1.1.2)(@glint/template@1.5.2)(ember-source@6.10.0-alpha.4(@glimmer/component@2.0.0)(rsvp@4.8.5))(webpack@5.101.3(esbuild@0.25.10))
       ember-data-5.3:
         specifier: npm:ember-data@~5.3.13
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -53,7 +53,8 @@
       "@types/eslint": "^8.37.0",
       "babel-plugin-module-resolver@5.0.1": "5.0.0",
       "browserslist": "^4.14.0",
-      "graceful-fs": "^4.0.0"
+      "graceful-fs": "^4.0.0",
+      "@embroider/addon-shim": "workspace:*"
```
ef4, line 57: How confident are we that this pattern works in pnpm? Because I tried this locally and still saw packages depending on a copy of addon-shim that came from the NPM registry, not the local one.

Also: I see why we would need this config to test this particular PR. But I don't think we should keep it after this fix is released. It's good that our tests are forced to interact with previously released versions, since that's more realistic.

ef4:
I think the description is now out of date and this doesn't contain an attempted fix, just the tests to show what's needed.

I think we can do the lazy strategy just like we did in ember-auto-import. The key is to pick an addon hook that ember-cli runs and embroider's prebuild does not, and wait until then to throw the assertion. 

One candidate would be `postprocessTree`. It will get called at broccoli-pipeline-construction time, so before anything actually tries to build. Embroider doesn't support it, so should never call it.

package.json
mansona: so this doesn't work unless you re-up your pnpm-lock at the same time, I assume it's just because pnpm isn't considering this particular type of override an input of it's pnpm-lock resolution cache-busting or something 🤷 

I'm happy to remove this right after we release when our dependencies can float to a version that fixes the problem, we won't be able to do that before we actually release

PR #2622, opened 2025-11-24
Protect against the compat build running multiple times in parallel

NullVoxPopuli:
when vite is invoked, `options` can be called multiple times.

This change prevents us calling `emberBuild` multiple times in those situations.

We use the `options` hook here because it's the earliest hook available to us, and we need the compat build stuff to happen before everything else.

### DIFF
--- a/packages/vite/src/build.ts
+++ b/packages/vite/src/build.ts
@@ -53,12 +53,11 @@ export function emberBuild(command: string, mode: string, resolvableExtensions: 
   });
 }
 
-let buildPromise: Promise<void> | undefined;
-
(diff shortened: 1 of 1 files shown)


packages/vite/src/build.ts
```diff
@@ -53,6 +53,8 @@ export function emberBuild(command: string, mode: string, resolvableExtensions:
   });
 }
 
+let buildPromise: Promise<void> | undefined;
```
ef4: Maybe this should be inside the scope of `compatPrebuild()`

PR #2624, opened 2025-11-25
Embroider router support for ember source less than 4.12

void-mAlex:
import getOwner from ember application for ember source pre 4.12

### DIFF
--- a/packages/router/src/index.ts
+++ b/packages/router/src/index.ts
@@ -5,12 +5,12 @@
 import EmberRouter from '@ember/routing/router';
 import { buildWaiter } from '@ember/test-waiters';
 import { isDestroying, isDestroyed } from '@ember/destroyable';
-import { macroCondition, getGlobalConfig, dependencySatisfies, importSync } from '@embroider/macros';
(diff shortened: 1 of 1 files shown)


packages/router/src/index.ts
```diff
@@ -2,12 +2,25 @@
...
 import EmberRouter from '@ember/routing/router';
 import { buildWaiter } from '@ember/test-waiters';
 import { isDestroying, isDestroyed } from '@ember/destroyable';
-import { macroCondition, getGlobalConfig } from '@embroider/macros';
+import { macroCondition, getGlobalConfig, dependencySatisfies, importSync } from '@embroider/macros';
 import type Resolver from 'ember-resolver';
+import { type getOwner as getOwenerType } from '@ember/owner';
+let getOwner: typeof getOwenerType;
+
+if (macroCondition(dependencySatisfies('ember-source', '>=4.12.0'))) {
```
mansona: doesn't this fail because ember-source isn't declared as a peer? I think what you need is `appEmberSatisfies()` from https://github.com/embroider-build/embroider/pull/2550 ... which looks like it's not documented 🫠

mansona (approved):
looks good

PR #2626, opened 2025-11-26
Fix appEmberSatisfies macro

ef4:
- You can't rely on require.resolve to resolve package names. Not every package guarantees a default entrypoint.
 - Doing require.resolve is wasted effort because we already found the ember-source package anyway.
 - The comment about Package#version being a range makes no sense. It's read from the same place that the later code here was trying to read it -- the package.json.


NullVoxPopuli (approved):
well that's way simpler than what I did 🥲

NullVoxPopuli:
ah, I was doing this:
```js
if (macroCondition(appEmberSatisfies('^6.8.0-beta.1 || >= 6.8.0'))) {
```
and must've gotten confused as to which version was what

PR #2628, opened 2025-11-26
App ember satisfies for dummy apps

void-mAlex:
dummy apps have no package.json to be read so can't be used

since it's a well known path we strip it to find the real package


BlueCutOfficial:
@void-mAlex apologies, I reran the CI on your PR because it was green and recent, and I just wanted to ensure that there is a general problem with `pnpm install --no-lockfile` that has nothing to do with the changes 🙏

PR #2639, opened 2025-12-03
Fix `getAppFiles` filters to prevent app.ts from ending up in compatModules

### LINKED ISSUE #2638: App files causing cycles in typescript
In [virtual-entrypoint.ts](https://github.com/embroider-build/embroider/blob/main/packages/core/src/virtual-entrypoint.ts#L317-L322), `getAppFiles` gets a list of all files in the application so they can be included in the `compatModules` import.

There are filters for known things like the `app.js` file itself, which we don’t want to end up in compat modules because it causes module cycles. But we’re only filtering `app.js`, and in a TypeScript app we have `app.ts` instead, which shows up in the app files, which causes module cycles.

We should make the filter more dynamic based on all resolvable extensions. See https://github.com/embroider-build/embroider/blob/main/packages/core/src/module-resolver-options.ts#L76-L83

BlueCutOfficial:
Closes #2638

### DIFF
--- a/packages/core/src/virtual-entrypoint.ts
+++ b/packages/core/src/virtual-entrypoint.ts
@@ -8,7 +8,7 @@ import walkSync from 'walk-sync';
 import type { V2AddonPackage } from '@embroider/shared-internals/src/package';
 import escapeRegExp from 'escape-string-regexp';
 import { optionsWithDefaults } from './options';
-import resolvableExtensions from './resolvable-extensions';
+import { resolvableExtensions } from './resolvable-extensions';
 
 export interface EntrypointResponse {
   type: 'entrypoint';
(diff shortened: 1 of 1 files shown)


packages/core/src/resolvable-extensions.ts
```diff
@@ -0,0 +1,8 @@
+export default function resolvableExtensions(): string[] {
```
NullVoxPopuli: Why default?

BlueCutOfficial: It's the only thing the file currently exports. Unless you already have something specific in mind, there's no apparent reason why it would export anything else. And the usage is internal, so we are not signing for a public API or anything. What advantage do you see in a named export apart from hypothetical flexibility that may or may not be used later?

NullVoxPopuli: oh I just mean that default exports lead to typos and harder to grep for things.

named exports make the authored name be more likely to be used everywhere in the project.


I reserved default exports only for things that would be called by a framework, which cannot know any named export generally (without additional conventions) -- and since we're not using a framework here -- named exports makes more sense

MrChocolatine: > default exports lead to typos and harder to grep for things.

Hallelujah.

PR #2640, opened 2025-12-03
Align the rollup versions to latest beta - fixes CI

BlueCutOfficial:
This PR fixes the CI that was crashing on `pnpm install --no-lockfile`. There is a new release of rolldown (`1.0.0-beta.53`) that addresses type-related things, and it triggers errors in Embroider's CI due to other versions explicitly installed at the same time.

This PR attempts to align the versions of rolldown to the latest beta.


ef4 (approved):
This looks good to me. We were only on the earlier beta because that's the last time we worked on rolldown support.

PR #2646, opened 2025-12-11
use a filter for transform in hbs and template-tag plugins instead of early-out return in the transform function

mansona:
This allows rolldown-vite to execute the filter in rustland before passing back to the JS environment. I don't know how much of a performance increase this might have but why not 🤷 

Edit: we also drop a dependency which is nice 😂


NullVoxPopuli (approved):
gotta go fast

PR #2651, opened 2025-12-24
Update addon-dev to support glint 1 and glint 2 and choose between them based on your available dependencies

NullVoxPopuli:
brb, need to ditch jest
- https://github.com/embroider-build/embroider/pull/2652


packages/addon-dev/jest.config.js
NullVoxPopuli, line 1: this was missed in https://github.com/embroider-build/embroider/pull/2652

driesdl (approved):

PR #2655, opened 2025-12-29
Add new ember-standard-module: `@ember/reactive`

NullVoxPopuli:
All my projects are on vite, so that's how I missed this.

Failures seem to represent what is going on in real world.
need to debug why the new module isn't accessible from classic build land.


urbany:
hi @NullVoxPopuli is this a fix for [this issue](https://github.com/tracked-tools/tracked-built-ins/issues/445)?

NullVoxPopuli:
It was an attempt, yes, but it's not quite there.

Right now the only workaround i know of for that tracked built ins issue is to move to vite

PR #2662, opened 2026-01-27
`setTesting()` from `@embroider/macros`

NullVoxPopuli:
Provides public API for managing the testing value for apps/libraries


packages/macros/src/addon/runtime.d.ts
NullVoxPopuli, line 1: this looks out of place, but we don't currently have a build for the runtime code.
we can add that and convert the runtime to TS, ofc
easy for me either way

NullVoxPopuli: when running `pnpm compile` without this file:
```
packages/macros/src/index.ts:62:28 - error TS7016: Could not find a declaration file for module './addon/runtime'. '<repo>/packages/macros/src/addon/runtime.js' implicitly has an 'any' type.

62 export { setTesting } from './addon/runtime';
                              ~~~~~~~~~~~~~~~~~


Found 1 error.
```

tests/app-template-minimal/package.json
```diff
@@ -18,7 +18,7 @@
   },
   "scripts": {
     "build": "vite build",
-    "build:test": "NODE_ENV=development vite build --mode test",
```
NullVoxPopuli, line 21: we stopped using --mode test ages ago.

doesn't have an effect on the setTesting behavior afaict -- but I don't even want to entertain support of `--mode test` rn 🙈

tests/scenarios/minimal-app-test.ts
```diff
@@ -23,6 +23,7 @@ minimalAppScenarios
     // because we're testing the development experience itself.
     app.linkDevDependency('testem', { baseDir: __dirname });
     app.linkDevDependency('@embroider/test-support', { baseDir: __dirname });
+    app.linkDevDependency('@ember/test-waiters', { baseDir: __dirname, resolveName: '@ember/test-waiters-4' });
```
NullVoxPopuli, line 26: test-waiters-3 triggers barrel file deprecation and made logs unreadable

tests/scenarios/minimal-app-test.ts
```diff
@@ -102,6 +103,41 @@ minimalAppScenarios
         },
       },
       tests: {
+        'debug-test.js': `
```
NullVoxPopuli, line 106: https://github.com/embroider-build/embroider/actions/runs/21770349571/job/62816545241?pr=2662

NullVoxPopuli: > [babel] /tmp/tmp-2399nVG17B6dowG9/src/config.js: setTesting(true) cannot change the testing state in compile-time mode. 
>
> The current global config has isTesting=false. 
> setTesting() calls are compiled away at build time, 
> so they cannot change the testing state. 
>
> If you need to change the testing state at runtime, 
> use runtime mode instead.


hmmmm

tests/scenarios/minimal-app-test.ts
```diff
@@ -194,10 +230,24 @@ minimalAppScenarios
 
       Qmodule('vite build', function () {
         test('run tests suite against vite build output', async function (assert) {
-          let result = await app.execute('pnpm vite build --mode test');
+          let result = await app.execute('pnpm cross-env NODE_ENV=development vite build --mode development');
           assert.equal(result.exitCode, 0, result.output);
           result = await app.execute('pnpm ember test --path dist --config-file ./testem.cjs');
           assert.equal(result.exitCode, 0, result.output);
+
```
NullVoxPopuli, line 237: Is there a(n existing) way to get assertions printed in the cli?

tests/scenarios/minimal-app-test.ts
```diff
@@ -194,10 +230,24 @@ minimalAppScenarios
 
       Qmodule('vite build', function () {
         test('run tests suite against vite build output', async function (assert) {
-          let result = await app.execute('pnpm vite build --mode test');
+          let result = await app.execute('pnpm cross-env NODE_ENV=development vite build --mode development');
```
NullVoxPopuli, line 233: Adding this made things pass. So that's good. The old command isn't usually what we want. 

It does make me wonder how we want to handle or document handling folks shipping tests in prod builds

tests/scenarios/minimal-app-test.ts
NullVoxPopuli: :shrug: build worked, and testem passes 
image

packages/macros/src/addon/runtime.d.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export function macroCondition(predicate: boolean): boolean;
+
+export function config(packageRoot: string): any;
+
+export function getGlobalConfig(): any;
+
+export function isTesting(): boolean;
+
+export function setTesting(isTesting: boolean): void;
```
ef4: This file should not exist. Nobody should be able to see typescript types for the runtime because the runtime should be invisible from outside the package.

packages/macros/src/babel/macros-babel-plugin.ts
```diff
@@ -107,6 +107,46 @@ export default function main(context: typeof Babel): unknown {
...
+            }
+
+            let arg = args[0];
+            let evaluator = new Evaluator({ state });
+            let result = evaluator.evaluate(arg);
+
+            if (!result.confident) {
+              throw error(
+                arg,
+                `setTesting() can only be called with a statically analyzable value in compile-time mode. The argument must be a literal boolean (true or false), not a variable or expression that can't be evaluated at build time.`
```
ef4: The message is a little too strict. It could suggest using a literal boolean *or* another macro. For example, `setTesting(getOwnConfig().something)` would be OK, and `result.confident` should already understand that.

packages/macros/.gitignore
```diff
@@ -7,3 +7,4 @@
 /tests/**/*.map
 !/src/vendor/*.js
 !/src/addon/*.js
+!/src/addon/runtime.d.ts
```
ef4: Delete

packages/macros/package.json
```diff
@@ -15,7 +15,10 @@
   "author": "Edward Faulkner",
   "main": "src/index.js",
   "exports": {
-    ".": "./src/index.js",
+    ".": {
+      "browser": "./src/addon/runtime.js",
+      "default": "./src/index.js"
+    },
```
ef4: We don't want this. The runtime isn't made to be a drop-in replacement for the default. Most it wouldn't do the right thing.

Something similar to this would be appropriate in a different PR that makes an entirely runtime version of macros, but it would have different content than our existing runtime.js. For example, it wouldn't have a bunch of `throw Oops`

packages/macros/src/addon/runtime.d.ts
NullVoxPopuli: this is for internal type checking

packages/macros/.gitignore
NullVoxPopuli: ok, then we get ts-ignores, coming up!

NullVoxPopuli: oh nvm, I just throw an Oops

packages/macros/src/addon/runtime.d.ts
NullVoxPopuli: done -- this wasn't needed once committing to setTesting not being actually accessible (for realsies) at runtime (as the other macros currently behave)

packages/macros/.gitignore
NullVoxPopuli: done

packages/macros/package.json
NullVoxPopuli: makes sense, thanks! -- reverted, oops-at-runtime now thrown

packages/macros/src/babel/macros-babel-plugin.ts
NullVoxPopuli: Updated, and added a test

PR #2666, opened 2026-02-03
upgrade release-plan

mansona:
This updates the release-plan package and also the github CI jobs to use our new shared actions

One thing to note with this PR: the new actions assume OIDC support 🤔 this means that we would need to configure OIDC for each of  the packages in the monorepo. 

Here is a checklist that we should do before we merge: 

- [x] @embroider/config-meta-loader
- [x] @embroider/reverse-exports
- [x] @embroider/util
- [x] @embroider/core
- [x] @embroider/addon-dev
- [x] @embroider/babel-loader-9
- [x] @embroider/hbs-loader
- [x] @embroider/shared-internals
- [x] @embroider/addon-shim
- [x] @embroider/broccoli-side-watch
- [x] @embroider/macros
- [x] @embroider/compat
- [x] @embroider/legacy-inspector-support
- [x] @embroider/router
- [x] @embroider/vite
- [x] @embroider/template-tag-codemod

That's a lot of packages 🙈 

Edit: turns out I had enough permissions to setup OIDC for each of these so I have now already done it all 👍

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -22,21 +22,20 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write
-      pull-requests: write
       id-token: write
       attestations: write
 
     steps:
-      - uses: actions/checkout@v4
-      - uses: ./.github/actions/setup
+      - uses: actions/checkout@v6
+      - uses: pnpm/action-setup@v4
+      - uses: actions/setup-node@v6
         with:
-          use_pinned_node: "true"
-      - uses: actions/setup-node@v4
-        with:
-          # This creates an .npmrc that reads the NODE_AUTH_TOKEN environment variable
+          node-version: 22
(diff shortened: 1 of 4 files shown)


.github/workflows/publish.yml
```diff
@@ -22,21 +22,20 @@ jobs:
...
+      - uses: actions/setup-node@v6
         with:
-          use_pinned_node: "true"
-      - uses: actions/setup-node@v4
-        with:
-          # This creates an .npmrc that reads the NODE_AUTH_TOKEN environment variable
+          node-version: 22
           registry-url: 'https://registry.npmjs.org'
+          cache: pnpm
+      - run: npm install -g npm@latest # ensure that the globally installed npm is new enough to support OIDC
```
NullVoxPopuli, line 36: 🙌 

all our tokens expire today, ya?

PR #2667, opened 2026-02-05
Fix the hbs filter

NullVoxPopuli:


packages/vite/src/hbs.ts
```diff
@@ -3,7 +3,12 @@ import type { PluginContext } from 'rollup';
 import type { Plugin } from 'vite';
 import { hbsToJS, templateOnlyComponentSource } from '@embroider/core';
 
-const hbsFilter = createFilter('**/*.hbs?([?]*)');
+const hbsPathFilter = createFilter('**/*.hbs');
+
+export function hbsFilter(id: string): boolean {
```
NullVoxPopuli, line 8: same technique as the gjsFilter

PR #2668, opened 2026-02-05
Rolldown + Rollup :: better filtering

NullVoxPopuli:
Ditches rollup/pluginutils and ignores rolldown/pluginutils, because neither has a way to explicitly match file extensions (this is kinda difficult in regex, and impossible with globs).

I had PR'd to rolldown/pluginutils for a utility specifically for extension matching, but they seem uninterested? (which is fine, but 🤷 , we still have a path forward) https://github.com/rolldown/rolldown/pull/8210

Adds a new utility that _builds_ a filter that maybe we can re-use for all our internal plugins.

Copied from https://github.com/sveltejs/vite-plugin-svelte/blob/170bacc73d95d268e3673a5ec339da187adb82e0/packages/vite-plugin-svelte/src/utils/id.js#L174

(thanks @dominikg! 🎉 )

Of note, Vite 5 (which we support) -- does not support the filter-object syntax like how this PR was trying to use .... so we have to have more complexity for supporting both :(    
maybe fine? I don't want to cause a breaking change atm

<details><summary>Old context back when I was exploring `*/pluginutils` libraries</summary>

Most of this diff is lockfile, because I had to swap `@rollup/pluginutils` for `@rolldown/pluginutils`

Docs: https://github.com/rolldown/rolldown/tree/main/packages/pluginutils

I learned about this new function from the rolldown discord.

So far:
- no existing utility is sufficient for matching just a file extension
  - I PR'd here: https://github.com/rolldown/rolldown/pull/8210
- we _may_ need a new feature in `filter: {}`? specifically for matching extensions?

</details>

Builds on: https://github.com/embroider-build/embroider/pull/2667

Supersedes: https://github.com/embroider-build/embroider/pull/2646

### DIFF
--- a/packages/vite/package.json
+++ b/packages/vite/package.json
@@ -47,7 +47,6 @@
     "@types/send": "^0.17.4",
     "esbuild": "^0.25.0",
     "html-minifier-terser": "^7.2.0",
-    "picomatch": "^4.0.3",
     "rolldown-vite": "^7.2.10",
     "rollup": "^4.53.3",
     "vite": "^7.2.6",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -902,7 +902,7 @@ importers:
         version: 3.1.2
       ember-cli:
         specifier: ^4.12.1
-        version: 4.12.3(@types/node@22.19.1)(babel-core@6.26.3)(ejs@3.1.10)(handlebars@4.7.8)(underscore@1.13.7)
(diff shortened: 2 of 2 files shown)


packages/vite/package.json
```diff
@@ -24,7 +24,6 @@
     "@babel/core": "^7.22.9",
     "@embroider/macros": "workspace:*",
     "@embroider/reverse-exports": "workspace:*",
-    "@rollup/pluginutils": "^5.1.0",
```
NullVoxPopuli, line 27: this package doesn't support the "match id ending with extension" use case

(in their defense, doing so with regex _is_ hard)

PR #2672, opened 2026-02-09
Disabling the gzip reporting, even on small projects showed a 12% speedup in build time.

NullVoxPopuli:
no behavioral difference.

this only affects output reporting after a build


NullVoxPopuli:
Going to close because I have a way more performant build setup over in my own blueprint generator:

https://github.com/NullVoxPopuli/ember.nvp

PR #2695, opened 2026-03-27
Conditional file writes for Ember modules

NullVoxPopuli:


kategengler (approved):
Does v3 of embroider need something similar to be compatible with v7?

mansona:
> Is ember-auto-import saving us there?

I'm almost certain that ember-auto-import is inert in Embroider v3 too 🤔 so this won't be the change that's relevant

It's likely that we just changed the architecture enough for Embroider v4 that these files aren't being requested so they aren't being generated 🤷 maybe we should confirm that it's really not breaking for Embroider v3?

Issue #2563, opened 2025-09-26
Add the ability to run template-tag-codemod on a classic addon and a v2 addon (old style)


mansona:
Since the [new v2 addon blueprint](https://github.com/ember-cli/ember-addon-blueprint) doesn't support HBS out of the box we need to give people a way to upgrade their existing addons to GJS/GTS before migrating to the new blueprint.

Currently the template-tag-codemod doesn't work on classic addons or the old-style v2 addon blueprint because of a number of problems it encounters and this issue is intended to be a list of all those problems so that we can fix them 👍 

## classic addons

- [ ] TBD

## old-style v2 addons

- [ ]  can can often get the "ember-auto-import is not insalled" error so solving that will make it easier
- [ ] it seems like there is a requirement for ember-source to be a dependency or devDependency for it to work correctly
- [ ] there is an error where @ember/optional-features is required and you also need a config (even though it doesn't make sense in a v2 addon)
- [ ] invokables defined in the v2 addon are not discovered so you need a custom resolver - maybe we could provide one that helps with this issue?

NullVoxPopuli:
>  there is an error where @ember/optional-features is required and you also need a config (even though it doesn't make sense in a v2 addon)

how did this come up? I don't recall running in to this 🤔 it's an ember-cli error, ya? and ember-cli isn't needed in the old-style v2 addons

>  it seems like there is a requirement for ember-source to be a dependency or devDependency for it to work correctly

can you expand more on this? 
the reqs here shouldn't be any different from `@ember/addon-blueprint`

>  invokables defined in the v2 addon are not discovered so you need a custom resolver - maybe we could provide one that helps with this issue?

can you expand on this? do you mean the appReexports plugin?

PR #2697, opened 2026-03-30
provide a resolver for v1 addons running template-tag-codemod

mansona:
I've recently pushed a bit further to understand what is necessary to get the template-tag-codemod to work on v1 addons while working with the ember-bootstrap maintainers

This PR adds a basic addon resolver that can be opted into by passing it as a command-line arg to the template tag codemod: 

```
pnpm dlx @embroider/template-tag-codemod convert --components addon/components/**/* --customResolver @embroider/template-tag-codemod/addon-resolver
```

This is not the only thing that is needed to successfully run the template-tag codemod on a v1 addon, I have updated https://github.com/embroider-build/embroider/issues/2563 to list other issues I had after solving the resolver problem

### DIFF
--- a/packages/template-tag-codemod/package.json
+++ b/packages/template-tag-codemod/package.json
@@ -13,7 +13,8 @@
   "author": "Edward Faulkner <edward@eaf4.com>",
   "exports": {
     ".": "./dist/src/index.js",
-    "./default-renaming": "./dist/src/default-renaming.js"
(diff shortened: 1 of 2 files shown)


mansona:
Just in case anyone is following along with this PR, I have extracted this into a package so people can use it by installing it locally: https://github.com/mansona/ember-template-tag-addon-resolver

I have added a basic testing system and technically it's only designed to work with v2 addons (old style) for now 😂

PR #2704, opened 2026-04-17
Add `@ember/reactive` to virtual deps

NullVoxPopuli:
This https://github.com/embroider-build/embroider/pull/2655
would have solved the problem for just this issue report in Discord:
```
(!) Unresolved dependencies
https://rollupjs.org/troubleshooting/#warning-treating-module-as-external-dependency
@ember/reactive/collections (imported by "src/components/listbox.gts")
```
but... originally I wanted to support renamed-modules in the dependencies pluign -- but there isn't actually a guarante that ember-source will be available to look at its renamed-modules ... I didn't want to keep updating "static" lists, so supporting renamed-modules makes sense... but I think actually adding to the existing static list is the best solve for now, because we may not have an ember-source to read (it's external wholly anyway)


packages/addon-dev/tests/dependencies.test.ts
```diff
@@ -0,0 +1,85 @@
...
+  process.chdir(dir);
+
+  try {
+    const bundle = await rollup({
+      input: './src/index.js',
+      plugins: [dependencies()],
+      ...rollupOptions,
+      onLog(level, log, defaultLog) {
+        if (['warn'].includes(level)) {
+          expect(log).toBe("we don't want warnings");
```
NullVoxPopuli, line 32: the key thing is that this errors when warnings are logged

PR #2709, opened 2026-04-24
remove pin on babel-plugin-ember-template-compilation

mansona:
The pin was added in https://github.com/embroider-build/embroider/pull/2426 which was intended to be a type-only change but it limits anyone upgrading to babel-plugin-ember-template-compilation 2.4.0 when using Embroider@v3, and this was the version that we made it legal to use things like `Array` in your templates.

Without this change people can't get https://github.com/emberjs/babel-plugin-ember-template-compilation/releases/tag/v2.4.0-babel-plugin-ember-template-compilation and thus can't use https://rfcs.emberjs.com/id/1070-default-globals-for-strict-mode/ in Embroider@v3 applications

Note: I added the `any` type to the one places where the types didn't match and I reasoned that this was a simple change to unblock a legacy support branch 👍 Sure it would be nice to track down the difference but the blast radius is likely not worth the effort

### DIFF
--- a/packages/compat/package.json
+++ b/packages/compat/package.json
@@ -40,7 +40,7 @@
     "@types/yargs": "^17.0.3",
     "assert-never": "^1.1.0",
     "babel-import-util": "^2.0.0",
-    "babel-plugin-ember-template-compilation": "~2.3.0",
+    "babel-plugin-ember-template-compilation": "^2.3.0",
     "babel-plugin-syntax-dynamic-import": "^6.18.0",
     "babylon": "^6.18.0",
     "bind-decorator": "^1.0.11",
--- a/packages/compat/src/resolver-transform.ts
+++ b/packages/compat/src/resolver-transform.ts
@@ -209,7 +209,7 @@ class TemplateResolver implements ASTPlugin {
       case 'component':
       case 'modifier':
       case 'helper': {
-        let name = this.env.meta.jsutils.bindImport(resolution.specifier, resolution.importedName, parentPath, {
+        let name = this.env.meta.jsutils.bindImport(resolution.specifier, resolution.importedName, parentPath as any, {
           nameHint: resolution.nameHint,
         });
         setter(parentPath.node, this.env.syntax.builders.path(name));
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -29,7 +29,7 @@
     "@embroider/macros": "^1.18.1",
     "@embroider/shared-internals": "workspace:*",
     "assert-never": "^1.2.1",
-    "babel-plugin-ember-template-compilation": "~2.3.0",
+    "babel-plugin-ember-template-compilation": "^2.3.0",
     "broccoli-node-api": "^1.7.0",
     "broccoli-persistent-filter": "^3.1.2",
     "broccoli-plugin": "^4.0.7",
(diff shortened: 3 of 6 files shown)


packages/compat/src/resolver-transform.ts
```diff
@@ -209,7 +209,7 @@ class TemplateResolver implements ASTPlugin {
       case 'component':
       case 'modifier':
       case 'helper': {
-        let name = this.env.meta.jsutils.bindImport(resolution.specifier, resolution.importedName, parentPath, {
+        let name = this.env.meta.jsutils.bindImport(resolution.specifier, resolution.importedName, parentPath as any, {
```
NullVoxPopuli, line 212: Why?

packages/compat/package.json
```diff
@@ -40,7 +40,7 @@
     "@types/yargs": "^17.0.3",
     "assert-never": "^1.1.0",
     "babel-import-util": "^2.0.0",
-    "babel-plugin-ember-template-compilation": "~2.3.0",
+    "babel-plugin-ember-template-compilation": "^2.3.0",
```
NullVoxPopuli, line 43: This is too old, ya? Should be v4

packages/core/package.json
```diff
@@ -29,7 +29,7 @@
     "@embroider/macros": "^1.18.1",
     "@embroider/shared-internals": "workspace:*",
     "assert-never": "^1.2.1",
-    "babel-plugin-ember-template-compilation": "~2.3.0",
```
NullVoxPopuli, line 32: All of these should be v4?

packages/compat/src/resolver-transform.ts
mansona: Because without it we get an error like this: https://github.com/embroider-build/embroider/actions/runs/24887341716/job/72870656087#step:3:112 and the original pin was put in place because of "internal type errors" so I figured fixing it like this on the stable branch was the right approach https://github.com/embroider-build/embroider/pull/2426

packages/compat/package.json
mansona: I very strongly disagree with this. This PR is targeting the `stable` branch and we should not be bumping dependencies by a major unless absolutely necessary. In fact bumping the major here would be a breaking change which is impossible becuase the next major version is taken already 🤷

packages/compat/src/resolver-transform.ts
mansona: Also I described this in the description of the PR, did you read it? 🤔 

> Note: I added the any type to the one places where the types didn't match and I reasoned that this was a simple change to unblock a legacy support branch 👍 Sure it would be nice to track down the difference but the blast radius is likely not worth the effort

packages/compat/package.json
void-mAlex: it's not impossible we'd just have to do two jumps
3->5 and 4->6
but I'd rather just get the carret and not have to do that

mansona: I mean that's a real stretch of what people expect from semver 🙈 I know technically you are correct but to think that @embroider/vite is compatible with @embroider/compat 4 and 6, but not 5 🙈 that will be more confusing than we would ever want

If this PR were blocked on doing that strategy I would be advocating that this feature isn't supported in Embroider@3 (which is a hard sell too)

PR #2710, opened 2026-04-24
fix CI for stable branch

mansona:
~I know this has no effect on CI because we run with no lock file, but I wanted to open a PR to stable to test if CI is broken or if #2709 has broken something~

It looks like CI was broken on `stable` so this PR is to attempt to fix it. I've opted to skip problematic failing type checks for simplicity here, but I wonder if this is suggesting that typescript apps won't be able to update to the current release of ember-source without getting errors? 🤔

### DIFF
--- a/tests/scenarios/router-test.ts
+++ b/tests/scenarios/router-test.ts
@@ -231,19 +231,13 @@ routerApp.forEachScenario(scenario => {
       app = await scenario.prepare();
     });
 
-    test(`type checks`, async function (assert) {
-      let result = await app.execute('pnpm tsc');
-      assert.equal(result.exitCode, 0, result.output);
-    });
-  });
-});
-
-routerApp.forEachScenario(scenario => {
-  Qmodule(scenario.name, function (hooks) {
-    let app: PreparedApp;
-    hooks.before(async () => {
-      app = await scenario.prepare();
-    });
+    // type checks no longer pass on ember-release
+    if (scenario.name !== 'release-router') {
+      test(`type checks`, async function (assert) {
+        let result = await app.execute('pnpm tsc');
+        assert.equal(result.exitCode, 0, result.output);
+      });
+    }
 
     test(`CLASSIC pnpm test:ember`, async function (assert) {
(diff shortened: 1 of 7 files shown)


tests/scenarios/router-test.ts
```diff
@@ -231,19 +231,13 @@ routerApp.forEachScenario(scenario => {
...
-  });
-});
-
-routerApp.forEachScenario(scenario => {
-  Qmodule(scenario.name, function (hooks) {
-    let app: PreparedApp;
-    hooks.before(async () => {
-      app = await scenario.prepare();
-    });
+    // type checks no longer pass on ember-release
```
NullVoxPopuli, line 234: Why not?

This comment isn't helpful in code, but is moreso for the pr description above the explanation

mansona: > Why not?

I have no idea 🤷 and I don't really have the time to dig into this 😞 Here is the failing test if you want to take a look: https://github.com/embroider-build/embroider/actions/runs/24888194247/job/72873491346#step:5:292

I'm assuming you are also in the same camp of not wanting to spend too much time fixing types for the **legacy** Embroider version? 

> This comment isn't helpful in code, but is moreso for the pr description above the explanation

I put this here because it explains why we are skipping it only on release, maybe I should say `type checks no longer pass on ember-release and we aren't currently investing the time to fix them - people should upgrade to Embroider@v4`

Do you have a suggestion of what I should do here?

NullVoxPopuli: that error occurs when the repos deps are messed up -- multiple copies of `@glimmer/*` are present in the dep graph, and older TSes discover all of them.

This isn't really related to embroider, but just fixing the dep graph.

Non-blocking (because who knows what caused your dep graph to be incorrect (tho, tbh, a lot of folks that encounter this just pin/override the glimmer deps to match what ember-source expects)), but that's the situation.

Thanks for the link! that was helpful <3

NullVoxPopuli (approved):

PR #2717, opened 2026-04-30
fix(template-tag-codemod): this becomes controller on component invokation

BlueCutOfficial:
**Context**: While working with template-tag-codemod to convert a codebase to template tags, I came across a case where the controller has a getter returning a component class. In the template, `this.theComponent` was correctly replaced in conditional blocks, but not on the invocation.

**This PR**: In the plugin that transforms `this` into `@controller`, adds a handler for `ElementNode` so it catches and transforms an expression like `<this.myComponent>`.


mansona (approved):
I'm happy with this change, the situation is simple and the test is perfect 🎉 nice work

PR #2718, opened 2026-04-30
fix(ts-app-template): use scripts for a Vite app

BlueCutOfficial:
**Context**: `ts-app-template` is the template of a TypeScript Ember app building with Vite for the scenario tester. However, the scripts in the `package.json` correspond to a classic app, so when you output a scenario to do some manual testing, the scripts don't work out of the box.

**This PR**: Changes the scripts in `ts-app-template` to correspond to an Embroider+Vite app.

### DIFF
--- a/tests/ts-app-template/package.json
+++ b/tests/ts-app-template/package.json
@@ -15,7 +15,7 @@
     "./*": "./app/*"
   },
   "scripts": {
-    "build": "ember build --environment=production",
+    "build": "vite build",
     "lint": "concurrently \"npm:lint:*(!fix)\" --names \"lint:\"",
     "lint:css": "stylelint \"**/*.css\"",
     "lint:css:fix": "concurrently \"npm:lint:css -- --fix\"",
@@ -24,9 +24,9 @@
     "lint:hbs:fix": "ember-template-lint . --fix",
     "lint:js": "eslint . --cache",
     "lint:js:fix": "eslint . --fix",
-    "start": "ember serve",
-    "test": "concurrently \"npm:lint\" \"npm:test:*\" --names \"lint,test:\"",
-    "test:ember": "ember test"
+    "start": "vite",
+    "test": "vite build --mode test && ember test --path dist",
+    "test:ember": "vite build --mode test && ember test --path dist"
   },
   "devDependencies": {
     "@babel/core": "^7.22.20",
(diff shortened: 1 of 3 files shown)


NullVoxPopuli (approved):

NullVoxPopuli (dismissed):
oops

tests/ts-app-template/package.json
```diff
@@ -24,9 +24,9 @@
     "lint:hbs:fix": "ember-template-lint . --fix",
     "lint:js": "eslint . --cache",
     "lint:js:fix": "eslint . --fix",
-    "start": "ember serve",
-    "test": "concurrently \"npm:lint\" \"npm:test:*\" --names \"lint,test:\"",
-    "test:ember": "ember test"
+    "start": "vite",
+    "test": "vite build --mode test && ember test --path dist",
+    "test:ember": "vite build --mode test && ember test --path dist"
```
NullVoxPopuli: these test commands should be have
```
NODE_ENV=development vite build --mode development
```
(or setting NODE_ENV in an .env.development file)

tests/ts-app-template/package.json
```diff
@@ -24,9 +24,9 @@
     "lint:hbs:fix": "ember-template-lint . --fix",
     "lint:js": "eslint . --cache",
     "lint:js:fix": "eslint . --fix",
-    "start": "ember serve",
-    "test": "concurrently \"npm:lint\" \"npm:test:*\" --names \"lint,test:\"",
-    "test:ember": "ember test"
+    "start": "vite",
+    "test": "vite build --mode test && ember test --path dist",
```
mansona: I am pretty sure I had to fix this incorrect mode in the recent CI change 🤔 we should also make sure that we have a correct .env.development file before merging

mansona (changes requested):

tests/ts-app-template/package.json
mansona: I think we need both development and the .env file 🤔

NullVoxPopuli: yes

BlueCutOfficial: I changed for what I see in the current app blueprint, does it look better?

mansona (approved):
thanks for this 🎉

tests/ts-app-template/package.json
mansona: perfect 🎉

PR #2719, opened 2026-05-01
Fix CI for ember-source 7

mansona:
ember-source looks to be shipping shared chunks now (probably because of the AMD module deprecation removal) so we just need to update the audit result to match 👍 

Edit: this PR started by just updating the jest suites... but it has decended into the madness of fixing Ember 7 on CI 😭 

I'm surprised this hasn't come up before now 🤔


tests/scenarios/scenarios.ts
```diff
@@ -99,6 +99,26 @@ function patchTestWaiters(project: Project) {
   project.addDevDependency(testHelpers);
 }
 
+function updateEmberQunit(project: Project) {
+  project.linkDevDependency('ember-qunit', { baseDir: __dirname, resolveName: 'ember-qunit-9' });
+
+  (project.files['tests'] as any)['test-helper.js'] = `import Application from 'app-template/app';
```
NullVoxPopuli: should we not make this default on disk, and use the project/scenario stuff to downgrade to less-than-default?

mansona: so we have a massive range that we're testing here, from Ember 3.28 all the way to 7 now. I think it's probably safer to get this merged as is (or close to that) since CI has not been running for over a month now and if we want to clean up we can do that a) from a green CI and b) without having to rush

mansona:
As far as I can see it, this PR now represents the **minimum number of changes necessary** to get CI passing. I think there could be a lot of improvements here (such as @NullVoxPopuli 's suggestion to just change the defaults in the template apps) but I would urge that we do not do that until **after** this is merged.

We are now in release week and any delay to fixing CI will make the bad situation of CI being broken for a month even worse 🙈

tests/scenarios/scenarios.ts
```diff
@@ -87,13 +88,101 @@ async function lts_6_12(project: Project) {
   });
 }
 
+export function patchTestWaiters(externalProject: Project) {
+  // this overrides the @ember/test-waiters dependency of @ember/test-helper and @embroider/router to make sure it is @ember/test-waiters@4
+  // both versions are in-range as a depedency, but for some reason scenario tester will always pick the lower one
```
NullVoxPopuli, line 93: I have a hunch that scenario-tester doesn't support `||`
e.g.:
```
    "@ember/test-waiters": "^3.0.2 || ^4.0.0",
```

NullVoxPopuli (approved):
makes sense

PR #2727, opened 2026-05-12
Update `send` package to fix security vulnerability

mkszepp:
Updating `send` to latest helps to fix a vulnerable in embroider apps with vite (see https://github.com/advisories/GHSA-m6fv-jmcg-4jfg)

Not sure if this change is breaking or not, because there is no node version in package.json defined...
`send` has dropped node < 18 with 1.x (see https://github.com/pillarjs/send/blob/master/HISTORY.md#100--2024-07-25)

grafik

As alternative, we can update to 0.19.x

Note: Vite 5 has already required Node.js 18+, so maybe its not breaking (embroider/vite supports `vite >= 5.2.0`

### DIFF
--- a/packages/vite/package.json
+++ b/packages/vite/package.json
@@ -44,7 +44,7 @@
     "fast-glob": "^3.3.2",
     "fs-extra": "^10.0.0",
     "jsdom": "^25.0.0",
-    "send": "^0.18.0",
+    "send": "^1.2.1",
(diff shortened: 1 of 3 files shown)


ef4:
Thanks for investigating the node support, I agree, this is non-breaking for us because our minimum vite is 5 and vite 5 already requires node 18+.

PR #2729, opened 2026-05-16
Reimplement @embroider/webpack on the modern (vite-style) architecture

NullVoxPopuli:
like how we have `@embroider/vite` set up, but with webpack's constraints

Tested myself locally
- boot
- watch mode

> [!WARNING]
> I don't know anything about webpack's APIs, other than them being quite unergonomic, and a pain to debug. Most of this PR was generated by claude. I suspect this PR has a lot more work left.


packages/webpack/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "include": [
+    "./src/**/*.ts"
+  ],
+  "extends": "../../tsconfig-base.json",
+  "compilerOptions": {
+    "composite": true,
```
NullVoxPopuli-ai-agent, line 7: Caveat I hit: this package had to become a **composite TS project emitting `dist/`** (mirroring `@embroider/vite`). The legacy layout compiled `.js` next to `.ts` with no tsconfig. The moment there's an `index.ts`, `tsc -b` emits `index.d.ts` next to it, which is then *also* matched as an input by the root `tsconfig.json` include glob -> `error TS5055: Cannot write file ... it would overwrite input file`, intermittently (only once stale artifacts exist). Composite + `outDir: dist`, plus referencing/excluding this project in the root tsconfig, was the fix.

packages/webpack/package.json
```diff
@@ -10,46 +10,34 @@
   },
   "license": "MIT",
   "author": "Edward Faulkner",
-  "main": "src/ember-webpack.js",
+  "main": "dist/src/index.js",
```
NullVoxPopuli-ai-agent, line 13: Consequence of the composite build: `main`/`types` point at `dist/src/*`. Bumped to `5.0.0` (from-scratch rewrite; legacy `Packager` removed). Dropped `thread-loader`/`csso`/`lodash`/`semver`/`supports-color`. **`thread-loader` removal is load-bearing**: see the virtual-module registry note on `webpack-resolver-plugin.ts` — loaders now always run in-process, which is what makes the resolver↔loader registry valid (like vite's in-memory `responseMetas`).

packages/webpack/src/compat-prebuild.ts
```diff
@@ -0,0 +1,59 @@
...
+import { join, dirname } from 'node:path';
+import type { Compiler } from 'webpack';
+
+// The webpack equivalent of vite's `compatPrebuild` plugin / `emberBuild`. It
+// forks `ember build` with EMBROIDER_PREBUILD=true, which (via compatBuild in
+// the app's ember-cli-build.js) runs embroider's stage1+stage2 and writes the
+// .embroider working directory (resolver.json, content-for.json, rewritten
+// packages) plus the prebuilt vendor/test-support assets into
+// tmp/compat-prebuild.
+let prebuildPromise: Promise<void> | undefined;
```
NullVoxPopuli-ai-agent, line 19: Problem: `resolver.json` / `content-for.json` only exist *after* the compat prebuild, but webpack plugins are applied *before* the build runs. Solution: a process-cached prebuild promise. The plugin taps `beforeRun`/`watchRun` only to sequence logs; the real ordering guarantee is that `ember()`'s async `entry` awaits this same cached promise before discovering anything.

packages/webpack/src/ember-webpack.ts
```diff
@@ -1,726 +1,248 @@
...
+  // set by classicEmberSupport(); ember() reads it to decide whether to run
+  // the compat prebuild + apply content-for (a fully-v2 app uses only ember()).
+  classic: boolean;
+}
 
-  private configureWebpack(appInfo: AppInfo, variant: Variant, variantIndex: number): Configuration {
-    const { entrypoints, publicAssetURL, packageName, resolverConfig } = appInfo;
+// classicEmberSupport() and ember() coordinate through one Shared object per
+// compiler, regardless of the order they appear in the plugins array.
+const sharedByCompiler = new WeakMap<Compiler, Shared>();
```
NullVoxPopuli-ai-agent, line 70: `classicEmberSupport()` and `ember()` need to coordinate (content-for flag, whether to run the prebuild) but webpack applies plugins independently and in arbitrary array order. Solution: a single `Shared` object per `Compiler` via a `WeakMap`; whichever plugin runs first creates it. This is the webpack stand-in for the way vite's plugins just share `process.cwd()`/module state.

packages/webpack/src/ember-webpack.ts
```diff
@@ -1,726 +1,248 @@
...
-              })
-            );
-          }
-          resolve(stats);
-        } catch (e) {
-          reject(e);
+      // embroider owns entry. We set it to an async function (webpack supports
+      // this) so the compat prebuild has produced resolver.json /
+      // content-for.json before we discover the html entrypoints.
+      opts.entry = (async () => {
```
NullVoxPopuli-ai-agent, line 159: `entry` is an async function (webpack supports this) that runs the prebuild and then discovers html entrypoints. I first tried adding entries dynamically with `EntryPlugin` in the `make` hook; that fails with **"No dependency factory available for dependency type: EntryDependency"** because only `EntryPlugin` registers that factory, at `thisCompilation`, which is too late when you call `compilation.addEntry` from `make`. The async `entry` fn avoids the whole problem.

packages/webpack/src/html-output-plugin.ts
```diff
@@ -0,0 +1,304 @@
...
+      } else if (src) {
+        let request = src.replace(/^\//, '');
+        if (RAW_VIRTUAL.has(request)) {
+          record.handled.push({ kind: 'raw-virtual', specifier: request, el: script, isStyle: false });
+        }
+        // non-virtual, non-module scripts with a src are left untouched
+      }
+    }
+
+    if (moduleScripts.length > 0) {
```
NullVoxPopuli-ai-agent, line 177: **The hardest bug, worth reading.** Originally every inline `<script type="module">` became its own webpack entrypoint. Webpack's module registry (`__webpack_require__.c`) is *per-runtime*, and separate entrypoints get separate runtimes. A singleton shared between a page's entrypoints — `@glimmer/validator` (older ember-source imports it from ~12 modules) or `qunit` — was then instantiated once per entrypoint -> runtime error "@glimmer/validator included twice" / "QUnit already defined". `pnpm build` (production) passed because it only has the single `index.html` entry; rollup/vite never hit it because a build has one module registry. Fix: **one webpack entry per html page**, with that page's module scripts as ordered `import`s — the first script tag receives the built assets, the rest are removed. This matches how rollup/vite/esbuild treat a document (one module graph per page). `optimization.runtimeChunk:'single'` also dedupes the registry but globally couples *different pages'* entrypoints and regressed `release` — rejected in favor of the per-page model, which is simply the correct structure.

packages/webpack/src/html-output-plugin.ts
```diff
@@ -0,0 +1,304 @@
...
+}
+
+function isAbsoluteURL(url: string) {
+  return /^(?:[a-z]+:)?\/\//i.test(url);
+}
+
+// Support the common eager forms used by ember app/test index.html:
+//   import.meta.glob("./**/*.{js,gjs,gts}", { eager: true });
+//   import.meta.glob('./**/*.js', { eager: true });
+function transformImportMetaGlob(code: string): string {
```
NullVoxPopuli-ai-agent, line 60: Caveat: `import.meta.glob` is a Vite/Rollup feature webpack has no equivalent for, and the app/test `index.html` use it to load tests. We transform only the **eager forms the templates actually use** (`import.meta.glob("...", { eager: true })`) into `require.context`. This is a targeted transform, *not* a general `import.meta.glob` implementation — lazy/option variants are unsupported. Also note the generated filter: the `require.context` must exclude our own generated entry file or it recurses into itself.

packages/webpack/src/html-output-plugin.ts
```diff
@@ -0,0 +1,304 @@
...
+import { join, dirname, posix } from 'path';
+import { createHash } from 'crypto';
+import { ResolverLoader, virtualContent } from '@embroider/core';
+import { applyContentFor } from './content-for';
+import { resolveVirtual } from './virtual-resolve';
+
+// These are the `@embroider/virtual/*` references that, in vite, are emitted
+// as raw assets by the resolver plugin's `buildEnd` hook (rather than being
+// bundled). We do the same here instead of pushing them through webpack.
+const RAW_VIRTUAL = new Set([
```
NullVoxPopuli-ai-agent, line 14: Caveat: only `vendor.js`/`vendor.css`/`test-support.js`/`test-support.css` are emitted *raw* (the webpack analog of vite's resolver `buildEnd` `emitVirtualFile`). `/@embroider/virtual/app.css` is deliberately **not** in this set — it isn't a module; it's a public-asset of the synthesized styles addon, emitted by `AssetsPlugin`, and the `<link>` just keeps pointing at it (exactly like vite). An earlier attempt to bundle `app.css` through a generated css entry was wrong and is why this list is explicit.

packages/webpack/src/webpack-resolver-plugin.ts
```diff
@@ -183,12 +205,16 @@ class WebpackRequestAdapter implements RequestAdapter<WebpackResolution> {
...
-    virtualFileName: string | undefined
+    virtual: VirtualResponse | false
   ): ExtendedResolveData {
     let specifier = request.specifier;
-    if (virtualFileName) {
+    if (virtual) {
+      // Hand the VirtualResponse across in-process and key the request only by
+      // the stable specifier, so the same logical virtual always produces the
+      // same webpack module (this is what rollup/vite get for free).
+      registerVirtualResponse(this.appRoot, virtual);
```
NullVoxPopuli-ai-agent, line 234: Adopted vite's **id-based** virtual-module model here. Webpack keys modules by the full request string, so encoding the (importer-dependent) `VirtualResponse` into the request made the *same logical* virtual (e.g. a fastboot switch) produce multiple webpack modules. Now the request carries only the stable `virtual.specifier`; the `VirtualResponse` is handed across **in-process** via a registry in `virtual-loader.ts`, exactly like vite's `responseMetas` + `load(id)`. This is only safe because `thread-loader` was dropped (loaders run in the same process as this plugin).

packages/webpack/src/webpack-resolver-plugin.ts
```diff
@@ -1,25 +1,35 @@
...
 
 export class EmbroiderPlugin {
-  #resolver: EmbroiderResolver;
+  #resolverLoader: ResolverLoader;
   #babelLoaderPrefix: string;
   #appRoot: string;
 
-  constructor(opts: EmbroiderResolverOptions, babelLoaderPrefix: string) {
-    this.#resolver = new EmbroiderResolver(opts);
+  // Note: the resolver config (resolver.json) is read lazily, the first time
```
NullVoxPopuli-ai-agent, line 21: The embroider resolver is constructed lazily (first `resolve`), not in the constructor: `resolver.json` is written by the compat prebuild *during* the build, after this plugin has been `apply`'d.

packages/webpack/src/virtual-resolve.ts
```diff
@@ -0,0 +1,63 @@
...
+  type Resolver,
+  type VirtualResponse,
+} from '@embroider/core';
+import { join } from 'path';
+
+type Captured = Resolution<VirtualResponse | undefined, Error>;
+
+// A tiny RequestAdapter used to ask the core resolver "if the app imported
+// this specifier, what virtual response would you produce?". This is the
+// webpack analog of vite's `ensureVirtualResolve`, used so we can emit the
```
NullVoxPopuli-ai-agent, line 15: This tiny `RequestAdapter` exists because webpack has no `buildEnd`-style "emit these virtual files" hook like vite's resolver plugin. It asks the core resolver "what `VirtualResponse` would this specifier produce" so we can emit vendor/test-support as real assets in `processAssets` — the webpack-shaped equivalent of vite's `emitVirtualFile`.

packages/webpack/src/content-for.ts
```diff
@@ -0,0 +1,28 @@
...
+import { locateEmbroiderWorkingDir } from '@embroider/core';
+
+// The webpack equivalent of vite's `embroider-content-for` plugin. The compat
+// prebuild writes a content-for.json keyed by origin-absolute html path (e.g.
+// "/index.html", "/tests/index.html"). Each value maps a content-for slot name
+// to the html that ember-cli would have injected there classically.
+export function applyContentFor(html: string, htmlPath: string, appRoot: string): string {
+  let key = htmlPath.startsWith('/') ? htmlPath : '/' + htmlPath;
+  let configPath = join(locateEmbroiderWorkingDir(appRoot), 'content-for.json');
+  if (!existsSync(configPath)) {
```
NullVoxPopuli-ai-agent, line 13: Tolerate a missing `content-for.json`: a fully-v2 app using only `ember()` (no `classicEmberSupport()`) never runs the compat prebuild, so there's nothing to substitute. Without this guard it threw `ENOENT` for that valid configuration.

NullVoxPopuli-ai-agent (commented):
Context: I'm the agent that wrote this rewrite. Posting the non-obvious caveats/problems and how they were solved — most are inline on the relevant lines; this summary covers the cross-cutting bits.

### Dead-ends investigated for the "@glimmer/validator included twice" CI failure (≤ lts_5_4)
Tried and rejected (all reverted; none are in the final diff):
- `resolve.symlinks: false` — embroider-conventional, but no effect here.
- `cleanUrl(fromFile)` in the resolver adapter to mirror rollup's importer query-stripping — plausible (the fastboot-switch specifier carries `?names=…`), but no effect; reverted.
- `optimization.runtimeChunk: 'single'` — dedupes the registry across entrypoints, but didn't fix the older versions *and* regressed `release` with "QUnit already defined" (it globally couples unrelated pages' entrypoints). Reverted.

The actual root cause and fix are documented inline on `html-output-plugin.ts` (one webpack entry per html page). The `@glimmer/validator.js?` in the stack traces was a red herring — webpack's loader query in the sourceURL, not a second module; a module dump proved there was always exactly one webpack module. The duplication was across per-entrypoint *runtimes*.

### Limitations / not covered (intentional scope)
- **Fastboot variant not implemented.** `experiments.topLevelAwait` is kept to leave room, but there's no fastboot scenario or fastboot-switch runtime wiring beyond what the core resolver provides.
- **No dev-server** equivalent to `vite dev` — this is `compatBuild(app, buildOnce)` + `webpack build` only.
- **Only the standard `app-template` wiring** is provided/validated; `ts-app-template`, `app-template-minimal`, `ts-app-template-classic` are not converted to a webpack variant.
- The `import.meta.glob`→`require.context` transform only handles the eager forms the templates use.

### Unrelated CI noise
`lts_6_12-typescript-app` and `release-compat-template-colocation-…-windows` failed on Chrome/dbus infra flakes in scenarios that don't use `@embroider/webpack` — not caused by this PR.

### Architecture note
Throughout, the guiding principle was "do what `@embroider/vite` does": two config-mutating plugins (`classicEmberSupport()` + `ember()`), id-based virtual modules, raw-emit of vendor/test-support, public-assets for `app.css`. Where webpack's model differs from rollup's (per-runtime module registry, no HTML/`import.meta.glob` handling, no `buildEnd` emit, request-string module identity) is exactly where the inline comments are.

tests/scenarios/webpack-app-test.ts
```diff
@@ -0,0 +1,123 @@
...
+
+// The modern @embroider/webpack mirrors @embroider/vite: the app keeps its real
+// index.html / tests/index.html, the compat prebuild produces the .embroider
+// working directory, and webpack does the bundling using @embroider/core's
+// Resolver + virtual content. This scenario swaps the default vite wiring of
+// the app-template for the webpack wiring and exercises a real build.
+// Expanded across the same Ember matrix as `vite-app-basics`
+// (wideAppScenarios / fullSupportMatrix), producing
+// `<emberVersion>-webpack-app-basics`.
+wideAppScenarios
```
NullVoxPopuli-ai-agent: Uses `wideAppScenarios` (`fullSupportMatrix`) so `webpack-app-basics` expands across the **exact same Ember matrix as `vite-app-basics`** (lts_3_28 … canary), each becoming its own CI job. Validation caveat: locally I only ran `lts_3_28` + `lts_5_4` + `release` × (`pnpm build` prod and `pnpm test:ember` dev+ember-test) — `lts_3_28`/`lts_5_4` were the worst CI failures, so they're the highest-signal; the full matrix is delegated to CI.

packages/webpack/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@embroider/webpack",
-  "version": "4.1.0",
+  "version": "5.0.0",
```
mansona, line 3: bad claude, don't change this because it's using release-plan

PR #2735, opened 2026-05-25
compile `@embroider/core` to es2022 for native #private fields

davidtaylorhq:
ModuleRequest is constructed once per resolveId call (30k+ in a large build) and its fields are #private. At the current es2019 target, TypeScript downlevels #private to WeakMap-backed helpers (__classPrivateFieldGet/Set), so every ModuleRequest field access pays a WeakMap lookup. These shims show up when profiling the performance of Embroider's `resolveId` hook.

In a ~7s Discourse build we can attribute roughly 300ms to this. So it's small, but not nothing.

Bumping `@embroider/core`'s target to es2022 emits native #private fields and drops the helper entirely.

These features require dropping node 12 and 14 support. This PR goes further and bumps node to `">= 20.19.*"` for parity with `@embroider/compat`. Dropped node versions have all been EOL for more than a year.


packages/core/tsconfig.json
mansona, line 6: I feel like this sort of change really needs to happen in the base config 🤔 what are your thoughts?

I know there are some packages that need wider node support, but maybe we can bump the default and then override those other packages 🤔

mansona:
@NullVoxPopuli we should probably discuss this very carefully before we add breaking to it 🤔 If i remember correctly @ef4 said that core is only ever used as a direct dependency so it safe to bump as necessary, but I just want to be careful when we do any breaking changes in the Embroider monorepo

packages/core/tsconfig.json
davidtaylorhq: It's certainly possible, but it would be a much larger diff, touching a lot more packages. Here's a sketch for comparison: https://github.com/embroider-build/embroider/pull/2736

FWIW, template-tag-codemod already has an override on the ts build target (es2021 in that case).

NullVoxPopuli:
@mansona our main constraint is macros and its dependencies: https://github.com/embroider-build/embroider/pull/2736/changes#r3298904461

anything else is fair game

or are you saying that it is maybe not a breaking change because the engines.node is actually out of date and this PR might just be updating to what we actually support? 🤔

ef4:
Not just macros. shared-internals also needs extended node support (it gets used by ember-auto-import).

packages/core/tsconfig.json
mansona: I was just going to do something to test on your other branch but it looks like we don't have the ability to do that just yet 🫠 give me one sec

ef4:
Marking this non-breaking because in practice we actually already required node 20 in embroider 4.x. It was an oversight that this wasn't already declared in core.

PR #2736, opened 2026-05-25
bump the monorepo to es2022 and Node >= 20.19

davidtaylorhq:
es2022 output is not safe on Node 12/14, so engines.node is raised to ">= 20.19.*" for the 11 packages still on the old range, matching the floor PR #2656 set for @embroider/compat and @embroider/template-tag-codemod.

`@embroider/compat` change is required due to the change in class-field semantics in es2022. (`useDefineForClassFields` default went from `false` to `true`)

(sketching out as a possibility for discussion - see #2735 for context)


packages/macros/package.json
NullVoxPopuli, line 1: macros cannot change with a breaking impact -- we have to keep the engines.node here the same

PR #2740, opened 2026-05-25
Cleanup unused amd support features

ef4:
This is a followup to https://github.com/embroider-build/embroider/pull/2103 that removes more things that have been inert since that landed.


mansona (approved):
Looks good! great that you found a completely unused babel plugin 😂

PR #2742, opened 2026-05-25
fix: handle Embroider macro helpers in template tag codemod

### LINKED ISSUE #2671: @embroider/template-tag-codemod breaks when encountering template macros
Running the template-tag codemod on a template that uses macros breaks with the following error:

```
> pnpm dlx @embroider/template-tag-codemod
Running addon prebuild...
Completed addon prebuild.
success route template: app/templates/application.hbs 
failure component: app/components/sample.hbs

Completed run, summarizing results:

Failed to handle component: app/components/sample.hbs because:
    Unhandled exception: Error: Unable to resolve @embroider/virtual/helpers/macroCondition from app/components/sample.js
1 successful files, 1 failed files
```

I also tested this with macros in the javascript file, but those get converted as expected. It seems like the issue is purely related to looking up the helper function.

## Reproduction
1. Clone this repository: https://github.com/SanderKnauff/macro-template-tag-codemod-repro
2. Run `pnpm install`
3. Run `pnpm dlx @embroider/template-tag-codemod`

The reproduction repository was setup using `pnpm dlx ember-cli@5.12` and `pnpm install -D @embroider/macros`

olenderhub:
Fixes #2671.

The template-tag codemod failed when migrating colocated component templates that use `@embroider/macros` template helpers. In the repro, `macroCondition` with `macroGetOwnConfig` caused the resolver to produce an unresolvable `@embroider/virtual/helpers/macroCondition` import from the backing component JS file.

This adds explicit handling for known template macro helpers:
- `macroCondition` and `dependencySatisfies` are imported from `@embroider/macros` in generated `.gjs/.gts`.
- `macroGetOwnConfig` is treated as a known template-only macro helper without generating an import.

A regression test covers the repro shape, and another covers the related `dependencySatisfies` case.

### DIFF
--- a/packages/compat/src/resolver-transform.ts
+++ b/packages/compat/src/resolver-transform.ts
@@ -120,6 +120,33 @@ function builtInKeywords(emberVersion: string): Record<string, BuiltIn | undefin
       importableHelper: ['uniqueId', '@ember/helper'],
     };
   }
+
+  // Template macro helpers from @embroider/macros.
+  //
+  // In app builds these helpers are normally handled by @embroider/macros'
+  // template transforms. The template-tag codemod needs to recognize them
+  // explicitly while generating scoped .gjs/.gts output, otherwise the
+  // resolver can produce unresolvable @embroider/virtual/helpers/* imports.
+  //
+  // Some of these names are real JS exports from @embroider/macros and need
+  // explicit imports in generated .gjs/.gts output:
+  //   macroCondition, dependencySatisfies
+  //
+  // Others are template-only macro helper names that are not JS exports.
+  // These must be treated as known keywords to suppress virtual imports,
+  // without generating import statements:
+  //   macroGetOwnConfig
+  const embroiderTemplateMacroHelpers: Record<string, BuiltIn> = {
+    dependencySatisfies: {
+      importableHelper: ['dependencySatisfies', '@embroider/macros'],
+    },
+    macroCondition: {
+      importableHelper: ['macroCondition', '@embroider/macros'],
+    },
+    macroGetOwnConfig: {},
+  };
+  Object.assign(builtInKeywords, embroiderTemplateMacroHelpers);
+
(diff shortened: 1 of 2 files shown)


ef4:
The javascript macros and the handlebars macros are historically not the same thing. This PR is switching *some* of the handlebars macros into javascript ones and leaving others alone.

The simplest fix here is to treat all the handlebars macros the same way you treated `macroGetOwnConfig` (as if they are keywords, because effectively they are).

Separately, it would be good to add test coverage to show that the importable JS macros all work correctly when used inside a template, and then we could sunset the handlebars macros and always use the importable JS ones.

olenderhub:
@ef4 Thanks, I updated the PR to treat the existing Handlebars/template macros consistently as known template keywords, without generating `@embroider/macros` imports.

I also looked briefly at the importable JS macro case inside `<template>`. I may be missing something, but it looks to me like that would need production support first: the Glimmer transform currently recognizes the template macro names:
https://github.com/embroider-build/embroider/blob/main/packages/macros/src/glimmer/ast-transform.ts

while the JS macro names seem to be handled by the Babel macro path:
https://github.com/embroider-build/embroider/blob/main/packages/macros/src/babel/macros-babel-plugin.ts

and the JS exports look like runtime stubs that expect to be compiled away:
https://github.com/embroider-build/embroider/blob/main/packages/macros/src/index.ts

So I left that out of this PR for now. Please let me know if I’m missing something.

Also, I noticed the `canary-minimal-app-windows` job failed:
https://github.com/embroider-build/embroider/actions/runs/26833296917/job/79120909574?pr=2742

I can’t see enough detail from the public job page to tell whether it’s related, but the focused template-tag-codemod tests are passing locally.

NullVoxPopuli (approved):

NullVoxPopuli:
toggling CI, as this seems like a good change for the codemod

PR #2744, opened 2026-05-28
Revert "Fix emitFile warning"

mansona:
Reverts embroider-build/embroider#2732

This change is not compatible with people using rolldown directly. We discussed this as a viable use case for people using Embroider when discussing moving the Embroider config out of ember-cli-build so I figure is is something that we **want** to support.

It's probably worth adding a smoke test for rolldown directly to the monorepo but for now I thought I would just revert this one PR since it was a cosmetic change that does have functional ramifications 👍

### DIFF
--- a/packages/vite/src/resolver.ts
+++ b/packages/vite/src/resolver.ts
@@ -103,7 +103,6 @@ export function resolver(params?: { rolldown?: boolean }): Plugin {
   }
 
   let mode = '';
-  let command = '';
 
   return {
     name: 'embroider-resolver',
@@ -111,7 +110,6 @@ export function resolver(params?: { rolldown?: boolean }): Plugin {
 
     configResolved(config) {
       mode = config.mode;
-      command = config.command;
       cacheDir = normalize(config.cacheDir);
     },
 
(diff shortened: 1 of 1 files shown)


ef4:
I suspect we can just change `command === 'build'` to `command !== 'dev'`.

davidtaylorhq:
When using this plugin under rolldown, the `configResolved` hook never fires. It's vite-specific. Rolldown doesn't have a concept of 'commands'. So yes I guess in that case, switching to `!== 'dev'` would solve the problem :+1:

As for Vite, I haven't tested it, but I assume that using the name of the command would be problematic under the `experimental.bundledDev` mode of Vite 8. With that flag enabled, the internals work like `vite build`, but the command would be `dev`.

davidtaylorhq:
> As for Vite, I haven't tested it, but I assume that using the name of the command would be problematic under the experimental.bundledDev mode of Vite 8. With that flag enabled, the internals work like vite build, but the command would be dev.

I checked, and the "command" when using `experimental.bundleMode=true` is `"serve"`. So `!== "dev"` would handle that case too 👌 

Here's a PR for that: https://github.com/embroider-build/embroider/pull/2745

mansona:
closing in favour of @davidtaylorhq 's PR 👍

PR #2745, opened 2026-05-28
Fix `@embroider/virtual` emits under Vite's `experimental.bundleMode`

davidtaylorhq:
In Vite 8's `experimental.bundleMode: true` configuration, the command is returned as `"serve"`. We need to emit these vendor/test-support files in that mode.

This also fixes things for people using the plugin directly in rolldown/rollup. In that situation, the command variable is `undefined`.

Followup to ae47e7af87d0e903c6e9c36ebe989416c44ccc16


mansona (approved):
nice fix 🎉

PR #2753, opened 2026-06-02
perf: dedupe duplicate addons and avoid array spread

runspired:
While working on a v2 addon with v1 compat in its test app build I was hitting stack overflow errors from embroider/macros containing 68,900 copies of itself in its v1 addon tree all pointing at the same file on disk. This fixes that issue in 2 ways plus adds a lot more defensive coding to embroider overall around common array perf traps.

- Way it fixes it # 1: dedupe paths to avoid having 68,900 copies in the first place.
- Way it fixes it # 2: dont use array spread, which for even moderately sized arrays is very slow and leads to stack overflow errors.

Notes:

A: I'm not convinced we are deduping aggressively enough yet - I might still have follow up to take better advantage of the set in addPlugin.
B: There's clearly something going wrong higher in the the compat build system that leads us to have so many copies in the first place. I'm seeing the build take several minutes before even reaching the compatBuild for macros specifically, mostly in creating babel trees - thousands of them.
C: The project this is in has about 3k modules between library code and test code. It pulls in *maybe* another 100 files max from a few small v1-addons, and another 500 from 2 v2-addons. And yet, I am seeing logs that 10k files are transformed.

### DIFF
--- a/packages/compat/src/standalone-addon-build.ts
+++ b/packages/compat/src/standalone-addon-build.ts
@@ -8,7 +8,6 @@ import broccoliMergeTrees from 'broccoli-merge-trees';
 import writeFile from 'broccoli-file-creator';
 import type { Node } from 'broccoli-node-api';
 import type CompatApp from './compat-app';
-import { join } from 'path';
 
 export function convertLegacyAddons(compatApp: CompatApp) {
   let packageCache = PackageCache.shared('embroider', compatApp.root);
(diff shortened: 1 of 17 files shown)


packages/compat/src/standalone-addon-build.ts
```diff
@@ -57,20 +56,22 @@ ${summarizePeerDepViolations(violations)}
       segments.pop();
     }
     segments.push('moved-package-target.js');
-    return writeFile(join(...segments), '');
```
NullVoxPopuli, line 60: this change is incorrect and won't work on windows - your bot got a little overzealous with getting rid of `...` :p

runspired: this is safe and will work on windows

runspired:
closing in favor of broken out PRs see #2755 #2757 #2758 #2759 #2760

PR #2755, opened 2026-06-03
perf: use a faster hash function for cacheKey (and cache it where possible)

runspired:
This is part of a series of optimizations I've found while working on diagnosing why the v1-compat portion of a v2 addon test build was crashing with a stack-overflow after 30min. This series of optimizations now has that same build completing start-to-finish (not just compat) in 45s. This particular optimization drops the time taken from 3min (after all other optimizations are appleid) to 45s. It overlaps partially with a potentially unsafe optimization (separate PR) that had reduced the time by 1min - so the total improvement here is a bit bigger if we don't accept the unsafe optimization PR.

### DIFF
--- a/packages/macros/src/macros-config.ts
+++ b/packages/macros/src/macros-config.ts
@@ -497,7 +497,7 @@ const NO_LOCK_FILE_FOUND = 'no-lockfile-found';
 let _LOCK_FILE_CONTENTS: string | Buffer | null = null;
 function getLockFile(appRoot: string): Buffer | string {
   if (!_LOCK_FILE_CONTENTS) {
-    let lockFilePath = findUp.sync(['yarn.lock', 'package-lock.json', 'pnpm-lock.yaml'], { cwd: appRoot });
+    let lockFilePath = findUp.sync(['pnpm-lock.yaml', 'package-lock.json', 'yarn.lock'], { cwd: appRoot });
 
     if (!lockFilePath) {
       lockFilePath = findUp.sync('package.json', { cwd: appRoot });
--- a/packages/macros/src/macros-config.ts
+++ b/packages/macros/src/macros-config.ts
@@ -1,9 +1,9 @@
 import fs from 'fs';
 import { join } from 'path';
-import crypto from 'crypto';
+import { xxh3 } from '@node-rs/xxhash';
 import findUp from 'find-up';
 import type { PluginItem } from '@babel/core';
-import { RewrittenPackageCache, getOrCreate } from '@embroider/shared-internals';
+import { AddonInstance, RewrittenPackageCache, getOrCreate } from '@embroider/shared-internals';
 import type { FirstTransformParams } from './glimmer/ast-transform';
 import { makeFirstTransform, makeSecondTransform } from './glimmer/ast-transform';
 import type State from './babel/state';
@@ -338,7 +338,7 @@ export default class MacrosConfig {
       },
       owningPackageRoot,
 
-      isDevelopingPackageRoots: [...this.isDevelopingPackageRoots],
+      isDevelopingPackageRoots: Array.from(this.isDevelopingPackageRoots),
 
       get appPackageRoot() {
         return self.appRoot;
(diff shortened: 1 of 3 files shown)


packages/macros/src/macros-config.ts
```diff
@@ -504,3 +483,65 @@ function isScalar(val: any): boolean {
...
+ * Caching the lockfile contents in memory *does* mean that if you change something in the lockfile
+ * you will need to restart your build for it to take effect. However, this is already the case today
+ * with embroider assets (and for many things using vite) so this feels like an acceptable tradeoff.
+ *
+ * If we determine that this is a problem in practice, we can still gain a perf boost by caching the
+ * path lookup, and comparing the full contents of the lockfile to the cached version to determine
+ * if we need to bust the cache, without needing to read the lockfile on every single babel transform.
+ */
+const NO_LOCK_FILE_FOUND = 'no-lockfile-found';
+let _LOCK_FILE_CONTENTS: string | Buffer | null = null;
```
NullVoxPopuli, line 497: we are not guaranteed that there is only one lockfile in a project -- but we can safely assume there is only one lockfile per app root --- so if this going to be cached in module space, it needs to be a Map<appRot, lockfileContents>

packages/macros/src/macros-config.ts
```diff
@@ -504,3 +483,65 @@ function isScalar(val: any): boolean {
...
+
+    _LOCK_FILE_CONTENTS = lockFilePath ? fs.readFileSync(lockFilePath) : NO_LOCK_FILE_FOUND;
+  }
+
+  return _LOCK_FILE_CONTENTS;
+}
+
+// Cached so we only pay the cost of hashing the (potentially large) lockfile once.
+// Used as the seed for per-addon hashes so that both inputs affect the final key.
+let _lockFileHash: bigint | null = null;
```
NullVoxPopuli, line 514: same here

packages/macros/src/macros-config.ts
```diff
@@ -504,3 +483,65 @@ function isScalar(val: any): boolean {
...
+  return _lockFileHash;
+}
+
+/**
+ * Generates a two-part cache key for babel based on the contents of the app's lockfile and the
+ * addons in the project. This is used to bust babel's cache when either of those things change.
+ */
+function getAddonCacheKey(options: { appRoot: string; project: AddonInstance['project'] | null | undefined }): string {
+  // Part 1: lockfile hash, computed once and reused as the seed for part 2.
+  // @embroider/macros provides a macro called dependencySatisfies which checks if a given
```
NullVoxPopuli, line 528: why are we summarizing what `@embroider/macros` does here? remove this <3

NullVoxPopuli: (nevermind the original code did this 🙈 )

packages/macros/src/macros-config.ts
```diff
@@ -504,3 +483,65 @@ function isScalar(val: any): boolean {
...
+ *
+ * If we determine that this is a problem in practice, we can still gain a perf boost by caching the
+ * path lookup, and comparing the full contents of the lockfile to the cached version to determine
+ * if we need to bust the cache, without needing to read the lockfile on every single babel transform.
+ */
+const NO_LOCK_FILE_FOUND = 'no-lockfile-found';
+let _LOCK_FILE_CONTENTS: string | Buffer | null = null;
+function getLockFile(appRoot: string): Buffer | string {
+  if (!_LOCK_FILE_CONTENTS) {
+    let lockFilePath = findUp.sync(['yarn.lock', 'package-lock.json', 'pnpm-lock.yaml'], { cwd: appRoot });
```
NullVoxPopuli: this list should be reversed, since we discourage yarn, and we want pnpm to be the fastest

packages/macros/src/macros-config.ts
NullVoxPopuli: live review / feedback -- not relevant, because we happen to not ever build for multiple app roots in the same process

NullVoxPopuli (changes requested):
macros has to support node 12 for now, so this xxh3 package can't be used 

there are probably other changes here that could be separately PRable

runspired:
this xxh3 package supports node12

NullVoxPopuli:
`@node-rs/xxhash` is doing `require('node:module')` which is not supported in node 12, despite `node:` prefixes being added in node 12.16 -- we test on node 12.22.

node-rs/xxhash should drop the `node:` prefix, and then it'll probably work

- https://npmx.dev/package-code/@node-rs/xxhash/v/1.7.6/index.js#L6
- https://github.com/embroider-build/embroider/actions/runs/26908057317/job/79399289468?pr=2755#step:9:26


image

NullVoxPopuli:
I opened https://github.com/napi-rs/node-rs/pull/1147 to see if they're willing to add support


If not, in order for the embroider project to use this particular package in macros, we need to fork ourselves and provide instructions for how to override packages with other packages (a la nolyfill).

an alternative would be to inline a faster hash function

PR #2757, opened 2026-06-03
perf: fix stack overflow

runspired:
This is part of a series of optimizations I've found while working on diagnosing why the v1-compat portion of a v2 addon test build was crashing with a stack-overflow after 30min. This series of optimizations now has that same build completing start-to-finish (not just compat) in 45s. This particular optimization fixes the stack overflow and accounts for most of the time drop from 30min-to-crash to 4min-to-build.

The underlying root cause was a reliance on array spread in code paths that receive unbounded array sizes. In this case the array was 68,900 items (various embroider macros instances). Why the array was so large is a tougher subject to address: I have one PR that I'll open for discussion that partially addresses that issue but is likely unsafe.

PR #2758, opened 2026-06-03
perf: faster path generation, safer tree building

runspired:
This is part of a series of optimizations I've found while working on diagnosing why the v1-compat portion of a v2 addon test build was crashing with a stack-overflow after 30min. This series of optimizations now has that same build completing start-to-finish (not just compat) in 45s. This particular optimization was a smaller (but distinct) part of dropping the time from 30min-to-crash to 4min-to-build, most of which was achieved with #2757

### DIFF
--- a/packages/compat/src/standalone-addon-build.ts
+++ b/packages/compat/src/standalone-addon-build.ts
@@ -69,7 +69,7 @@ ${summarizePeerDepViolations(violations)}
     writeFile('index.json', JSON.stringify(index, null, 2)),
   ];
 
-  let finalTrees = exteriorTrees.slice().concat(additionalTrees, fakeTargets);
+  let finalTrees = exteriorTrees.concat(additionalTrees, fakeTargets);
 
   return broccoliMergeTrees(finalTrees);
 }
--- a/packages/compat/src/standalone-addon-build.ts
+++ b/packages/compat/src/standalone-addon-build.ts
@@ -8,7 +8,6 @@ import broccoliMergeTrees from 'broccoli-merge-trees';
 import writeFile from 'broccoli-file-creator';
 import type { Node } from 'broccoli-node-api';
 import type CompatApp from './compat-app';
-import { join } from 'path';
 
 export function convertLegacyAddons(compatApp: CompatApp) {
   let packageCache = PackageCache.shared('embroider', compatApp.root);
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
Bench from: https://github.com/embroider-build/embroider/pull/2756

## Before
```
benchmark                         avg (min … max) p75 / p99    (min … top 1%)
------------------------------------------------- -------------------------------
release-typescript-app: pnpm build    8.90 s/iter    8.93 s           █
                                (8.78 s … 9.02 s)    9.02 s ▅ ▅▅  ▅ ▅ █▅ ▅     ▅▅
                          (  2.46 kb … 355.81 kb) 178.92 kb █▁██▁▁█▁█▁██▁█▁▁▁▁▁██
```                          

## After
```
benchmark                         avg (min … max) p75 / p99    (min … top 1%)
------------------------------------------------- -------------------------------
release-typescript-app: pnpm build    8.75 s/iter    8.77 s █                   █
                                (8.63 s … 9.10 s)    8.81 s █▅   ▅     ▅▅▅▅ ▅   █
                          ( 37.15 kb … 212.16 kb) 110.68 kb ██▁▁▁█▁▁▁▁▁████▁█▁▁▁█
```

                       
slightly better memory, slightly better time

packages/compat/src/standalone-addon-build.ts
```diff
@@ -57,20 +56,22 @@ ${summarizePeerDepViolations(violations)}
       segments.pop();
     }
     segments.push('moved-package-target.js');
-    return writeFile(join(...segments), '');
+    return writeFile(segments.join('/'), '');
```
NullVoxPopuli, line 59: note for other reviewers, this change from path.join to string.join is safe because this segments array is created via splitting on `/` instead of path.sep

NullVoxPopuli (approved):

NullVoxPopuli (changes requested):
concat copies, so we don't need to slice first

packages/compat/src/standalone-addon-build.ts
```diff
@@ -57,20 +56,22 @@ ${summarizePeerDepViolations(violations)}
...
     }),
     new Funnel(compatApp.synthesizeVendorPackage(interiorTrees), {
       destDir: '@embroider/synthesized-vendor',
     }),
     writeFile('index.json', JSON.stringify(index, null, 2)),
-    ...fakeTargets,
-  ]);
+  ];
+
+  let finalTrees = exteriorTrees.slice().concat(additionalTrees, fakeTargets);
```
NullVoxPopuli:
```suggestion
  let finalTrees = exteriorTrees.concat(additionalTrees, fakeTargets);
```

packages/compat/src/standalone-addon-build.ts
runspired: we should consider using `path.sep` for both, happy to make that change

NullVoxPopuli (approved):

PR #2763, opened 2026-06-10
vite: remove `test/index.html` from the scope of dep-optimizer scan

johanrd:
## Problem

When `optimizeDeps.entries` is unset, **Vite** defaults to scanning `**/*.html` for the dependency optimizer. It ignores `**/__tests__/**` and `**/coverage/**` — but **not** Ember's `tests/index.html`, which `import.meta.glob`s every test module into the graph. Embroider sets no `entries`, so this default applies.

This causes unnecessary scan, and by that also risks a single unresolvable import in **any** (even unrun) test file aborts the whole scan:

```
(!) Failed to run dependency scan. Skipping dependency pre-bundling.
```

The app then gets **no** pre-bundled deps → a cascade of runtime `new dependencies optimized … reloading` on cold start. 

Reproduction repository: https://github.com/johanrd/embroider-repro-optimize-deps/tree/repro/scan-skip-tests

> [!NOTE]
> **Vite does warn at startup though, so a user has a path to resolve it**, but the warning may be hard to connect to the symptom (easy to hit via a half-removed addon leaving a dangling test/`_app_` import).

## Fix (this PR = option A)

Default `optimizeDeps.entries` to `['index.html']` so the dev-server scan covers the app only. It's only set when unset, so apps can still override it.

```js
config.optimizeDeps.entries ??= ['index.html'];
```

## Options considered

| | Status quo (no change) | **A — `['index.html']` (this PR)** | B — hard-copied vite settings with excluded `tests/index.html` |
|---|---|---|---|
| Config | *none: i.e. [Vite default](http://vite.dev/config/dep-optimization-options#optimizedeps-entries): `**/*.html, '**/*.html', '!**/__tests__/**', '!**/coverage/**'`* | `['index.html']` | `['**/*.html', '!**/__tests__/**', '!**/coverage/**', '!tests/index.html']` |
| Fixes the bug | ❌ one bad test import disables app pre-bundling | ✅ | ✅ |
| `pnpm start` startup | scans + pre-bundles test-only deps too | ✅ faster — app deps only | ✅ faster — app deps only |
| Multi-`.html`-entry apps | ✅ all scanned | ⚠️ only `index.html` scanned (should support most common Ember SPAs) | ✅ preserved |
| Test-only deps | pre-bundled up front | lazy on first `/tests` visit | lazy on first `/tests` visit |
| Maintenance | n/a | ✅ drift-proof (no Vite internals copied) | ⚠️ brittle — duplicates Vite's default glob + ignore list; drifts if Vite changes them¹ |

¹ Setting `entries` is all-or-nothing: it replaces Vite's default **and** drops Vite's `__tests__`/`coverage` auto-ignore, so B must re-declare those — a copy of Vite internals that can silently break on a Vite minor.

In this PR went with **A**: simplest, drift-proof, and behaviorally identical to B for a normal app (just `index.html` + `tests/index.html`). B may be a safer option to avoid breaking behavior, but also puts us at drift risk.

Cowritten by claude

### DIFF
--- a/packages/vite/src/ember.ts
+++ b/packages/vite/src/ember.ts
@@ -45,6 +45,11 @@ export function ember(params?: {
         config.optimizeDeps.exclude ||= [];
         config.optimizeDeps.extensions ||= [];
 
+        // Pin the dep-optimizer scan to the app entry. With optimizeDeps.entries
+        // unset, Vite defaults to scanning **/*.html — it ignores __tests__/ and
+        // coverage/, but NOT Ember's tests/index.html. Overridable.
+        config.optimizeDeps.entries ??= ['index.html'];
+
         if (!config.resolve.extensions) {
           config.resolve.extensions = extensions;
         }
--- a/packages/vite/tests/optimize-deps-entries.test.js
+++ b/packages/vite/tests/optimize-deps-entries.test.js
@@ -1,0 +1,50 @@
+/* eslint-disable-next-line import/no-extraneous-dependencies */
+import { it, expect, describe, vi } from 'vitest';
+
+// `resolver()` builds a ResolverLoader(process.cwd()), which requires a real
+// embroider app at cwd. We only need the `vite-plugin-ember-config` plugin from
+// ember(), so stub the resolver plugin out.
+vi.mock('../src/resolver', () => ({ resolver: () => ({ name: 'stub-resolver' }) }));
+
+import { ember } from '../src/ember';
+
+// Regression test: ember() defaults optimizeDeps.entries to the app entry, so a
+// bad import in a test file can't abort the scan and disable app pre-bundling.
+
+function getConfigHook() {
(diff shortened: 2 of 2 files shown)


mansona (changes requested):
Alas we can never merge anything like this. We can discuss some of the issues and come up with a different solution to solve the problem.

Essentially we must optimize the app and the tests html the exact same way otherwise there is a chance that you can have duplicates of the ember packages. This causes real serious problems like `instanceof` calls not working as expected, and in extreme cases it can cause duplications of the glimmer tracking system which throws errors

johanrd:
@mansona ok, thanks for the feedback.

I guess I personally never visit `/test` on normal `pnpm start`, so I didn't consider that.

NullVoxPopuli:
it _is_ the easiest way to dev with tests <3

PR #2766, opened 2026-06-16
Igonre ember-composable-helpers' build

NullVoxPopuli:


mansona:
we should probably have a test scenario that exercises this no? 🤔

NullVoxPopuli:
100% -- I put this up to see if one already existed

mansona:
> I put this up to see if one already existed

If there was one already it would have failed before this PR since it always fails in Vite 🤔 technically I would love to see a failing test PR where this fixes it

Issue #2670, opened 2026-02-06
Route that ends with  `.map` that is included in `splitAtRoutes` causes error


kategengler:
The error is 

```
•TypeError: Failed to fetch dynamically imported module:
https://local.example.org:4200/app/-embroider-route-entrypoint.js:route=authenticated.map?import
```

and it happens when navigating to the `authenticated.map` route. 

Turning off `splitAtRoutes` prevents the error as does renaming the `map` route.

embroider v4 / with vite. All embroider deps are up-to-date.

johanrd:
Made a reproduction app here: https://github.com/johanrd/embroider-repro-optimize-deps/tree/repro/map-route-name

For the record i have only been able to reproduce in dev mode.

PR #2770, opened 2026-06-17
Fix: split child route "parent.map" fails to load in vite dev

johanrd:
Failing test + targeted fix for embroider-build/embroider#2670.

Repro: https://github.com/johanrd/embroider-repro-optimize-deps/tree/repro/map-route-name

Cowritten by claude.

## Problem

A code-split route whose name ends in the segment `map` (e.g. `authenticated.map`) cannot be loaded in `vite dev`. Production builds are unaffected.

```
TypeError: Failed to fetch dynamically imported module:
  .../app/-embroider-route-entrypoint.js:route=authenticated.map?import
```

## Root cause

`handleRouteEntrypoint` appends the raw route name, so a `map` route's id ends in `.map`. Vite's dev sourcemap middleware (same code in Vite 6, 7 and 8) strips that `.map`, looks up `…:route=<parent>` — a real module, because the parent is also split — and serves that module's sourcemap instead of the route's JavaScript.

## Fix

The route is carried in the resolver `meta` and nothing decodes the `:route=…` tail, so append a trailing `.js` to the id. Every id then ends in `.js` and every sourcemap URL ends in `.map`, so the collision becomes unrepresentable rather than merely avoided. A `\0` / `virtual-module:` prefix would be the more canonical Vite convention, but a much larger change for the same outcome.

## Test

The bug only reproduces through a real browser dynamic import; an HTTP/audit fetch against `vite dev` does **not** trigger it. So the check rides the existing `vite-app-test` scenario, which already boots a browser against `vite dev`: split `parent` / `parent.map`, add a `map` route, `visit('/parent/map')`. Fails before, passes after, no new CI job.

`compat-route-split-test`'s three route-entrypoint id assertions now assert `route=people.js`. Two of them dropped a `?import` tail, because Vite only adds that marker to ids it doesn't recognize as JS (`path.extname` of the old id was `.js:route=people`) and the id now ends in `.js`.

### DIFF
--- a/packages/core/src/module-resolver.ts
+++ b/packages/core/src/module-resolver.ts
@@ -464,7 +464,9 @@ export class Resolver {
       browser: true,
       conditions: ['default', 'imports'],
     });
-    let target = matched ? `${matched}:route=${routeName}` : `-embroider-route-entrypoint.js:route=${routeName}`;
+    // Keep the id from ending in a route-controlled extension like `.map`,
+    // which Vite's dev sourcemap middleware would mis-serve. See #2670.
+    let target = matched ? `${matched}:route=${routeName}.js` : `-embroider-route-entrypoint.js:route=${routeName}.js`;
     let specifier = resolve(pkg.root, target);
 
     return logTransition(
--- a/packages/core/src/module-resolver.ts
+++ b/packages/core/src/module-resolver.ts
@@ -464,9 +464,10 @@ export class Resolver {
       browser: true,
       conditions: ['default', 'imports'],
     });
-    // Keep the id from ending in a route-controlled extension like `.map`,
-    // which Vite's dev sourcemap middleware would mis-serve. See #2670.
-    let target = matched ? `${matched}:route=${routeName}.js` : `-embroider-route-entrypoint.js:route=${routeName}.js`;
+    // The trailing `.js` stops the id from ending in a route-controlled extension
+    // like `.map`, which Vite's dev sourcemap middleware hijacks. See #2670.
+    let entrypoint = matched ?? '-embroider-route-entrypoint.js';
+    let target = `${entrypoint}:route=${routeName}.js`;
     let specifier = resolve(pkg.root, target);
 
     return logTransition(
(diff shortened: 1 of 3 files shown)


tests/scenarios/compat-route-split-test.ts
NullVoxPopuli, line 1: why did these change?

NullVoxPopuli (commented):
I like how small and targeted this is, but do we have any tests proving that sourcemaps work (whether inline or from a map file?)

tests/scenarios/compat-route-split-test.ts
johanrd: @NullVoxPopuli After the change to module-resolver.ts, the resolver now emits `…:route=people.js` instead of `…:route=people`, and that flips whether Vite appends ?import.

johanrd: also some precedence in 087b8a84 ("update tests that were leaking vite internals")

johanrd:
@NullVoxPopuli thanks. Sourcemaps aren't changed by this PR, I see you have started your own explorations of tests proving that sourcemaps work (https://github.com/embroider-build/embroider/pull/2782)

NullVoxPopuli:
gonna re-run ci

packages/core/src/module-resolver.ts
```diff
@@ -464,7 +464,10 @@ export class Resolver {
       browser: true,
       conditions: ['default', 'imports'],
     });
-    let target = matched ? `${matched}:route=${routeName}` : `-embroider-route-entrypoint.js:route=${routeName}`;
+    // The trailing `.js` stops the id from ending in a route-controlled extension
+    // like `.map`, which Vite's dev sourcemap middleware hijacks. See #2670.
+    let entrypoint = matched ?? '-embroider-route-entrypoint.js';
+    let target = `${entrypoint}:route=${routeName}.js`;
```
NullVoxPopuli, line 470: adding an extension here seems technically incorrect, since "route name" is extensionless, but I understand that having extensions makes things work easier in vite

NullVoxPopuli (approved):

Issue #2603, opened 2025-11-05
Parralel builds fail after upgrading @embroider/vite from 1.2.1 to 1.3.6


pieter-v:
We have a monorepo containing several web applications. Each web app builds successfully on its own, but when the builds are executed in parallel, they fail (see stack trace below).

This issue did not occur in version 1.2.1, where parallel builds worked without any problems.

```
error during build:
[commonjs--resolver] ENOTEMPTY: directory not empty, rename '/private/var/folders/dw/pfvq8t0d1rqbdfqrp8fq68rc0000gn/T/embroider-vite-jump-tmp-wUlvTt' -> '/private/var/folders/dw/pfvq8t0d1rqbdfqrp8fq68rc0000gn/T/embroider-vite-jump-b9840194'
    at renameSync (node:fs:1014:11)
    at maybeCaptureNewOptimizedDep (file:///.../node_modules/.pnpm/@embroider+vite@1.3.6_@embroider+core@4.2.5_@glint+template@1.7.2__@glint+template@1.7._b46bb3134dc0b15be6137748bd4027d3/node_modules/@embroider/vite/dist/src/resolver.js:213:13)
    at resolveId (file:///.../node_modules/.pnpm/@embroider+vite@1.3.6_@embroider+core@4.2.5_@glint+template@1.7.2__@glint+template@1.7._b46bb3134dc0b15be6137748bd4027d3/node_modules/@embroider/vite/dist/src/resolver.js:51:34)
    at async Object.resolveId (file:///.../node_modules/.pnpm/@embroider+vite@1.3.6_@embroider+core@4.2.5_@glint+template@1.7.2__@glint+template@1.7._b46bb3134dc0b15be6137748bd4027d3/node_modules/@embroider/vite/dist/src/resolver.js:123:30)
    at async PluginDriver.hookFirstAndGetPlugin (file:///.../node_modules/.pnpm/rollup@4.52.5/node_modules/rollup/dist/es/shared/node-entry.js:22308:28)
    at async resolveId (file:///.../node_modules/.pnpm/rollup@4.52.5/node_modules/rollup/dist/es/shared/node-entry.js:20785:26)
    at async ModuleLoader.resolveId (file:///.../node_modules/.pnpm/rollup@4.52.5/node_modules/rollup/dist/es/shared/node-entry.js:21221:15)
[...]

PR #2771, opened 2026-06-17
fix: tolerate ENOTEMPTY on jumpRoot rename race in @embroider/vite resolver

hugorodro:
## Summary

`maybeCaptureNewOptimizedDep` in `@embroider/vite`'s resolver already tolerates `EEXIST` from `renameSync(tmp, jumpRoot)` because the jump root is content-addressable — if another concurrent build won the race, its directory is equivalent to ours and we can safely continue.

On Linux, the same race surfaces as `ENOTEMPTY` instead of `EEXIST` (rename onto a now-non-empty directory). Today that errno escapes and aborts the build.

This widens the recovery to cover both errnos, treating them the same way.

```ts
} catch (err) {
-   if (err.code === 'EEXIST') {
+   if (err.code === 'EEXIST' || err.code === 'ENOTEMPTY') {
      // somebody raced us and made it first. It's content-addressable so
-     // that's fine.
+     // that's fine. ENOTEMPTY happens on Linux when renaming onto a
+     // now-non-empty directory; same race, different errno.
    } else {
      throw err;
    }
```

Refs #2603 — multiple users in that issue (including on `@embroider/vite@1.7.x`) hit `ENOTEMPTY` in parallel/CI builds with the same `embroider-vite-jump-tmp-* -> embroider-vite-jump-<hash>` stack trace.

## Verification

We've been running this patch in production via `pnpm patch` on `@embroider/vite@1.7.3` across our monorepo CI for parallel builds, and it cleanly resolves the intermittent `ENOTEMPTY` failures we were seeing on Linux runners while keeping behavior identical for the success and `EEXIST` paths.

### DIFF
--- a/packages/vite/src/resolver.ts
+++ b/packages/vite/src/resolver.ts
@@ -258,9 +258,10 @@ async function maybeCaptureNewOptimizedDep(
     try {
       renameSync(tmp, jumpRoot);
     } catch (err) {
-      if (err.code === 'EEXIST') {
+      if (err.code === 'EEXIST' || err.code === 'ENOTEMPTY') {
         // somebody raced us and made it first. It's content-addressable so
-        // that's fine.
+        // that's fine. ENOTEMPTY happens on Linux when renaming onto a
+        // now-non-empty directory; same race, different errno.
       } else {
         throw err;
       }


packages/vite/src/resolver.ts
```diff
@@ -258,9 +258,10 @@ async function maybeCaptureNewOptimizedDep(
     try {
       renameSync(tmp, jumpRoot);
     } catch (err) {
-      if (err.code === 'EEXIST') {
+      if (err.code === 'EEXIST' || err.code === 'ENOTEMPTY') {
         // somebody raced us and made it first. It's content-addressable so
-        // that's fine.
+        // that's fine. ENOTEMPTY happens on Linux when renaming onto a
```
NullVoxPopuli, line 263: Specifically ARW Linux, tho. I run linux on x86 all the time and no issue

ef4 (approved):

PR #2773, opened 2026-06-23
Preserve source maps in keep-assets

NullVoxPopuli:
keep-assets drops sourcemaps in two places -- `transform` swaps a kept asset's module for a placeholder and returns no map, and `renderChunk` prepends the `import` statements without one either. So importing a kept asset (a colocated css file, say) warns `SOURCEMAP_BROKEN`, empties the js chunk's map, and the css itself ships with no `.map` -- even when an upstream plugin (ember-scoped-css) handed rollup a real `{ code, map }` from its load hook.

Now:

- `transform` captures the combined map for text assets we know how to annotate (a comment-syntax table, currently just css) and stashes it in module meta. Rollup's synthesized identity map and broken-chain maps are recognized and ignored, so assets with no real upstream map are emitted byte-for-byte, exactly as before.
- `generateBundle` emits the companion `.map` and appends the `sourceMappingURL` comment, honoring each output's `sourcemap` setting (`false` / `true` / `'inline'` / `'hidden'`). Absolute paths in `sources` are rebased so build-machine paths don't leak into published maps.
- the placeholder module returns `{ mappings: '' }` (rollup's convention for "deliberately no map") -- except when a real upstream map exists on a file type we can't annotate, where we return no map at all so the drop stays loud.
- `renderChunk` rewrites via magic-string so the chunk's map survives the injected imports.

binary assets are untouched (no stray `.map`).

found this while teaching ember-scoped-css to emit css sourcemaps -- it hands rollup a real map now, but keep-assets was throwing it away.

---

Disclosure: the review-response commits and this description were written by an AI (Claude Fable 5, via Claude Code), directed and reviewed by @NullVoxPopuli.


packages/addon-dev/src/rollup-keep-assets.ts
```diff
@@ -50,36 +51,90 @@ export default function keepAssets({
...
           // side-effect only
-          return `${marker}("${ref}")`;
+          replacement = `${marker}("${ref}")`;
         }
+
+        // The original module content (the asset) is gone from the JS graph —
+        // it's been emitted as a file and replaced by this placeholder. There's
+        // nothing to map it to, so hand back an empty map rather than letting
+        // rollup warn that this transform dropped the source map.
+        return { code: replacement, map: { mappings: '' } };
```
NullVoxPopuli: what is mappings? why empty?

NullVoxPopuli: map is emitted it renderChunk -- does that relate to `mappings: ''`

packages/addon-dev/src/rollup-keep-assets.ts
```diff
@@ -50,36 +51,90 @@ export default function keepAssets({
...
+        }
+      }
+
+      magic.prepend(imports());
+
+      // Returning a map keeps the chunk's source map intact across the import
+      // injection and marker replacement above.
+      return {
+        code: magic.toString(),
+        map: magic.generateMap({ hires: true }),
```
NullVoxPopuli: map generated here

packages/addon-dev/src/rollup-keep-assets.ts
```diff
@@ -50,36 +51,90 @@ export default function keepAssets({
...
+        // This is gated to `.css` on purpose: the `sourceMappingURL` annotation
+        // is a CSS/JS comment, so appending it to other kept assets (images,
+        // arbitrary text, etc.) would corrupt them. CSS is also the only text
+        // asset that realistically arrives here with a meaningful map.
+        if (
+          !ourMeta?.binaryLoaded &&
+          assetFileName.toLowerCase().endsWith('.css')
+        ) {
+          let map = this.getCombinedSourcemap();
+          if (map && map.mappings) {
```
NullVoxPopuli: do we need all of this if we have the maps in renderChunk?

packages/addon-dev/src/rollup-keep-assets.ts
NullVoxPopuli-ai-agent: `mappings` is the VLQ-encoded segment data of a sourcemap — the part that actually maps generated positions back to original positions. `{ mappings: '' }` is rollup's documented convention for "this transform deliberately produced code with no mapping back to the source" (see [Source Code Transformations](https://rollupjs.org/plugin-development/#source-code-transformations)). Returning no map at all means "this plugin lost the map", which is what triggers the `SOURCEMAP_BROKEN` warning and poisons the sourcemap chain for the module.

It's the right value here: after this transform, the module's code is a synthetic placeholder (`__keep_assets_marker__(...)`) with no meaningful correspondence to the original CSS text — the CSS's real map was already captured by the `.map` asset emitted above.

And no, it's unrelated to the `renderChunk` map. This is a *module-level* map (transform maps compose per module, before bundling); `renderChunk`'s map is *chunk-level* (composed onto the already-bundled chunk's map afterward). Different pipeline stages, and neither substitutes for the other.

packages/addon-dev/src/rollup-keep-assets.ts
NullVoxPopuli-ai-agent: Right — but this map only describes what `renderChunk` itself did to the JS chunk (the prepended imports + marker rewrites), and rollup composes it with the chunk's existing map. It doesn't cover the CSS asset at all — see the thread above on the transform hook for why both are needed.

packages/addon-dev/src/rollup-keep-assets.ts
NullVoxPopuli-ai-agent: Yes — they cover different artifacts. The `renderChunk` map only keeps the **JS chunk's** map accurate. The kept CSS is emitted via `emitFile` as an *asset*, and assets sit entirely outside rollup's sourcemap pipeline — rollup never writes a `.map` for an emitted asset, no matter what `renderChunk` returns. By chunk time the CSS module's content (and its map) is gone from the JS graph anyway.

So without this block, the map an upstream plugin attached to the CSS (the ember-scoped-css case in the PR description) is silently dropped: `styles.css` ships with no map and devtools can't trace it back. This block preserves the *CSS asset's* map; `renderChunk` preserves the *JS chunk's* map.

NullVoxPopuli-ai-agent (commented):
Reviewing with the maintainer hat on. The direction is right, and the tests are a welcome addition — asserting actual trace-mapping positions rather than just map presence is exactly the right bar. CI is green. Four things I'd want resolved before merge, in rough priority order (the first two I verified empirically against rollup 4):

**1. Every plugin-loaded CSS file gets a `.map` and mutated bytes — not just CSS with real maps.** `this.getCombinedSourcemap()` never returns a falsy map: when no upstream plugin supplied one, rollup synthesizes a hires *identity* map (non-empty `mappings`, `sources: [<absolute module id>]`, full `sourcesContent`). So the `map.mappings` guard at `rollup-keep-assets.ts:70` passes for any CSS an earlier plugin loaded as a bare string — the existing `virtual-css` scenario in `tests/scenarios/v2-addon-dev-test.ts` is exactly this shape. Those addons now publish a self-referential `.map` (leaking absolute build-machine paths in `sources`, and roughly doubling the CSS's size contribution via `sourcesContent`) plus CSS whose bytes changed (appended `sourceMappingURL` comment). Suggest detecting the synthesized identity map — e.g. skip when `map.sourcesContent?.[0] === code` — so emission only happens when an upstream plugin genuinely produced a map, and adding a test for "upstream loads CSS *without* a map", which is the one CSS shape the suite doesn't cover.

**2. Calling `getCombinedSourcemap()` has side effects of its own.** If any upstream transform in the chain returned code without a map, the call itself fires `SOURCEMAP_BROKEN` (attributed to that upstream plugin) even when the output is built with `sourcemap: false` — new warning noise on previously-quiet builds — and the collapsed map comes out with `sources: []`, which this code would then emit as an authoritative `.css.map`. Whatever guard falls out of (1) should also avoid collapsing the chain when the result won't be used, and avoid emitting these junk maps.

**3. The blanket `map: { mappings: '' }` silences real drops.** For non-CSS text assets that arrive with a genuine upstream map (kept `.js`, `.scss`, …), the map is still discarded — but the `SOURCEMAP_BROKEN` warning that used to make that loud is now suppressed too. Relatedly, the `.endsWith('.css')` gate is extension-sniffing inside glob-generic infrastructure; extending this later means more hard-coded cases. An extension→comment-syntax table (`/*# … */` vs `//# …`) or an explicit plugin option would age better. At minimum, only return the empty map when the module's map was actually preserved (or never existed).

**4. Output-phase config is ignored.** Emitting during `transform` means `output.sourcemap: false` — and per-output settings in multi-output builds — can't suppress the `.map` assets or the CSS annotation. Stashing the combined map in module `meta` (next to the existing `keep-assets` meta) and emitting from `generateBundle` would let `outputOptions.sourcemap` gate it per output, and is also the natural place to rebase `sources`: passthrough maps are currently emitted verbatim, never rebased to the emitted file's location, and only resolve today because the test happens to embed `sourcesContent`.

Smaller:

- `magic.generateMap({ hires: true })` builds a per-character map of the whole chunk for what is a few marker edits plus one prepend. `hires: 'boundary'` gives near-identical accuracy at a fraction of the size — but it needs magic-string ≥ 0.30.2 (0.30.0 silently coerces the string to plain `true`), so bump the declared range if you take this.
- The `exec` loop with manual `start`/`end` arithmetic and the `update`/`remove` branch can be `magic.replaceAll(pattern, (_m, ref) => { … return getName(relativeName) ?? ''; })` — available in ^0.30.0, produces the same edits and map, and removes the offset math (an empty-string overwrite is equivalent to `remove` for map purposes).
- Tests: the second inline stub plugin and the two hand-inlined rollup builds are near-copies of `upstream()`/`build()` (including a re-declared entry-id constant). A parameterized `stubModules(record)` + `build({ plugins })` would keep the next fixture from drifting.

One thing I looked at and dismissed: a checked-in `foo.css.map` matching the `include` globs colliding with the generated companion map — rollup 4 silently keeps the first emitted asset for a duplicate `fileName`, so it's a wrong-content edge case in an exotic config, not a build break.

NullVoxPopuli:
I do not like giant walls of text :( 

https://noslopgrenade.com/

NullVoxPopuli-ai-agent:
Follow-up on the updated commits: they address the review point-for-point, and the restructured tests are an improvement. Two scope trims worth considering before this merges, since the diff has grown a fair bit:

1. **Drop the `.js`/`.mjs`/`.cjs` entries from `annotationFormats`.** The motivating case is CSS, and no test exercises a kept JS asset with an upstream map — those three entries are untested, speculative surface. Keeping the table structure but shrinking it to `.css` means extending later is still a one-line change, while the PR only claims behavior it actually verifies.

2. **Consider whether the `generateBundle` layer is worth its weight for this PR.** The meta-stashing plus the `generateBundle` hook (with `inline`/`hidden` handling) is the single biggest block (~55 source lines + ~20 test lines), and what it buys is per-output `sourcemap` config fidelity — which the default `addon.output()` (`sourcemap: true`) never exercises. It's implemented correctly and tested, so keeping it is defensible; but if reviewers prefer a tighter first landing, emitting from `transform` unconditionally and adding output-config fidelity as a follow-up would shrink this PR by roughly a third of its source delta. My mild preference is to keep it (shipping known config-ignoring behavior just to save lines is a poor trade), but it's the one honest place to cut.

Everything else in the growth is load-bearing: the identity-map detection prevents a real regression (junk `.map`s + mutated bytes for every plugin-loaded CSS, including the `virtual-css` scenario shape), the `map = undefined` path keeps genuine map drops loud, and `rebaseSources` stops absolute CI paths leaking into published tarballs.

NullVoxPopuli:
an alternative to this PR is to discourage folks from using keepAssets or rollup in general and move to rolldown/tsdown

(rolldown/tsdown doesn't need keepAssets, they have CSS import support built in)


baseline tho, we need source map testing: https://github.com/embroider-build/embroider/pull/2782

that will allow me to add CSS to those tests (in this #2773 PR) -- which will demonstarte the situation a bit better.

(but sourcemaps are presently empty when using keepAssets (w/ css) -- so it's pretty obvious right now lol)

NullVoxPopuli:
j/k, rolldown/tsdown have the same bug haha

PR #2774, opened 2026-06-24
[draft] add a maybe-babel implementation

mansona:
This is a basic implementation of 2 different options for maybeBabel that people could use for themselves. These functions have no effect unless imported and passed as the `filter` option to the `@rollup/plugin-babel` plugin: 

```js
import { defineConfig } from 'vite';
import { extensions, classicEmberSupport, ember } from '@embroider/vite';

// importing both so you can swap between them and test each one
import { maybeBabelFilter, maybeBabelRegexFilter } from '@embroider/vite';

export default defineConfig({
  plugins: [
    classicEmberSupport(),
    ember(),
    // extra plugins here
    babel({
      babelHelpers: 'runtime',
      extensions: extensions.filter(ext => ext !== '.json'),
      filter: maybeBabelFilter,
    }),
  ],
});
```

### DIFF
--- a/packages/vite/src/maybe-babel.ts
+++ b/packages/vite/src/maybe-babel.ts
@@ -1,0 +1,89 @@
+/**
+ * Most of this code was taken from https://github.com/discourse/discourse/blob/7f591bc7c590eb4f9f970a5deb33a25bfee3575a/frontend/discourse/lib/maybe-babel.mjs
+ * and previous iterations of the same code. This is currently provided as an experiment for people to try out and report back their findings
+ */
+
+import { parse as oxcParse } from 'oxc-parser';
+import { walk } from 'zimmerframe';
+import { extensions } from './ember.js';
+import { and, code, id, include, not, or } from '@rolldown/pluginutils';
+
+const babelRequiredImports = [
+  // Templates
+  '@ember/template-compiler',
+  '@ember/template-compilation',
+  'ember-cli-htmlbars',
+  'ember-cli-htmlbars-inline-precompile',
+  'htmlbars-inline-precompile',
+
+  // Macros
+  '@embroider/macros',
+  '@glimmer/env',
+  '@ember/debug',
+  '@ember/application/deprecations',
+];
+
+export async function maybeBabelFilter(id: string, code: string) {
+  const estree = await oxcParse(id, code);
+
+  let hasDecorators = false;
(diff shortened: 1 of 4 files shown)


packages/vite/src/maybe-babel.ts
```diff
@@ -0,0 +1,89 @@
...
+      Decorator(_node: unknown, { stop }: { stop: () => void }) {
+        hasDecorators = true;
+        stop();
+      },
+      ImportDeclaration(node: any, { stop }: { stop: () => void }) {
+        if (babelRequiredImports.includes(node.source.value)) {
+          hasBabelRequiredImport = true;
+          stop();
+        }
+      },
```
gossi, line 46: other detection methods would be required. The example I have in mind is `scoped-css`.

But this already has a rollup/rolldown plugin. So [inter-plugin communication](https://rolldown.rs/apis/plugin-api/inter-plugin-communication) can offer a way for other plugins to note, they have a babel part.

packages/vite/src/maybe-babel.ts
```diff
@@ -0,0 +1,89 @@
...
+  '@ember/application/deprecations',
+];
+
+export async function maybeBabelFilter(id: string, code: string) {
+  const estree = await oxcParse(id, code);
+
+  let hasDecorators = false;
+  let hasBabelRequiredImport = false;
+
+  walk(
```
NullVoxPopuli, line 32: why do you need this? Discourse did not need it

NullVoxPopuli: https://github.com/discourse/discourse/blob/main/frontend/discourse/lib/maybe-babel.mjs

mansona: if you look down the file both implementations are in there, so we can experiment and see what the differences are 😂 I've seen the first one being faster in small apps 🤷

NullVoxPopuli (changes requested):
turns out! there is an easier way!

mansona:
> turns out! there is an easier way!

care to explain? 😂

NullVoxPopuli:
soooooooon

NullVoxPopuli:
https://github.com/embroider-build/embroider/pull/2781

NullVoxPopuli:
PR to the PR https://github.com/embroider-build/embroider/pull/2784

PR #2779, opened 2026-06-26
Macros: v2 addon

NullVoxPopuli:
For compatless apps, this is needed so that the embroiderResolver can resolve the runtime module in runtime mode correctly

todo:
- [x] define a test that has a v2 addon deep in dependencies that checks the isTesting() value (e.g.: v2.0.0 of ember-window-mock)
- [x] define a test susing runtime mode in compat to see if things still work
- [x] try on ember-auto-import with a patch that sets v2 addon
  - [x] https://github.com/embroider-build/ember-auto-import/pull/719
    - looks like it's working
- [x] create repro/app for old ember
  is this relevant?: https://github.com/emberjs/ember.js/blob/295185931a184306b083af1bb1a3fc408ae85174/lib/index.cjs#L12
  - [x] ember-auto-import v1
  - [x] no ember-auto-import (3.8? lol)

### DIFF
--- a/tests/app-template-minimal/package.json
+++ b/tests/app-template-minimal/package.json
@@ -18,7 +18,7 @@
   },
   "scripts": {
     "build": "vite build",
-    "build:test": "NODE_ENV=development vite build --mode development",
+    "build:test": "vite build --mode development",
     "start": "vite",
     "test": "pnpm test:ember",
     "test:ember": "pnpm build:test && pnpm dist:ember:test",
--- a/tests/scenarios/helpers/browser.ts
+++ b/tests/scenarios/helpers/browser.ts
@@ -1,0 +1,47 @@
+import puppeteer, { type Browser } from 'puppeteer-core';
+import { existsSync } from 'fs';
+
+/**
+ * We use puppeteer-core (rather than full puppeteer) so we don't download a
+ * bundled Chromium. Instead we drive whatever Chrome/Chromium is already on the
+ * machine — the same browser testem uses in CI. Honor the usual env overrides
+ * first, then fall back to the well-known install locations per platform.
+ */
+export function findChrome(): string {
+  let fromEnv = process.env.PUPPETEER_EXECUTABLE_PATH || process.env.CHROME_BIN;
+  if (fromEnv && existsSync(fromEnv)) {
+    return fromEnv;
+  }
+
+  let candidates = [
+    // macOS
+    '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome',
+    '/Applications/Chromium.app/Contents/MacOS/Chromium',
+    // Linux
+    '/usr/bin/google-chrome',
+    '/usr/bin/google-chrome-stable',
+    '/usr/bin/chromium',
+    '/usr/bin/chromium-browser',
+    // Windows
+    'C:\\Program Files\\Google\\Chrome\\Application\\chrome.exe',
+    'C:\\Program Files (x86)\\Google\\Chrome\\Application\\chrome.exe',
+  ];
+  for (let candidate of candidates) {
+    if (existsSync(candidate)) {
+      return candidate;
+    }
+  }
+
+  throw new Error(
+    'Could not find a Chrome/Chromium executable to drive. ' +
+      'Set PUPPETEER_EXECUTABLE_PATH (or CHROME_BIN) to its path.'
+  );
+}
+
+export async function launchBrowser(): Promise<Browser> {
+  return puppeteer.launch({
+    executablePath: findChrome(),
+    headless: true,
+    args: ['--no-sandbox', '--disable-dev-shm-usage', '--disable-gpu', '--mute-audio'],
+  });
+}
--- a/tests/scenarios/macro-deep-v2-addon-compat-test.ts
+++ b/tests/scenarios/macro-deep-v2-addon-compat-test.ts
@@ -1,0 +1,100 @@
+import { appScenarios, baseV2Addon } from './scenarios';
+import type { PreparedApp } from 'scenario-tester';
+import QUnit from 'qunit';
+import merge from 'lodash/merge';
+
+const { module: Qmodule, test } = QUnit;
+
+appScenarios
+  .only('canary')
+  .map('macro-deep-v2-addon-compat-istesting', project => {
+    let addon = baseV2Addon();
+    addon.pkg.name = 'macros-consumer-addon';
+    // an app-js initializer that re-exports a module which imports undeclared macros
+    (addon.pkg as any)['ember-addon']['app-js']['./initializers/macros-consumer.js'] =
+      './app/initializers/macros-consumer.js';
+    merge(addon.files, {
+      app: {
+        initializers: {
+          'macros-consumer.js': `export { default } from 'macros-consumer-addon/macros-init';`,
+        },
+      },
+      'macros-init.js': `
+        import { isTesting } from '@embroider/macros';
+
+        export const isTestingAtModuleLoad = isTesting();
+
+        export default {
+          name: 'macros-consumer',
+          initialize() {},
+        };
+      `,
+    });
+
+    let deep = baseV2Addon();
+    deep.pkg.name = 'deep-macros-addon';
+    merge(deep.files, {
+      'is-testing-at-load.js': `
+        import { isTesting } from '@embroider/macros';
+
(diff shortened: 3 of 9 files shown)


tests/scenarios/macro-deep-v2-addon-compat-test.ts
```diff
@@ -0,0 +1,100 @@
...
+  .forEachScenario(scenario => {
+    Qmodule(scenario.name, function (hooks) {
+      let app: PreparedApp;
+
+      hooks.before(async () => {
+        app = await scenario.prepare();
+      });
+
+      test('pnpm test', async function (assert) {
+        let result = await app.execute('pnpm test');
```
mansona, line 96: This is great that it's testing the expected behaviour of the isTesting() macro in a test 👍 but we need the other side of the equation. We need something that will run the app (not in the tests) and check that some behaviour (hand wave 👋 ) is different from the test mode.

I recommend that you use the same infra that we have for testing the vite dev mode because it will run the vite dev server and you can visit the app and the tests separately and verify the behviour that way 👍

mansona (dismissed):

tests/scenarios/macro-deep-v2-addon-compat-test.ts
NullVoxPopuli: pushed. we had no existing infra that I could find that actually looked at the output in-browser. only the CLI.

NullVoxPopuli:
back in draft so I can abstract the vite + testem-proxy way of running qunit via vite dev + /tests for this PR

tests/app-template-minimal/package.json
```diff
@@ -18,7 +18,7 @@
   },
   "scripts": {
     "build": "vite build",
-    "build:test": "NODE_ENV=development vite build --mode development",
```
NullVoxPopuli, line 21: without cross-env (or the .env file), this fails on windows if invoked directly

tests/scenarios/helpers/browser.ts
```diff
@@ -0,0 +1,47 @@
...
+ * machine — the same browser testem uses in CI. Honor the usual env overrides
+ * first, then fall back to the well-known install locations per platform.
+ */
+export function findChrome(): string {
+  let fromEnv = process.env.PUPPETEER_EXECUTABLE_PATH || process.env.CHROME_BIN;
+  if (fromEnv && existsSync(fromEnv)) {
+    return fromEnv;
+  }
+
+  let candidates = [
```
NullVoxPopuli, line 16: this should be default puppeteer behavior, tbh

NullVoxPopuli: (is closer to testem)

NullVoxPopuli: this exists because I need to visit the app and ask the DOM questions, which I am uncomfortable with forcing testem in that role

NullVoxPopuli: (because `isTesting()` is not (always) a compiled thing, I need to assert actual runtime output)

driesdl (approved):

NullVoxPopuli-ai-agent:
The remaining "repro/app for old ember" boxes can be checked. I tested the change against four classic ember-cli apps, and all of them pass.

Repros: https://github.com/NullVoxPopuli-ai-agent/embroider-macros-v2-addon-repros

| App | ember-cli | ember-source | ember-auto-import | Extra |
| --- | --- | --- | --- | --- |
| [`repro-app`](https://github.com/NullVoxPopuli-ai-agent/embroider-macros-v2-addon-repros/tree/main/repro-app) | 5.12.0 | 5.12 | 2.13.1 | a v2 addon that depends on `@embroider/macros` and uses `isTesting`, `macroCondition`, `importSync` |
| [`old-app`](https://github.com/NullVoxPopuli-ai-agent/embroider-macros-v2-addon-repros/tree/main/old-app) | 3.28.6 | 3.28 | 1.12.2 | macros used in `app/app.js` |
| [`old-app-noeai`](https://github.com/NullVoxPopuli-ai-agent/embroider-macros-v2-addon-repros/tree/main/old-app-noeai) | 3.28.6 | 3.28 | none | macros used in `app/app.js` |
| [`older-app`](https://github.com/NullVoxPopuli-ai-agent/embroider-macros-v2-addon-repros/tree/main/older-app) | 3.8.3 | 3.8 | none | macros used in `app/app.js` |

Each app ran `ember build`, `ember build --environment=production`, and `ember test` twice: once with published `@embroider/macros@1.20.5` as is, and once with `"version": 2, "type": "addon"` added to its `ember-addon` field. Results were identical.

Why classic builds do not care:

- ember-cli never reads `ember-addon.version`. It finds addons by the `ember-addon` keyword and loads `ember-addon.main` ([package-info.js#L244](https://github.com/ember-cli/ember-cli/blob/master/lib/models/package-info-cache/package-info.js#L244)). So `src/ember-addon-main.js` still runs, and the classic `addon` tree still defines `@embroider/macros/runtime` and `@embroider/macros/es-compat2`.
- ember-auto-import 1.x also only checks the keyword ([package.ts#L332 on v1.12.2](https://github.com/embroider-build/ember-auto-import/blob/v1.12.2/packages/ember-auto-import/ts/package.ts#L332)).
- The "needs ember-auto-import 2" error comes from `@embroider/addon-shim`. Macros does not use the shim, so no new requirement appears.
- Under ember-auto-import 2, code inside v2 addons imports the macros runtime by relative file path, because ember-auto-import calls `babelPluginConfig()` without an owning package. The `exports` map is never consulted for it.

The ember.js `lib/index.cjs#L12` link is not relevant. That line is `addonV1Shim(...)`. ember-source needs ember-auto-import 2 because of the shim, and macros keeps its own full v1 `main`.

Two smaller observations:

1. Under ember-auto-import 2, the tests bundle now contains `@embroider/macros/src/index.js` as an AMD module named `@embroider/macros`. Before the change, ember-auto-import skipped the package as a v1 addon. The file only holds the type stubs that throw, and macros compile away before runtime, so this adds a few KB to test builds only.
2. On embroider-build/ember-auto-import#719, the `v2-addon` scenario only runs on `release`, and every `release*` job fails with "Could not find module `ember`" from ember-source. So that CI run did not verify the v2 addon plus macros case. `repro-app` above covers it, with the addon listing `@embroider/macros` as a dependency, which the ember-auto-import fixture does not.

Not tested: Ember 3.8 with ember-auto-import 1.x. The 3.28 app with ember-auto-import 1.12.2 covers the same code path.

NullVoxPopuli:
I guess that's technically valid
- https://github.com/NullVoxPopuli-ai-agent/embroider-macros-v2-addon-repros/blob/main/set-macros-version.mjs#L11
- very nontraditional way of patching type=module, but it _is_ what this PR is doing

NullVoxPopuli-ai-agent:
Node 20 results for the [same four repro apps](https://github.com/NullVoxPopuli-ai-agent/embroider-macros-v2-addon-repros), with the PR change applied: the macros change itself needs no `require(esm)` support. The one `require(esm)` requirement I found comes from `testem`, and it exists with or without this PR.

| Node | `ember build` (dev + prod) | `ember test` |
| --- | --- | --- |
| 24.20.0 | pass (4/4 apps) | pass (4/4 apps) |
| 20.19.5 (first Node 20 with `require(esm)` on by default) | pass (4/4 apps) | pass (4/4 apps) |
| 20.9.0 (no `require(esm)`) | pass (4/4 apps) | pass in the ember-cli 3.8 app only |

Details on the Node 20.9.0 `ember test` failures:

- The error is `ERR_REQUIRE_ESM`: `testem/lib/utils/fileutils.js` does `require("execa")`, and the three newer apps resolve `testem@3.20.2`, which declares `"execa": "^9.6.1"`. `execa@9` is ESM-only.
- The failure is identical with unpatched `@embroider/macros@1.20.5`, so this PR does not cause it. It is a `testem` floating-dependency issue that hits any fresh install.
- The ember-cli 3.8 app resolves `testem@2.17.0` with `execa@1` (CJS), so its `ember test` passes on 20.9.0.

On the macros package itself: every file that Node loads (`src/ember-addon-main.js`, `src/node.js`, `src/babel.js`, `src/index.js`) is compiled CJS, and `package.json` has no `"type": "module"`. The only ESM files are `src/addon/*`, and only bundlers read those. So this PR adds no `require(esm)` requirement, and Node 20.9.0 builds confirm that.

evoactivity (approved):

Issue #1773, opened 2024-01-30
[type=module] Quest Issue


NullVoxPopuli:
## For Libraries

- [ ] https://github.com/embroider-build/embroider/issues/1672
- [ ] https://github.com/embroider-build/embroider/issues/1674

### Notes

- [`keywords`' ember-addon is required](https://github.com/embroider-build/embroider/issues/1761)
- [addon shim required for "app-js"](https://github.com/embroider-build/embroider/issues/1264)

## For Apps

- [ ] https://github.com/embroider-build/embroider/issues/1678

Issue #1674, opened 2023-11-19
type=module packages are not translated correctly if they are v2 addons


NullVoxPopuli:
This is an issue for tracking exploration in https://github.com/embroider-build/embroider/pull/1572
And this is separate from this type=module issue: https://github.com/embroider-build/embroider/issues/1672

It is not possible to reproduce this behavior in the a monorepo setup for some reason, even with injected dependencies.

Repro:
- create v2 addon
- add type=module to package.json
- publish it (or use `@universal-ember/preem` @ `0.1.0`)
- create separate app in separate repo/folder
- add the above package
- start the app
- visit `/`
- receive error:
  ```
  Module not found: Error: Can't resolve '.' in '/home/nullvoxpopuli/Development/NullVoxPopuli/package-majors/node_modules/.embroider/rewritten-packages/@glimmer/component.23618b61'
  Did you mean 'index.js'?
  BREAKING CHANGE: The request '.' failed to resolve only because it was resolved as fully specified
  (probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
  The extension in the request is mandatory for it to be fully specified.
  Add the extension to the request.
  ```
  note that the error is calling out `@glimmer/component`.
  

best lead we have is this https://github.com/embroider-build/embroider/pull/1572#issuecomment-1716345434

ef4:
I suspect that the code that breaks here is already changed on embroider main, so we should probably recheck.

NullVoxPopuli:
Testing this again with the changes on main -- my steps:
```bash
npx ember-cli@beta new my-app --embroider --pnpm
# ember-cli@5.5.0-beta.1
cd my-app
pnpm add @universal-ember/preem@0.1.7-type-module
```
add app/components/demo.gjs
with content, 
```js
import { ThemeToggle } from '@universal-ember/preem';

export default ThemeToggle;
```
and app/templates/application.hbs
```hbs
<Demo />
```
and then
```
pnpm start
# error, I need to install template-imports
# ctrl+c
pnpm add ember-template-imports
pnpm start
# visit http://localhost:4200
# no error in console!
```

NullVoxPopuli:
Re-opening to communicate that in embroider@3, this isn't fixed

PR #2780, opened 2026-07-08
Support type=module v2 addons in webpack builds

### LINKED ISSUE #1672: `importSync` breaks the builds of apps consuming type=module packages
repro:
- create a v2 addon
- add type=module to the package's package.json
- add a `src/example.css` file
- add code to `src/index.js`
  ```js

  export const two = 2;
 
  importSync('./example.css');
  ```
  Note, in my actual project, importSync is wrapped in a `macroCondition`.
  if it were not wrapped in a `macroCondition` I would just use regular `import './example.css';`
- build the addon
- change the test-app ember-cli-build.js to directly use embroider (compat or strict doesn't matter):
  ```js
    skipBabel: [
      {
         package: 'qunit',
      },
    ],
  });
  ```
- start the app
- receive error
  ```bash
  Module not found: Error: Can't resolve '../../node_modules/.pnpm/@embroider+macros@1.13.3/node_modules/@embroider/macros/src/addon/es-compat2' in '/home/nvp/Development$TMPDIR/type-module-test/type-module-test/dist/index.js'
  Did you mean 'es-compat2.js'?
  BREAKING CHANGE: The request '../../node_modules/.pnpm/@embroider+macros@1.13.3/node_modules/@embroider/macros/src/addon/es-compat2' failed to resolve only because it was resolved as fully specified
  (probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
  The extension in the request is mandatory for it to be fully specified.
  Add the extension to the request.
  ``` 

To see that type=module appears to work:
- remove `importSync`
- rebuild addon
- restart app

### LINKED ISSUE #1774: [type=module] _ember_component_template_only is not a function
Actual error is a bit less terse
```
Error while processing route: index _ember_component_template_only__WEBPACK_IMPORTED_MODULE_5__ is not a function @webpack://docs-app/../../../../dist/browser/index.js?:133:5
```

<details><summary>full thing</summary>

```
Uncaught TypeError: _ember_component_template_only__WEBPACK_IMPORTED_MODULE_5__ is not a function
    <anonymous> index.js:133
    js chunk.94cc0614ce424338e4a6.js:981
    __webpack_require__ chunk.0db5a460aa4562fdfd20.js:31
    <anonymous> application.js:12
    js chunk.9eda96c766a243e217dc.js:58
    __webpack_require__ chunk.0db5a460aa4562fdfd20.js:31
    <anonymous> docs-app.js:55
    exports loader.js:106
    requireModule loader.js:27
    Ember 16
    _routeEnteredOrUpdated router_js.js:1505
    routeEnteredOrUpdated router_js.js:1515
    setupContexts router_js.js:1461
    finalizeTransition router_js.js:1369
    promise router_js.js:1314
    invokeCallback rsvp.js:434
    publish rsvp.js:420
    invoke backburner.js.js:274
    flush backburner.js.js:188
    flush backburner.js.js:344
    _end backburner.js.js:773
    end backburner.js.js:576
    _run backburner.js.js:817
    run backburner.js.js:611
    Ember 2
```

image

image

</details>

Relevant compiled output:
```js
/* harmony import */ var _ember_component_template_only__WEBPACK_IMPORTED_MODULE_5__ = __webpack_require__(/*! @ember/component/template-only */ "../../../../node_modules/.pnpm/@embroider+babel-loader-9@3.1.1_@embroider+core@3.4.3_supports-color@8.1.1_webpack@5.90.0/node_modules/@embroider/babel-loader-9/index.js?{\"variant\":{\"name\":\"dev\",\"runtime\":\"browser\",\"optimizeForProduction\":false},\"appBabelConfigPath\":\"/home/nvp/Development/NullVoxPopuli/kolay/docs-app/node_modules/.embroider/rewritten-app/_babel_config_.js\",\"cacheDirectory\":\"/tmp/embroider/webpack-babel-loader\"}!../../../../node_modules/.pnpm/@embroider+webpack@3.2.1_@embroider+core@3.4.3_webpack@5.90.0/node_modules/@embroider/webpack/src/virtual-loader.js?f=%2F%40embroider%2Fext-cjs%2F%40ember%2Fcomponent%2Ftemplate-only&a=%2Fhome%2Fnvp%2FDevelopment%2FNullVoxPopuli%2Fkolay%2Fdocs-app!");
```

Repro is easy:
- https://github.com/universal-ember/kolay/pull/11
- clone
- checkout `add-in-viewport-log`
- cd docs-app
- pnpm start

## V2 Addon Checklist

- [ ] package.json#keywords has `ember-addon`
- [ ] package.json#ember-addon exists and is
  ```js
    "ember-addon": {
      "version": 2,
      "type": "addon",
      "main": "addon-main.cjs"
    },
  ```
- [ ] exports includes both addon-main as well as the runtime file
  ```js
  "exports": {
    ".": {
      "default": "./dist/browser/index.js"
    },
    "./addon-main.cjs": {
      "require": "./addon-main.cjs"
    },
  ```

NullVoxPopuli-ai-agent:
Part of the type=module quest issue: #1773. Fixes #1672. Should also fix #1774 and likely #1674 (explanation below). Supersedes the macros half of #1906 and resolves the blocker that PR was stuck on.

## Problem

When a v2 addon sets `"type": "module"` in its package.json, webpack applies strict ESM semantics to all of the addon's `.js` files, which breaks them in three ways that non-type=module addons never hit:

1. **Lost `__esModule` interop for externalized modules** (#1774). Default-importing one of the CommonJS modules we externalize — e.g. the `/@embroider/ext-cjs/` virtual modules that stand in for `@ember/component/template-only` when it isn't resolvable statically — yields the module's exports object instead of its default export, because webpack (correctly, per spec) skips `__esModule` interop for strict-ESM importers. At runtime: `_ember_component_template_only__WEBPACK_IMPORTED_MODULE_x__ is not a function`.

2. **Fully-specified resolution** (#1672). Import specifiers inside the addon must carry file extensions, so directory imports fail, and so does the relative extensionless `es-compat2` import that `@embroider/macros` emits when compiling `importSync()`:
   ```
   Module not found: Error: Can't resolve '…/@embroider/macros/src/addon/es-compat2'
   BREAKING CHANGE: The request … failed to resolve only because it was resolved as fully specified
   ```

3. **`require()` is not allowed in strict ESM**, so the `require()` calls that `importSync()` compiles to are not treated as webpack dependencies and get left for the runtime AMD loader, which can't resolve relative paths. This was the remaining blocker on #1906's test (`Could not find module ./side-effecting.js imported from (require)`).

This is also very likely the mechanism behind #1674: a request originating in a strict-ESM addon file carries `fullySpecified` resolve options that survive being rehomed into a rewritten package (see the analysis in https://github.com/embroider-build/embroider/pull/1572#issuecomment-1716345434), producing the `Can't resolve '.'` error. With the addon's files back in `javascript/auto`, those resolve options are never created in the first place.

## Solution

- **`@embroider/webpack`**: add a module rule that opts `.js` files owned by v2 addons back into `type: 'javascript/auto'` with `resolve: { fullySpecified: false }`. This is exactly the treatment every non-type=module v2 addon already gets (their `.js` files default to `javascript/auto` because their package.json has no `type` field), so type=module addons now behave identically to all other v2 addons. Non-addon packages are unaffected and keep spec-compliant strict-ESM behavior.
- **`@embroider/macros`**: emit fully-specified (extension-bearing) paths for our `es-compat2` and `runtime` helper modules (embroider-mode only; the classic AMD specifier is unchanged). This keeps the emitted code correct under strict ESM semantics in any bundler, independent of the webpack rule.

## Testing

- New scenario `v2-addon-type-module-test.ts`: a `type=module` v2 addon (with `addon-main.cjs`) covering a template-only component, a `@glimmer/component` subclass, an internal directory import, and `importSync()` of a relative module. Runs in both `safe` and `optimized` embroider modes. All pass on `release`, `lts_3_28`, and `lts_5_12`; a control run without the two source fixes fails with the strict-ESM errors quoted above.
- Macros jest suite: 356 passing (two harness/assertion updates for the now-extension-bearing emitted paths).
- No regressions: `v2-addon-test.ts` and `macro-test.ts` release scenarios (including `CLASSIC=true` runs) all pass.
- Verified end-to-end against the quest repro repo (https://github.com/NullVoxPopuli/embroider-type-module-testing): the previously-failing `app-embroider-3-lib-import-sync` app builds and passes with these packages linked in.

The same interop/fully-specified fix for ember-auto-import is embroider-build/ember-auto-import#718.

### DIFF
--- a/packages/macros/src/babel/state.ts
+++ b/packages/macros/src/babel/state.ts
@@ -69,8 +69,12 @@ const runtimeAddonPath = resolve(join(__dirname, '..', 'addon'));
 
 function pathToAddon(this: State, moduleName: string): string {
   if (!this.opts.owningPackageRoot) {
-    // running inside embroider, so make a relative path to the module
-    return explicitRelative(dirname(this.sourceFile), join(runtimeAddonPath, moduleName));
+    // running inside embroider, so make a relative path to the module. The
+    // extension matters because this import may get emitted into a file
+    // whose package sets `"type": "module"`, where the bundler will apply
+    // strict ESM semantics and refuse to resolve import paths that aren't
+    // fully-specified.
+    return explicitRelative(dirname(this.sourceFile), join(runtimeAddonPath, `${moduleName}.js`));
   } else {
     // running inside a classic build, so use a classic-compatible runtime
     // specifier.
--- a/packages/macros/tests/babel/helpers.ts
+++ b/packages/macros/tests/babel/helpers.ts
@@ -44,8 +44,7 @@ export function makeRunner(transform: Transform) {
     }
     return runDefault(code, {
       dependencies: {
-        [explicitRelative(dirname(optsWithDefaults.filename), runtimeFilename.replace(/\.[tj]s$/, ''))]:
-          cachedMacrosPackage,
+        [explicitRelative(dirname(optsWithDefaults.filename), runtimeFilename)]: cachedMacrosPackage,
       },
     });
   };
--- a/packages/macros/tests/babel/import-sync.test.ts
+++ b/packages/macros/tests/babel/import-sync.test.ts
@@ -11,7 +11,7 @@ describe('importSync', function () {
       import { importSync } from '@embroider/macros';
       importSync('foo');
       `);
-      expect(code).toMatch(/import esc from "\.\.\/\.\.\/src\/addon\/es-compat2"/);
+      expect(code).toMatch(/import esc from "\.\.\/\.\.\/src\/addon\/es-compat2\.js"/);
       expect(code).toMatch(/esc\(require\(['"]foo['"]\)\)/);
       expect(code).not.toMatch(/window/);
     });
--- a/packages/webpack/src/ember-webpack.ts
+++ b/packages/webpack/src/ember-webpack.ts
@@ -11,7 +11,7 @@
 
 import type { AppMeta, BundleSummary, Packager, PackagerConstructor, Variant, ResolverOptions } from '@embroider/core';
 import { HTMLEntrypoint, getAppMeta, getPackagerCacheDir, getOrCreate } from '@embroider/core';
-import { locateEmbroiderWorkingDir, RewrittenPackageCache, tmpdir } from '@embroider/shared-internals';
+import { locateEmbroiderWorkingDir, PackageCache, RewrittenPackageCache, tmpdir } from '@embroider/shared-internals';
 import type { Configuration, RuleSetUseItem, WebpackPluginInstance } from 'webpack';
 import webpack from 'webpack';
 import type { Stats } from 'fs-extra';
@@ -140,6 +140,11 @@ const Webpack: PackagerConstructor<Options> = class Webpack implements Packager 
     warmUp(this.extraThreadLoaderOptions);
   }
 
+  private fileIsInV2Addon(filename: string): boolean {
+    let pkg = PackageCache.shared('embroider', this.appRoot).ownerOfFile(filename);
+    return Boolean(pkg?.isV2Addon());
+  }
+
   get bundleSummary(): BundleSummary {
     let bundleSummary = this._bundleSummary;
     if (bundleSummary === undefined) {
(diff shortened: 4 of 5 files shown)


NullVoxPopuli-ai-agent:
Note on CI: every `release-*` job on this PR (all 38, including suites this PR doesn't touch like `release-compat-renaming`) fails during build setup with

```
TypeError: Cannot read properties of undefined (reading 'cleanup')
TypeError: Cannot read properties of undefined (reading 'templateCompiler')
```

That's pre-existing drift, not caused by this change: the `release` scenario floats `ember-source@latest` / `ember-cli@latest`, which moved to the 7.x line after stable's last green CI run on April 29. The pinned scenarios — including the new `lts_3_28-v2-addon-type-module` and `lts_5_12-v2-addon-type-module` — pass.

Locally (against the committed lockfile, ember-source 6.12), `release-v2-addon-type-module` passes in both safe and optimized modes, and fails without the two source fixes with the strict-ESM errors described above.

NullVoxPopuli:
looks like we haven't fixed (don't plan to fix?) release scenarios

NullVoxPopuli-ai-agent:
Added explicit coverage for the "type=module v2 addon depends on a v1 addon" case: `esm-v2-addon` now declares `inner-v1-addon` as a dependency and both default-imports and named-imports from it. In embroider the v1 addon gets rewritten to v2 during compat, so this exercises the request path where an import originating in a strict-ESM file lands in a rewritten package — the suspected #1674 mechanism. Passes in safe + optimized on release, lts_3_28, and lts_5_12 locally.

PR #2788, opened 2026-08-01
Fix: virtual modules fail to resolve under vite's bundledDev

johanrd:
`experimental.bundledDev: true` crashes the dev server:

```
UNRESOLVED_IMPORT: Could not resolve '/…/app/-embroider-route-entrypoint.js:route=index'
in app/-embroider-route-entrypoint.js:route=index?rolldown-lazy=1
```

bundledDev puts rolldown in lazy dev mode, which wraps each lazy entry in a `?rolldown-lazy=` stub that re-imports the resolved id. That hands an id we already virtualized back to `resolveId` as a literal source, and we only virtualize `@embroider/core/route/…` specifiers, so nothing claims it. Not route-specific — it applies to any id we virtualize.

It hits us and not most plugins because a split route has no file behind it. Vue Router and SvelteKit lazily import a real component file, which resolves fine; we synthesize the route entrypoint, and it is the combination of a *generated* module that is *dynamically imported* that breaks.

Fix: make resolution idempotent for our own ids, using the `responseMetas` map that `load` already relies on.

Test: a unit test resolves a route entrypoint, then resolves the result again with a `?rolldown-lazy=1` importer — `null` before the fix, the same id after.

This does not make bundledDev usable on its own. The lazy entry itself still fails to resolve, which is a different code path that this PR does not touch, reported upstream as vitejs/vite#23124.

Cowritten/coinvestigated by Claude

### DIFF
--- a/packages/vite/src/resolver.ts
+++ b/packages/vite/src/resolver.ts
@@ -40,6 +40,13 @@ export function resolver(params?: { rolldown?: boolean }): Plugin {
       return await observeDepScan(context, source, importer, options);
     }
 
+    // Ids we already virtualized come back as literal sources when rolldown's
+    // lazy dev mode re-imports a lazy entry through its `?rolldown-lazy=` stub.
+    let alreadyVirtualized = responseMetas.get(normalizePath(source));
+    if (alreadyVirtualized) {
+      return { id: source, meta: { 'embroider-resolver': alreadyVirtualized } };
+    }
+
     let request = ModuleRequest.create(RollupRequestAdapter.create, {
       context,
       source,
--- a/packages/vite/src/resolver.ts
+++ b/packages/vite/src/resolver.ts
@@ -40,13 +40,6 @@ export function resolver(params?: { rolldown?: boolean }): Plugin {
       return await observeDepScan(context, source, importer, options);
     }
 
-    // Ids we already virtualized come back as literal sources when rolldown's
-    // lazy dev mode re-imports a lazy entry through its `?rolldown-lazy=` stub.
-    let alreadyVirtualized = responseMetas.get(normalizePath(source));
-    if (alreadyVirtualized) {
-      return { id: source, meta: { 'embroider-resolver': alreadyVirtualized } };
-    }
-
     let request = ModuleRequest.create(RollupRequestAdapter.create, {
       context,
       source,
@@ -157,6 +150,11 @@ export function resolver(params?: { rolldown?: boolean }): Plugin {
     },
 
     async resolveId(source, importer, options) {
+      let seen = responseMetas.get(normalizePath(source));
+      if (seen) {
+        return { id: source, meta: { 'embroider-resolver': seen } };
+      }
+
       let resolution = await resolveId(this, source, importer, options);
       if (typeof resolution === 'string') {
         return resolution;
--- a/packages/vite/tests/resolver.test.js
+++ b/packages/vite/tests/resolver.test.js
@@ -1,0 +1,48 @@
+import { describe, expect, it, vi } from 'vitest';
+
+const virtual = { type: 'route-entrypoint', specifier: '/app/-embroider-route-entrypoint.js:route=application' };
+const virtualId = '/app/-embroider-route-entrypoint.js:route=application';
+
+class FakeResolverLoader {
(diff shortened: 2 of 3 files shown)


packages/vite/src/resolver.ts
```diff
@@ -40,6 +40,13 @@ export function resolver(params?: { rolldown?: boolean }): Plugin {
       return await observeDepScan(context, source, importer, options);
     }
 
+    // Ids we already virtualized come back as literal sources when rolldown's
+    // lazy dev mode re-imports a lazy entry through its `?rolldown-lazy=` stub.
+    let alreadyVirtualized = responseMetas.get(normalizePath(source));
```
NullVoxPopuli: Why is this the solution? I suspect somewhere, we are trying to look for qps, and accidentally ignoring whole requests with unexpected qps

NullVoxPopuli: I guess, when does this cache get populated?

packages/vite/src/resolver.ts
```diff
@@ -40,6 +40,13 @@ export function resolver(params?: { rolldown?: boolean }): Plugin {
       return await observeDepScan(context, source, importer, options);
     }
 
+    // Ids we already virtualized come back as literal sources when rolldown's
```
NullVoxPopuli: I don't think this comment is fully accurate.


if we have a response for resolveId, it may not be virtual.

So, I think the entire comment should go, and the variable should be renamed to "seen".

This is where responseMetas is set... https://github.com/embroider-build/embroider/blob/bf48da2a6c1e91396ae21e1f00b0c09851edb0b7/packages/vite/src/resolver.ts#L158


so.. would it not make more sense to early return before even calling this function?: https://github.com/embroider-build/embroider/blob/bf48da2a6c1e91396ae21e1f00b0c09851edb0b7/packages/vite/src/resolver.ts#L153


e.g.:
```ts

    async resolveId(source, importer, options) {
      let seen = responseMetas.get(normalizePath(source));
    
      if (seen) {
        return { id: source, meta: { 'embroider-resolver': alreadyVirtualized } };
      }

      let resolution = await resolveId(this, source, importer, options);
      if (typeof resolution === 'string') {
        return resolution;
      }
      if (resolution && resolution.meta?.['embroider-resolver']) {
        responseMetas.set(normalizePath(resolution.id), resolution.meta['embroider-resolver'] as ResponseMeta);
      }
      return resolution;
    },
```

packages/vite/src/resolver.ts
NullVoxPopuli: found out

packages/vite/src/resolver.ts
johanrd: @NullVoxPopuli thanks, updated now

tests/scenarios/vite-internals-test.ts
```diff
@@ -412,6 +412,57 @@ function runViteInternalsTest(scenario: Scenario) {
...
+          // entry, via its `?rolldown-lazy=` stub.
+          writeFileSync(
+            resolve(app.dir, 'vite.bundled-dev.config.mjs'),
+            `
+              import { defineConfig } from 'vite';
+              import { extensions, classicEmberSupport, ember } from '@embroider/vite';
+              import { babel } from '@rollup/plugin-babel';
+              import { resolve } from 'node:path';
+
+              const reResolveVirtualId = {
```
NullVoxPopuli: Is this plugin needed?
Looks like all it does is log?

johanrd: it did also perform the two this.resolve() calls the test asserted on.

But you are probably that it was the wrong shape. Moved to a unit test that mocks `ResolverLoader` the way scripts-plugin.test.js already does.

packages/vite/tests/resolver.test.js
NullVoxPopuli, line 1: we need a scenario test with bundledDev enabled <3

NullVoxPopuli: (or for all vite 8 matrix pivot that enables bundledDev and runs all the same tests)

johanrd: Included now again

PR #2793, opened 2026-08-10
Kill the compat prebuild watcher when a vite dev server closes

gedaiu:
`compatPrebuild` forks a long lived `ember build --watch` and keeps no reference to it, so it has nothing to terminate. Vite rebuilds plugin instances when it restarts a dev server, and the new instance forks another watcher (`buildPromise` is undefined in the fresh closure) while the previous one keeps running. Every restart orphans a watcher, and each one holds a whole broccoli build resident.

Measured in an Ember monorepo with 8 vite apps: roughly 560 MB per orphan. With a `server.watch.ignored` entry that stopped ignoring `node_modules/.embroider`, which is the prebuild's own output, the restart became self feeding and reached 76 orphaned processes holding 31 GB, exhausting a 123 GB machine. Even after removing that trigger, an ordinary boot of 9 apps restarts about 4 times and leaks roughly 2 GB per `vite` start.

This stores the forked watcher and kills it in `closeBundle`, which vite calls when it tears a server down, including on restart. `vite build` forks no watcher, so it does nothing there.

Why this belongs inside the plugin: an external plugin cannot do it. A stale watcher and a freshly forked one are both `ember` children of the same vite process, and vite interleaves the teardown of the old server with the startup of the new one. I measured a plugin that reaped `ember` children killing the live watcher instead, after which editing an app source file no longer triggered a rebuild. Only the plugin instance knows which child is its own.

Tests in `packages/vite/tests/compat-prebuild.test.js` cover the close, the restart sequence (the previous watcher is killed, the replacement is left alone) and the production build case. All three fail without this change. The full `packages/vite` suite passes, `tsc -b` is clean, and prettier and eslint are clean.

The same code is present in 1.7.9.

### DIFF
--- a/packages/vite/tests/compat-prebuild.test.js
+++ b/packages/vite/tests/compat-prebuild.test.js
@@ -1,0 +1,89 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { EventEmitter } from 'node:events';
+
+// `emberBuild` resolves ember-cli relative to cwd, which is not installed for this package,
+// and forks a real long running build. Both are stubbed so the test can drive the plugin's
+// lifecycle directly.
+vi.mock('child_process', () => ({ fork: vi.fn() }));
+
+vi.mock('node:module', async importOriginal => {
+  return {
+    ...(await importOriginal()),
+    createRequire: () => ({ resolve: () => '/fake/node_modules/ember-cli/lib/cli/index.js' }),
+  };
+});
+
+const { fork } = await import('child_process');
+const { compatPrebuild } = await import('../src/build');
+
+function fakeChild() {
+  const child = new EventEmitter();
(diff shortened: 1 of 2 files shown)


packages/vite/tests/compat-prebuild.test.js
```diff
@@ -0,0 +1,89 @@
...
+  return {
+    ...(await importOriginal()),
+    createRequire: () => ({ resolve: () => '/fake/node_modules/ember-cli/lib/cli/index.js' }),
+  };
+});
+
+const { fork } = await import('child_process');
+const { compatPrebuild } = await import('../src/build');
+
+function fakeChild() {
```
NullVoxPopuli, line 19: By mocking all these things, we become less confident in the tests

PR #2797, opened 2026-08-12
Accept Babel 8 in @embroider/core and @embroider/vite

wagenet:
Widens the `@babel/*` ranges in `@embroider/core` and `@embroider/vite` so apps that have moved to Babel 8 can install them:

| package | before | after |
| --- | --- | --- |
| `@embroider/core` | `@babel/core`, `@babel/parser`, `@babel/traverse` at `^7.14.5` | `^7.14.5 \|\| ^8.0.0` |
| `@embroider/vite` | `@babel/core` at `^7.22.9` | `^7.22.9 \|\| ^8.0.0` |

The lockfile still resolves Babel 7 everywhere, so this repo's existing suites are unaffected. Only the declared specifier moves, plus one code fix and one new scenario.

## Why these two packages

Their Babel surface is small enough to audit exhaustively.

`@embroider/core` touches Babel in exactly one file, `src/describe-exports.ts`: it parses a module, walks it with `@babel/traverse`, and reads `ExportNamedDeclaration` / `ExportDefaultDeclaration` plus four `@babel/types` predicates. None of that AST changed between 7 and 8. (`@babel/parser` is declared as a dependency but never imported directly.)

`@embroider/vite` has one call, `transformAsync(src, { filename })` in `src/esbuild-resolver.ts`.

## The code fix

`describeExports` called `parse(code, opts)` and used the return value. Babel 8 turns that into a hard error:

```
Error: Starting from Babel 8.0.0, the 'parse' function expects a callback.
If you need to call it synchronously, please use 'parseSync'.
```

So the range alone was not enough. This switches the call to `parseSync`, which has existed since Babel 7.0 and behaves identically under both majors. It also swaps the `TransformOptions` type import for a type derived from `parseSync`'s own signature, because Babel 8 renamed that type to `InputOptions`.

This is the code path behind FastBoot shadowing and app-js re-export resolution in `module-resolver.ts`, which is easy to miss in a downstream smoke test if the app doesn't hit those paths.

## The scenario

`tests/scenarios/babel-8-test.ts` adds `release-minimal-babel-8`: the minimal v2 app template, built and run with Babel 8 resolved for the app, `@embroider/core` and `@embroider/vite`. It runs `vite build` and the browser test suite, and asserts the versions actually in play so the scenario can't quietly pass on Babel 7.

The minimal template is deliberate. The classic pipeline pulls Babel 7 back in through `ember-cli-babel`, so a compat scenario couldn't answer the question without fixing `ember-cli-babel` first.

Two details in how it's wired. `scenario-tester` links embroider's packages straight from the workspace, nested `node_modules` and all, so pointing only the app's `@babel/core` at 8 would have left embroider on 7 and proved nothing. The `useBabel8` helper re-links each embroider package's own copy. And since a minimal v2 app never reaches `describeExports`, there's a small direct check of it against the Babel the app resolved. Reverting `parseSync` to `parse` makes that check fail with the error above.

## Three things Babel 8 needs that are not embroider's ranges

These all surfaced while getting the scenario green, and they're the reason the range change alone isn't the whole story for anyone doing this downstream.

`decorator-transforms` has to be at least 2.4.0. Earlier versions call `assertVersion("^7.0.0-0")` and hard fail. 2.4.0 is the first release whose peer range accepts 8.

The app's `@babel/plugin-transform-runtime` options have to change. Babel 8 removed `regenerator` entirely, and `useESModules` is now implied by `@babel/runtime`'s exports map. The stock minimal template passes both, so the scenario overrides `babel.config.cjs` with the same plugin chain minus those two options. Worth knowing that the app template itself isn't Babel 8 ready as written.

Node has to be at least 20.19.5. Babel 8 is ESM-only and embroider is CJS, so this rests on `require(esm)`. Node 20.19.0 through 20.19.4 reject the require when the same ESM module is already mid-evaluation, which is exactly what happens here: `vite.config.mjs` imports `@rollup/plugin-babel`, which imports `@babel/core`, and while that graph is still evaluating `@embroider/core` requires `@babel/core` synchronously. The failure is `ERR_REQUIRE_CYCLE_MODULE`. 20.19.5 fixed it.

That last one has a wrinkle worth your call. This repo's `.npmrc` pins `use-node-version=20.19.0`, so running the scenario locally with the pinned Node hits the broken range and the scenario skips with a message. CI is fine, because the setup action strips the pin and `actions/setup-node` with `node-version: 20` resolves to 20.20.x. I left both alone rather than bump your pin or add a Node override to the scenario matrix, but either would be an improvement over a scenario that only really runs in CI. Note also that Babel 8's own `engines` asks for `^22.18.0 || >=24.11.0`, stricter than what actually works here.

## Testing

`release-minimal-babel-8`, run against `@babel/core@8.0.1`, `@babel/parser@8.0.4`, `@babel/traverse@8.0.4`:

- Node 22.22.3: 4 tests, all passing.
- Node 20.20.2 (what CI's `node-version: 20` resolves to): 4 tests, all passing.
- Node 20.19.0 (the pinned local version): skips with an explanatory assertion.

Repo defaults (Babel 7, pinned Node 20.19.0, normal lockfile install):

- `pnpm compile`: clean.
- `pnpm lint`: clean.
- `@embroider/core`, `@embroider/macros`, `@embroider/shared-internals`, `@embroider/vite`, `@embroider/reverse-exports`, `@embroider/broccoli-side-watch`: all passing.

Two suites fail on my machine with 5s and 15s vitest timeouts, `packages/addon-dev/tests/declarations.test.ts` and `packages/compat/tests/audit.test.ts`. Both fail the same way on a clean checkout of this branch point, so they are unrelated to this change. They look like this machine being slow rather than anything real.

## What this does not claim

`@embroider/compat` and `@embroider/webpack` keep their Babel 7 ranges. Compat calls the sync-style `transform()` in four places in `compat-adapters/`, which fails the same way `parse()` did, and it ships `@babel/preset-env` and several `@babel/plugin-*` at `^7`. Plugins have to match the core's major, so widening compat's `@babel/core` without moving its plugins in lockstep would let a broken combination install. Same reasoning for `@embroider/template-tag-codemod`. Those can follow separately.

Two more things to be aware of downstream:

Plugins fed through the config pipeline must match the core's major. Babel 8's parser emits `TSEnumBody` where 7 put members on `TSEnumDeclaration`, so a v8 TS plugin against a v7 core (or the reverse) crashes. Widening these ranges does not cause that. It just lets you pick one major consistently.

Jest cannot load Babel 8. `jest-runtime` implements its own `require` and does not support requiring ESM, so `@embroider/core` under Jest with Babel 8 fails with `SyntaxError: Unexpected token 'export'` from inside `@babel/core/lib/index.js`. That is a Jest limitation, not an Embroider one. Vitest is fine.

---

Opened by @wagenet. The investigation, the code change, and the testing in this PR were done by Claude (Anthropic), working in @wagenet's checkout. The text above is Claude's, not @wagenet's. Happy to answer questions or restructure it however you prefer.


NullVoxPopuli:
Converted to draft while @wagenet and i sync in slack

packages/core/src/describe-exports.ts
```diff
@@ -1,12 +1,13 @@
...
-import { parse } from '@babel/core';
+import { parseSync } from '@babel/core';
 import type { NodePath } from '@babel/traverse';
 import traverse from '@babel/traverse';
 import { types as t } from '@babel/core';
 import assertNever from 'assert-never';
 
-export function describeExports(code: string, babelParserConfig: TransformOptions): { names: Set<string> } {
-  let ast = parse(code, babelParserConfig);
+type ParseOptions = NonNullable<Parameters<typeof parseSync>[1]>;
```
NullVoxPopuli, line 7: the name of the types changed between 7 and 8, so here the type is derived instead

tests/scenarios/scenarios.ts
```diff
@@ -159,6 +159,10 @@ async function release(project: Project) {
   project.linkDevDependency('ember-page-title', { baseDir: __dirname, resolveName: 'ember-page-title-9' });
   project.removeDevDependency('ember-cli-app-version');
   project.linkDevDependency('@glimmer/component', { baseDir: __dirname });
+}
+
+async function release(project: Project) {
+  releaseDeps(project);
 
   updateEmberQunit(project);
```
NullVoxPopuli, line 167: I made a separate utility function here because updateEmberQunit in particular is not compatible with the minimal app setup, which imports the env from package.json#imports

.github/workflows/ci.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         with:
           node-version: ${{matrix.node}}
           cache: pnpm
-      - run: pnpm install --no-lockfile
```
NullVoxPopuli, line 76: this is not how we want to test extended node support.

because then our build tools (for our repo) need to also support every version)


instead, what we want to do is:
- build everything early on and upload to cache
- then: download everything each step, and restore from cache.

This simulates an build from npm, and will allow us to actually test what we want

ef4:
This PR is confused about how package managers work.

Using `7 || 8` in `dependencies` doesn't give people some nuanced broad support for both. It's just going to solve to 8 the vast majority of the time. So either 8 works for all our supported users (in which case, we should completely drop 7) or it doesn't (in which case it's not safe to say "7 || 8" because nothing is going to tell the package manager *when* it's OK to use 8.

But also: the usage in `@embroider/core` is definitely entirely internal (all calls bottom out with `{configFile: false}` and don't accept user-provided arguments, so there's no way for users to influence the transformation that's running). The app's babel version has nothing to do with what `@embroider/core` is doing. So there's no reason for apps to care what version appears in core's dependencies (beyond node_modules size optimization, which this PR also fails to achieve since the package manager is still free to pick 8 even when the app already has 7).

The usage in `@embroider/vite` isn't entirely internal, but that implies to me that we have a bug and `@babel/core` should have been in `peerDependencies` there. And in that case, `7 || 8` is fine since the choice between them is made by the app's dependencies.

packages/core/src/describe-exports.ts
NullVoxPopuli: since the move to treating `@embroider/core`s babel as internal, this file's change is no longer needed, and has been extracted to a separate PR

NullVoxPopuli: https://github.com/embroider-build/embroider/pull/2819

ef4:
I'm documenting this as a bug fix to explain why people are seeing a new peerDep in a non-major release. It was always intended that `@embroider/vite` is using the same babel you've configured in your app's own babel config. It was a bug that those could vary. This fixes the bug, while also (as an enhancement) extending the allowed range.

PR #2800, opened 2026-08-26
addon-dev: read externals from ember-source's renamed-modules

NullVoxPopuli:
`addon.dependencies()` now marks a package as external when ember-source lists it in `ember-addon.renamed-modules`. The static `emberVirtualPackages` list stays as the fallback, so nothing changes for addons that build without ember-source installed.

This removes the rollup warning for modules newer than the static list:

```
(!) Unresolved dependencies
@ember/reactive/collections (imported by "src/services/known-modules.ts")
@ember/reactive (imported by "src/services/known-modules.ts")
```

## How it works

At `buildStart`, the plugin resolves `ember-source/package.json` from the addon's working directory. If found, every package name derived from the `renamed-modules` keys joins the external check, and the file joins rollup's watch list. If ember-source is not resolvable, the plugin behaves exactly as before.

## Context

#2655 and #2704 added `@ember/reactive` to the static list. #2704's description already named this approach as the original goal, with the concern that ember-source may be absent at addon build time. The opportunistic resolve plus static fallback answers that concern, and no list update is needed for future ember-source modules.

This change does not touch `ember-standard-modules.ts`, so classic-build behavior is unaffected (the failure mode that stopped #2655).

## Testing

- New unit tests in `packages/addon-dev/tests/dependencies.test.ts`: a fake ember-source with `renamed-modules` gets its modules externalized with zero rollup warnings, and the static list still applies when ember-source is absent.
- Verified against a real addon (`ember-repl` in NullVoxPopuli/limber, ember-source 7.x alpha): the warning disappears and the `@ember/reactive` imports stay intact in `dist`.

### DIFF
--- a/packages/addon-dev/src/rollup-addon-dependencies.ts
+++ b/packages/addon-dev/src/rollup-addon-dependencies.ts
@@ -1,5 +1,7 @@
 import type { Plugin } from 'rollup';
 import { readJsonSync } from 'fs-extra';
+import { createRequire } from 'module';
+import { join } from 'path';
 import {
   emberVirtualPackages,
   emberVirtualPeerDeps,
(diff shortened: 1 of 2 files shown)


NullVoxPopuli:
gonna close and add to here: https://github.com/embroider-build/embroider/blob/5449eb2944cd13548794ceff97df1d86a8c5d53d/packages/shared-internals/src/ember-standard-modules.ts

PR #2802, opened 2026-09-01
Don't re-run the declarations type-check on every watch rebuild

wagenet:
`addon.declarations()` spawns your type-check command from `buildStart` and awaits it in `writeBundle`. Rollup calls `buildStart` on every build, so an addon whose `start` script is `rollup --config --watch` runs a full `tsc` / `ember-tsc --declaration` on every rebuild.

In one v2 addon we measured (~200 source files) that's 4.14s added to a rebuild whose bundle step takes well under a second. On every keystroke.

The plugin already reads `this.meta.watchMode`, using it to downgrade a failed type-check from `error` to `warn`, but never to decide whether to run at all. This PR uses it for that.

## Semantics: why "once" and not "skip"

Skipping entirely in watch mode was my first instinct and I think it's wrong. Consider who actually reads `declarations/`:

* an app's runtime build (vite/webpack): never. Types are stripped; `.d.ts` is not an input.
* the addon author's own editor: no, it type-checks the addon's source directly.
* a **consuming** app's editor and its `tsc`/`glint` run: yes, resolved through the addon's `types`/`exports`.

So during a watch session the consumer needs those files to *exist*. It does not need them regenerated per keystroke, because nothing reads them between one keystroke and the next. That splits the first build from the rebuilds:

* **first build: emit.** Otherwise `pnpm start` on a fresh clone leaves `declarations/` empty and every consuming app's editor lights up red, which is a much worse failure than staleness.
* **rebuilds: skip.**

Hence the default, `watch: 'once'`. A run that fails is retried on the next rebuild, so if you start watch with a type error and then fix it you still get declarations instead of silently going without for the rest of the session.

`'always'` restores today's behavior. `'never'` is for addons that run their own `tsc --watch` alongside rollup.

## API

The second argument now takes an options object in addition to the command string it already took:

```js
addon.declarations('declarations')                       // 'once' (new default)
addon.declarations('declarations', 'tsc --declaration')  // unchanged
addon.declarations('declarations', { watch: 'always' })
addon.declarations('declarations', { command: 'tsc --declaration', watch: 'never' })
```

## `this.meta.watchMode` availability

It lives on `MinimalPluginContext`, so it is present in `buildStart`. Rollup has exposed it since 2.53. addon-dev's peer range is `^4.6.0`, and the version this repo tests against (3.30) has it as well. The existing `error`-vs-`warn` branch already depends on it.

## Is this breaking?

I'd call it a minor. Reasoning below, and I'm happy to be overruled.

Nothing changes outside watch mode. `rollup -c` in CI, in a release script, in `prepack` all behave exactly as before, so no published package's contents change.

What changes is the watch dev loop, in two ways:

1. Type errors are no longer reported on each rebuild. An author who uses watch as their only type feedback loses that. They still get it from their editor, from CI, and from `watch: 'always'`.
2. `declarations/` can go stale during a session. This matters if you edit addon A and immediately want app B's editor to see a new export. Restart watch, or set `watch: 'always'`.

Both are dev-time only, and both have an opt-out, so this reads to me as an enhancement rather than a break.

There's a real counter-argument: keep `'always'` as the default and make `'once'` opt-in. That's strictly additive and sidesteps the semver question entirely. I didn't do it because the default is where the cost lands. The monorepo that prompted this has 17 packages calling `declarations()`, plus two blueprints that generate more, and none of them will ever go looking for an opt-in flag. Say the word and I'll flip it.

## Tests

Four new watch-mode tests drive rollup's real `watch()` through an initial build plus one rebuild, counting how many times the command actually spawned. Two more cover non-watch mode and the options-object form.

I also raised this package's `testTimeout` to 60s. The glint tests here spawn a real type-checker and take a few seconds, so on a loaded machine they were tripping vitest's 5s default. That's pre-existing, but adding more subprocess-spawning tests made it easy to hit.

### DIFF
--- a/packages/addon-dev/src/rollup-declarations.ts
+++ b/packages/addon-dev/src/rollup-declarations.ts
@@ -8,19 +8,63 @@ import { packageUp } from 'package-up';
 let glint1 = 'glint --declaration';
 let glint2 = 'ember-tsc --declaration';
 
-export default function rollupDeclarationsPlugin(
-  declarationsDir: string,
+export type DeclarationsOptions = {
   /**
    * The command to use to generate types.
    * Defaults to:
    * - glint --declaration     # for glint v1
    * - ember-tsc --declaration # for glint v2
    */
-  command?: string
-): Plugin {
-  let glintPromise: Promise<void>;
+  command?: string;
 
-  let commandToRun = command;
+  /**
+   * How often to run the command while rollup is in watch mode. Outside of
+   * watch mode it always runs.
+   *
+   * - 'once' (default): run on the first build, then leave the emitted
+   *   declarations in place for the rest of the watch session. A consuming
+   *   app's editor and type-checker need the files to exist on disk; they
+   *   don't need them regenerated on every keystroke, and the addon author
+   *   already sees type errors in their editor.
+   *   A failed run is retried on the next rebuild, so fixing the error that
+   *   broke the first attempt still gets you declarations.
+   * - 'always': type-check and re-emit on every rebuild. Always correct but
+   *   slow; on a few hundred source files this adds seconds to every rebuild.
+   * - 'never': don't emit declarations in watch mode at all. For addons that
+   *   run their own `tsc --watch` alongside rollup.
+   */
+  watch?: 'once' | 'always' | 'never';
+};
+
+export default function rollupDeclarationsPlugin(
+  declarationsDir: string,
+  commandOrOptions?: string | DeclarationsOptions
+): Plugin {
+  let options: DeclarationsOptions =
+    typeof commandOrOptions === 'string'
+      ? { command: commandOrOptions }
+      : commandOrOptions ?? {};
+
+  let watchBehavior = options.watch ?? 'once';
+
+  let glintPromise: Promise<void> | undefined;
+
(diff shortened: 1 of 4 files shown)


packages/addon-dev/src/rollup-declarations.ts
```diff
@@ -8,19 +8,63 @@ import { packageUp } from 'package-up';
...
+   * - 'once' (default): run on the first build, then leave the emitted
+   *   declarations in place for the rest of the watch session. A consuming
+   *   app's editor and type-checker need the files to exist on disk; they
+   *   don't need them regenerated on every keystroke, and the addon author
+   *   already sees type errors in their editor.
+   *   A failed run is retried on the next rebuild, so fixing the error that
+   *   broke the first attempt still gets you declarations.
+   * - 'always': type-check and re-emit on every rebuild. Always correct but
+   *   slow; on a few hundred source files this adds seconds to every rebuild.
+   * - 'never': don't emit declarations in watch mode at all. For addons that
```
NullVoxPopuli, line 33: never probably doesn't need to be supported?, and folks can just _omit_ the `addon.declarations()` plugin entirely

PR #2803, opened 2026-09-02
Don't reload virtual modules on file content changes

12finger:
## Problem

With `@embroider/vite` in dev, **every stylesheet edit triggers a full page reload** instead of a hot style update. In an app whose CSS is imported for side effects from `app/app.js` (e.g. `import './app.css'`, or a Sass entry via `vite.config` `css.preprocessorOptions`), that module is a self-accepting CSS module and Vite *does* emit a clean `hmr update` for it — but a full-reload lands on top of it every time:

```
file changed app/styles/_header.scss
[vite] (client) page reload /app/-embroider-entrypoint.js   <- unwanted
[vite] (client) hmr update /app/styles/app.css              <- what should happen alone
```

## Cause

`configureServer` in `packages/vite/src/resolver.ts` installs a watcher that reloads **every** virtual module whose `watches` paths contain the changed file, on **any** watcher event:

```ts
server.watcher.on('all', (_eventName, path) => {
  for (let [id, watches] of virtualDeps) {
    for (let watch of watches) {
      if (path.startsWith(watch)) {
        server.moduleGraph.onFileChange(id);
        let m = server.moduleGraph.getModuleById(id);
        if (m) server.reloadModule(m);
      }
    }
  }
});
```

The only virtual modules that register a non-empty `watches`:

- **the app / engine entrypoint** (`virtual-entrypoint.ts` → `watches: [fromDir]`) — its source is a list of `import`s produced by globbing `fromDir`
- **template-only-component shims** (`virtual-content.ts` → `watches: [specifier, hbs?]`) — its source is the constant `templateOnlyComponentSource()`

Both derive their output purely from **which files exist**, never from file *contents*. So only `add` / `unlink` events can invalidate them. A plain `change` event cannot change what they generate — but the watcher still calls `server.reloadModule` on the entrypoint, which has no HMR boundary, so it escalates to a full page reload.

Stylesheets are just the most visible victim (they live under the app dir, so they're inside the entrypoint's watch, and Vite otherwise HMRs them cleanly). Editing a component `.js`/`.hbs` hits the same redundant entrypoint reload — it's only less noticeable there because the end result is a full reload either way.

## Fix

Ignore `change` events in that watcher; keep reacting to `add` / `unlink` (and dir variants), which is what actually affects virtual-module output. Vite's own module graph already handles content changes — CSS HMR, JS reload, and template handling via the hbs plugin / `ember-vite-hmr`.

## Verification

Tested against a real app (Ember 6.8, Embroider core 4.6 / vite 1.7.10, Vite 8):

| edit | before | after |
|---|---|---|
| `app/styles/_header.scss` (content) | full page reload | `hmr update` only, no reload ✅ |
| `app/services/foo.js` (content) | full page reload | full page reload (unchanged) ✅ |
| add `app/components/probe.js` | entrypoint reload | entrypoint reload (still picked up) ✅ |
| delete `app/components/probe.js` | entrypoint reload | entrypoint reload (still picked up) ✅ |

Happy to add a scenario test if you can point me at the right harness — I didn't find existing coverage for the dev-server watcher / HMR behavior under `tests/scenarios/vite-*`.

### DIFF
--- a/packages/vite/src/resolver.ts
+++ b/packages/vite/src/resolver.ts
@@ -117,7 +117,18 @@ export function resolver(params?: { rolldown?: boolean }): Plugin {
 
     configureServer(s) {
       server = s;
-      server.watcher.on('all', (_eventName, path) => {
+      server.watcher.on('all', (eventName, path) => {
+        // The virtual modules that register `watches` (the app/engine
+        // entrypoint and template-only-component shims) derive their contents
+        // purely from *which* files exist under the watched paths, never from
+        // those files' contents. So only add/unlink events can invalidate
+        // them. A plain 'change' event cannot, and reloading the virtual
+        // module for it forces a redundant full page reload -- most visibly on
+        // every stylesheet edit, where it pre-empts Vite's own CSS HMR. Let
+        // Vite's module graph handle content changes.
+        if (eventName === 'change') {
+          return;
+        }
         for (let [id, watches] of virtualDeps) {
           for (let watch of watches) {
             if (path.startsWith(watch)) {


NullVoxPopuli:
is there a way I can test the before/after behavior? and better yet, is there a way to add an automated tests for that using our existing test setup?

12finger:
I currently have this fix added to my app via patching by means of `patch-package`.

I am afraid i am not the seasoned dev who can churn out a test for this fix, or present a minimal reproducible app for this, easily.
This fix was discovered when moving upwars toward Ember 6.8, this i now burried many commits back in my git history.


_some background:_
@NullVoxPopuli I am developing an Ember app which dates back to 3.20, which i started back in 2020.
I stayed put for many years, and sticked with 3.20 for yeeeeeears. I also stayed with now looong abandonned Emberfire adapter, and many other things like abandonned EmberPaper.. 
I watched the EmberInitiative come up and pick up steem.. But i stopped checking on their details a while ago..

Over the years i continued developing my app, added new features, but all in all it became clear that moving forward there needed to be found a solution for the stale Firebase solution that was baked into my app, and kept me from updating my underlying Ember stack.

Fast forward 2026: I made some tiny steps dabbeling in AI and used Githubs Copilot and Claude Code (=CC) to my pleasant surprise on some minor refactors and features. I thereby gathered confidence in their powers, and then found the solution to the millon dollar question: how, and what can be done to migrate to something future-proof regarding Ember+Firebase.
The answer was simple yet astonishing: Embroider has matured so much, that one could simply use the vanilla Firebase SDK with some light adapter around it, and would also get the nice realtime updates.
And with CC's massive help i was able to make the bumps to Embroider+Webpack, to 3.28, then to 4 then to 5 then.. all the way to currently Ember 6.12 running on Embroider+Vite.
CC wrote the firebase refactor and the adapter in no time..

So yes, i am running (what used to be) a (legacy) app, based on node-sass and a fullblown .sass indented SASS syntax with many postcss plugins ontop of that.
I wanted to keep as many goodies i got used to, and carefully adapt new conventions along the way, slowly.

That said, i have to issue the disclaimer, that 'my' above PR is not _mine_.  CC wrote it for me, as i got very surprised that a three line fix burried deep down in the internals of embroider enables CSS hot reloading for my 6 years old app on VITE. I thought it might be of interest for others with legacy apps and maybe help ease their pain and share the finding that helped me. 

If you @NullVoxPopuli genuinely think that this fix might be of real help to the wider [(legacy app plagued ;)] community, then i am happy to offer to try to water down and shrink my app to a hollow 'nothing', which resambles a reproducible minial app, that showcases the before/after.

packages/vite/src/resolver.ts
```diff
@@ -117,7 +117,18 @@ export function resolver(params?: { rolldown?: boolean }): Plugin {
 
     configureServer(s) {
       server = s;
-      server.watcher.on('all', (_eventName, path) => {
+      server.watcher.on('all', (eventName, path) => {
```
NullVoxPopuli, line 120: @copilot what other eventNames are possible? (link to docs, so I can reference)

12finger:
I am sorry i made a mess with git branching and force pushs onto this PR's branch! 😞
So i closed this PR.

For simplicity and better overview, and to avoid force pushing alltogether i opened a new PR/issue:

@NullVoxPopuli  please have a look here:  https://github.com/embroider-build/embroider/pull/2814

PR #2809, opened 2026-09-07
upgrade to broccoli v4

### LINKED ISSUE #2722: braces high vulnerability in @embroider/compat v3.9.4
@embroider/compat 3.9.4 depends on
broccoli 3.5.2 depends on
sane 4.0.0 depends on
anymatch 2.0.0 depends on
micromatch 3.1.10 depends on
braces 2.3.1

braces 2.3.1 has a high vulnerability "Uncontrolled resource consumption in braces"

Updating broccoli to 4.0.0 resolves this vulnerability.

https://github.com/advisories/GHSA-grv7-fg5c-xmjg

Geodewd549:
Let's upgrade our dependency on broccoli to v4 in order to fix some assorted security vulnerabilities.

I believe this should fix #2722.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -253,8 +253,8 @@ importers:
         specifier: ^1.0.11
         version: 1.0.11
       broccoli:
-        specifier: ^3.5.2
-        version: 3.5.2
+        specifier: ^4.0.0
+        version: 4.0.0
       broccoli-concat:
         specifier: ^4.2.5
         version: 4.2.7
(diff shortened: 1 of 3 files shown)


NullVoxPopuli:
You have a lil conflict there <3

pnpm-lock.yaml
NullVoxPopuli, line 1: There is more change in this lockfile than i expect. Can you checkout from stable, run pnpm install, and commit that? 

There should be no changes unrelated to the broccoli change

Geodewd549: whoops, my bad, i didn't realize the pnpm command that i ran grabbed some extra stuff. good catch!

NullVoxPopuli:
I don't remember if ci on the target branch is green. We'll see what happens

NullVoxPopuli (approved):

NullVoxPopuli:
Looks like only the `release-` failed, which i think is expected for now

mansona (changes requested):
We can't upgrade broccoli on this legacy branch because it would need to be a major version bump and that is already taken. This is similar to the reasoning in https://github.com/emberjs/rfcs/blob/main/text/1187-deprecate-embroider-3.md

